PR #3, opened 2017-01-17
Upgrade eslint-plugin-html

### LINKED ISSUE #2: eslint-plugin-html v2 compatibility
I am about to release a new major version of `eslint-plugin-html` and it will not export processors anymore, so [this](https://github.com/vuejs/eslint-plugin-vue/blob/6a3a6db540e823b51af1e02950896ac9c2b49219/index.js#L5) won't work anymore. Of course if you don't change your `eslint-plugin-html` dependency version everything will be fine, but I don't plan to maintain the v1, and the v2 has long awaited features like the `eslint --fix` support, which might interest you.

My plugin isn't exposing processors anymore because the current eslint system is too limited (e.i.: doesn't support --fix), and I wanted users to be able to configure the html file extensions they want (most of my plugin PR was about adding new file extensions). Instead, it patches some ESLint functions as soon as it is required. It seems scarry but it already did it from the beginning and it worked correctly since then.

So one solution would be to simply `require("eslint-plugin-html");` in your plugin and that's it. `.vue` files are already considered as HTML by default, but if you want to override the default to *only* lint `.vue` files, you can add the following setting in `eslint-config-vue`:

```json
    "settings": {
        "html/html-extensions": [".vue"],
        "html/xml-extensions": []
    }
```

See [documentation](https://github.com/BenoitZugmeyer/eslint-plugin-html/blob/v2/README.md#htmlhtml-extensions) and [changelog](https://github.com/BenoitZugmeyer/eslint-plugin-html/blob/v2/CHANGELOG.md). I released the v2 as a beta tag, so you can try it with `npm i eslint-plugin-html@beta`.

Let me know if you have any concern about those changes.

znck:
Fixes #2

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,11 @@
 'use strict'
 
+require('eslint-plugin-html')
+
 module.exports = {
-  plugins: [
-    'html'
+  settings: [
+    "html/html-extensions": [".vue"],
+     "html/xml-extensions": []
   ],
   rules: {
     'jsx-uses-vars': require('eslint-plugin-react/lib/rules/jsx-uses-vars')
--- a/index.js
+++ b/index.js
@@ -1,9 +1,9 @@
 'use strict'
 
 module.exports = {
-  processors: {
-    '.vue': require('eslint-plugin-html').processors['.vue']
-  },
+  plugins: [
+    'html'
+  ],
   rules: {
     'jsx-uses-vars': require('eslint-plugin-react/lib/rules/jsx-uses-vars')
   }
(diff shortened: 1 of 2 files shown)


znck:
👍

index.js
```diff
@@ -1,9 +1,12 @@
 'use strict'
 
+require('eslint-plugin-html')
+
 module.exports = {
-  processors: {
-    '.vue': require('eslint-plugin-html').processors['.vue']
-  },
+  settings: [
```
trainiac: I think you want curlies here.

Thanks for this plugin!

znck: Yeah definitely.

chrisvfritz: Just updated!

PR #16, opened 2017-06-03
Add information about migration of this project

michalsnik:
I think we should make it clear that this plugin changed repository and all further questions should be asked there.

This repository will soon contain completely new plugin focused on real vue.js projects' purposes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,33 +1,8 @@
 # eslint-plugin-vue
 
-ESLint plugin for Vue.js projects
+:exclamation: This plugin has been moved to [eslint-plugin-vue-libs](http://github.com/vuejs/eslint-plugin-vue-libs)
 
-## Usage
-
-1. `npm install --save-dev eslint-plugin-vue`
-2. create a file named `.eslintrc` in your project:
-
-```js
-{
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I'm not sure this is necessary, since it's only a migration from the perspective of official Vue lib maintainers. For those using this in their projects, it serves the same purpose and should support all the features it did previously, plus many more. Does that make sense?

michalsnik:
Hmm, technically yes. But we don't know yet if it will have all the settings it had previously. From what I understood this plugin was made for internal projects purposes thus it's in my opinion too opinionated and we should reevaluate all settings and make sure it's as generic as can be. It had only one rule so we'll probably leave it, but other settings - we'll see. If we'll leave any they'll for sure will be added in recommended configuration and not the base one. I'm on it.

michalsnik:
On the other hand users don't necessarily need to know what's going on with `eslint-plugin-vue-libs `as it has been made only for internal purposes and the one that they should be interested in is this one. And we'll probably publish v3.0.0 after integratin eslint-plugin-vue-trail anyway, so fine. Makes sense @chrisvfritz 👍

PR #17, opened 2017-06-03
Simplify configuration

michalsnik:
General purpose of this PR is to clean and simplify current code base to get a solid ground for upcoming `v3.0.0-beta` release.

Things, that I'm working on in this PR:
- [x] Separate and simplify configs (base + recommended in dedicated directory)
- [x] Improve rules' resolver
- [x] Update package.json
- [x] Update Readme
- [x] Update `update-rules` script
- [x] Replace TravisCI with CircleCI
- [x] Add `jsx-uses-vars` rule, that was previously available in `eslint-plugin-vue` (+ documentation and tests)

I intentionally removed mysticatea's config from `.eslintrc` so that we can make a deliberate decision about preferred styleguide in order to make this plugin's style match other internal projects. I think we should use something more general, like `airbnb-base`. But I'd like to hear your thoughts as well @chrisvfritz @mysticatea @yyx990803


michalsnik:
Could you please review this PR in your spare time @mysticatea?

config/base.js
```diff
@@ -0,0 +1,21 @@
...
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+    ecmaFeatures: {
+      jsx: true,
+    },
+  },
+
+  env: {
+    browser: true,
```
mysticatea: Personally, I dislike `browser` environment because this environment includes many global variables which has general names: `name`, `length`, `status`, `event`, `open`, `parent`, etc.... It will possibly cause the problems about `no-undef` / `no-unused-vars` / `no-shadow` rules.

So I prefer that it defines fewer global variables than `browser` environment manually (then I use `window.name` instead of `name`) ([browser.js](https://github.com/mysticatea/eslint-config/blob/8d348f853045eeab9b01dd8662a1fe6efa5170a6/browser.js) is my config for frontend. This defines classes and whitelisted variables of `browser` environment).

However, I know I'm in the minority.

.eslintrc.js
```diff
@@ -0,0 +1,22 @@
...
+  root: true,
+  parserOptions: {
+    ecmaVersion: 6,
+  },
+  env: {
+    node: true,
+    mocha: true,
+  },
+  extends: [
+    "plugin:eslint-plugin/recommended",
```
michalsnik: Ah, true. This might be a good idea 👍

circle.yml
```diff
@@ -0,0 +1,13 @@
...
+
+dependencies:
+  pre:
+    - nvm install 7
+
+test:
+  override:
+    - npm test
+    - nvm use 6 && npm test
+    - nvm use 7 && npm test
```
michalsnik: I can add 8 additionally. These tests run pretty fast and we'll be sure it's actually bulletproof.

config/base.js
michalsnik: I'd keep this simple 'browser' setting for now, at least for v3.0.0-beta to get a better picture about possible issues with this during feedback gathering period. WDYT @mysticatea ?

michalsnik:
I addressed your concerns @mysticatea. If it's ok I can merge this and release `v3.0.0-beta` finally.

mysticatea:
Oh, `.eslintrc` without extensions has been deprecated. Please use a extension.

package.json
```diff
@@ -2,51 +2,58 @@
...
+    "url": "git+https://github.com/vuejs/eslint-plugin-vue.git"
+  },
+  "bugs": {
+    "url": "https://github.com/vuejs/eslint-plugin-vue/issues"
+  },
+  "engines": {
+    "node": ">=4"
+  },
   "peerDependencies": {
     "eslint": "^3.18.0"
```
mysticatea: Today, ESLint 4.0.0 has been released. I'd like to include 4.x in peerDependencies to avoid installation failure.

michalsnik: Good point. `peerDependencies` updated

michalsnik:
I also added `package-lock.json` generated by npm 5.

michalsnik:
@mysticatea Please approve if you agree with my latest updates and I'll release beta. I removed recommendations for `html-end-tags` and `no-duplicate-attributes` because of those slight problems with parser we were discussing. We'll mark them `recommended` as soon as your new parser fixes mentioned issues.

PR #20, opened 2017-06-12
Update dependancy versioning to fix compatability with eslint v4.0.0

AnthonyPorthouse:
This fixes an error where eslint-plugin-html versions in the v2 range have issues with eslint v4's api changes.

This simply adds the new versions optional as valid dependancy ranges.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
   },
   "homepage": "https://github.com/vuejs/eslint-plugin-vue#readme",
   "peerDependencies": {
-    "eslint": "^2.0.0 || ^3.0.0"
+    "eslint": "^2.0.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {
-    "eslint-plugin-html": "^2.0.0",
-    "eslint-plugin-react": "^6.9.0"
+    "eslint-plugin-html": "^2.0.0 || ^3.0.0",
+    "eslint-plugin-react": "^6.9.0 || ^7.0.0"
   }
 }


package.json
```diff
@@ -21,10 +21,10 @@
   },
   "homepage": "https://github.com/vuejs/eslint-plugin-vue#readme",
   "peerDependencies": {
-    "eslint": "^2.0.0 || ^3.0.0"
+    "eslint": "^2.0.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {
-    "eslint-plugin-html": "^2.0.0",
-    "eslint-plugin-react": "^6.9.0"
+    "eslint-plugin-html": "^2.0.0 || ^3.0.0",
```
mysticatea: `eslint-plugin-html` v3 looks to have huge breaking changes which affect to several core rules. 
I think better that we should not include `eslint-plugin-html` v3 into this PR. (we are preparing 3.0.0 beta which does not use `eslint-plugin-html`. So we cannot major-release with this.)

AnthonyPorthouse: Thanks for your input @mysticatea, do you have any sort of eta on the 3.0.0 beta as currently this causes breakage within the workflow of anyone upgrading eslint

michalsnik: Beta is coming probably tomorrow @AnthonyPorthouse. I have it ready, just waiting for final approval :)

michalsnik:
@AnthonyPorthouse beta has just been released, you can install it by running:
```
npm install eslint-plugin-vue@beta
```

or simply by updating version in `package.json` to `3.0.0`.

Let us know if you're having any issues with this. I tested it yesterday on hackernews example using eslint 4 and it worked nice.

michalsnik:
Documentation is still on `dev` branch, but will probably soon be merged to `master`. @AnthonyPorthouse

AnthonyPorthouse:
Thanks @michalsnik you've been a great help.

PR #22, opened 2017-06-14
Update master branch

michalsnik:
This PR merges development branch which was used to prepare `v3.0.0` release.

Now that it has been released I think we should update master with additional informations in readme that master is all about new version, which is still tagged as `@beta` release and is not yet official.

Given our extensive documentation in `3.0.0` thanks to @mysticatea  we could encourage people to try it and give us a feedback.


yyx990803:
Let's update README with a in-beta section at the top and merge it - after that I can make an announcement on Twitter and ask users to try it out.

PR #25, opened 2017-06-15
Update dependencies

### LINKED ISSUE #21: Compatibility with ESLint 4
Hi, `eslint-plugin-html` author here. ESLint released its v4, and I released a v3 of my plugin to support it.

I've got some users who are using the `vue` plugin reporting issues because the `html` plugin version is fixed to v2. Would you mind changing the `html` plugin version to v3?

As I understand, you are planing to move away from the `html` plugin. I don't mind, but I think that in the meantime, users should be allowed to upgrade to ESLint v4.

michalsnik:
This PR fixes #21 regarding `eslint-plugin-html` and `eslint@4.x` compatibility.

We decided to do a minor release for 2.x version while working on 3.x anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,10 @@
   },
   "homepage": "https://github.com/vuejs/eslint-plugin-vue#readme",
   "peerDependencies": {
-    "eslint": "^2.0.0 || ^3.0.0"
+    "eslint": "^2.0.0 || ^3.0.0 || ^4.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


michalsnik:
CircleCI is failing because we added it for the sake of `3.x` that's coming up. And the old version doesn't use it.

yyx990803:
I moved the original `master` into a new branch `2.x` so we'll have to redo this PR to that branch instead.

PR #34, opened 2017-06-18
Fix: upgrade vue-eslint-parser

### LINKED ISSUE #29: Self closing elements break v-if rules
Self-closing elements break the following rules:

- vue/no-invalid-template-root
- vue/no-invalid-v-else-if
- vue/no-invalid-v-else

# Expected

This should not be reported:

```html
<template>
  <c1 v-if="1" />
  <c2 v-else-if="1" />
  <c3 v-else />
</template>
```

# Happening

This works:

```html
<template>
  <c1 v-if="1"></c1>
  <c2 v-else-if="1"></c2>
  <c3 v-else></c3>
</template>
```

This doesn't:

```html
<template>
  <c1 v-if="1" />
  <c2 v-else-if="1" />
  <c3 v-else />
</template>
```

Here are the report messages:

```
  26:3  error  The template root requires the next element which has 'v-else' directives if it has 'v-if' directives     vue/no-invalid-template-root
  27:7  error  'v-else-if' directives require being preceded by the element which has a 'v-if' or 'v-else-if' directive  vue/no-invalid-v-else-if
  28:7  error  'v-else' directives require being preceded by the element which has a 'v-if' or 'v-else' directive        vue/no-invalid-v-else
```

mysticatea:
This upgrading will solve 2 problems.

1. Ensure making correct ASTs about self-closing elements (fixes #29). The first impression was difficult, but actually this was easy. (diff: https://github.com/mysticatea/vue-eslint-parser/commit/b954bee331fa38be86b3db4bf88b599c0ed4dbbe)
2. Improve the integration with `eslint-plugin-import` (refs https://github.com/vuejs/eslint-plugin-vue/issues/21#issuecomment-308957697). I avoid that error as I use `parser.parseForESLint()` alias. (diff: https://github.com/mysticatea/vue-eslint-parser/commit/62acbe86a3c1b57a0c929f8910448497f67ef6ab)

Additionally, I added some tests about above.


michalsnik:
Good work @mysticatea. I didn't know about `parseForESLint`, seems like a decent solution. Btw. pretty clever way of testing this integration, not so obvious at first sight but does the job 👍

Issue #13, opened 2017-05-28
Rule proposition: order-in-components


michalsnik:
As I mentioned in #10, I think that creating separate issue for each rule proposition has many benefits, so this is my first proposition:

Rule name: `order-in-components`
Description: Keep order of properties in components according to https://docs.gitlab.com/ce/development/fe_guide/style_guide_js.html#ordering

Gitlab's style guide lists all properties, I would however consider grouping lifecycle hooks in order to make it simpler, like so:
- name
- props
- mixins
- data
- components
- computedProps
- methods
- lifecycle hooks

Introducing order rule we can make sure that every component is written in exactly the same way, which would make it more readable and consistent across codebase.

This rule should be configurable, so that anyone can amend proposed order in `.eslintrc`.

yyx990803:
I personally prefer having all assets at top of the component (like dependencies), and then group all state-related options (props, data, computed), hence:

- name
- assets (mixins, components, directives, filters)
- props
- data
- computed
- methods
- lifecycle hooks
- render (if used)

michalsnik:
Alright @yyx990803, it makes sense 👍 We can make this order the default one, but we'll also make it possible for devs to change the order in `.eslintrc` so our recomendation is not the only right choice.

michalsnik:
I started to work on this, and there are far more properties that we should take into consideration. Order I came up with is as follows:

1. Options / Misc (`name`, `delimiters`, `functional`, `model`)
2. `el`
3. Options / Assets (`components`, `directives`, `filters`)
4. Options / Composition (`parent`, `mixins`, `extends`, `provide`, `inject`)
5. `props`
6. `propsData`
7. `data`
8. `computed`
9. `methods`
10. `watch`
11. lifecycle hooks
12. `template`
13. `render`
14. `renderError`

What is in braces should be treated equally in terms of order, the rest should be in the exact order.
I kept grouping of some properties (`misc`, `assets` and `composition`) in tact with official documentation, but had to tear `DOM` properties. I'd like to know your thought on this guys. I'll start implementing it and if we came up with different order we'll simply change the default configuration. If there are any other properties that I should take into consideration let me know!
cc @yyx990803 @chrisvfritz 

Another thing is what kind components definitions are we going to support. My first thoughts are:
1. `export.default {}` in `*.vue` file
2. `Vue.component('xxx', {})` in any file

[...]

PR #42, opened 2017-06-21
New rule `order-in-components`

michalsnik:
This PR adds new rule regarding this proposition: #13 

This rule will be applied on:
- `export default {}` inside `.vue` and `.jsx` files
- `new Vue({})` in any file
- `Vue.component('xxx', {})` or `component('xxx', {})` in any file

Thing to be noted - we won't support dynamically loaded components, as it's not so obvious to tell if it actually is or is not a vue component.

### DIFF
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -36,8 +36,9 @@ const groups = {
   ]
 }
 
-function isComponentFile (path) {
-  return path.endsWith('.vue') || path.endsWith('.jsx')
+function isComponentFile (node, path) {
+  const isVueFile = path.endsWith('.vue') || path.endsWith('.jsx')
+  return isVueFile && node.declaration.type === 'ObjectExpression'
 }
 
 function isVueComponent (node) {
@@ -113,7 +114,7 @@ function create (context) {
   return {
     ExportDefaultDeclaration (node) {
       // export default {} in .vue || .jsx
-      if (!isComponentFile(filePath)) return
+      if (!isComponentFile(node, filePath)) return
       checkOrder(node.declaration.properties, orderMap, context)
     },
     CallExpression (node) {
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -41,11 +41,26 @@ function isComponentFile (path) {
 }
 
 function isVueComponent (node) {
-  return true
+  const callee = node.callee
+
+  const isFullVueComponent = node.type === 'CallExpression' &&
+    callee.type === 'MemberExpression' &&
+    callee.object.type === 'Identifier' &&
+    callee.object.name === 'Vue' &&
+    callee.property.type === 'Identifier' &&
+    callee.property.name === 'component'
+
+  const isDestructedVueComponent = callee.type === 'Identifier' &&
+    callee.name === 'component'
+
+  return isFullVueComponent || isDestructedVueComponent
 }
 
 function isVueInstance (node) {
-  return true
+  const callee = node.callee
+  return node.type === 'NewExpression' &&
+    callee.type === 'Identifier' &&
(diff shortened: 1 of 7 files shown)


lib/rules/order-in-components.js
```diff
@@ -0,0 +1,143 @@
...
+  return {
+    ExportDefaultDeclaration (node) {
+      // export default {} in .vue || .jsx
+      if (!isComponentFile(filePath)) return
+      checkOrder(node.declaration.properties, orderMap, context)
+    },
+    CallExpression (node) {
+      // Vue.component('xxx', {}) || component('xxx', {})
+      if (!isVueComponent(node)) return
+      checkOrder(node.arguments.slice(-1)[0].properties, orderMap, context)
```
mysticatea: I guess that this line can throw `TypeError` if `component()` with no argument exists.

lib/rules/order-in-components.js
```diff
@@ -0,0 +1,143 @@
...
+    },
+    NewExpression (node) {
+      // new Vue({})
+      if (!isVueInstance(node)) return
+      checkOrder(node.arguments[0].properties, orderMap, context)
```
mysticatea: Same here.

lib/rules/order-in-components.js
```diff
@@ -0,0 +1,143 @@
...
+  const filePath = context.getFilename()
+
+  const extendedOrder = order.map(property => groups[property] || property)
+  const orderMap = getOrderMap(extendedOrder)
+
+  return {
+    ExportDefaultDeclaration (node) {
+      // export default {} in .vue || .jsx
+      if (!isComponentFile(filePath)) return
+      checkOrder(node.declaration.properties, orderMap, context)
```
mysticatea: I guess that this line can cause `TypeError` if default exports without object literals (e.g. `export default 777`) exists.

lib/rules/order-in-components.js
michalsnik: Good point, but I couldn't imagine such a component though. Fixed it anyway :)

lib/rules/order-in-components.js
michalsnik: Very good point! Fixed :)

lib/rules/order-in-components.js
michalsnik: Fixed

lib/rules/order-in-components.js
mysticatea: Many people use ESLint in editor integrations. In that case, ESLint runs on every character typing, so TypeErrors mess development experience. 😉

lib/rules/order-in-components.js
```diff
@@ -0,0 +1,148 @@
...
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  create,
+  meta: {
+    docs: {
+      description: 'Keep order of properties in components',
+      category: 'Best Practices',
+      recommended: true
```
mysticatea: I think that it needs major version if we keep `recommended: true`.

michalsnik: yeah, I was thinking the same

michalsnik:
I removed `recommended` setting and will release it as minor atm. We'll update this setting in a while @mysticatea

hirokiosame:
@michalsnik Will this eventually have support for components that were assigned to a variable, and then exported? eg.
```
const ComponentName = {
     name: 'ComponentName',
     render(h) { return h('div'); }
};

export default ComponentName;
```

hirokiosame:
@michalsnik

michalsnik:
I'm not sure yet @hirokiosame. Theoretically we could support this case, and one when you only use that component in current template, but don't even export it. But this needs further investigation, so feel free to create a proposition as separate issue.

PR #45, opened 2017-06-22
Update: improve warnings about key on template elements

### LINKED ISSUE #43: Don't require key with v-for on template tags
* **ESLint Version:** 3.19.0
* **eslint-plugin-vue Version:** 3.1.3
* **Node Version:** 7.9.0

```vue

<template>
	<div>
		<template v-for="crumb in crumbArr" :key="crumb">
			<custom-el :crumb="crumb" />
		</template>
	</div>
</template>
```

For it to work.

No ESLint error. Vue error: `<template> cannot be keyed. Place the key on real elements instead.`

mysticatea:
Fixes #43.

This PR does 2 things.

1. Fixes `no-invalid-v-for` and `require-v-for-key` rules to check whether child elements of `<template>` elements have `v-bind:key` attribute. [semver-minor]
2. Add new `no-template-key` rule to disallow `key` attribute on `<template>` elements. I think that `no-template-key` rule should be recommended, however I set `recommended: false` for now because [semver policy of ESLint](https://github.com/eslint/eslint#semantic-versioning-policy) needs a major version to update the recommended config. [semver-minor]

### DIFF
--- a/docs/rules/no-template-key.md
+++ b/docs/rules/no-template-key.md
@@ -2,11 +2,11 @@
 
 Vue.js disallows `key` attribute on `<template>` elements.
 
-## 📖 Rule Details
+## :book: Rule Details
 
 This rule reports the `<template>` elements which have `key` attribute.
 
-👎 Examples of **incorrect** code for this rule:
+:-1: Examples of **incorrect** code for this rule:
 
 ```html
 <template>
@@ -18,7 +18,7 @@ This rule reports the `<template>` elements which have `key` attribute.
 </template>
 ```
 
-👍 Examples of **correct** code for this rule:
+:+1: Examples of **correct** code for this rule:
 
 ```html
 <template>
@@ -29,6 +29,6 @@ This rule reports the `<template>` elements which have `key` attribute.
 </template>
 ```
 
-## 🔧 Options
+## :wrench: Options
 
 Nothing.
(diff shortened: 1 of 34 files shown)


docs/rules/no-template-key.md
```diff
@@ -0,0 +1,34 @@
+# Disallow 'key' attribute on '<template>' (no-template-key)
+
+Vue.js disallows `key` attribute on `<template>` elements.
+
+## 📖 Rule Details
+
+This rule reports the `<template>` elements which have `key` attribute.
+
+👎 Examples of **incorrect** code for this rule:
```
kazupon: I think GitHub emoji code (e.g. `:tada:`) is maintainaceble(read/write) than emoji binary code. 👀

michalsnik: Agree with @kazupon, I did so while importing `elint-plugin-vue-trial` too

mysticatea: In my custom, I don't use the notation since it doesn't work outside of GitHub (i.g. jsdoc, editor previews, etc...). Should I use it?

mysticatea:
I solved merge conflicts.

armano2:
thats why: https://github.com/vuejs/eslint-plugin-vue/issues/43
`<template> cannot be keyed. Place the key on real elements instead.`

PR #54, opened 2017-06-26
Disable "html-no-self-closing" rule in recommended configuration (Resolves #48)

### LINKED ISSUE #48: html-no-self-closing shouldn't be part of the recommended config

`html-no-self-closing` should not have been included. I don't see why it should be the default. It's a preference. People used to XHTML won't like that.

**What actually happened?**

`html-no-self-closing` was enabled when using `plugin:vue/recommended`.

michalsnik:
Resolves #48


JamesCrann:
Will this disable errors for all element types? I would have thought it should still validate non void elements?

michalsnik:
Released in `v3.3.0`

PR #60, opened 2017-06-29
Add rule `max-attributes-per-line` (resolves #47)

### LINKED ISSUE #47: Rule proposition: Maximum number of props per line
We often struggle to read templates with a different number of props per line.

This rule would limit the maximum number of props on a single line can to improve readability.
There should also be the possibility to disable this rule when a tag has only one prop.

**Proposal**
`'vue/max-props-per-line': {max: Number, when: 'always|multiline'}`

The default could be: `'vue/max-props-per-line': {max: 1, when: 'always'}`

1. Enforcing one prop per line always `'vue/max-props-per-line': {max: 1, when: 'always'}`
```
<component 
  prop="something"
  foo="bar"
  />

<component
  prop="something"
  />
```

2. Enforcing one prop per line only when it has multilines it would accept: `'vue/max-props-per-line': {max: 1, when: 'multiline'}`

```
<component prop="something" foo="bar" />

<component
  prop="something"
  foo="bar" 
  />
```

filipalacerda:
WIP for https://github.com/vuejs/eslint-plugin-vue/issues/47

Adds rule to verify the number of props per line.


filipalacerda:
@mysticatea @michalsnik 
I opened this PR as a WIP to check the rule definition with you before starting implementing, hope that's ok.

Can you please check https://github.com/vuejs/eslint-plugin-vue/pull/60/files#diff-c73e4e318ae2ee54c86d3c86d4f0434f and see if the rule definition matches what you were expecting or if I need to change it?

Thanks!

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,88 @@
+# Define the number of attributes allows per line (max-attributes-per-line)
+
+Limits the maximum number of props per line to improve readability
```
michalsnik: I'd say `maximum number of attributes/properties` and add `.` at the end of the sentence.

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+## Rule Details
+
+This rule aims to enforce a number of props per line in templates. It checks all the elements in a template and verifies that the number of props per line does not exceed the defined maximum.
```
michalsnik: Again `attributes`, unless we only care about components' props. But I think we should care about every attribute. Prop is just another type of property/attribute.

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,88 @@
...
+A prop is considered to be in a new line when there is a line break between two props.
```
michalsnik: Maybe something like: `There is predefined number of props that are acceptable in one-line case, as well as how many attributes are acceptable per line in multi-line case.`?

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,88 @@
...
+```
+{
+  "vue/max-attributes-per-line": ["error", {
+    "singleline": 2,
+    "multiline": 1
+  }]
+}
+```
+### `singleline`
+Number of maximum number of props when a tag is in a single line.
```
michalsnik: When the opening tag is in a single line to be precise.

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,88 @@
...
+This rule aims to enforce a number of props per line in templates. It checks all the elements in a template and verifies that the number of props per line does not exceed the defined maximum.
+A prop is considered to be in a new line when there is a line break between two props.
+
+The default is one prop per line
+
+Examples of **incorrect** code for this rule:
+
+```js
+
+<component foo="1" bar="3"></component>
```
michalsnik: I think default should be:
- singleline: 3
- multiline: 1

Then examples could look like this:
*Incorrect*:
```
<component lorem="1" ipsum="2" dolor="3" sit="4">
</component>

<component
  lorem="1" ipsum="2"
  dolor="3"
  sit="4"
>
</component>
```

*Correct*:
```
<component lorem="1" ipsum="2" dolor="3">
</component>

<component
  lorem="1"
  ipsum="2"
  dolor="3"
  sit="4"
>
</component>
```

We should probably add additional option for specifying if we accept attribute in first line or not too, maybe something like `firstline: 1`, that would make it possible to have one property in the first line, or more if specified. WDYT?

filipalacerda:
@michalsnik thank you so much for your feedback 😊 going to update it asap! :)

docs/rules/max-attributes-per-line.md
filipalacerda: I am a little lost here, should I add it or replace something with it?

docs/rules/max-attributes-per-line.md
filipalacerda: Makes sense to me!

docs/rules/max-attributes-per-line.md
michalsnik: Instead of text on line 9.

filipalacerda: 👍

docs/rules/max-attributes-per-line.md
```diff
@@ -1,33 +1,42 @@
...
 
 ## Rule Details
 
-This rule aims to enforce a number of props per line in templates. It checks all the elements in a template and verifies that the number of props per line does not exceed the defined maximum.
-A prop is considered to be in a new line when there is a line break between two props.
+This rule aims to enforce a number of attributes per line in templates. It checks all the elements in a template and verifies that the number of attributes per line does not exceed the defined maximum.
+An attribute is considered to be in a new line when there is a line break between two attributes.
 
-The default is one prop per line
+There is predefined number of attributes that are acceptable in one-line case, as well as how many attributes are acceptable per line in multi-line case.
```
michalsnik: Let's update a little bit more:
```
There is a configurable number of attributes that are acceptable in one-line case (default 3), as well as how many attributes are acceptable per line in multi-line case (default 1).
```
And then we can remove the next sentence.

docs/rules/max-attributes-per-line.md
```diff
@@ -38,42 +47,68 @@ Examples of **correct** code for this rule:
...
     "singleline": 2,
     "multiline": 1
   }]
 }
 ```
+
+### `firstline`
+For muliline declarations, defines if it accepts an attribute in the first line.
+
+The following patterns is considered a warning:
```
michalsnik: Let's be consistent :)
```
Example of **incorrect** code for this setting:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -38,42 +47,68 @@ Examples of **correct** code for this rule:
...
+```js
+// ["error", { "firstline": 1, "singleline": 2, "multiline": 1}]
+<component
+  foo="John" bar="Smith"
+  foobar={5555555}>
+</component>;
+```
+
+The following pattern is not considered a warning:
```
michalsnik:
```
Example of **correct** code for this setting:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -38,42 +47,68 @@ Examples of **correct** code for this rule:
...
+  >
+</component>;
+```
+
+
 ### `singleline`
-Number of maximum number of props when a tag is in a single line.
+Number of maximum attributes per line when the opening tag is in a single line.
 
 The following patterns is considered a warning:
```
michalsnik: As above: 
```
Example of **incorrect** code for this setting:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -38,42 +47,68 @@ Examples of **correct** code for this rule:
...
   "vue/max-attributes-per-line": ["error", {
+    "firstline": 1,
     "singleline": 2,
     "multiline": 1
   }]
 }
 ```
+
+### `firstline`
+For muliline declarations, defines if it accepts an attribute in the first line.
```
michalsnik:
```
For multi-line declarations, defines how many attributes are acceptable to be put in the first line. (Default 0)
```
We can discuss default settings later :)

docs/rules/max-attributes-per-line.md
filipalacerda: Fixed ✅

docs/rules/max-attributes-per-line.md
filipalacerda: Fixed ✅

docs/rules/max-attributes-per-line.md
filipalacerda: Fixed them all ✅

docs/rules/max-attributes-per-line.md
filipalacerda: Fixed ✅

docs/rules/max-attributes-per-line.md
filipalacerda: ✅

filipalacerda:
@michalsnik can you please review this? Thank you!

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,111 @@
...
+const utils = require('../utils')
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Define the number of attributes allows per line',
+      category: 'Stylistic Issues',
+      recommended: false
+    },
+    fixable: 'code',
```
mysticatea: It should be `fixable: null`.
`fixable: 'code'` means that this rule modifies source code with `eslint --fix`.

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+      }
+    }]
+  },
+
+  create: function (context) {
+    const configuration = context.options[0]
+    const multilineMaximum = configuration.multiline
+    const singlelinemMaximum = configuration.singleline
+    const canHaveFirstLine = configuration.firstline !== 0
```
mysticatea: I think `boolean` or `string` option is more proper in this case.
How about the below?

**A:** Simple boolean option.

```js
{
    "singleline": 3,
    "multiline": 1,
    "allowFirstLine": false // or true
}
```

**B:** I think that the first attribute position is a taste for each people. It might be good if people can specify the position. But I'm not sure it's a responsibility of this rule.

```js
{
    "singleline": 3,
    "multiline": 1,
    "firstAttributeOfMultiline": "newline" // or "beside"
}
```

**C:** I'm wondering that it's a setting of multiline. Should it locate in the multiline?

```js
{
    "singleline": 3,
    "multiline": {"max": 1, "allowFirstLine": false}
}
```

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,111 @@
...
+          minimum: 1
+        }
+      }
+    }]
+  },
+
+  create: function (context) {
+    const configuration = context.options[0]
+    const multilineMaximum = configuration.multiline
+    const singlelinemMaximum = configuration.singleline
```
mysticatea: Documents say the options have each default value. But this looks to not have default values.

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,111 @@
...
+      'VStartTag' (node) {
+        const numberOfAttributes = node.attributes.length
+
+        if (!numberOfAttributes) return
+
+        if (isSingleLine(node) && numberOfAttributes > singlelinemMaximum) {
+          context.report({
+            node,
+            loc: node.loc,
+            message: 'It has more than {{singlelinemMaximum}} attributes per line.',
```
mysticatea: I think better if this rule reports the violated attributes instead of elements which have the violated attributes.
Because editors can indicate more proper locations with red lines.

tests/lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,91 @@
...
+      options: [{ singleline: 3, multiline: 1, firstline: 0 }]
+    },
+    {
+      code: `<template><component
+        name="John Doe" age="30"
+        job="Vet">
+        </component>
+      </template>`,
+      options: [{ singleline: 3, multiline: 2, firstline: 0 }]
+    }
```
mysticatea: I want to see tests with no options (use default values) as well.

filipalacerda:
Thanks @mysticatea! Will update asap :)

lib/rules/max-attributes-per-line.js
filipalacerda: :woman_facepalming: forgot about that! Fixed!

lib/rules/max-attributes-per-line.js
filipalacerda: It is only for multiline, I will go with C

lib/rules/max-attributes-per-line.js
filipalacerda: @mysticatea 
I am looking at the docs and I am a little lost here. I thought they were assumed by default, meaning, if no options were provided they would fallback to the ones in the schema.

Should I do something like:
`const configuration = context.options[0] || { firstline: 0, singleline: 1, multiline: 1}`
?

lib/rules/max-attributes-per-line.js
filipalacerda: Agreed! :)

filipalacerda: Question:

with the default of 3, if it has more than 3, we would trigger an error saying:"attribute(s) x and z should be in a new line."
But if the multiline max is 1 putting those last ones in a new line will trigger another error.

tests/lib/rules/max-attributes-per-line.js
filipalacerda: Fixed ✅

filipalacerda:
@mysticatea updated some of the things you mentioned but I need your help regarding the single line error: https://github.com/vuejs/eslint-plugin-vue/pull/60#discussion_r127166697

And I need you to confirm the defaults part, not sure if it is the right way of doing it 😅 

Thanks for your help!

lib/rules/max-attributes-per-line.js
mysticatea: I have imaged like [that logic](https://gist.github.com/mysticatea/7d7fc98bcb9ec23091074edd7c17fa7b). (sorry, I have not run that code.)

lib/rules/max-attributes-per-line.js
michalsnik: I'd go with `C` but still keep it as number, e.g.:
```
{
    "singleline": 3,
    "multiline": {"max": 1, "maxInFirstLine": 2}
}
```
So it's possible to have two properties in the first line but the rest in new lines in capacity of 1 per line, like here:
```
<input name="someField" class="input"
  v-model="message"
  placeholder="Lorem ipsum"
>
```
What do you think @filipalacerda @mysticatea ?

michalsnik: Ah, I think I get it know @mysticatea. You meant that the value for a singleline option should be considered the max for the first line, when `allowFirstLine` is true? If that's the case I'm all for it.

lib/rules/max-attributes-per-line.js
michalsnik: Or we can simply put:
```There are 3 attributes in this line, but the maximum is 2```

File: lib/rules/max-attributes-per-line.js (hunk 13 above)
Comment (michalsnik) [inline lib/rules/max-attributes-per-line.js]: I think it should be something like:
```
const configuration = context.options[0] || {}
const multilineMaximum = configuration.multiline || 1
const singlelinemMaximum = configuration.singleline || 1
```
so you can pass only `multiline` option and get fallback for `singleline`

File: lib/rules/max-attributes-per-line.js (hunk 14 above)
Comment (filipalacerda) [inline lib/rules/max-attributes-per-line.js]: I would rather go with @michalsnik suggestion, since as a user I think it will help me more to know the exact numbers :)

**Comment by filipalacerda** [comment]
@mysticatea @michalsnik all points were addressed :)

Can you please check again?

File: tests/lib/rules/max-attributes-per-line.js [hunk 16]
```diff
@@ -71,7 +71,7 @@ ruleTester.run('max-attributes-per-line', rule, {
   invalid: [
     {
       code: `<template><component name="John Doe" age="30" job="Vet" petname="Snoopy"></component></template>`,
-      errors: ['It has more than 3 attributes per line.']
+      errors: ['There are 4 in this line, but the maximum is 3.']
```
michalsnik: Please add `attributes` keyword to the message :) So it says:
`There are 4 attributes in this line, but the maximum is 3.`

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,116 @@
...
+              type: 'boolean'
+            }
+          }
+        }
+      }
+    }]
+  },
+
+  create: function (context) {
+    const configuration = context.options[0] || { singleline: 3, multiline: { max: 1, allowFirstLine: false }}
```
michalsnik: Please update settings so it's possible to override only part of it without the need to update everything. You can use `Object.assign` to merge passed settings with default ones:
```
const configuration = Object.assign({}, context.options[0] || {}, {
  singleline: 3,
  multiline: { max: 1, allowFirstLine: false }
})
```

And I'd also consider accepting setting: `multiline: 4` as just a number, then we would assume `allowFirstLine` is false. What do you think?

tests/lib/rules/max-attributes-per-line.js
filipalacerda: :woman_facepalming:

lib/rules/max-attributes-per-line.js
filipalacerda: I think that by having only one way of setting a rule we make user's lives easier, but I can change it if you don't agree :)

What do you think @michalsnik ?

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,116 @@
...
+    },
+    fixable: null,
+    schema: [{
+      type: 'object',
+      properties: {
+        'singleline': {
+          type: 'number',
+          minimum: 1
+        },
+        'multiline': {
```
mysticatea: I'd like to change this schema to https://jsonschemalint.com/#/version/draft-06/markup/json?gist=a6a3592eaf8365d069031288f29e5445

- That schema allows `{singleline: 1, multiline: 3}` if somebody don't mind `{allowFirstLine: false}`. It's simple.
- On the other hand, that schema allows `{singleline: {max: 1}, multiline: {max: 3}}` if somebody want to write as details. It's more consistent than `{singleline: 1, multiline: {max: 3}}`.

lib/rules/max-attributes-per-line.js
mysticatea: > `const configuration = context.options[0] || { singleline: 3, multiline: { max: 1, allowFirstLine: false }}`

In that case, if somebody write `["error", {"singleline": 1}]` then it's going to throw a TypeError at `const multilineMaximum = configuration.multiline.max`. I'd like to define `parseOptions` function:

```js
function parseOptions(options) {
    const result = { singlelineMax: 3, multilineMax: 1, allowFirstLine: false }

    if (options) {
        if (typeof options.singleline === "number") {
            result.singlelineMax = options.singleline
        } else if (options.singleline && options.singleline.max) {
            result.singlelineMax = options.singleline.max
        }

        if (typeof options.multiline === "number") {
            result.multilineMax = options.multiline
        } else if (options.multiline) {
            if (options.multiline.max) {
                result.multilineMax = options.multiline.max
            }
            if (options.multiline.allowFirstLine) {
                result.allowFirstLine = options.multiline.allowFirstLine
            }
        }
    }

    return result
}
```

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,116 @@
...
+    utils.registerTemplateBodyVisitor(context, {
+      'VStartTag' (node) {
+        const numberOfAttributes = node.attributes.length
+
+        if (!numberOfAttributes) return
+
+        if (isSingleLine(node) && numberOfAttributes > singlelinemMaximum) {
+          context.report({
+            node,
+            loc: node.loc,
```
mysticatea: I still think better if this rule reports the violated attributes instead of elements which have the violated attributes.
This is consistent with the core rules which have similar functionally.

```js
/*eslint 
    max-statements-per-line: error,
    newline-per-chained-call: error,
    object-property-newline: error,
    array-element-newline: error
 */

aaa(); bbb(); ccc();
test.aaa().bbb().ccc().ddd()
var obj = {aaa:1, bbb:2, ccc:3}
var ary = [1, 2, 3]
```

image

lib/rules/max-attributes-per-line.js
mysticatea: @michalsnik This is suggestion which is based on this line of the implementation: `const canHaveFirstLine = configuration.firstline !== 0`. It uses `multiline.max` for the first line if `allowFirstLine: true`.

filipalacerda:
@michalsnik @mysticatea I am little lost 😅 , can you confirm this is what you are suggesting:

1) Update config to accept:
```
{
  singleline: 1,
  multiline: { max: 3, allowFirstLine: false}
}
```
and 
```
{
  singleline: 1,
  multiline: 3
}
```

In the latter, assumes the default `false` for `allowFirstLine`

2) Show which attributes are wrong in the single line case

3) Update defaults to allow for only one of them being provided

Is this it?
Thanks!

mysticatea:
@filipalacerda 

1. Yes. I hope that the codes in my comments help you.
2. I think that this rule should report attributes instead of elements at all. [Like this](https://gist.github.com/mysticatea/7d7fc98bcb9ec23091074edd7c17fa7b).
3. Yes.

filipalacerda:
@mysticatea @michalsnik 

I need help :) how do I set the schema to accept 2 different configs for `multiline`? 🤔 

```
schema: [
  'multiline': {
    type: // can be an object or a number
   }
]
```

Should I not set a schema? I do not see it in most of the other rules 🤔

filipalacerda:
found it! ignore me! 🙈

filipalacerda:
@mysticatea @michalsnik all fixed :) Can you please check again? Thanks!

armano2:
@filipalacerda i like this but i see one issue here you are allways reporting VStartTag instead of VAttribute

```
    message: 'Attribute age should be on a new line.',
    line: 1,
    column: 11,
    nodeType: 'VStartTag',
```

i will love to have accurate reports for editors to match this:
```diff
diff --git a/tests/lib/rules/max-attributes-per-line.js b/tests/lib/rules/max-attributes-per-line.js
index ab98db5..f3a65b9 100644
--- a/tests/lib/rules/max-attributes-per-line.js
+++ b/tests/lib/rules/max-attributes-per-line.js
@@ -83,7 +83,11 @@ ruleTester.run('max-attributes-per-line', rule, {
   invalid: [
     {
       code: `<template><component name="John Doe" age="30" job="Vet" petname="Snoopy"></component></template>`,
-      errors: ['Attribute petname should be on a new line.']
+      errors: [{
+        message: 'Attribute petname should be on a new line.',
+        type: 'VAttribute',
+        line: 1
+      }]
     },
     {
       code: `<template><component job="Vet"
@@ -91,12 +95,24 @@ ruleTester.run('max-attributes-per-line', rule, {
         age="30">
         </component>
       </template>`,
-      errors: ['Attribute job should be on a new line.']
+      errors: [{
+        message: 'Attribute job should be on a new line.',
+        type: 'VAttribute',
+        line: 1
+      }]
     },
     {
       code: `<template><component name="John Doe" age="30" job="Vet"></component></template>`,
       options: [{ singleline: 1, multiline: { max: 1, allowFirstLine: false }}],
-      errors: ['Attribute age should be on a new line.', 'Attribute job should be on a new line.']
+      errors: [{
+        message: 'Attribute age should be on a new line.',
+        type: 'VAttribute',
+        line: 1
+      }, {
+        message: 'Attribute job should be on a new line.',
+        type: 'VAttribute',
+        line: 1
+      }]
     },
     {
       code: `<template><component name="John Doe"
@@ -104,7 +120,11 @@ ruleTester.run('max-attributes-per-line', rule, {
         </component>
       </template>`,
       options: [{ singleline: 3, multiline: { max: 1, allowFirstLine: false }}],
-      errors: ['Attribute name should be on a new line.']
+      errors: [{
+        message: 'Attribute name should be on a new line.',
+        type: 'VAttribute',
+        line: 1
+      }]
     },
     {
       code: `<template><component
@@ -113,7 +133,11 @@ ruleTester.run('max-attributes-per-line', rule, {
         </component>
       </template>`,
       options: [{ singleline: 3, multiline: { max: 1, allowFirstLine: false }}],
-      errors: ['Attribute age should be on a new line.']
+      errors: [{
+        message: 'Attribute age should be on a new line.',
+        type: 'VAttribute',
+        line: 2
+      }]
     },
     {
       code: `<template><component
@@ -122,7 +146,11 @@ ruleTester.run('max-attributes-per-line', rule, {
         </component>
       </template>`,
       options: [{ singleline: 3, multiline: { max: 2, allowFirstLine: false }}],
-      errors: ['Attribute petname should be on a new line.']
+      errors: [{
+        message: 'Attribute petname should be on a new line.',
+        type: 'VAttribute',
+        line: 3
+      }]
     },
     {
       code: `<template><component
@@ -131,10 +159,15 @@ ruleTester.run('max-attributes-per-line', rule, {
         </component>
       </template>`,
       options: [{ singleline: 3, multiline: { max: 2, allowFirstLine: false }}],
-      errors: [
-        'Attribute petname should be on a new line.',
-        'Attribute extra should be on a new line.'
-      ]
+      errors: [{
+        message: 'Attribute petname should be on a new line.',
+        type: 'VAttribute',
+        line: 3
+      }, {
+        message: 'Attribute extra should be on a new line.',
+        type: 'VAttribute',
+        line: 3
+      }]
     }
   ]
 })
```

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        if (!isSingleLine(node)) {
+          if (!canHaveFirstLine && node.attributes[0].loc.start.line === node.loc.start.line) {
+            showErrors([node.attributes[0]], node)
+          }
+
+          const propsPerLine = [[node.attributes[0]]]
+
+          node.attributes.reduce(function (previous, current) {
```
michalsnik: - Use arrow function
- I'd create `groupAttrsByLine` or `collectAttributes` function that would take `node.attributes` as parameter and return 2d array, using such function in this place would make this code more readable:
```
const attrsGroupedByLine = groupAttrsByLine(node.attributes);

attrsGroupedByLine
  .filter(attrs => attrs.length > multilineMaximum)
  .forEach(attrs => showErrors(attrs.splice(multilineMaximum), node)
```

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,163 @@
...
+            }
+          })
+        }
+      }
+    })
+
+    // ----------------------------------------------------------------------
+    // Helpers
+    // ----------------------------------------------------------------------
+    function isSingleLine (node) {
```
michalsnik: Consider moving this function to utils

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,163 @@
...
+
+      return defaults
+    }
+
+    function showErrors (attributes, node) {
+      attributes.forEach((prop) => {
+        context.report({
+          node,
+          loc: node.loc,
+          message: 'Attribute {{propName}} should be on a new line.',
```
michalsnik: Add `""` around propName, it will make there errors even more readable :)

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,163 @@
...
+          node.attributes.reduce(function (previous, current) {
+            if (previous.loc.end.line === current.loc.start.line) {
+              propsPerLine[propsPerLine.length - 1].push(current)
+            } else {
+              propsPerLine.push([current])
+            }
+            return current
+          })
+
+          propsPerLine.forEach(function (attributes) {
```
michalsnik: You can use arrow function, you don't need the context here

lib/utils/index.js
```diff
@@ -241,5 +241,14 @@ module.exports = {
     assert(typeof name === 'string')
 
     return VOID_ELEMENT_NAMES.has(name.toLowerCase())
+  },
+
+    /**
+     * Check whether the component is declared in a single line or not.
+     * @param {ASTNode} node
+     * @returns {boolean
```
armano2: you are missing here `}`

lib/utils/index.js
```diff
@@ -398,5 +398,13 @@ module.exports = {
         cb(node.arguments.slice(-1)[0])
       }
     }
+  },
+      /**
+     * Check whether the component is declared in a single line or not.
+     * @param {ASTNode} node
+     * @returns {boolean
```
armano2: you are missing here `}`

filipalacerda:
@michalsnik all done :) Can you please check again?

@armano2 I opted by this report since it matches the error style thrown in the other rules. (At least at the time I opened this MR) I would rather keep it, if you do not mind :)

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+  lorem="1" ipsum="2"
+  dolor="3"
+  sit="4"
+>
+</component>
+```
+
+Examples of **correct** code for this rule:
+
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+
+This rule aims to enforce a number of attributes per line in templates.
+It checks all the elements in a template and verifies that the number of attributes per line does not exceed the defined maximum.
+An attribute is considered to be in a new line when there is a line break between two attributes.
+
+There is a configurable number of attributes that are acceptable in one-line case (default 3), as well as how many attributes are acceptable per line in multi-line case (default 1).
+
+Examples of **incorrect** code for this rule:
+
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
+# Define the number of attributes allows per line (max-attributes-per-line)
+
+Limits the maximum number of attributes/properties per line to improve readability.
+
+
+## Rule Details
```
armano2:
```
## :book: Rule Details
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+
+This rule aims to enforce a number of attributes per line in templates.
+It checks all the elements in a template and verifies that the number of attributes per line does not exceed the defined maximum.
+An attribute is considered to be in a new line when there is a line break between two attributes.
+
+There is a configurable number of attributes that are acceptable in one-line case (default 3), as well as how many attributes are acceptable per line in multi-line case (default 1).
+
+Examples of **incorrect** code for this rule:
```
armano2:
```
:-1: Examples of **incorrect** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+
+<component
+  lorem="1" ipsum="2"
+  dolor="3"
+  sit="4"
+>
+</component>
+```
+
+Examples of **correct** code for this rule:
```
armano2:
```
:+1: Examples of **correct** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }]
+}
+```
+
+### `allowFirstLine`
+For multi-line declarations, defines if allows attributes to be put in the first line. (Default false)
+
+Example of **incorrect** code for this setting:
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+// [{ "multiline": { "allowFirstLine": false }}]
+<component foo="John" bar="Smith"
+  foobar={5555555}>
+</component>;
+```
+
+Example of **correct** code for this setting:
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+  >
+</component>;
+```
+
+
+### `singleline`
+Number of maximum attributes per line when the opening tag is in a single line. (Default is 3)
+
+Example of **incorrect** code for this setting:
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+// [{"singleline": 2,}]
+<component foo="John" bar="Smith" foobar={5555555}></component>;
+```
+
+Example of **correct** code for this setting:
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+// [{"singleline": 3,}]
+<component foo="John" bar="Smith" foobar={5555555}></component>;
+```
+
+
+### `multiline`
+Number of maximum attributes per line when a tag is in multiple lines. (Default is 1)
+
+Example of **incorrect** code for this setting:
+```js
```
armano2:
```
```html
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+// [{"multiline": 1}]
+<component foo="John" bar="Smith"
+  foobar={5555555}>
+</component>;
+```
+
+Example of **correct** code for this setting:
+```js
```
armano2:
```
```html
```
or vue

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+### `singleline`
+Number of maximum attributes per line when the opening tag is in a single line. (Default is 3)
+
+Example of **incorrect** code for this setting:
+```js
+// [{"singleline": 2,}]
+<component foo="John" bar="Smith" foobar={5555555}></component>;
+```
+
+Example of **correct** code for this setting:
```
armano2:
```
:+1: Examples of **correct** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+      allowFirstLine: false
+    }
+  }]
+}
+```
+
+### `allowFirstLine`
+For multi-line declarations, defines if allows attributes to be put in the first line. (Default false)
+
+Example of **incorrect** code for this setting:
```
armano2:
```
:-1: Examples of **incorrect** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+```js
+// [{"singleline": 3,}]
+<component foo="John" bar="Smith" foobar={5555555}></component>;
+```
+
+
+### `multiline`
+Number of maximum attributes per line when a tag is in multiple lines. (Default is 1)
+
+Example of **incorrect** code for this setting:
```
armano2:
```
:-1: Examples of **incorrect** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+```js
+// [{"multiline": 1}]
+<component foo="John" bar="Smith"
+  foobar={5555555}>
+</component>;
+```
+
+Example of **correct** code for this setting:
```
armano2:
```
:+1: Examples of **correct** code for this rule:
```

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+    "singleline": 3,
+    "multiline": {
+      max: 1,
+      allowFirstLine: false
+    }
+  }]
+}
+```
+
+### `allowFirstLine`
```
armano2: it should be 4th lvl header because its part of `Options`

docs/rules/max-attributes-per-line.md
```diff
@@ -0,0 +1,122 @@
...
+  lorem="1"
+  ipsum="2"
+  dolor="3"
+  sit="4"
+>
+</component>
+
+```
+
+### Options
```
armano2:
```
## :wrench: Options
```

armano2:
@filipalacerda i added few comments about documentation

filipalacerda:
@michalsnik @mysticatea @armano2 

I have made all the changes you requested, do you think this is ready to go?

lib/rules/max-attributes-per-line.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      return defaults
+    }
+
+    function showErrors (attributes, node) {
+      attributes.forEach((prop) => {
+        context.report({
+          node,
+          loc: node.loc,
```
mysticatea: I still think better if this rule reports the violated attributes instead of elements which have the violated attributes.

armano2: @filipalacerda can you look on it?

filipalacerda: I will try, but probably only next weekend! If any of you wants/is able to take over please feel free to do so, I can give you access to the fork 😊

mysticatea:
@michalsnik Could you re-review and merge?

PR #63, opened 2017-07-01
Fix formatting README.md & no-template-key.md

armano2:
github com-vuejs-eslint-plugin-vue


michalsnik:
Thanks for submitting this PR @armano2, but we use https://github.com/vuejs/eslint-plugin-vue/blob/master/tools/update-rules.js to generate table in readme. And this script uses description directly from rules' files, so you should change description messages in each rule and then execute: `npm run update` which will automatically update readme.

armano2:
i fixed docs/rules/no-template-key.md :dancer:

michalsnik:
Yeah, but you changed Readme too, and it will be overwritten in the next release so we should update rules descriptions instead, so those changes will be permanent.

armano2:
ok i will change this

Issue #65, opened 2017-07-04
vue/no-invalid-v-for shouldn't catch slots


Akryum:
**Tell us about your environment**

* **ESLint Version:** 4.1.1
* **eslint-plugin-vue Version:** 3.5.0
* **Node Version:** 8.1.3

<details>
<summary>Please show your full configuration:</summary>

```javascript
module.exports = {
  root: true,
  extends: [
    'eslint:recommended',
    'plugin:vue/recommended',
  ],
  'parserOptions': {
    'parser': 'babel-eslint',
    'sourceType': 'module',
  },
  env: {
    es6: true,
    node: true,
    browser: true,
  },
  // add your custom rules here
  'rules': {
    'accessor-pairs': 'error',
    'array-callback-return': 'warn',
    'arrow-spacing': ['error', { 'before': true, 'after': true }],
    'block-spacing': ['error', 'always'],
    'brace-style': ['error', '1tbs', { 'allowSingleLine': false }],
    'camelcase': ['error', { 'properties': 'never' }],
    'comma-dangle': ['error', 'always-multiline'],
    'comma-spacing': ['error', { 'before': false, 'after': true }],
    'comma-style': ['error', 'last'],
    'constructor-super': 'error',
    'curly': ['error', 'multi-line'],
    'dot-location': ['error', 'property'],
    'eol-last': 'error',
    'eqeqeq': ['error', 'always', { 'null': 'ignore' }],
    'func-call-spacing': ['error', 'never'],
    'handle-callback-err': ['error', '^(err|error)$' ],
    'indent': ['error', 2, { 'SwitchCase': 1 }],
    'key-spacing': ['error', { 'beforeColon': false, 'afterColon': true }],
    'keyword-spacing': ['error', { 'before': true, 'after': true }],
    'new-cap': ['error', { 'newIsCap': true, 'capIsNew': false }],
    'new-parens': 'error',
    'no-array-constructor': 'error',
    'no-caller': 'error',
    'no-class-assign': 'error',
    'no-compare-neg-zero': 'error',
    'no-cond-assign': 'error',
    'no-console': 'off',
    'no-const-assign': 'error',
    'no-constant-condition': ['error', { 'checkLoops': false }],
    'no-control-regex': 'error',
    'no-delete-var': 'error',
    'no-dupe-args': 'error',
    'no-dupe-class-members': 'error',
    'no-dupe-keys': 'error',
    'no-duplicate-case': 'error',
[...]

PR #68, opened 2017-07-04
#65 - vue/no-invalid-v-for shouldn't catch slots

armano2:
#65


armano2:
I think we should add few additional internal vue elements like `<router-view>` `<transition>` `<transition-group>`

mysticatea:
Thank you for contributing!

Oh, [the `lib/html-elements.json`](https://github.com/vuejs/eslint-plugin-vue/blob/9fce35dba3883cf8c564eb18693c3882e71d21c9/lib/utils/html-elements.json) seems wrong. [The `<slot>` elements are valid HTML elements](https://html.spec.whatwg.org/#the-slot-element). Could you replace `content`, `shadow`, and `element` (those were dropped from standard) by `slot` in the `lib/html-elements.json`?

> I think we should add few additional internal vue elements like <router-view> <transition> <transition-group> but they should be not allowed?

Those are embedded custom components. I believe that embedded custom components also need `key` attributes in iterations.

image

armano2:
I found out that there is few more missing html5 valid tags:
`slot`, `blockquote`, `iframe`, `noframes`, `picture`, `tfoot`

armano2:
base on https://github.com/vuejs/vue/blob/dev/src/platforms/web/util/element.js `content`, `shadow`, and `element` can be still in use

Issue #58, opened 2017-06-28
Rule proposition: `no-side-effects-in-computed-properties`


michalsnik:
This rule would check if there are no side effects inside computed properties.

So this example would throw an error:
```
export default {
  computed: {
    hello() {
      this.lastName = 'Dolor' // <- error "Unexpected side effect in computed property 'hello'"
      return `Hello ${this.firstName} ${this.lastName}`
    }
  },
  data() {
    return {
      firstName: 'Lorem',
      lastName: 'Ipsum'
    }
  }
}
```

What do you think @mysticatea @chrisvfritz ?

chrisvfritz:
I like it!

LinusBorg:
Would this rule also check for things like:
```javascript
export default {
  computed: {
    hello() {
      this.sideEffect() // <- error "Unexpected side effect in computed property 'hello'"
      return `Hello ${this.firstName} ${this.lastName}`
    }
  },
methods: {
  sideEffect() {
    // some side effect.
  }
}

```
any method call whose return value is not used in the CP could be considered a side-effect.

michalsnik:
We could @LinusBorg, so the assumptions would be to throw an error when:
- `this.xxx  <=|+=|++|-=|-->`
- `this.xxx.yyy... <=|+=|++|-=|-->`
- `this.xxx.<push|pop|shift|unshift|reverse|splice|sort|copyWithin|fill>`
- any of the above is present in method that is being called inside computed property

Am I missing something?

LinusBorg:
> But we still need to remember that it should be possible to use map or other non-mutable methods on properties.

When using map etc., we would use the return value, right?
```javascript
this.something.map() // doesn't make sense
const result = this.something.map() // makes sense
```

The same is true for  method calls - if we don't do something with the return value, it's safe to assume that the method we call has a side-effect.

michalsnik:
> When using map etc., we would use the return value, right?
I would check all `CallExpressions`, in terms of non-mutable methods like `map` I don't have to care about assignments, as it would not proudce any side effect in any case. And in terms of mutable methods like `splice` we also don't need to check for any assignment because it will mutate the initial value anyway, so presence of any `this.something.<any mutable method>` should trigger error.

> The same is true for method calls - if we don't do something with the return value, it's safe to assume that the method we call has a side-effect.
[...]

PR #69, opened 2017-07-04
Add `no-side-effects-in-computed-properties` rule

michalsnik:
This PR implements rule proposed in #58 

This rule checks for existance of any of the following inside all computed properties:
- `AssignmentExpression` that starts with `this.` (e.g. `this.x = 'asd'`, `this.x += 1`)
- `UpdateExpression` that starts with `this.` (e.g `this.x++`)
- `CallExpression` that starts with `this.` and calls one of the mutable methods (`push`, `pop`, `shift`, `unshift`, `reverse`, `splice`, `sort`, `copyWithin`, `fill`) not followed by any of those that remove the reference (like: `slice`, `map`, `filter`, `concat`)

Feel free to comment if I missed anything. At this point I'd like to stick with those simple checks and then we can investigate it further if we can check anything more than that. If believe this is a good starting point.

Left to do:
- Add one more case - simple CallExpression `this.anything(` that is not assigned anywhere should be considered unexpected side effect
- Add more test scenarios
- Clean current solution (I'm considering using regex for checking `AssignmentExpression` and `UpdateExpression` too, instead of parsing those simple chunks of AST)


lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,127 @@
...
+      p.key.type === 'Identifier' &&
+      p.key.name === 'computed' &&
+      p.value.type === 'ObjectExpression'
+    )[0]
+
+  if (!computedPropertiesNode) { return [] }
+
+  const computedProperties = computedPropertiesNode.value.properties
+
+  return computedProperties.map(cp => {
```
armano2: in case when there is only
```js
computed: {
  foo: {
    set () {}
  }
}
```
returned value is going to be 
```js
return [
  {
    key: '....',
    value: undefined
  }
]
```

And code when its used has to check for that case or we should filter it here

mysticatea:
We should be careful about nested functions if we use `ThisExpression` to lint.
`this`s of nested functions often are bound to different instances.

lib/rules/no-side-effects-in-computed-properties.js
michalsnik: Good point, thanks @armano2

armano2: @michalsnik i think its better to check it here cus we are going propably to add rule about checking if computed propery has `get` and we can do that by just checking it `.value` is not `undefined`

michalsnik:
Agree @mysticatea, but given the way vue works and the `this` bound to a component in default scope I think other `this`s inside computed property or any other place in vue component should be considered harmful. Messing different contexts seems like a very bad idea to me and I think users should above else use parameters instead (e.g. when using forEach loop). I'd go with current implementation and see how it works in the wild. Please take a look once again and approve if you agree :)

lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,129 @@
...
+      p.value.type === 'ObjectExpression'
+    )[0]
+
+  if (!computedPropertiesNode) { return [] }
+
+  const computedProperties = computedPropertiesNode.value.properties
+
+  return computedProperties
+    .map(cp => {
+      const key = cp.key.name
```
armano2: `cp.key` can be undefined if you are using babel and [spread transform](https://babeljs.io/docs/plugins/transform-object-rest-spread/)

```
TypeError: Cannot read property 'name' of undefined
```

example:
```js
export default {
  computed: {
    ...mapGetters({})
  }
}
```

lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,129 @@
...
+      let value
+
+      if (cp.value.type === 'FunctionExpression') {
+        value = cp.value.body
+      } else if (cp.value.type === 'ObjectExpression') {
```
armano2: coverage from unit test never enters this part `ObjectExpression`

lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,129 @@
...
+
+function getComputedProperties (componentProperties) {
+  const computedPropertiesNode = componentProperties
+    .filter(p =>
+      p.key.type === 'Identifier' &&
+      p.key.name === 'computed' &&
+      p.value.type === 'ObjectExpression'
+    )[0]
+
+  if (!computedPropertiesNode) { return [] }
```
armano2: unit test never enters this `if`

lib/rules/no-side-effects-in-computed-properties.js
armano2: @michalsnik `cp.key` can be undefined if you are using [`experimentalObjectRestSpread`](https://babeljs.io/docs/plugins/transform-object-rest-spread/)
Solution: `if cp.type === 'Property'`

```
TypeError: Cannot read property 'name' of undefined
```

example:
```js
export default {
  computed: {
    ...mapGetters({})
  }
}
```

lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,129 @@
...
+    while (memberExpression.type === 'MemberExpression') {
+      if (memberExpression.property.type === 'Identifier') {
+        members.push(memberExpression.property.name)
+      }
+      memberExpression = memberExpression.object
+    }
+
+    if (memberExpression.type === 'ThisExpression') {
+      members.push('this')
+    } else if (memberExpression.type === 'Identifier') {
```
armano2: unit test never enters here

lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -0,0 +1,129 @@
...
+    {
+      // this.xxx <=|+=|-=>
+      'AssignmentExpression > MemberExpression' (node) {
+        if (parseMemberExpression(node)[0] === 'this') {
+          forbiddenNodes.push(node)
+        }
+      },
+      // this.xxx <++|-->
+      'UpdateExpression > MemberExpression' (node) {
+        if (parseMemberExpression(node)[0] === 'this') {
```
armano2: all tests are for `this.`

lib/rules/no-side-effects-in-computed-properties.js
michalsnik: Very good point, thank you @armano2 for pointing that!

michalsnik:
Thanks for suggestions @armano2! I moved two helper functions to utils and added separate tests for them. Please take another look.

michalsnik:
Okay, let's push this forward. We're still in beta phase and it's not enabled by default so I think it's good to go.

Issue #53, opened 2017-06-26
Rule proposition: No async in computed property functions


chrisvfritz:
A common issue that beginners run into is relying on async code in a computed property function, even though we warn about it in the docs. This might not be possible to identify 100% of the time, but it would be nice for the linter to warn about:

- `async`/`await`
- passing a function as a function argument (covers promises and callbacks)

Am I missing any cases where either of the above would actually be valid inside a computed property function?

michalsnik:
Sounds nice! I think this one will definitely fit in `Best practices` category.

Could you please add few examples of code that in your opinion should cause warnings @chrisvfritz ?
This way it will be easier to get anyone's head around this and later to define proper test scenarios.

chrisvfritz:
Test scenarios:

``` js
computed: {
  foo: async function () {
    return await someFunc()
  }
}
```

``` js
computed: {
  foo () {
    return new Promise((resolve, reject) => {})
  }
}
```

``` js
computed: {
  foo () {
    return fetch(url).then(response => {})
  }
}
```

``` js
computed: {
  foo () {
    return asyncFunc(function () {})
  }
}
```

mysticatea:
I have a concern about callbacks. Can it distinguish sync or async? For example, `return this.users.map(u => u.name)`. It may be `Array#map` or `Promise#map` (Bluebird).

chrisvfritz:
@mysticatea I think it'd be better to err on the side of undermatching than overmatching, but I'm thinking about a variety of strategies that could be combined for _pretty good_ detection. For example, if there's a `then` method or any argument called `response` or `res`.

michalsnik:
I'd say we should cover first 3 examples, and the last one would be to detect presence of any of `then`, `catch` or `finally` in all `CallExpression`s inside computed property's body. Bur I'm against checking arguments `response` and `res` as they might be used even if the body has nothing to do with async computing. For example application has a module that gathers users' responses for a given topic, than we could have a computed property `activeResponse` in which we'd like to call a utility function with `response` argument, or short res. I had similar case with `route` in other plugin. I thought it's enough to check for that name and then someone had an app with generating routes on google maps :D

chrisvfritz:
@michalsnik Good thinking. 🙂

armano2:
It's all or did i miss something?

```js
  invalid: [
    {
      filename: 'test.vue',
      code: `export default {
        computed: {
          foo: async function () {
            return await someFunc()
          }
        }
[...]

PR #72, opened 2017-07-08
Add `no-async-in-computed-properties` rule

armano2:
This PR implements rules proposed in #53 

**DONE:**
* [x] Create documtation
* [x] Add tests
* [x] `async` `await`
* [x] `new Promise`
* [x] `.then` `.catch` `.finally`
* [x] `Promise.all()` `Promise.race()` `Promise.reject()` `Promise.resolve()`
* [x] `setTimeout()` `setInterval()` `setImmediate()` `requestAnimationFrame()`

### DIFF
--- a/docs/rules/no-async-in-computed-properties.md
+++ b/docs/rules/no-async-in-computed-properties.md
@@ -1,0 +1,64 @@
+# Check if there are no asynchronous actions inside computed properties (no-async-in-computed-properties)
+
+Computed properties should be synchronous. Asynchronous actions inside them may not work as expected and can lead to an unexpected behaviour, that's why you should avoid them.
+If you need async computed properties you might want to consider using additional plugin [vue-async-computed]
+
+## :book: Rule Details
+
+This rule is aimed at preventing asynchronous methods from being called in computed properties.
+
+:-1: Examples of **incorrect** code for this rule:
+
+```js
+export default {
+  computed: {
+    pro () {
+      return Promise.all([new Promise((resolve, reject) => {})])
+    },
+    foo: async function () {
+      return await someFunc()
+    },
+    bar () {
+      return fetch(url).then(response => {})
+    },
+    yiel: function* () {
+      yield 1
+      yield* g1()
+    },
+    tim () {
+      setTimeout(() => { }, 0)
+    },
+    inter () {
+      setInterval(() => { }, 0)
+    },
+    anim () {
+      requestAnimationFrame(() => {})
+    }
+  }
+}
+```
+
+:+1: Examples of **correct** code for this rule:
+
+```js
+export default {
+  computed: {
+    foo () {
+      var bar = 0
+      try {
+        bar = bar / this.a
+      } catch (e) {
+        return 0
+      } finally {
+        return bar
+      }
+    }
+  }
+}
+```
+
+## :wrench: Options
+
+Nothing.
+
+[vue-async-computed]: https://github.com/foxbenjaminfox/vue-async-computed
--- a/docs/rules/no-async-in-computed-properties.md
+++ b/docs/rules/no-async-in-computed-properties.md
@@ -21,10 +21,6 @@ export default {
     bar () {
       return fetch(url).then(response => {})
     },
-    yiel: function* () {
-      yield 1
-      yield* g1()
-    },
     tim () {
       setTimeout(() => { }, 0)
     },
--- a/lib/rules/no-async-in-computed-properties.js
+++ b/lib/rules/no-async-in-computed-properties.js
@@ -61,8 +61,6 @@ function create (context) {
   const expressionTypes = {
     promise: 'asynchronous action',
     await: 'await operator',
-    yield: 'yield keyword',
-    generator: 'generator function expression',
     async: 'async function declaration',
     new: 'Promise object',
     timed: 'timed function'
@@ -73,12 +71,6 @@ function create (context) {
       forbiddenNodes.push({
         node: node,
         type: 'async'
-      })
-    }
-    if (node.generator) {
-      forbiddenNodes.push({
-        node: node,
-        type: 'generator'
       })
     }
   }
@@ -113,14 +105,6 @@ function create (context) {
             type: 'timed'
           })
         }
-      },
-
-      YieldExpression (node) {
-        // await nodes are YieldExpression's with babel-eslint < 7.0.0
-        forbiddenNodes.push({
-          node: node,
-          type: 'yield'
-        })
       },
 
       AwaitExpression (node) {
--- a/lib/rules/no-async-in-computed-properties.js
+++ b/lib/rules/no-async-in-computed-properties.js
@@ -1,0 +1,173 @@
+/**
+ * @fileoverview Check if there are no asynchronous actions inside computed properties.
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../utils')
+
+const PROMISE_FUNCTIONS = [
+  'then',
+  'catch',
+  'finally'
+]
+
+const PROMISE_METHODS = [
+  'all',
+  'race',
+  'reject',
+  'resolve'
+]
+
+const TIMED_FUNCTIONS = [
+  'setTimeout',
+  'setInterval',
+  'setImmediate',
+  'requestAnimationFrame'
+]
+
+function isTimedFunction (node) {
+  return (
+    node.type === 'CallExpression' &&
+    node.callee.type === 'Identifier' &&
+    TIMED_FUNCTIONS.indexOf(node.callee.name) !== -1
+  ) || (
+    node.type === 'CallExpression' &&
+    node.callee.type === 'MemberExpression' &&
+    node.callee.object.type === 'Identifier' &&
+    node.callee.object.name === 'window' && (
+      TIMED_FUNCTIONS.indexOf(node.callee.property.name) !== -1
+    )
+  ) && node.arguments.length
+}
+
+function isPromise (node) {
+  if (node.type === 'CallExpression' && node.callee.type === 'MemberExpression') {
+    return ( // hello.PROMISE_FUNCTION()
+      node.callee.property.type === 'Identifier' &&
+      PROMISE_FUNCTIONS.indexOf(node.callee.property.name) !== -1
+    ) || ( // Promise.PROMISE_METHOD()
+      node.callee.object.type === 'Identifier' &&
+      node.callee.object.name === 'Promise' &&
+      PROMISE_METHODS.indexOf(node.callee.property.name) !== -1
+    )
+  }
+  return false
+}
+
+function create (context) {
+  const forbiddenNodes = []
+
+  const expressionTypes = {
+    promise: 'asynchronous action',
+    await: 'await operator',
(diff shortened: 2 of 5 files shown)


armano2:
@michalsnik There is one commit there with you as author (i took some of your changes) to align with it 🗡 

I cleaned up commits from merges and most of rules are working now 🔢

michalsnik:
Great @armano2, thank you so much for all the hard work! I'll review it in details later today, but so far it looks really good! :) I'm glad you figured out more cases, these all make perfect sense to me :)

docs/rules/no-async-in-computed-properties.md
```diff
@@ -0,0 +1,56 @@
...
+This rule is aimed at preventing asynchronous methods from being called in computed properties.
+
+
+:-1: Examples of **incorrect** code for this rule:
+
+```js
+export default {
+  computed: {
+    pro () {
+      setTimeout(() => { }, 0)
```
michalsnik: Let's put each method causing errors in separate property

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
+/**
+ * @fileoverview Check if there are no async inside computed properties.
```
michalsnik: `asynchronous actions`

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+const TIMED_FUNCTIONS = [
+  'setTimeout',
+  'setInterval',
+  'setImmediate',
+  'requestAnimationFrame'
+]
+
+function isTimedFunction (node) {
+  return (
+    node.callee.type === 'Identifier' &&
```
michalsnik: Add additional `node.type === 'CallExpression'`. It's good to always check the given node too, and not depend on implementation that is outside this function.

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+    TIMED_FUNCTIONS.indexOf(node.callee.name) !== -1
+  ) || (
+    node.callee.type === 'MemberExpression' &&
+    node.callee.object.name === 'window' && (
```
michalsnik: There is a missing check: `node.callee.object.type === 'Identifier' &&`

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+      TIMED_FUNCTIONS.indexOf(node.callee.property.name) !== -1 ||
+      TIMED_FUNCTIONS.indexOf(node.callee.property.value) !== -1
```
michalsnik: Why do you check value here? Isn't `name` enough?

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+    )
+  ) && node.arguments.length
+}
+
+function isPromise (node) {
+  if (node.type === 'CallExpression' && node.callee && node.callee.type === 'MemberExpression') {
```
michalsnik: If it's a `CallExpression` there is `callee` property inside, no need to confirm that.

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+    return ( // hello.PROMISE_FUNCTION()
+      node.callee.property &&
```
michalsnik: You should rather check type instead: `node.callee.property.type === 'Identifier'`

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+      PROMISE_FUNCTIONS.indexOf(node.callee.property.name) !== -1
+    ) || ( // Promise.PROMISE_METHOD()
+      node.callee.object.type === 'Identifier' &&
+      node.callee.object.name === 'Promise' &&
+      PROMISE_METHODS.indexOf(node.callee.property.name) !== -1
+    ) || ( // somePromise.ANYTHING()
+      node.callee.object && isPromise(node.callee.object)
```
michalsnik: Ok this recursion is for checking chained promises, right?

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+    if (node.async) {
+      forbiddenNodes.push({
+        node: node,
+        type: 'async'
+      })
+    }
+    if (node.generator) {
+      forbiddenNodes.push({
+        node: node,
+        type: 'yield*'
```
michalsnik: I think we should call this type simply `generator`

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+
+      computedProperties.forEach(cp => {
+        forbiddenNodes.forEach(el => {
+          if (
+            cp.value &&
+            el.node.loc.start.line >= cp.value.loc.start.line &&
+            el.node.loc.end.line <= cp.value.loc.end.line
+          ) {
+            let message = `Unexpected asynchronous action in "{{name}}" computed property.`
+            if (el.type === 'await') {
```
michalsnik: I see a lot of repetition here. Please think about creating a map with all types and corresponding names. Then return the message without repeating it :)

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,178 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  create,
+  meta: {
+    docs: {
+      description: 'Check if there are no async inside computed properties.',
```
michalsnik: `asynchronous actions`

tests/lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,335 @@
...
+          foo: function* () {
+            yield 1
+            yield* g1()
+          }
+        }
+      }
+      `,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+      errors: [{
+        message: 'Unexpected yield* expression in "foo" computed property.',
```
michalsnik: `generator function expression` would be more correct

lib/rules/no-async-in-computed-properties.js
armano2: yes

armano2:
Thank you for code review i adapted all requests into code except chained promises

lib/rules/no-async-in-computed-properties.js
armano2: @michalsnik there can be expression like
```js
el.somethink.somethink.catch()
```

michalsnik: I don't think so @armano2, in your example it would call `isPromise` with `memberExpression` -> `el.foo.bar` which will automatically return false. What you're checking is:
```
something.crazy.then().test()
```
But I don't see a real case with this kind of node. I don't think you'll ever end up with custom method after promise method so I think we can even remove this last condition. Add more tests if you want, delete it and check if they all pass.

docs/rules/no-async-in-computed-properties.md
```diff
@@ -0,0 +1,64 @@
+# Check if there are no asynchronous action inside computed properties (no-async-in-computed-properties)
```
michalsnik: `actions`

lib/rules/no-async-in-computed-properties.js
armano2: ok

armano2: @michalsnik you are completely right :D
thank you for explanation

lib/rules/no-async-in-computed-properties.js
```diff
@@ -0,0 +1,174 @@
...
+    )
+  }
+  return false
+}
+
+function create (context) {
+  const forbiddenNodes = []
+
+  const messages = {
+    promise: 'Unexpected asynchronous action in "{{name}}" computed property.',
```
michalsnik: I was thinking about:
```
const expressionTypes = {
  promise: 'asynchronous action',
  await: 'await operator',
  yield: 'yield keyword',
  generator: 'generator function expression',
  async: 'async function declaration',
  new: 'Promise object',
  timed: 'timed function'
}
```
and then you could pass message as:
```
Unexpected {{expressionName}} in "{{propertyName}}" computed property.
```
with:
```
data: {
  propertyName: cp.key,
  expressionName: expressionTypes[cp.key]
}
```
What do you think? That would prevent you from duplicating the message.

armano2: I was not sure if those messages are correct at all. I'm terrible at describing stuff 🥇

armano2: @michalsnik updated

michalsnik:
I put `do not merge` label on all of your rules. And I'll start merging/finishing reviewing them after #69 is merged, to have a better picture of your changes.

armano2:
ok

michalsnik:
Alright, you can fix conflicts now :)

armano2:

mysticatea:
I'm not sure this rule should disallow `yield` expressions.
This is not related to async. But it might be dangerous since the value is cached...

```js
export default {
    computed: {
        *numbers() {
            yield 1
            yield 2
            yield 3
        }
    }
}
```

armano2:
> await nodes are YieldExpression's with babel-eslint < 7.0.0

i don't like this to but `babel-eslint < 7.0.0` 👎

mysticatea:
@armano2 I see. Hmm, how about disallowing `async` functions instead of `await` expressions? `babel-eslint@6` looks to have `node.async === true` for `async` functions correctly. The `await` expressions on outside of `async` functions are syntax error.

armano2:
@mysticatea yes its true and we `node.async === true` is already supported

Doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield

> The yield keyword causes generator function execution to **pause** and the value of the expression following the yield keyword is returned to the generator's caller. It can be thought of as a generator-based version of the return keyword.

For me yield can be called async-like.
> A yield, which causes the generator to once again pause and return the generator's new `value`. The next time `next()` is called, **execution resumes with the statement** immediately after the `yield`.

but if you want i can remote it from support.

michalsnik:
Hmm, I'm ok with both versions but to be honest, who would ever want to set generator as computed property? It doesn't make sense at all. I think we can even remote this check and maybe add another rule for that but personally I don't think it's even necessary.. What do you think guys?

armano2:
@michalsnik yield/generator removed 🍡

michalsnik:
Okay, one last thing @armano2 let's change the `await` check to `async` function detection and it's good to go.

armano2:
@michalsnik but we are checking `async` function declaration already here: https://github.com/vuejs/eslint-plugin-vue/pull/72/files#diff-ceb2d093cf4ac93cc581e9af530bfebaR70

PR #76, opened 2017-07-14
Fix options documentation in order-in-components

armano2:


docs/rules/order-in-components.md
```diff
@@ -64,24 +64,26 @@ export default {
...
+      'data',
+      'computed',
+      'watch',
+      'LIFECYCLE_HOOKS',
+      'methods',
+      'render',
+      'renderError'
+    ]
+  }
+]
```
michalsnik: I think this is actually wrong. Current version seems fine @armano2, why do you want to change it?

armano2: `lifecycle_hooks` is invalid because it's should be `LIFECYCLE_HOOKS`

Its defined here [rules/order-in-components.js#L25](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/order-in-components.js#L25) and used [rules/order-in-components.js#L111](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/order-in-components.js#L111)

if you want to make it work we will have to change
```js
const extendedOrder = order.map(property => groups[property] || property)
```
to
```js
const extendedOrder = order.map(property => groups[property.toUpperCase()] || property)
```

michalsnik: Ah this one. Okay. Then please update only this attribute in this PR. The rest is fine. In your pasted example there is a `{` missing.

michalsnik:
Perfect, thanks!

Issue #75, opened 2017-07-14
Rule proposition: `return-in-computed-property`.


armano2:
**Enforces that a return statement is present in computed property (computed-property-return)**

This example would throw an error:
```js
export default {
  computed: {
    a () { }
  }
}
```
What do you think @mysticatea @michalsnik?

michalsnik:
Released in `v3.7.0` 🚀  Thanks @armano2 ! Very good job :)

PR #78, opened 2017-07-15
Add `return-in-computed-property` rule.

armano2:
This PR implements rules proposed in #75

**DONE:**
- [x] Basic check for `return`
- [x] Check if all paths `return` value
- [x] Option: `treatUndefinedAsUnspecified` - disallows implicitly returning undefined with a `return;` statement.

### DIFF
--- a/tests/lib/rules/return-in-computed-property.js
+++ b/tests/lib/rules/return-in-computed-property.js
@@ -1,0 +1,138 @@
+/**
+ * @fileoverview Enforces that a return statement is present in computed property (return-in-computed-property)
+ * @author Armano
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/return-in-computed-property')
+
+const RuleTester = require('eslint').RuleTester
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester()
+ruleTester.run('return-in-computed-property', rule, {
+
+  valid: [
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          computed: {
+            foo () {
+              return true
+            },
+            bar: function () {
+              return false
+            },
+            bar3: {
+              set () {
+                return true
+              },
+              get () {
+                return true
+              }
+            }
+          }
+        }
+      `,
+      parserOptions: { ecmaVersion: 8, sourceType: 'module' }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          computed: {
+            foo: {
+              get () {
+                return
+              }
+            }
+          }
+        }
+      `,
+      parserOptions: { ecmaVersion: 8, sourceType: 'module' },
+      options: [{ treatUndefinedAsUnspecified: false }]
+    }
+  ],
+
+  invalid: [
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          computed: {
+            foo () {
+            }
+          }
+        }
+      `,
+      parserOptions: { ecmaVersion: 8, sourceType: 'module' },
+      errors: [{
+        message: 'Expected to return a value in "foo" computed property.',
+        line: 4
+      }]
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          computed: {
+            foo: function () {
+            }
+          }
+        }
+      `,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+      errors: [{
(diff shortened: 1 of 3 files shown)


armano2:
@michalsnik can you review this pull request?

michalsnik:
Sure, later today I'll try to push my PR and both of yours forward. Busy time :)

armano2:
It's my first time working with rules from eslint (developing them) but they seems not to complex to make.

sorry for pushing you to test my code 🗡

docs/rules/computed-property-return.md
```diff
@@ -0,0 +1,44 @@
+# Enforces that a return statement is present in computed property (computed-property-return)
```
michalsnik: I'd call this rule `return-in-computed-property` to make it more descriptive. WDYT?

docs/rules/computed-property-return.md
```diff
@@ -0,0 +1,44 @@
...
+}
+```
+
+## :wrench: Options
+
+This rule has an object option:
+- `"treatUndefinedAsUnspecified"`: `true` (default) allows implicitly returning undefined with a `return;` statement.
+
+```
+vue/order-in-components: [2, {
```
michalsnik: Whoops

docs/rules/computed-property-return.md
```diff
@@ -0,0 +1,44 @@
...
+      return false
+    }
+  }
+}
+```
+
+## :wrench: Options
+
+This rule has an object option:
+- `"treatUndefinedAsUnspecified"`: `true` (default) allows implicitly returning undefined with a `return;` statement.
```
michalsnik: I think we should force to return something explicitly, thus in my opinion `return;` should also trigger an error and this option should not be necessary.
We could consider two types of messages:
- There is no `return` statement inside `foo` computed property
- There is an incomplete `return` statement inside `foo` computed property
 What do you think @armano2 @chrisvfritz ?

armano2: damn description is invalid, i forgot to change it, by default its treating undefined as unspecified than `return;` will thrown an error, but i prefer to leave a choice to disable it because undefined can be useful in some cases.

lib/rules/computed-property-return.js
```diff
@@ -0,0 +1,112 @@
...
+        if (!isValidReturn() && funcInfo.codePath.currentSegments.some((segment) => segment.reachable)) {
+          forbiddenNodes.push({
+            hasReturn: funcInfo.hasReturn,
+            node: funcInfo.node,
+            type: 'return'
+          })
+        }
+      }
+    },
+    utils.executeOnVueInstance(context, properties => {
```
michalsnik: Shouldn't it be `executeOnVueComponent` ? And this one in `computed-data` rule instead?

docs/rules/computed-property-return.md
armano2: i'm bad at naming stuff, your name is way better

lib/rules/computed-property-return.js
armano2: no right now there is 2 functions, `executeOnVueInstance` and `executeOnVueComponent` we have to distinct both of those types, 

see: https://github.com/vuejs/eslint-plugin-vue/pull/78/commits/281f230ef92b3d034c28c7014f11da66269682d1

michalsnik: Yeah, but I thought this matters for `data` as function case, as it's ok to have data as object in vue instance, but not in component. And this rule is about `return` statement in computed property which seems irrelevant.

armano2: `executeOnVueInstance` supports `components` & `new vue`

i just renamed function to simplified merges and improve readability.

armano2:

chrisvfritz:
Will this also flag computed properties that technically have a return somewhere in their content, but not directly? For example:

``` js
computed: {
  foo: function () {
    function bar () {
      return this.baz * 2 
    }
    bar()
  }
}
```

armano2:
@chrisvfritz yes (test for this case added)

tests/lib/rules/return-in-computed-property.js
```diff
@@ -0,0 +1,158 @@
...
+            }
+          }
+        }
+      `,
+      parserOptions: { ecmaVersion: 8, sourceType: 'module' },
+      errors: [{
+        message: 'Expected to return a value in "foo" computed property.',
+        line: 4
+      }]
+    }
```
mysticatea: Could you add invalid tests for `treatUndefinedAsUnspecified` option?

armano2: ok

armano2: @mysticatea added

PR #83, opened 2017-07-16
Add .editorconfig file

armano2:
Add default editorconfig to help develop unified code under most of IDE

see: http://editorconfig.org/


michalsnik:
Thanks @armano2

Issue #79, opened 2017-07-15
Rule proposition: `no-shared-component-data` must be a function.


armano2:
When using the data property on a component (i.e. anywhere except on new Vue), the value must be a function that returns an object.

based on [style guide](https://vue-style-guide-dev.surge.sh/v2/guide/style-guide.html#Component-data)

Invalid:
```js
Vue.component('some-comp', {
  data: {
    foo: 'bar'
  }
})
```
```js
export default {
  data: {
    foo: 'bar'
  }
}
```

Valid:
```js
new Vue({
  data: function () {
    return {
      foo: 'bar'
    }
  }
})
```
```js
new Vue({
  data: {
    foo: 'bar'
  }
})
```
```js
Vue.component('some-comp', {
  data: function () {
    return {
      foo: 'bar'
    }
  }
})
```
```js
export default {
  data: function () {
    return {
      foo: 'bar'
    }
  }
}
```

armano2:
Can I start working on this rule?

michalsnik:
I definitely agree with this one 👍 It's something strongly advised in vue documentation and I think it's a must have rule in our set.

armano2:
it's already implemented and `ready for review` 🍡

michalsnik:
Released in `v3.8.0`. Thanks @armano2 !

PR #84, opened 2017-07-16
Add `no-shared-component-data` rule.

armano2:
This PR implements rules proposed in #79 

**DONE:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule

**Importnant:**

This rule splits `executeOnVueComponent` into `executeOnVueInstance` and `executeOnVueComponent` and adds alias `executeOnVue`

- if you want to process only Vue components use `utils.executeOnVueComponent`
- if you want to process only `new Vue` use `utils.executeOnVueInstance`
- If you want to process both use `utils.executeOnVue`

### DIFF
--- a/docs/rules/no-shared-component-data.md
+++ b/docs/rules/no-shared-component-data.md
@@ -1,0 +1,33 @@
+# Enforces component's data property to be a function (no-shared-component-data)
+
+When using the data property on a component (i.e. anywhere except on `new Vue`), the value must be a function that returns an object.
+
+## :book: Rule Details
+
+When the value of `data` is an object, it’s shared across all instances of a component.
+
+:-1: Examples of **incorrect** code for this rule:
+
+```js
+Vue.component('some-comp', {
+  data: {
+    foo: 'bar'
+  }
+})
+```
+
+:+1: Examples of **correct** code for this rule:
+
+```js
+Vue.component('some-comp', {
+  data: function () {
+    return {
+      foo: 'bar'
+    }
+  }
+})
+```
+
+## :wrench: Options
+
+Nothing.
--- a/lib/rules/no-shared-component-data.js
+++ b/lib/rules/no-shared-component-data.js
@@ -11,24 +11,22 @@ function create (context) {
   // Public
   // ----------------------------------------------------------------------
 
-  return Object.assign({},
-    utils.executeOnVueComponent(context, (obj) => {
-      obj.properties
-        .filter(p =>
-          p.type === 'Property' &&
-          p.key.type === 'Identifier' &&
-          p.key.name === 'data' &&
-          p.value.type !== 'FunctionExpression' &&
-          p.value.type !== 'Identifier'
-        )
-        .forEach(cp => {
-          context.report({
-            node: cp.value,
-            message: '`data` property in component must be a function'
-          })
+  return utils.executeOnVueComponent(context, (obj) => {
+    obj.properties
+      .filter(p =>
+        p.type === 'Property' &&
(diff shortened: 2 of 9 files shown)


lib/rules/component-data.js
```diff
@@ -0,0 +1,60 @@
...
+  // ----------------------------------------------------------------------
+
+  return Object.assign({},
+    utils.executeOnVueComponent(context, properties => {
+      const dataPropertiesNode = properties
+        .filter(p =>
+          p.key.type === 'Identifier' &&
+          p.key.name === 'data' &&
+          p.value.type !== 'FunctionExpression' &&
+          p.value.type !== 'ArrowFunctionExpression'
```
armano2: I'm not really sure if we should allow to use `ArrowFunctionExpression` in `data` due to changed context of function.

armano2:

mysticatea:
I'm considering the rule name. In general, eslint rules which disallow something starts with `no-`.

I don't have the best idea, but maybe `no-shared-component-data`?
I guess that it's shared by component instances if the `data` is not a function.

armano2:
@mysticatea `no-shared-component-data` make more sense than `component-data`, i will rename it

lib/rules/no-shared-component-data.js
```diff
@@ -0,0 +1,51 @@
...
+        .filter(p =>
+          p.type === 'Property' &&
+          p.key.type === 'Identifier' &&
+          p.key.name === 'data' &&
+          p.value.type !== 'FunctionExpression' &&
+          p.value.type !== 'Identifier'
+        ).forEach(cp => {
+          context.report({
+            node: cp.value,
+            message: 'Component `data` must be a function.'
```
michalsnik: What do you think about changing the message to:
`data` property in component must be a function

lib/rules/no-shared-component-data.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces that a component data must be a function',
```
michalsnik: `component's`

docs/rules/no-shared-component-data.md
```diff
@@ -0,0 +1,33 @@
+# Enforces that a component data must be a function (no-shared-component-data)
```
michalsnik: What about: `Enforces component's data property to be a function` ?

lib/rules/no-shared-component-data.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * @fileoverview When using the data property on a component (i.e. anywhere except on new Vue), the value must be a function that returns an object.
```
michalsnik: I'd use the same title as in documentation

lib/rules/no-shared-component-data.js
```diff
@@ -0,0 +1,51 @@
...
+  return Object.assign({},
+    utils.executeOnVueComponent(context, (obj) => {
+      obj.properties
+        .filter(p =>
+          p.type === 'Property' &&
+          p.key.type === 'Identifier' &&
+          p.key.name === 'data' &&
+          p.value.type !== 'FunctionExpression' &&
+          p.value.type !== 'Identifier'
+        ).forEach(cp => {
```
michalsnik: Move `.forEach` to new line for consistency

armano2:
@michalsnik suggestions applied

armano2:
@michalsnik fixed

Issue #19, opened 2017-06-05
Rule proposition: prop-specificity


chrisvfritz:
A rule to require a certain level of prop specificity could be useful. Two levels of specificity that come to mind:

1. At least type(s) (e.g. `myProp: String` or `myProp: [String, Number]`) - object value for `props`
2. At least definition object (e.g. `myProp: { type: String }`), making it easier to add other restrictions (e.g. `required`, `validate`) - requires object value for `props` and also for any properties on that object

michalsnik:
Sounds legit, but I'd consider this as purely stylistic rule, not necessarily a good practice as it's a matter of preference. What do you think @mysticatea ?

chrisvfritz:
I thinks it's a grey area. Greater specificity catches more errors, but errors are not inevitable with lower specificity.

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

armano2:
Validation of props helps to catch issues easier and its highly recommend for bigger projects due to a lot of components and often inexperienced developers.

We should have this rule but it's should not be enabled by default.

PR #85, opened 2017-07-16
Add `require-prop-types` rule.

armano2:
This PR implements rules proposed in #19

**DONE:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule

### DIFF
--- a/lib/rules/require-prop-types.js
+++ b/lib/rules/require-prop-types.js
@@ -1,0 +1,94 @@
+/**
+ * @fileoverview Prop definitions should be detailed
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../utils')
+
+function create (context) {
+  // ----------------------------------------------------------------------
+  // Helpers
+  // ----------------------------------------------------------------------
+
+  function objectHasType (node) {
+    const typeProperty = node.properties
+      .find(p =>
+        utils.getStaticPropertyName(p.key) === 'type' &&
+        (
+          p.value.type !== 'ArrayExpression' ||
+          p.value.elements.length > 0
+        )
+      )
+    return Boolean(typeProperty)
+  }
+
+  function checkProperties (items) {
+    for (const cp of items) {
+      if (cp.type !== 'Property') {
+        return
+      }
+      let hasType = true
+      if (cp.value.type === 'ObjectExpression') { // foo: {
+        hasType = objectHasType(cp.value)
+      } else if (cp.value.type === 'ArrayExpression') { // foo: [
+        hasType = cp.value.elements.length > 0
+      } else if (cp.value.type === 'FunctionExpression' || cp.value.type === 'ArrowFunctionExpression') {
+        hasType = false
+      }
+      if (!hasType) {
+        context.report({
+          node: cp,
+          message: 'Prop "{{name}}" should define at least it\'s type.',
+          data: {
+            name: cp.key.name
+          }
+        })
+      }
+    }
+  }
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVue(context, (obj) => {
+    const node = obj.properties
+      .find(p =>
+        p.type === 'Property' &&
+        p.key.type === 'Identifier' &&
+        p.key.name === 'props'
+      )
+
+    if (!node) return
+
+    if (node.value.type === 'ObjectExpression') {
+      checkProperties(node.value.properties)
+    } else {
+      context.report({
+        node,
+        message: 'Props should at least define their types.'
+      })
+    }
+  })
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prop definitions should be detailed',
+      category: 'Best Practices',
+      recommended: false
+    },
+    fixable: null,  // or "code" or "whitespace"
+    schema: [
+      // fill in your schema
+    ]
+  },
+
+  create
+}
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -273,17 +273,59 @@ module.exports = {
   },
 
   /**
+   * Gets the property name of a given node.
+   * @param {ASTNode} node - The node to get.
+   * @return {string|null} The property name if static. Otherwise, null.
+   */
+  getStaticPropertyName (node) {
+    let prop
+    switch (node && node.type) {
+      case 'Property':
+      case 'MethodDefinition':
+        prop = node.key
+        break
+      case 'MemberExpression':
+        prop = node.property
+        break
+      case 'Literal':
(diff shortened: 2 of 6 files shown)


armano2:
@michalsnik its ready to review

armano2:
@mysticatea i like more `require-prop-types` -> renamed

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,113 @@
...
+const utils = require('../utils')
+
+function create (context) {
+  // variables should be defined here
+
+  // ----------------------------------------------------------------------
+  // Helpers
+  // ----------------------------------------------------------------------
+
+  function getPropTypes (componentProperties) {
```
michalsnik: In my opinion this function contains too much logic. Can you please try to split it into smaller chunks - easier to understand and test? We should make the code as simple as possible, so that it's easier for anyone to contribute later on.
Such functions should preferable contain a small comment above on what they do, what parameters are required and what's the supposed output.

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,113 @@
...
+            )[0])
+        } else if (cp.value.type === 'ArrayExpression') { // foo: [
+          hasType = cp.value.elements.length > 0
+        } else if (cp.value.type === 'FunctionExpression' || cp.value.type === 'ArrowFunctionExpression') {
+          hasType = false
+        }
+        return { key, node: cp, hasType }
+      })
+    } else if (type === 'ArrayExpression') { // props: [
+      // nodes = node.elements
```
michalsnik: What about this?

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,113 @@
...
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVueComponent(context, (obj) => {
+    const data = getPropTypes(obj.properties)
+    if (data.type === 'ObjectExpression') {
+      data.nodes.forEach(cp => {
+        if (!cp.hasType) {
+          context.report({
+            node: cp.node || data.node,
+            message: 'Prop "{{name}}" definitions should always be as detailed with at least type(s).',
```
michalsnik: What about this message instead?
`Prop "{{name}}" should define at least it's type.`

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,113 @@
...
+            data: {
+              name: cp.key
+            }
+          })
+        }
+      })
+    } else if (data.type === 'ArrayExpression') {
+      context.report({
+        node: data.node,
+        message: 'Props definitions should always be an object and detailed with at least type(s).'
```
michalsnik: Maybe this instead?
`Props' should at least define their types.`
or
`Prop types are required`

armano2:
@mysticatea suggestions applied 🐶

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,95 @@
...
+
+const utils = require('../utils')
+
+function create (context) {
+  // ----------------------------------------------------------------------
+  // Helpers
+  // ----------------------------------------------------------------------
+
+  function objectHasType (nodes) {
+    return !!(nodes
```
michalsnik: KISS:
```
const typeProperty = nodes.filter(p =>
// ...
return Boolean(typeProperty)
```

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,95 @@
...
+
+  function checkProperties (items) {
+    items.map(cp => {
+      if (cp.type !== 'Property') {
+        return
+      }
+      const key = cp.key.name
+      let hasType = true
+      if (cp.value.type === 'ObjectExpression') { // foo: {
+        hasType = objectHasType(cp.value.properties)
```
michalsnik: If this function name is to be accurate you should pass `cp.value` as it's an `ObjectExpression`.

armano2:
@michalsnik i found out that we are not supporting
```
['type']: ....
```
in rules 🗡

lib/utils/index.js
```diff
@@ -273,6 +273,48 @@ module.exports = {
   },
 
   /**
+   * Gets the property name of a given node.
+   * @param {ASTNode} node - The node to get.
+   * @return {string|null} The property name if static. Otherwise, null.
+   */
+  getStaticPropertyName (node) {
```
michalsnik: Please add unit tests for this method

armano2: @michalsnik added

mysticatea: It's sweet memories, I wrote it at [eslint/lib/ast-utils](https://github.com/eslint/eslint/blame/2874d75ed8decf363006db25aac2d5f8991bd969/lib/ast-utils.js#L878).
I hope `eslint/lib/ast-utils` to separate to another package (though it's an item of ESLint 5 wish list).

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,94 @@
...
+    })
+  }
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVue(context, (obj) => {
+    const node = obj.properties
+      .filter(p =>
```
mysticatea: We can use `Array#find` instead.

> http://node.green/#ES2015-built-ins-typed-arrays--TypedArray--prototype-find

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,94 @@
...
+        (
+          p.value.type !== 'ArrayExpression' ||
+          p.value.elements.length > 0
+        )
+      )[0]
+    return Boolean(typeProperty)
+  }
+
+  function checkProperties (items) {
+    items.map(cp => {
```
mysticatea: I'm wondering that it uses `Array#map` to iterate the array. I like `for-of` :smiley:

lib/utils/index.js
armano2: this method is not same but almost, and ast-utils is not exported than i will have to import file directly what can caus some errors between versions of eslint.

lib/rules/require-prop-types.js
```diff
@@ -0,0 +1,94 @@
...
+const utils = require('../utils')
+
+function create (context) {
+  // ----------------------------------------------------------------------
+  // Helpers
+  // ----------------------------------------------------------------------
+
+  function objectHasType (node) {
+    const typeProperty = node.properties
+      .filter(p =>
```
mysticatea: We can use `Array#find` instead.

armano2: i have to start using http://node.green/ to be sure what i can/can't use 🍏

lib/rules/require-prop-types.js
armano2: this map is incorrect 👍 we are even not returning any values 🍡

armano2:
@mysticatea suggestions applied, thank you, and thank you for sharing "http://node.green/"

Issue #86, opened 2017-07-17
Rule proposition: `no-dupe-keys`.


armano2:
There is many fields in vue instance/component where we can specify properties visible inside component, `data`, `props`, `computed`, `methods`.
Invalid Code:
```js
export default {
  props: ['foo'],
  props: {
    foo: String  
  },
  computed: {
    foo () {
    }
  },
  data() {
    return {
      foo: null
    }
  },
  data: {
    foo: null
  },
  methods: {
    foo () {
    }
  }
}
```

Also there is few reserved names witch should not be overridden by vue instance:
* properties: `$data`, `$props`, `$el`, `$options`, `$parent`, `$root`, `$children`, `$slots`, `$scopedSlots`, `$refs`, `$isServer`, `$attrs`, `$listeners`
* methods: `$watch`, `$set`, `$delete`, `$on`, `$once`, `$off`, `$emit`, `$mount`, `$forceUpdate`, `$nextTick`, `$destroy` .....

also user should be able to specify his own reserved names to remove conflicts with used plugins.

mysticatea:
Thank you for the proposal!

Sounds reasonable to me.
`no-dupe-keys` core rule cannot catch duplication of Vue's definition. We should catch it.

Also, the names which start with `_` in `data` are reserved by Vue.js.

armano2:
@mysticatea i think we should rename this rule to same as core `no-dupe-keys` instead of `no-duplicate-field-names`

> Also, the names which start with _ in data are reserved by Vue.js.

👍 i didn't know that but i will add this.

michalsnik:
I like this proposition a lot, but.. I think this rule should be splitted into two rules.
1. `no-dupe-keys`
2. `no-reservered-keys`
So that each rule does only one thing. What do you think @mysticatea @armano2 ?

armano2:
eslint recommends to make functions to small chunks and split them to separate rules.
> One rule should do one think

@michalsnik i like it 👍

PR #88, opened 2017-07-17
Add rules: `no-dupe-keys` and `no-reserved-keys`

armano2:
This PR implements rules proposed in #86

**DONE:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule

### DIFF
--- a/lib/rules/no-reservered-keys.js
+++ b/lib/rules/no-reservered-keys.js
@@ -1,0 +1,74 @@
+/**
+ * @fileoverview Prevent overwrite reserved keys
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const RESERVED_KEYS = require('../utils/vue-reserved.json')
+const GROUP_NAMES = ['props', 'computed', 'data', 'methods']
+
+function create (context) {
+  const options = context.options[0] || {}
+  const reservedKeys = new Set(RESERVED_KEYS.concat(options.reserved || []))
+  const groups = new Set(GROUP_NAMES.concat(options.groups || []))
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVue(context, (obj) => {
+    const properties = utils.iterateProperties(obj, groups)
+    for (const o of properties) {
+      if (o.groupName === 'data' && o.name[0] === '_') {
+        context.report({
+          node: o.node,
+          message: "Keys starting with with '_' are reserved in '{{name}}' group.",
+          data: {
+            name: o.name
+          }
+        })
+      } else if (reservedKeys.has(o.name)) {
+        context.report({
+          node: o.node,
+          message: "Key '{{name}}' is reserved.",
+          data: {
+            name: o.name
+          }
+        })
+      }
+    }
+  })
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent overwrite reserved keys.',
+      category: 'Possible Errors',
+      recommended: false
+    },
+    fixable: null,  // or "code" or "whitespace"
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          reserved: {
+            type: 'array'
+          },
+          groups: {
+            type: 'array'
+          }
+        },
+        additionalProperties: false
+      }
+    ]
+  },
+
+  create
+}
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -272,6 +272,48 @@ module.exports = {
   },
 
   /**
+   * Gets the property name of a given node.
+   * @param {ASTNode} node - The node to get.
+   * @return {string|null} The property name if static. Otherwise, null.
+   */
+  getStaticPropertyName (node) {
+    let prop
+    switch (node && node.type) {
+      case 'Property':
+      case 'MethodDefinition':
+        prop = node.key
+        break
+      case 'MemberExpression':
+        prop = node.property
+        break
+      case 'Literal':
+      case 'TemplateLiteral':
+      case 'Identifier':
+        prop = node
+        break
+      // no default
+    }
+
+    switch (prop && prop.type) {
+      case 'Literal':
+        return String(prop.value)
+      case 'TemplateLiteral':
+        if (prop.expressions.length === 0 && prop.quasis.length === 1) {
+          return prop.quasis[0].value.cooked
+        }
+        break
+      case 'Identifier':
+        if (!node.computed) {
+          return prop.name
+        }
+        break
+      // no default
+    }
+
+    return null
+  },
+
+  /**
    * Get all computed properties by looking at all component's properties
    * @param {ObjectExpression} Object with component definition
    * @return {Array} Array of computed properties in format: [{key: String, value: ASTNode}]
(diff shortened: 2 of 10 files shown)


armano2:
@michalsnik its ready to review

michalsnik:
Let's do it one after another. It's really good to see you work so fast and being so committed! 💪 , however we didn't even approved the proposition yet. I hope you don't mind if we'll wait a while until merging all of your PRs? We'd like to focus and gracefully add more and more rules while constantly keeping everything well organized.

armano2:
@michalsnik we are just need them in out project to improve quality of code.

pull requests for js validation should be merged/reviewed after your #69 is going to be merged and conflicts are going to be resolved :dancer:

michalsnik:
I added suggestion in #86

armano2:
@michalsnik i found out that we are not supporting
```
['type']: ....
```
in rules 🗡

armano2:
@michalsnik rule splited to:
- `no-dupe-keys`
- `no-reservered-keys`

lib/utils/index.js
```diff
@@ -384,5 +426,48 @@ module.exports = {
         cb(node.arguments[0])
       }
     }
+  },
+
+  iterateProperties (node, scopeNames, cb) {
```
michalsnik: Please add comments that will describe those new functions

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+
+const RESERVER_NAMES = new Set(require('../utils/vue-reserved.json'))
+const SCOPE_NAMES = new Set(['props', 'computed', 'data', 'methods'])
+
+function create (context) {
+  const reservedNames = RESERVER_NAMES
+  const scopeNames = SCOPE_NAMES
+
+  const options = context.options[0] || {}
+  if (options.reserved) {
```
michalsnik: You can actually simplify config with:
```
const options = context.options[0] || {}
const reservedKeys = new Set([...RESERVED_KEYS, ...options.reserved || []])
const scopes = new Set([...SCOPE_NAMES, ...options.scopes || []])
```
Plus you can then set `RESERVED_KEYS` and `SCOPE_NAMES` as arrays above.
I changed names of variables so keep it in mind :)

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+'use strict'
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const RESERVER_NAMES = new Set(require('../utils/vue-reserved.json'))
```
michalsnik: `RESERVED`, not `RESERVER`

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+    })
+  }
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVueComponent(context, (obj) => {
+    utils.iterateProperties(obj, scopeNames, (name, node, groupName) => {
+      if (groupName === 'data' && name.substr(0, 1) === '_') {
```
michalsnik: `name[0]` would also work

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+        context.report({
+          node: node,
+          message: "Field '{{name}}' which start with _ in data is reserved.",
```
michalsnik: What do you think about this message: `Keys starting with with "_" are reserved in "data" group.` ?

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+          data: {
+            name
+          }
+        })
+      } else if (reservedNames.has(name)) {
+        context.report({
+          node: node,
+          message: "Reserved key '{{name}}'.",
```
michalsnik: `Key '{{name}}' is reserved`

lib/rules/no-reservered-keys.js
```diff
@@ -0,0 +1,84 @@
...
+        })
+      }
+    })
+  })
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent overwrite reserved keys',
```
michalsnik: I added suggestion for description above

lib/utils/index.js
michalsnik: Also, what do you think about calling `scopeNames` -> `groups` instead? Scope is actually the same so being semantically correct we'd call it rather a group. At least I think so

lib/utils/index.js
```diff
@@ -384,5 +426,48 @@ module.exports = {
...
+        } else if (node.value.type === 'ObjectExpression') {
+          this.iterateObjectExpression(node.value, name, cb)
+        } else if (node.value.type === 'FunctionExpression') {
+          this.iterateFunctionExpression(node.value, name, cb)
+        }
+      })
+  },
+
+  iterateArrayExpression (node, groupName, cb) {
+    node.elements.forEach(item => {
```
michalsnik: I'd check type of the passed node in each of those functions first.

tests/lib/rules/no-dupe-keys.js
```diff
@@ -0,0 +1,146 @@
...
+      filename: 'test.vue',
+      code: `
+        export default {
+          ...foo(),
+          props: {
+            ...foo(),
+            foo: String
+          },
+          computed: {
+            ...mapGetters({
```
michalsnik: We don't support checking `mapGetters` and `mapState` case yet right? It's probably something worth considering but only once we have `vuex` environment available in this plugin.

armano2: @michalsnik we are not supporting it yet, and i'm waiting with it for respond about environments 💃 this is just example with spread operator.

lib/rules/no-reservered-keys.js
armano2: spread operator is not working with node 4

armano2:
@michalsnik suggestions applied

lib/utils/index.js
```diff
@@ -398,5 +440,69 @@ module.exports = {
         cb(node.arguments.slice(-1)[0])
       }
     }
+  },
+
+  iterateProperties (node, groups, cb) {
```
mysticatea: There is the reason you avoid generator functions by callback style?
I like `for-of`.

armano2:
Node 4 :D

lib/utils/index.js
armano2: Node 4

mysticatea:
Node 4 supports generators and for-of :smile:

armano2:
Aaaa I was almost 100% sure than it's not supported....

Than I can change it after launch

armano2:
@mysticatea updated

mysticatea:
@armano2 Thank you for prompt update! I'm sorry that I give up today... (AM 1:30 of Monday in my timezone).

lib/utils/index.js
```diff
@@ -273,6 +273,48 @@ module.exports = {
   },
 
   /**
+   * Gets the property name of a given node.
+   * @param {ASTNode} node - The node to get.
+   * @return {string|null} The property name if static. Otherwise, null.
+   */
+  getStaticPropertyName (node) {
```
michalsnik: Can you please add test for this method?

armano2: @michalsnik added

PR #90, opened 2017-07-17
Add support for v-on="$listeners" (#89)

### LINKED ISSUE #89: Add support for v-on="$listeners"
The [no-invalid-v-on](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/no-invalid-v-on.md) rule is breaking for the new `v-on="$listeners"` feature in [Vue v2.4.0](https://github.com/vuejs/vue/releases/tag/v2.4.0)

* **ESLint Version:** v3.19.0
* **eslint-plugin-vue Version:** v3.5.1
* **Node Version:** v7.9.0

```
<template>
	<div>
		<input v-on="$listeners">
	</div>
</template>
```

No warning/errors.

```

WARNING in ./src/components/Input/src/Input.vue

  ✘  https://google.com/#q=vue%2Fno-invalid-v-on  'v-on' directives require event names
  src/components/Input/src/Input.vue:12:4
  			v-on="$listeners"
```

armano2:
this pull request fixes #89

### DIFF
--- a/lib/rules/no-invalid-v-on.js
+++ b/lib/rules/no-invalid-v-on.js
@@ -33,15 +33,6 @@ const VERB_MODIFIERS = new Set([
 function create (context) {
   utils.registerTemplateBodyVisitor(context, {
     "VAttribute[directive=true][key.name='on']" (node) {
-      if (!node.key.argument) {
-        if (!node.value || node.value.type !== 'VExpressionContainer' || node.value.expression.name !== '$listeners') {
-          context.report({
-            node,
-            loc: node.loc,
-            message: "'v-on' directives require event names."
-          })
-        }
-      }
       for (const modifier of node.key.modifiers) {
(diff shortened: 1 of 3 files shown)


nickmessing:
Theoretically we could use something like:
```html
<template>
  <button v-on="listeners">Hey</button>
</template>
<script>
export default {
  data: () => ({ // doesn't have to be in data, just an example
    listeners: {
      click: someFn,
      keyup: someOtherFn
    }
  })
}
</script>
```

And it is valid in Vue starting with 2.4.0. This is a bad pattern most of the time but sometimes it could be useful like having a computed property that removes `click` from `$listeners` but has anything else.

Do we consider that?

armano2:
Not in this PR, also for now Vue itself do not support this syntax (or maybe i understand this incorrectly)

nickmessing:
@armano2, vue does support that so I am not sure this rule is valid anymore, here's an example: https://jsfiddle.net/z8y5fpx6/

armano2:
thank you for clarification, part of this rule has been removed

Issue #93, opened 2017-07-18
Rule Proposition: `name` property casing


filipalacerda:
Per feedback in https://github.com/vuejs/eslint-plugin-vue/issues/77#issuecomment-316057943:

Rule proposition:
### `name` property casing
Define a style for the `name` property casing for consistency purposes: 

#### Configuration
`'vue/name-casing': ['camelCase'|'kebab-case'|'PascalCase']`
##### `camelCase`
```
{
  name: 'myComponent'
}
```
##### `kebab-case`
```
{
  name: 'my-component'
}
```
##### `PascalCase`
```
{
  name: 'MyComponent'
}
```

cc @armano2

armano2:
@filipalacerda rule added, can you please look into

https://github.com/armano2/eslint-plugin-vue/blob/patch-14-name-casing/docs/rules/name-casing.md
is that enough?

filipalacerda:
@armano2 the readme LGTM 🚀

mysticatea:
Thank you for the proposal!

Sounds good to me. I prefer `PascalCase` by default and the default should be declared in documents.

I think the rule name `name-casing` is too general. It should be specialized for the component name.

Maybe `component-name-casing`?
But it might mislead to the rule which is for use-side (e.g. `<template><MyComponent /></template>`).

Hmm...
I'm not sure proper name...

armano2:
@mysticatea what do you think about `no-invalid-component-name-casing`?

mysticatea:
@armano2 I have learned that we should avoid double negation, while I have written core rules.

How about `name-property-casing`?

mysticatea:
Aha, I have already use many double negation 😂

armano2:
@mysticatea `name-property-casing` seem good and there is not going to be confusion about template <=> script

armano2:
@mysticatea #94 is merged i thik we can close this ticket

michalsnik:
Released in `v3.8.0` 🚀 Thanks!

PR #94, opened 2017-07-18
Add `name-property-casing` rule.

armano2:
This PR implements rules proposed in #93

**DONE:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule


docs/rules/name-property-casing.md
```diff
@@ -0,0 +1,37 @@
...
+  name: 'myComponent'
+}
+```
+
+## :wrench: Options
+
+Default casing is set to `PascalCase`
+
+```
+'vue/name-property-casing': [2, 'camelCase'|'kebab-case'|'PascalCase']
```
michalsnik: `camelCase|kebab-case|PascalCase`

lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * @fileoverview Name property casing for consistency purposes
```
michalsnik: Use the same title as in documentation

lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,79 @@
...
+  const options = context.options[0]
+  const caseType = ['camelCase', 'kebab-case', 'PascalCase'].indexOf(options) !== -1 ? options : 'PascalCase'
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return utils.executeOnVueComponent(context, (obj) => {
+    const node = obj.properties
+      .filter(item => item.type === 'Property' && item.key.name === 'name' && item.value.type === 'Literal')[0]
```
michalsnik: Please wrap this line to make it more readable

lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,79 @@
...
+    if (node) {
```
michalsnik: Add empty line above `if`

michalsnik: Btw. you can also do:
```
if (!node) { return; }
```
To not introduce unnecessary indentation.

lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,79 @@
...
+  return str.replace(/(?:^\w|[A-Z]|\b\w)/g, (letter, index) => index === 0 ? letter.toLowerCase() : letter.toUpperCase()).replace(/[\s-]+/g, '')
+}
+
+function pascalCase (str) {
+  str = camelCase(str)
+  return str.length > 0 ? str.charAt(0).toUpperCase() + str.slice(1) : ''
+}
+
+function convertCase (str, caseType) {
+  if (caseType === 'kebab-case') {
```
michalsnik: Ideal case for `switch/case`. Or better you can create map with converters:

```
const convertersMap = {
  'kebab-case': kebabCase,
  'camelCase': camelCase,
  'PascalCase': pascalCase
}

function getConverter(name) {
  return convertersMap[name] || pascalCase
}
```

And use it like so:
```
function convertCase (str, caseType) {
  return getConverter(caseType)(str)
}
```

WDYT?

lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,79 @@
...
+  return camelCase(str)
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+function create (context) {
+  const options = context.options[0]
+  const caseType = ['camelCase', 'kebab-case', 'PascalCase'].indexOf(options) !== -1 ? options : 'PascalCase'
```
michalsnik: Please move the array with possible caseTypes out of the `create` function, it's how it's done in most rules I think.

tests/lib/rules/name-property-casing.js
```diff
@@ -0,0 +1,125 @@
...
+        message: 'Property name "foo!bar" is not camelCase.',
+        type: 'Literal',
+        line: 3
+      }]
+    },
+    {
+      filename: 'test.js',
+      code: `
+        new Vue({
+          name: 'foo!bar'
```
michalsnik: Please add one more test with `foo_bar`

armano2:
@michalsnik i applied your suggestions & i wrapped few additional line of code.

michalsnik:
I'll release it later :)

F3n67u:
Make name-property-casing fixable is very dangerous, because the fix code only fix the name and leave the reference code same. I think the fix should leave to manual fix, we just show the alert and do nothing else.

Issue #92, opened 2017-07-18
Rule Proposition: Props Casing


filipalacerda:
Per feedback in https://github.com/vuejs/eslint-plugin-vue/issues/77#issuecomment-316057943:

Rule proposition:
### Props Casing
Define a style for the props casing in templates

We (at GitLab) often see inconsistency in the way we provide props through templates:
```
<component myProp="prop"></component>
```
or
```
<component my-prop="prop"></component>
```

#### Configuration
`'vue/props-casing': ['camelCase'|'kebab-case'|'PascalCase']`
##### `camelCase`
```
<component myProp="prop"></component>
```
##### `kebab-case`
```
<component my-prop="prop"></component>
```
##### `PascalCase`
```
<component MyProp="prop"></component>
```

cc @armano2

armano2:
@filipalacerda i think better name for this rule is `html-attributes-casing`

armano2:
@filipalacerda i think better name for this rule is `html-attributes-casing`

filipalacerda:
@armano2 they can be vue props tho 🤔

armano2:
yes but we don't know if they are props or regular attributes and even if they are props they are still attributes

mysticatea:
Thank you for the proposal!

Sounds good to me, but I have some notes.

- This rule should apply to only HTML context; should not apply to SVG/MathML context because XML is case-sensitive.
- I'm not sure the three options are proper. Because we are using all small letters in the attributes of HTML in practice. E.g. `onclick`, `contenteditable`, etc...

PR #95, opened 2017-07-18
Add rule `attribute-hyphenation`.

armano2:
This PR implements rules proposed in #92

**TODO:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule
- [x] Do not update `data-` attribute
- [x] Do not update Svg & MathML attributes
- [x] Add auto-fix
- [x] Do not update `aria-` attribute
- [x] Add tests for auto-fix
- [x] wait for mysticatea/vue-eslint-parser#12
- [x] Use use `node.namespace` to determine if node is Svg & MathML instead of checking node names
- [x] Improve tests

This PR contains changes proposed in #120 and should be merged after it.

### DIFF
--- a/docs/rules/attribute-hyphenation.md
+++ b/docs/rules/attribute-hyphenation.md
@@ -1,0 +1,53 @@
+# Define if attributes on cusom components can be hyphened. (attribute-hyphenation)
+
+## :wrench: Options
+
+Default casing is set to `always`
+
+```
+'vue/attribute-hyphenation': [2, 'always'|'never']
+```
+
+### `"always"` - Use hyphenated name. (It errors on upper case letters.)
+
+:+1: Examples of **correct** code`:
+
+```html
+<template>
+  <foo my-prop="prop">
+    <a onClick="return false"></a>
+  </foo>
+</template>
+```
+
+:-1: Examples of **incorrect** code`:
+
+```html
+<template>
+  <foo myProp="prop">
+    <a onClick="return false"></a>
+  </foo>
+</template>
+```
+
+### `"never"` - Don't use hyphenated name. (It errors on hyphens except `data-` and `aria-`.)
+
+:+1: Examples of **correct** code`:
+
+```html
+<template>
+  <foo myProp="prop">
+    <a onClick="return false"></a>
+  </foo>
+</template>
+```
+
+:-1: Examples of **incorrect** code`:
+
+```html
+<template>
+  <foo my-prop="prop">
+    <a onClick="return false"></a>
+  </foo>
+</template>
+```
--- a/lib/rules/attribute-hyphenation.js
+++ b/lib/rules/attribute-hyphenation.js
@@ -1,0 +1,76 @@
+/**
+ * @fileoverview Define a style for the props casing in templates.
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../utils')
+const casing = require('../utils/casing')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+function create (context) {
+  const sourceCode = context.getSourceCode()
+  const options = context.options[0]
+  const useHyphenated = options !== 'never'
+
+  const caseConverter = casing.getConverter(useHyphenated ? 'kebab-case' : 'camelCase')
+
+  function reportIssue (node, name) {
+    const text = sourceCode.getText(node.key)
+
+    context.report({
+      node: node.key,
+      loc: node.loc,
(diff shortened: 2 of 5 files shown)


armano2:
Based on https://github.com/vuejs/eslint-plugin-vue/issues/92#issuecomment-316506944 we are missing some cases

armano2:
@michalsnik rule is done

docs/rules/html-attributes-casing.md
```diff
@@ -0,0 +1,35 @@
+# Define a style for the attributes casing in templates. (html-attributes-casing)
```
mysticatea: Thank you for the contributing! I'm sorry for the delay.

I'm not sure whether this rule makes sense. Below is the result of my investigation.

- In all known HTML/SVG/MathML elements, all of `attrname`, `attrName` and `AttrName` are the same attribute but `attr-name` is another.
- In custom components,
    - about defined properties,
        - for example `attrName` property, people can choose `attrName` or `attr-name` but `attrname` and `AttrName` do not work.
        - for example `AttrName` property, people can choose `AttrName` or `attr-name` but `attrname` and `attrName` do not work.
    - about undefined properties, the same behaivor to known HTML/SVG/MathML elements.

So I think the options `camelCase` and `PascalCase` are not useful because it might be uncontrollable due to the definition of libraries. So maybe.... What do you think below?

- Change rule name to `attribute-hyphenation`
- Options are: `"always"` (default) or `"never"`.
    - `"always"` ... Use hyphenated name.
    - `"never"` ... Don't use hyphenated name.

(I think this rule don't need `html-` prefix because this rule can work on AST. I think that rules which use details of tokens should have `html-` prefix.)

lib/rules/html-attributes-casing.js
```diff
@@ -0,0 +1,85 @@
...
+      fix: fixer => fixer.replaceText(node.key, newName)
+    })
+  }
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  utils.registerTemplateBodyVisitor(context, {
+    'VStartTag' (obj) {
```
mysticatea: I think that you can use `VAttribute` handler to avoid extra looping.

lib/rules/html-attributes-casing.js
```diff
@@ -0,0 +1,85 @@
...
+      if (!utils.isSvgElementName(obj.id.name) && !utils.isMathMLElementName(obj.id.name)) {
```
mysticatea: This rule should check only custom components.

lib/utils/mathml-elements.json
```diff
@@ -0,0 +1,206 @@
+[
```
mysticatea: We can use `node.namespace` to check SVG/MathML now.

docs/rules/html-attributes-casing.md
armano2: I like it:
1. change name of rule to attribute-hyphenation
2. apply rule only on components `isCustomComponent`
3. change options to `always` (default) or `never`
4. add `aria-` to exceptions

lib/utils/mathml-elements.json
armano2: @mysticatea :+1:

docs/rules/attribute-hyphenation.md
```diff
@@ -0,0 +1,53 @@
+# Define if attributes on cusom components can be hyphened. (attribute-hyphenation)
+
+## :wrench: Options
+
+Default casing is set to `allways`
```
afontcu: always

docs/rules/attribute-hyphenation.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+### `"always"` - Use hyphenated name. (It errors on upper case letters.)
+
+:+1: Examples of **correct** code`:
+
+```html
+<template>
+  <foo my-prop="prop">
+    <a onClick="retrun false"></a>
```
afontcu: return

docs/rules/html-attributes-casing.md
armano2: @mysticatea there is bug: mysticatea/vue-eslint-parser#12 with parsing `argument` for `VDirectiveKey`

docs/rules/attribute-hyphenation.md
armano2: @afontcu fixed

docs/rules/attribute-hyphenation.md
armano2: @afontcu fixed

PR #108, opened 2017-07-25
Specify ecmaVersion to 2017

weizhenye:
For now using async/await syntax in `.vue` files will get `Parsing error: The keyword 'await' is reserved (Fatal)`.

Specify `ecmaVersion` to 2017 should solve this problem.

### DIFF
--- a/lib/config/base.js
+++ b/lib/config/base.js
@@ -4,7 +4,7 @@ module.exports = {
   parser: require.resolve('vue-eslint-parser'),
 
   parserOptions: {
-    ecmaVersion: 6,
+    ecmaVersion: 2017,
     sourceType: 'module',
     ecmaFeatures: {
       jsx: true,


armano2:
I think its not good idea to set up by default `ecmaVersion` to 2017, if you need it you can specify it by yourself like this.

```js
module.exports = {
  parserOptions: {
    parser: 'babel-eslint',
    ecmaVersion: 2017
  },
  extends: [
    'plugin:vue/recommended' // or 'plugin:vue/base'
  ]
}
```

armano2:
@weizhenye I think it's not good idea to set up by default `ecmaVersion` to 2017, if you need it you can specify it by yourself like this:

```js
module.exports = {
  parserOptions: {
    parser: 'babel-eslint',
    ecmaVersion: 2017
  },
  extends: [
    'plugin:vue/recommended' // or 'plugin:vue/base'
  ]
}
```

weizhenye:
That requires users to install another library, for what can be done by eslint itself. Only specify ecmaVersion to 2017 in my own `.eslintrc` file will not work for `.vue` file.

ES2017 is already released in June 2017, is there any inappropriate circumstances to not use ES2017?

armano2:
Library? AHH babel... You don't need it xd there

armano2:
```js
module.exports = {
  parserOptions: {
    ecmaVersion: 2017
  },
  extends: [
    'plugin:vue/recommended' // or 'plugin:vue/base'
  ]
}
```

weizhenye:
Oh, sorry, that's my mistake. My `.eslintrc` is similar to above config except adding `airbnb-base`, it keeps showing error in Atom, but no error with running command directly. I'll check my editor settings, sorry for disturbing.

PR #113, opened 2017-07-29
Add new options for `no-duplicate-attributes` #112

### LINKED ISSUE #112: no-duplicate-attributes should have css exceptions
* **ESLint Version:** 4.3.0
* **eslint-plugin-vue Version:** 3.8.0
* **Node Version:** 8.1.4

```vue

<span class="label label-rounded" :class="statusColor"></span>
<span style="font-weight:bold;" :style="{ color: activeColor, fontSize: fontSize + 'px' }"></span>

```

Duplicate attribute 'class'. (vue/no-duplicate-attributes)
Duplicate attribute 'style'. (vue/no-duplicate-attributes)

Class and Style, specifically, should be allowed to be duplicated since they tend to have a mix of static and dynamic content. You often have structural styling that gets set in stone at development time, but want reactive styles or classes applied at runtime.

Perhaps no-duplicate-attributes should take an option for ignoring css attributes (class and style).

armano2:
`allowCoexistClass` - Enables [`v-bind:class`](https://vuejs.org/v2/guide/class-and-style.html) directive can co-exist with the plain `class` attribute.
`allowCoexistStyle` - Enables [`v-bind:style`](https://vuejs.org/v2/guide/class-and-style.html) directive can co-exist with the plain `style` attribute.

This PR should fix bug: fixes #112 .


mysticatea:
Thank you for contributing! I'm sorry for the delay.

Please see https://github.com/vuejs/eslint-plugin-vue/issues/112#issuecomment-319361968

lib/rules/no-duplicate-attributes.js
```diff
@@ -37,27 +37,48 @@ function getName (attribute) {
...
+
+  const directiveNames = new Set()
+  const attributeNames = new Set()
+
+  function isDuplicate (name) {
+    if (allowCoexistStyle && name === 'style') {
+      return directiveNames.has(name)
+    }
+    if (allowCoexistClass && name === 'class') {
+      return directiveNames.has(name)
```
mysticatea: I'd like to see the test of the following case:

```html
<div :class="a" class="b"></div>
```

I guess that the `class="b"` is warned even if `allowCoexistClass` is true.

lib/rules/no-duplicate-attributes.js
```diff
@@ -37,27 +37,48 @@ function getName (attribute) {
  * @returns {Object} AST event handlers.
  */
 function create (context) {
-  const names = new Set()
+  const options = context.options[0] || {}
+  const allowCoexistStyle = Boolean(options.allowCoexistStyle !== false)
+  const allowCoexistClass = Boolean(options.allowCoexistClass !== false)
```
mysticatea: I think that this cast is redundant since the result of `!==` expression is always boolean.

armano2: yup, that's correct

lib/rules/no-duplicate-attributes.js
armano2: @mysticatea you are completely right on this, :+1:

armano2:
@mysticatea changes applied and i added few more tests

PR #114, opened 2017-07-29
Add rule `require-render-return`.

### LINKED ISSUE #107: Rule Proposal: `vue/require-render-return`

Enforces render function to always return value (require-render-return)

`[ ]` Enforces code style
`[ ]` Other (please specify:)

Provide 2-3 code examples that this rule will warn about:
```js
export default {
  render () {
  }
}
```
```js
export default {
  render (h) {
    if (foo) {
      return
    }
  }
}
```

armano2:
This PR implements rules proposed in #107

**DONE:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule

fixes #107

### DIFF
--- a/lib/rules/require-render-return.js
+++ b/lib/rules/require-render-return.js
@@ -14,7 +14,9 @@ function create (context) {
   // ----------------------------------------------------------------------
 
   return Object.assign({},
-    utils.findMissingReturns(forbiddenNodes, true),
+    utils.executeOnFunctionsWithoutReturn(true, node => {
+      forbiddenNodes.push(node)
+    }),
     utils.executeOnVue(context, obj => {
       const node = obj.properties.find(item => item.type === 'Property' &&
         utils.getStaticPropertyName(item) === 'render' &&
--- a/lib/rules/require-render-return.js
+++ b/lib/rules/require-render-return.js
@@ -1,0 +1,57 @@
+/**
+ * @fileoverview Enforces render function to always return value.
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../utils')
+
+function create (context) {
+  const forbiddenNodes = []
+
+  // ----------------------------------------------------------------------
+  // Public
+  // ----------------------------------------------------------------------
+
+  return Object.assign({},
+    utils.findMissingReturns(forbiddenNodes, true),
+    utils.executeOnVue(context, obj => {
+      const node = obj.properties.find(item => item.type === 'Property' &&
+        utils.getStaticPropertyName(item) === 'render' &&
+        (item.value.type === 'ArrowFunctionExpression' || item.value.type === 'FunctionExpression') &&
+        !item.value.expression // render: () => test
+      )
+      if (!node) return
+
+      forbiddenNodes.forEach(el => {
+        if (
+          el.loc.start.line >= node.value.loc.start.line &&
+          el.loc.end.line <= node.value.loc.end.line
+        ) {
+          context.report({
+            node: node.key,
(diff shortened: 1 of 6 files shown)


lib/rules/require-render-return.js
```diff
@@ -0,0 +1,47 @@
...
+  },
+
+  create (context) {
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+
+    return utils.executeOnVue(context, obj => {
+      obj.properties
+        .filter(item => item.type === 'Property' &&
```
mysticatea: I think that `Array#find` fits more in this case.

tests/lib/rules/require-render-return.js
```diff
@@ -0,0 +1,91 @@
...
+      }`,
+      parserOptions
+    },
+    {
+      filename: 'test.vue',
+      code: `export default {
+        render: () => null
+      }`,
+      parserOptions
+    }
```
mysticatea: Could you add tests that this rule does not report functions which are not the `render`?

armano2:
@mysticatea merged and few additional tests added

tests/lib/rules/require-render-return.js
```diff
@@ -0,0 +1,111 @@
...
+        message: 'Expected to return a value in render function.',
+        type: 'Identifier',
+        line: 2
+      }]
+    },
+    {
+      code: `Vue.component('test', {
+        render: function () {
+          if (a) {
+            return
```
michalsnik: and what about:
```
render() {
  if (a) {
    return `<div>a</div>`
  } else {
    return `<span>a</span>
  }
}
```

armano2: ohh i forgot about this case, thank you :dango:

armano2:
@michalsnik issue with `if {return}else{return}` fixed

lib/utils/index.js
```diff
@@ -520,5 +520,56 @@ module.exports = {
         }
       }
     }
+  },
+
+  /**
+   * Find all function witch do not allways return values
```
michalsnik: Should be: `Find all functions which do not always return values`

lib/utils/index.js
```diff
@@ -520,5 +520,56 @@ module.exports = {
...
+   * @param {Array} nodes an array nodes
+   * @param {boolean} treatUndefinedAsUnspecified
+   */
+  findMissingReturns (nodes, treatUndefinedAsUnspecified) {
```
michalsnik: You should avoid mutating arguments. I suggest to modify this function to be more explicit (we can even change the name a bit, to be consistent with `executeOnVue` and so on):
```
executeOnFunctionsWithoutReturn (treatUndefinedAsUnspecified, cb) {
  // ...

  return {
    // ...
    'ArrowFunctionExpression:exit' (node) {
      if (isValidReturn()) return
      cb(funcInfo.node)
    },
    'FunctionExpression:exit' (node) {
      if (isValidReturn()) return
      cb(funcInfo.node)
    }
  }
}
```

Then you could use it like this:
```
return Object.assign({},
  utils.executeOnFunctionsWithoutReturn(treatUndefinedAsUnspecified, node => {
    forbiddenNodes.push(node)
  }),
  utils.executeOnVue(context, properties => {
    // ...
  })
)
```

I think it will be much easier to understand, and what do you think @armano2 ?

armano2: seems logical, and more consistent with `executeOnVue` 👍

PR #115, opened 2017-07-29
Add documentation for some internal function

armano2:


tests/lib/rules/name-property-casing.js
```diff
@@ -83,8 +93,12 @@ ruleTester.run('name-property-casing', rule, {
           name: 'foo  bar'
         }
       `,
-      options: ['PascalCase'],
```
mysticatea: Is this removal intentional?

armano2: yes, `PascalCase` is set by default and we was missing test case without options set

mysticatea: OK, thank you!

PR #119, opened 2017-07-31
Add rule `vue/require-valid-default-prop`.

### LINKED ISSUE #117: Rule Proposal: `vue/require-valid-default-prop`

Enforces prop default values to be valid. In vue we have 2 ways to define `default` prop value:
- by function
  ```js
  {
    foo: {
      default () {}
    }
  }
  ```
- by literal value
  ```js
  {
    foo: {
      default: ''
    }
  }
  ```
Literal value is not is not valid when type is specified and set to Array or Object

This rule should also check type of literal values and determine if types match.
When variable is passed as default we should omit checking

`[ ]` Enforces code style
`[ ]` Other (please specify:)

## Valid:
```js
Vue.component('example', {
  props: {
    // basic type check (`null` means accept any type)
    propA: Number,
    // multiple possible types
    propB: [String, Number],
    // a number with default value
    propD: {
      type: Number,
      default: 100
    },
    // object/array defaults should be returned from a factory function
    propE: {
      type: Object,
      default: function () {
        return { message: 'hello' }
      }
    }
  }
})
```

## Invalid:
```js
Vue.component('example', {
  props: {
    propA: {
      type: String,
      default: {}
    },
    propB: {
      type: String,
      default: []
    },
    propC: {
      type: Object,
      default: []
    },
    propD: {
      type: Array,
      default: []
    },
    propE: {
      type: Object,
      default: { message: 'hello' }
    }
  }
})
```

see more at: https://vuejs.org/v2/guide/components.html#Prop-Validation

armano2:
This PR implements rules proposed in #117

**TODO:**
- [x] Create documtation
- [x] Add tests
- [x] Implement rule
- [x] Add test & implementation for array of types

fixes #117


tests/lib/rules/require-valid-default-prop.js
```diff
@@ -0,0 +1,378 @@
...
+      parserOptions,
+      errors: errorMessage('function')
+    },
+    {
+      filename: 'test.vue',
+      code: `export default {
+        props: {
+          foo: {
+            type: Array,
+            default: []
```
michalsnik: Did you think about supporting also:
```
{
  type: Array,
  default() {
    return 100
  }
}
```
? It should also throw an error I think

tests/lib/rules/require-valid-default-prop.js
```diff
@@ -0,0 +1,378 @@
...
+      }`,
+      parserOptions
+    },
+    {
+      filename: 'test.vue',
+      code: `Vue.component('example', {
+        props: {
+          foo: null,
+          foo: Number,
+          foo: [String, Number],
```
michalsnik: What if I use it like this:
```
foo: {
  type: [Object, Number],
  default: 10
},
bar: {
  type: [Object, Number],
  default: {}
}
```
? Will it throw an error?

armano2: `foo` should not throw an error (it's working)
`bar` should throw an error (it's not working)

tests/lib/rules/require-valid-default-prop.js
armano2: for now i'm not checking what function is returning or even if function is returning something

tests/lib/rules/require-valid-default-prop.js
armano2: I'm missing tests for that, i will add them

armano2:
@michalsnik i added few more tests and i applied you suggestions about doc/message

but, we should not check what is inside default function.
its producing to many cases, but if you want me to i can try to handle them.

PR #120, opened 2017-08-01
Add namespace check of svg & mathML instead of tag names

armano2:
**Old:**
- `isHtmlElementName(string)`
- `isSvgElementName(string)`

**New:**
- `isKnownHtmlElementNode(ASTNode)`
- `isSvgElementNode(ASTNode)`
- `isMathMLElementNode(ASTNode)`

and broken / deprecated / unused `getTemplateBodyTokenStore` is removed in this commit

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -183,7 +183,7 @@ module.exports = {
     assert(node && node.type === 'VElement')
 
     return (
-      !(this.isHtmlElementNode(node) || this.isSvgElementNode(node) || this.isMathMLElementNode(node)) ||
+      !(this.isKnownHtmlElementNode(node) || this.isSvgElementNode(node) || this.isMathMLElementNode(node)) ||
       this.hasAttribute(node, 'is') ||
       this.hasDirective(node, 'bind', 'is')
     )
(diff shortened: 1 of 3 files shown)


lib/utils/index.js
```diff
@@ -200,34 +182,44 @@ module.exports = {
...
-   * Check whether the given name is a HTML element name or not.
-   * @param {string} name The name to check.
-   * @returns {boolean} `true` if the name is a HTML element name.
+   * Check whether the given node is a HTML element or not.
+   * @param {ASTNode} node The node to check.
+   * @returns {boolean} `true` if the node is a HTML element.
    */
-  isHtmlElementName (name) {
-    assert(typeof name === 'string')
+  isHtmlElementNode (node) {
```
mysticatea: Thank you for good change.

However, it might cause confusing if `isHtmlElementNode` is not checking their namespace but `isSvgElementNode` and `isMathMLElementNode` are checking their namespace. How about renaming to `isKnownHtmlElementNode`?

armano2:
I think we should add namespace check to it to... `a` can be in both HTML and SVG

_Sent from my Xiaomi Redmi 4A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

lib/utils/index.js
armano2: @mysticatea I think we should add namespace check to it to... `a` can be in both HTML and SVG

mysticatea: @armano2 Oh, good catch! Also `<style>` and `<script>` exist in SVG.

PR #121, opened 2017-08-02
⚠️ Add support for deprecated state in update-rules ⚠️

armano2:
⚠️


michalsnik:
Good idea @armano2, please add also information in Readme what does the warning mean as we already have a sentence about other states. Also run `npm run update` so we can see the new table generated in this PR.

armano2:
@michalsnik information what :warning: means added and script runned.

michalsnik:
Thank you @armano2 ! :)

PR #123, opened 2017-08-02
Allow to use @vue/component to set enable parsing vue objects

### LINKED ISSUE #109: How to make parser work on plain js files.
I looked into my projects and i found out one missing functionality here, sometimes we can create abstract components in js files with no template, with or without render function and in that case eslint is not going to work with this file because he don't know that this is a component.

**Solution:**
I propose to add a way to instruct parser to treat files with comment `/* @vue/component */` as vue components.

armano2:
fixes #109

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -431,38 +431,30 @@ module.exports = {
    */
   executeOnVueComponent (context, cb) {
     const filePath = context.getFilename()
+    const sourceCode = context.getSourceCode()
     const _this = this
-    const componentComments = []
+    const componentComments = sourceCode.getAllComments().filter(comment => this.isVueComponentComment(comment.value))
     const foundNodes = []
 
-    function isDuplicateNode (node) {
-      return foundNodes.some(el => el.loc.start.line === node.loc.start.line)
+    const isDuplicateNode = (node) => {
+      if (foundNodes.some(el => el.loc.start.line === node.loc.start.line)) return true
+      foundNodes.push(node)
+      return false
     }
 
     return {
-      LineComment (node) {
-        // line comment with @vue/component
-        if (!_this.isVueComponentComment(node.value)) return
-        componentComments.push(node)
-      },
-      BlockComment (node) {
-        // block comment with @vue/component
-        if (!_this.isVueComponentComment(node.value)) return
-        componentComments.push(node)
-      },
       ObjectExpression (node) {
         if (!componentComments.some(el => el.loc.end.line === node.loc.start.line - 1) || isDuplicateNode(node)) return
-        foundNodes.push(node)
         cb(node)
       },
       'ExportDefaultDeclaration:exit' (node) {
         // export default {} in .vue || .jsx
-        if (!_this.isVueComponentFile(node, filePath) || isDuplicateNode(node)) return
+        if (!_this.isVueComponentFile(node, filePath) || isDuplicateNode(node.declaration)) return
         cb(node.declaration)
       },
       'CallExpression:exit' (node) {
         // Vue.component('xxx', {}) || component('xxx', {})
-        if (!_this.isVueComponent(node) || isDuplicateNode(node)) return
+        if (!_this.isVueComponent(node) || isDuplicateNode(node.arguments.slice(-1)[0])) return
         cb(node.arguments.slice(-1)[0])
       }
     }
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -433,7 +433,7 @@ module.exports = {
     const filePath = context.getFilename()
     const sourceCode = context.getSourceCode()
     const _this = this
-    const componentComments = sourceCode.getAllComments().filter(comment => this.isVueComponentComment(comment.value))
+    const componentComments = sourceCode.getAllComments().filter(comment => /@vue\/component/g.test(comment.value))
     const foundNodes = []
 
     const isDuplicateNode = (node) => {
@@ -458,15 +458,6 @@ module.exports = {
         cb(node.arguments.slice(-1)[0])
       }
     }
-  },
-
-  /**
-   * Check whether the given comment is a Vue component based
-   * @param {string} value Conent of comment.
-   * @return {boolean}
-   */
-  isVueComponentComment (value) {
-    return /@vue\/component/g.test(value)
   },
 
   /**
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -432,19 +432,49 @@ module.exports = {
   executeOnVueComponent (context, cb) {
     const filePath = context.getFilename()
     const _this = this
+    const componentComments = []
+    const foundNodes = []
+
+    function isDuplicateNode (node) {
+      return foundNodes.some(el => el.loc.start.line === node.loc.start.line)
+    }
 
     return {
+      LineComment (node) {
+        // line comment with @vue/component
+        if (!_this.isVueComponentComment(node.value)) return
+        componentComments.push(node)
+      },
+      BlockComment (node) {
+        // block comment with @vue/component
+        if (!_this.isVueComponentComment(node.value)) return
+        componentComments.push(node)
+      },
+      ObjectExpression (node) {
+        if (!componentComments.some(el => el.loc.end.line === node.loc.start.line - 1) || isDuplicateNode(node)) return
+        foundNodes.push(node)
+        cb(node)
+      },
       'ExportDefaultDeclaration:exit' (node) {
         // export default {} in .vue || .jsx
-        if (!_this.isVueComponentFile(node, filePath)) return
+        if (!_this.isVueComponentFile(node, filePath) || isDuplicateNode(node)) return
         cb(node.declaration)
       },
       'CallExpression:exit' (node) {
         // Vue.component('xxx', {}) || component('xxx', {})
-        if (!_this.isVueComponent(node)) return
+        if (!_this.isVueComponent(node) || isDuplicateNode(node)) return
         cb(node.arguments.slice(-1)[0])
       }
     }
+  },
+
+  /**
+   * Check whether the given comment is a Vue component based
+   * @param {string} value Conent of comment.
+   * @return {boolean}
+   */
+  isVueComponentComment (value) {
+    return /@vue\/component/g.test(value)
   },
 
   /**
--- a/tests/lib/utils/vue-component.js
+++ b/tests/lib/utils/vue-component.js
@@ -1,0 +1,277 @@
+/**
+ * @author Armano
+ */
+'use strict'
+
+const utils = require('../../../lib/utils/index')
+
+// ------------------------------------------------------------------------------
+// Requirements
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-dupe-keys.js
```diff
@@ -77,8 +77,8 @@ ruleTester.run('no-dupe-keys', rule, {
 
   invalid: [
     {
-      filename: 'test.vue',
-      code: `
+      // filename: 'test.vue',
```
michalsnik: Please create another tests for that, don't comment actual ones. Also try case with block comment too :)

armano2: ok

armano2: @michalsnik new tests added

armano2:
@michalsnik new tests added

armano2:
@michalsnik do you have any suggestions? I'm terrible at describing stuff xd

michalsnik:
I was thinking about adding information in `Usage` section, that we currently detect components based on certain requirements (list them) and if someone wants custom `.js` files to be checked in terms of the available rules he/she needs to add special comment, so that we can assume it's also to be considered as a component.

Btw. Now I realised that there might be other cases, when you for example want to create few small components inside one `.js` files and compose them, you could get:
```
// utils/example.js
const ChildComponent = {
  template: `<div>{{msg}}</div>`,
  data () {
    return { msg: 'Hello child' }
  }
}

export default {
 template: `<child-component />`,
  components: {
    ChildComponent
  }
}
```

such a situation might also have place in `.vue` files. My proposition would be to detect if `@vue/component` is exactly one line above the `ObjectExpression` - if yes -> treat the `ObjectExpression` as a component. This should solve all cases and shouldn't be hard to implement :) What do you think @armano2 ?

The above file could look like this:
```
// @vue/component
const ChildComponent = {
  template: `<div>{{msg}}</div>`,
  data () {
    return { msg: 'Hello child' }
  }
}

export default {
 template: `<child-component />`,
  components: {
    ChildComponent
  }
}
```

armano2:
hmm i like it, and i was thinking about it before.

component definition should work only if next block is
`export default {`,
`export const/var/let [name] = {`
`const/var/let [name] ={`

what do you think?

armano2:
@michalsnik hmm i like it, and i was thinking about it before.

component definition should work only if next block is
`export default {`,
`export const/var/let [name] = {`
`const/var/let [name] = {`

what do you think?

michalsnik:
There are also async components, I wonder if simple check for an object right below comment wouldn't be enough. Then it would be also possible to do:
```
Vue.component('async-example', function (resolve, reject) {
  setTimeout(function () {
    // @vue/component
    resolve({
      template: '<div>I am async!</div>'
    })
  }, 1000)
})
```

armano2:
@michalsnik i'm going to create separate test file for all cases to have them all in one place and i'm going to include async component to

armano2:
@michalsnik can you check if i didn't forgo about something?

https://github.com/armano2/eslint-plugin-vue/blob/patch-28-comment-vue-component/tests/lib/utils/vue-component.js

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,198 @@
...
+      parserOptions
+    },
+    {
+      filename: 'test.js',
+      code: `foo({ })`,
+      parserOptions
+    },
+    {
+      filename: 'test.js',
+      code: `Vue.component('async-example', function (resolve, reject) { })`,
```
michalsnik: `resolve({})` missing

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,198 @@
...
+      code: `export const foo = {}`,
+      parserOptions
+    },
+    {
+      filename: 'test.js',
+      code: `var foo = {}`,
+      parserOptions
+    },
+    {
+      filename: 'test.js',
```
michalsnik: Add some cases with `.jsx` and `.vue` files too that are not to be treated as components by default.

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,198 @@
...
+      errors: [makeError(2)]
+    },
+    {
+      filename: 'test.js',
+      code: `Vue.component({})`,
+      parserOptions,
+      errors: [makeError(1)]
+    },
+    {
+      filename: 'test.vue',
```
michalsnik: Add the same test for `test.jsx` file

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,198 @@
...
+      filename: 'test.js',
+      code: `Vue.component('async-example', function (resolve, reject) { })`,
+      parserOptions
+    }
+  ],
+  invalid: [
+    {
+      filename: 'test.js',
+      code: `Vue.component('async-example', function (resolve, reject) {
+        resolve({})
```
michalsnik: This should actually be not treated as component, we're not supporting this case by default, only with optional comment.

armano2:
@michalsnik i applied all requested changes

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,257 @@
...
+const ruleTester = new RuleTester()
+ruleTester.run('vue-component', rule, {
+
+  valid: [
+    {
+      filename: 'test.js',
+      code: `export default { }`,
+      parserOptions
+    }
+  ].concat(notFoundTests('js')).concat(notFoundTests('jsx')).concat(notFoundTests('vue')),
```
michalsnik: I'd call it `validTests` and `invalidTests` for consistency.

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,257 @@
...
+      parserOptions
+    }
+  ]
+}
+
+function foundTests (ext) {
+  return [
+    {
+      filename: `test.${ext}`,
+      code: `// ${ext}
```
michalsnik: I guess you added those comments for easier debugging?
If yes please keep them in once place, as for now I can see them on top and bottom of the code example.

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,257 @@
...
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const rule = {
+  create (context) {
+    return utils.executeOnVueComponent(context, obj => {
+      context.report({
+        node: obj,
+        message: 'Internal test: component.'
```
michalsnik: I think `Component detected` message would be a little bit more descriptive.

lib/utils/index.js
```diff
@@ -432,22 +432,52 @@ module.exports = {
...
+      },
+      ObjectExpression (node) {
+        if (!componentComments.some(el => el.loc.end.line === node.loc.start.line - 1) || hasNode(node)) return
+        foundNodes.push(node)
+        cb(node)
+      },
       'ExportDefaultDeclaration:exit' (node) {
         // export default {} in .vue || .jsx
-        if (!_this.isVueComponentFile(node, filePath)) return
+        if (!_this.isVueComponentFile(node, filePath, context) || hasNode(node)) return
```
michalsnik: Why did you passed `context` to `isVueComponentFile`? And why did you added `hasNode` here? I thought current check is ok and we only needed to add comment detection before objects?

lib/utils/index.js
```diff
@@ -432,22 +432,52 @@ module.exports = {
...
         cb(node.declaration)
       },
       'CallExpression:exit' (node) {
         // Vue.component('xxx', {}) || component('xxx', {})
-        if (!_this.isVueComponent(node)) return
+        if (!_this.isVueComponent(node) || foundNodes.some(el => el.loc.start.line === node.loc.start.line)) return
```
michalsnik: Why did you added here additional check too?

tests/lib/utils/vue-component.js
```diff
@@ -0,0 +1,257 @@
...
+      filename: `test.${ext}`,
+      code: `// ${ext}
+      // @vue/component
+      export default { }`,
+      parserOptions,
+      errors: [makeError(3)]
+    },
+    {
+      filename: `test.${ext}`,
+      code: `// ${ext}
```
michalsnik: Please consider formatting those code example as follows:
```
code: `
  /* @vue/component */
  export default {}
`
```
I think they would be much easier to read :)

lib/utils/index.js
armano2: to prevent duplicates when someone

lib/utils/index.js
armano2: context removed and to prevent duplicates from callback

tests/lib/utils/vue-component.js
armano2: @michalsnik changed

tests/lib/utils/vue-component.js
armano2: Yes that was the point, and i moved them to down (i forgot to this will all of them)

armano2:
@michalsnik thank you for suggestions, changes applied

lib/utils/index.js
michalsnik: Ah I see, good thinking :)

michalsnik:
Cool @armano2 Now please add info in readme and we're good to go :) Good work!

armano2:
@michalsnik now the hardest part 👎

armano2:
@michalsnik now the hardest part 👎

i wrote something, but i don't like it :(

michalsnik:
@armano2 And what do you think about:

### Attention

All component-related rules are being applied to code that passes any of the following checks:
- `Vue.component()` expression
- `export default {}` in `.vue` or `.jsx` file

If you however want to take advantage of our rules in any of your custom objects that are Vue components, you might need to use special comment `// @vue/component` that marks object in the next line as a Vue component in any file, e.g.:

```
// @vue/component
const CustomComponent = {
  name: 'custom-component',
  template: '<div></div>'
}

Vue.component('AsyncComponent', (resolve, reject) => {
  setTimeout(() => {
    // @vue/component
    resolve({
      name: 'async-component',
      template: '<div></div>'
    })
  }, 500)
})
```

armano2:
@michalsnik i like it way more

i'm just terrible at writing *documentations*

armano2:
There is one issue with this solution, it will not work with eslint 4


https://github.com/eslint/eslint/blob/8796d55a75a2d54b384c2ce3249861d5fe127b12/docs/user-guide/migrating-to-4.0.0.md#-linecomment-and-blockcomment-events-will-no-longer-be-emitted-during-ast-traversal

armano2:
@michalsnik  There is one issue with this solution, it will not work with eslint 4

https://github.com/eslint/eslint/blob/8796d55a75a2d54b384c2ce3249861d5fe127b12/docs/user-guide/migrating-to-4.0.0.md#-linecomment-and-blockcomment-events-will-no-longer-be-emitted-during-ast-traversal

PR #128, opened 2017-08-04
Chore: remove package-lock.json

mysticatea:
This PR removes `package-lock.json`.

In [eslint](https://github.com/eslint/eslint) repo, we have disabled `package-lock.json` since the following reasons:

- The lock file does not affect to `npm install eslint`, but it affects CI builds. This implies that we cannot test with actual dependencies that users install. It reduces the value of CI builds.
- It sometimes causes merge conflictions.
- It's fragile.

I propose to remove `package-lock.json` in this repo as well.


armano2:
I had a lot of issues with it to, and we disabled it in all of our projects.

PR #129, opened 2017-08-04
New: `html-self-closing` rule (fixes #31)

### LINKED ISSUE #31: html-no-self-closing: Ignore components option
I think it is pretty common in Vue templates (for example inside `.vue` files, not in HTML page) to write self-closing component, for example: `<ui-dropdown :options="options" />`. It also don't cause any potential issue since the component will be rendered in HTML even before being put in the DOM thanks to the template compiler.

It would be nice if the recommended config had an option set to the `html-no-self-closing` rule to ignore non-HTML tags.

This should then be reported:

```html
<div />
<ui-dropdown />
```

This shouldn't be reported:

```html
<div></div>
<ui-dropdown />
```

mysticatea:
Fixes #31.

This PR adds `html-self-closing-style` rule which replaces `html-no-self-closing` rule.
This rule would help us to unify self-closing style.

### DIFF
--- a/lib/rules/html-self-closing-style.js
+++ b/lib/rules/html-self-closing-style.js
@@ -68,12 +68,15 @@ function getKind (node) {
 
 /**
  * Check whether the given element is empty or not.
- * This ignores whitespaces.
+ * This ignores whitespaces, doesn't ignore comments.
  * @param {VElement} node The element node to check.
  * @returns {boolean} `true` if the element is empty.
  */
-function isEmpty (node) {
-  return node.children.every(child => child.type === 'VText' && child.value.trim() === '')
+function isEmpty (node, sourceCode) {
+  const start = node.startTag.range[1]
+  const end = (node.endTag != null) ? node.endTag.range[0] : node.range[1]
+
+  return sourceCode.text.slice(start, end).trim() === ''
 }
 
 /**
@@ -83,6 +86,7 @@ function isEmpty (node) {
  * @returns {object} AST event handlers.
  */
 function create (context) {
+  const sourceCode = context.getSourceCode()
   const options = parseOptions(context.options[0])
 
   utils.registerTemplateBodyVisitor(context, {
@@ -90,7 +94,7 @@ function create (context) {
       const kind = getKind(node)
       const mode = options[kind]
 
-      if (mode === 'always' && !node.startTag.selfClosing && isEmpty(node)) {
+      if (mode === 'always' && !node.startTag.selfClosing && isEmpty(node, sourceCode)) {
         context.report({
           node,
           loc: node.loc,
--- a/lib/rules/html-self-closing-style.js
+++ b/lib/rules/html-self-closing-style.js
@@ -1,0 +1,175 @@
+/**
+ * @author Toru Nagashima
+ * @copyright 2016 Toru Nagashima. All rights reserved.
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+/**
+ * Kind strings.
+ * This strings wil be displayed in error messages.
+ */
+const KIND = Object.freeze({
+  NORMAL: 'HTML elements',
+  VOID: 'HTML void elements',
+  COMPONENT: 'Vue.js custom components',
+  SVG: 'SVG elements',
+  MATH: 'MathML elements'
+})
+
+/**
+ * Normalize the given options.
+ * @param {Object|undefined} options The raw options object.
+ * @returns {Object} Normalized options.
+ */
+function parseOptions (options) {
+  return {
+    [KIND.NORMAL]: (options && options.html && options.html.normal) || 'never',
+    [KIND.VOID]: (options && options.html && options.html.void) || 'never',
+    [KIND.COMPONENT]: (options && options.html && options.html.component) || 'always',
+    [KIND.SVG]: (options && options.svg) || 'always',
+    [KIND.MATH]: (options && options.math) || 'always'
+  }
+}
+
+/**
+ * Get the kind of the given element.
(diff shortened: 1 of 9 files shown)


lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+/**
+ * Check whether the given element is empty or not.
+ * This ignores whitespaces.
+ * @param {VElement} node The element node to check.
+ * @returns {boolean} `true` if the element is empty.
+ */
+function isEmpty (node) {
+  return node.children.every(child => child.type === 'VText' && child.value.trim() === '')
```
armano2: You are checking only if element has VText and its not empty and you should also check if there are other elements inside. with somethink like this:
`child.type !== 'VText' || child.value.trim() === ''`

e.g. `html.component = allways`
`<component />` - it will not report error now
`<component><div /></component>` - it will report error now

mysticatea:
Thank you for the review.

I added more tests.

Also, I found a problem about comments then I fixed it.

lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param {RuleContext} context - The rule context.
+ * @returns {object} AST event handlers.
+ */
+function create (context) {
+  const sourceCode = context.getSourceCode()
+  const options = parseOptions(context.options[0])
+
+  utils.registerTemplateBodyVisitor(context, {
+    'VElement' (node) {
+      const kind = getKind(node)
```
michalsnik: I'd keep `kind` simple as: `normal`, `void`, `component` and so on, and only reference the longer (more descriptive) name in `message` using e.g. `getKindName(kind) {}` function.

Btw. I think we should rename `kind` and call it for example `htmlElementType` or `nodeType`, what do you think? `kind` itself doesn't mean much, kind of what?

tests/lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,291 @@
...
+  <img>
+  <img/>
+  <x-test></x-test>
+  <x-test/>
+  <svg><path></path></svg>
+  <svg><path/></svg>
+  <math><mspace></mspace></math>
+  <math><mspace/></math>
+</template>`
+const anyWith = (opts) => Object.assign(
```
michalsnik: Please add empty line above

tests/lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    // other cases are in `invalid` tests.
+  ],
+  invalid: [
+    // default
+    {
+      code: '<template><div/></template>',
+      output: '<template><div></div></template>',
+      errors: ['Disallow self-closing on HTML elements.']
```
michalsnik: What do you think about adding element name in the error?
```
Disallow self-closing on HTML elements (div).
```

tests/lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,291 @@
...
+  <x-test></x-test>
+  <x-test/>
+  <svg><path/></svg>
+  <svg><path/></svg>
+  <math><mspace></mspace></math>
+  <math><mspace/></math>
+</template>`,
+      options: [anyWith({ svg: 'always' })],
+      errors: [
+        { message: 'Require self-closing on SVG elements.', line: 8 }
```
michalsnik: What about error on line 9?

tests/lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,291 @@
...
+      code: ALL_CODE,
+      output: `<template>
+  <div></div>
+  <div/>
+  <img>
+  <img/>
+  <x-test></x-test>
+  <x-test/>
+  <svg><path></path></svg>
+  <svg><path></path></svg>
```
michalsnik: Again line 8 and 9 are the same, is it intentional?

lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,179 @@
...
+})
+
+/**
+ * Normalize the given options.
+ * @param {Object|undefined} options The raw options object.
+ * @returns {Object} Normalized options.
+ */
+function parseOptions (options) {
+  return {
+    [KIND.NORMAL]: (options && options.html && options.html.normal) || 'never',
```
mysticatea: Or sindresorhus' [dot-prop](https://www.npmjs.com/package/dot-prop). However, I'm not sure if this is enough worth adding a new dependency. This may be not stylish, but I think enough readable.

lib/rules/html-self-closing-style.js
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Check whether the given element is empty or not.
+ * This ignores whitespaces, doesn't ignore comments.
+ * @param {VElement} node The element node to check.
+ * @returns {boolean} `true` if the element is empty.
+ */
+function isEmpty (node, sourceCode) {
+  const start = node.startTag.range[1]
+  const end = (node.endTag != null) ? node.endTag.range[0] : node.range[1]
```
mysticatea: I love `== null`, this is equivalent to `(x === null || typeof x === "undefined")` and readable.

lib/rules/html-self-closing-style.js
mysticatea: > I'd keep kind simple as: normal, void, component and so on, and only reference the longer (more descriptive) name in message using e.g. getKindName(kind) {} function.

Hmm, I'm not sure if it increases maintainability/readability.

> Btw. I think we should rename kind and call it for example htmlElementType or nodeType, what do you think? kind itself doesn't mean much, kind of what?

It's nice. My English skill is poor.

tests/lib/rules/html-self-closing-style.js
mysticatea: No. I guess you see `output` property which is the result of auto-fix.

tests/lib/rules/html-self-closing-style.js
mysticatea: Yes. It's the result of auto-fix.

mysticatea:
Thank you for review.

I updated this PR.

PR #131, opened 2017-08-04
Fix issue return in computed properties & render to match only corret nodes

### LINKED ISSUE #130: Issue in return-in-computed-property & require-render-return
I tried to test new rules in one of my big projects and i have few false positive errors in case when there is CodePath/Function inside with no return.

Affected rules:
- `return-in-computed-property` 
- `require-render-return`

Example of false positive error:
```js
export default {
  computed: {
    foo () {
      const options = []
      this.matches.forEach((match) => {
        options.push(match)
      })
      return options
    }
  }
}
```

armano2:
fixes #130

### DIFF
--- a/lib/rules/require-render-return.js
+++ b/lib/rules/require-render-return.js
@@ -20,16 +20,12 @@ function create (context) {
     utils.executeOnVue(context, obj => {
       const node = obj.properties.find(item => item.type === 'Property' &&
         utils.getStaticPropertyName(item) === 'render' &&
-        (item.value.type === 'ArrowFunctionExpression' || item.value.type === 'FunctionExpression') &&
-        !item.value.expression // render: () => test
+        (item.value.type === 'ArrowFunctionExpression' || item.value.type === 'FunctionExpression')
       )
       if (!node) return
 
       forbiddenNodes.forEach(el => {
-        if (
-          el.loc.start.line >= node.value.loc.start.line &&
-          el.loc.end.line <= node.value.loc.end.line
-        ) {
+        if (node.value === el) {
           context.report({
             node: node.key,
             message: 'Expected to return a value in render function.'
--- a/lib/rules/return-in-computed-property.js
+++ b/lib/rules/return-in-computed-property.js
@@ -25,11 +25,7 @@ function create (context) {
 
       computedProperties.forEach(cp => {
         forbiddenNodes.forEach(el => {
-          if (
-            cp.value &&
-            el.loc.start.line >= cp.value.loc.start.line &&
-            el.loc.end.line <= cp.value.loc.end.line
-          ) {
+          if (cp.value && cp.value.parent === el) {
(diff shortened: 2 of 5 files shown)


armano2:
ok this is still not good :(

lib/rules/require-render-return.js
```diff
@@ -28,7 +28,8 @@ function create (context) {
       forbiddenNodes.forEach(el => {
         if (
           el.loc.start.line >= node.value.loc.start.line &&
-          el.loc.end.line <= node.value.loc.end.line
+          el.loc.end.line <= node.value.loc.end.line &&
+          node.value === el
```
michalsnik: I think simple:
```
el.loc.start.line === node.value.loc.start.line
```
would be enough, or even just:
```
node.value === el
```
wdyt? (I was sceptic but looks like this kind of comparison works just fine for ASTs)

Plus I think that [this line](https://github.com/armano2/eslint-plugin-vue/blob/ffffa70df6f7bfa2967e048604e917f93c5f9ba2/lib/rules/require-render-return.js#L24)  should be moved to [extecuteOnFunctionsWithoutReturn](https://github.com/armano2/eslint-plugin-vue/blob/ffffa70df6f7bfa2967e048604e917f93c5f9ba2/lib/utils/index.js#L556):

```diff
- if (!isValidReturn()) {
+ if (!isValidReturn() && !node.expression) {
```

lib/rules/return-in-computed-property.js
```diff
@@ -28,7 +28,8 @@ function create (context) {
           if (
             cp.value &&
             el.loc.start.line >= cp.value.loc.start.line &&
-            el.loc.end.line <= cp.value.loc.end.line
+            el.loc.end.line <= cp.value.loc.end.line &&
+            cp.value.parent === el
```
michalsnik: Same here

lib/rules/require-render-return.js
armano2: @michalsnik thank you for tips 🍡

lib/rules/require-render-return.js
```diff
@@ -20,15 +20,15 @@ function create (context) {
...
+        (item.value.type === 'ArrowFunctionExpression' || item.value.type === 'FunctionExpression')
       )
       if (!node) return
 
       forbiddenNodes.forEach(el => {
         if (
           el.loc.start.line >= node.value.loc.start.line &&
-          el.loc.end.line <= node.value.loc.end.line
+          el.loc.end.line <= node.value.loc.end.line &&
+          el.loc.start.line === node.value.loc.start.line
```
michalsnik: I meat we can remove all other conditions and leave only this one. Or even just leave:
```
node.value === el
```
:D up to you

armano2: yup, i need a bit more ☕️

PR #132, opened 2017-08-05
Update: deprecate `no-invalid-*` rules and add `valid-*` rules (fixes #103)

### LINKED ISSUE #103: Rename double negations.
I have made a ton of rules which have double negation in their name, e.g. `vue/no-invalid-v-for`.
However, double negations are a bad choice.

Do we have the chance that we rename those rules before the official release?
For example:

- `vue/no-invalid-v-for` → `vue/validate-v-for` or something like.

Though, I'm not sure that it's valuable even if compatibility confusion happens... :thinking:

mysticatea:
Fixes #103.

This PR renames `no-invalid-*` rules to avoid double negation.
New naming is `valid-*` similar to some core rules; `valid-jsdoc` and `valid-typeof`.

This PR deprecates `no-invalid-*` rules and doesn't remove those. So this is not a breaking change. We can remove those in future.

Additionally, this PR creates the table of deprecated rules and removes deprecated rules from rule tables.


mysticatea:
For reference, new rules are added by the script below:

<details>

```js
const fs = require('fs')
const path = require('path')

for (const root of ['docs/rules', 'lib/rules', 'tests/lib/rules']) {
  for (const filename of fs.readdirSync(root)) {
    if (!filename.startsWith('no-invalid-')) {
      continue
    }
    const src = path.join(root, filename)
    const dst = path.join(root, 'valid-' + filename.slice('no-invalid-'.length))
    const content = fs.readFileSync(src, 'utf8')
      .replace(/no-invalid-/g, 'valid-')
      .replace(/disallow invalid/g, 'enforce valid')
      .replace(/Disallow invalid/g, 'Enforce valid')

    fs.writeFileSync(dst, content)
  }
}
```

</details>

tools/update-rules.js
```diff
@@ -50,18 +50,38 @@ const rulesTableContent = categories.map(category => `
...
       return `| ${mark} | ${link} | ${description} |`
     })
     .join('\n')
 }
-`).join('\n')
+`).join('\n') + `
+### Deprecated
+
+> - :warning: We are going to remove deprecated rules in future. Please migrate to its replaced rules.
+> - :innocent: We don't fix bugs which are in deprecated rules since we don't have enough resource.
```
michalsnik: `resources`

tools/update-rules.js
```diff
@@ -50,18 +50,38 @@ const rulesTableContent = categories.map(category => `
...
+      const description = meta.docs.description || '(no description)'
       return `| ${mark} | ${link} | ${description} |`
     })
     .join('\n')
 }
-`).join('\n')
+`).join('\n') + `
+### Deprecated
+
+> - :warning: We are going to remove deprecated rules in future. Please migrate to its replaced rules.
```
michalsnik: `We're going to remove deprecated rules in the next major release. Please migrate to successor/new rules.`

mysticatea:
Thank you for review!

I updated this PR.

PR #135, opened 2017-08-06
Add "require-default-prop" rule (fixes #122)

### LINKED ISSUE #122: Rule Proposal: `vue/require-default-prop`

This rule aims to ensure that any non-required `prop` declaration of a component has a corresponding `default` value.

`[ ]` Enforces code style
`[ ]` Other (please specify:)

Provide 2-3 code examples that this rule will warn about:

```js
export default {
  props: {
     foo: {
         type: String
     }
  }
}
```

```js
export default {
  props: {
     foo: {
         type: Array,
         required: false
     }
  }
}
```

based on conversation from: https://github.com/vuejs/eslint-plugin-vue/issues/117#issuecomment-319317832

michalsnik:
This PR implements `require-default-prop` rule, that was proposed in #122.

cc @armano2

### DIFF
--- a/lib/rules/require-default-prop.js
+++ b/lib/rules/require-default-prop.js
@@ -1,0 +1,98 @@
+/**
+ * @fileoverview Require default value for props
+ * @author Michał Sajnóg <msajnog93@gmail.com> (http://github.com/michalsnik)
+ */
+'use strict'
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Require default value for props',
+      category: 'Best Practices',
+      recommended: false
+    },
+    fixable: null,  // or "code" or "whitespace"
+    schema: []
+  },
+
+  create: function (context) {
+    // ----------------------------------------------------------------------
+    // Helpers
+    // ----------------------------------------------------------------------
+
+    /**
+     * Checks if the passed prop is required
+     * @param {Property} prop - Property AST node for a single prop
+     * @return {boolean}
+     */
+    function propIsRequired (prop) {
+      const propRequiredNode = prop.value.properties
+        .find(p =>
+          p.key.name === 'required' &&
+          p.value.type === 'Literal' &&
+          p.value.value === true
+        )
+
+      return Boolean(propRequiredNode)
+    }
+
+    /**
+     * Checks if the passed prop has a defualt value
+     * @param {Property} prop - Property AST node for a single prop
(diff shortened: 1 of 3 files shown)


lib/rules/require-default-prop.js
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Finds all props that don't have a default value set
+     * @param {Property} propsNode - Vue component's "props" node
+     * @return {boolean}
+     */
+    function findPropsWithoutDefaultValue (propsNode) {
+      return propsNode.value.properties
+        .filter(prop => {
+          if (prop.value.type !== 'ObjectExpression') return true
```
armano2: In case of spread operator `value` is going to be undefined:
`TypeError: Cannot read property 'type' of undefined`
```js
export default {
  props: {
    a: {
      ...test
    }
  }
}
```

lib/rules/require-default-prop.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Checks if the passed prop has a defualt value
+     * @param {Property} prop - Property AST node for a single prop
+     * @return {boolean}
+     */
+    function propHasDefault (prop) {
+      const propDefaultNode = prop.value.properties
+        .find(p => p.key.name === 'default')
```
armano2: In case of spread operator `key` is going to be undefined:
`TypeError: Cannot read property 'name' of undefined`

```js
export default {
  props: {
    ...test
  }
}
```

lib/rules/require-default-prop.js
```diff
@@ -0,0 +1,98 @@
...
+
+    return utils.executeOnVue(context, (obj) => {
+      const propsNode = obj.properties
+        .find(p =>
+          p.type === 'Property' &&
+          p.key.type === 'Identifier' &&
+          p.key.name === 'props'
+        )
+
+      if (!propsNode || propsNode.value.type !== 'ObjectExpression') return
```
armano2: you can move `propsNode.value.type !== 'ObjectExpression'` into `find`

armano2:
@michalsnik there is few issues when someone is going to use spread operator in props

lib/rules/require-default-prop.js
michalsnik: Ah, I totally forgot about this case.. Thanks for the heads up @armano2 :)

michalsnik:
@armano2 thanks for the review! Can you please check once again?

Btw. regarding:
> I was thinking how we should treat type: null, it means any type, than undefined is valid for this case.

I don't check `type` of any property in this rule. Only whether it is required and if has the default value set.

PR #138, opened 2017-08-06
Add "no-multi-spaces" rule (fixes #133)

### LINKED ISSUE #133: Rule Proposal: `vue/no-extra-whitespaces`

This rule warns about the usage of useless spaces.

`[x]` Enforces code style
`[ ]` Warns about a potential error
`[ ]` Other (please specify:)

```html
<template>
  <div class="wrapper">
    <div class="example1" ></div>
    <div class="example2"  @click="handleClick">
      <!-- Notice there are two spaces between `class` and `@click` -->
    </div>
    <div class="example3 ">
      <!-- Is it possible to catch extra spaces in attribute value? -->
    </div>
  </div>
</template>
```

However, the following should be considered fine:

```html
<template>
  <div class="wrapper">
    <self-closing />
    <div 
      class="multiple-line"
      @click="handleClick"
      >
    </div>
  </div>
</template>
```

The suggested rule name is likely not good enough.

armano2:
This PR implement rule proposed in #133

### DIFF
--- a/lib/rules/no-multi-spaces.js
+++ b/lib/rules/no-multi-spaces.js
@@ -24,15 +24,6 @@ module.exports = {
    * @returns {Object} AST event handlers.
    */
   create (context) {
-    function formatValue (token) {
-      switch (token.type) {
-        case 'HTMLSelfClosingTagClose': return '/>'
-        case 'HTMLTagClose': return '>'
-      }
-
-      return token.value
-    }
-
     // ----------------------------------------------------------------------
     // Public
     // ----------------------------------------------------------------------
@@ -46,6 +37,7 @@ module.exports = {
           })
           return
         }
+        const sourceCode = context.getSourceCode()
         const tokenStore = context.parserServices.getTemplateBodyTokenStore()
         const tokens = tokenStore.getTokens(node.templateBody, { includeComments: true })
 
@@ -62,7 +54,7 @@ module.exports = {
               message: "Multiple spaces found before '{{displayValue}}'.",
               fix: (fixer) => fixer.replaceTextRange([prevToken.range[1], token.range[0]], ' '),
               data: {
-                displayValue: formatValue(token)
+                displayValue: sourceCode.getText(token)
               }
             })
           }
--- a/lib/rules/no-multi-spaces.js
+++ b/lib/rules/no-multi-spaces.js
@@ -1,0 +1,74 @@
+/**
+ * @fileoverview This rule warns about the usage of extra whitespaces between attributes
+ * @author Armano
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'This rule warns about the usage of extra whitespaces between attributes',
+      category: 'Stylistic Issues',
+      recommended: false
+    },
+    fixable: 'whitespace',  // or "code" or "whitespace"
+    schema: []
+  },
+
+  /**
+   * @param {RuleContext} context - The rule context.
+   * @returns {Object} AST event handlers.
+   */
+  create (context) {
+    function formatValue (token) {
+      switch (token.type) {
+        case 'HTMLSelfClosingTagClose': return '/>'
+        case 'HTMLTagClose': return '>'
+      }
+
+      return token.value
+    }
+
+    // ----------------------------------------------------------------------
+    // Public
+    // ----------------------------------------------------------------------
+
+    return {
+      Program (node) {
+        if (context.parserServices.getTemplateBodyTokenStore == null) {
+          context.report({
+            loc: { line: 1, column: 0 },
+            message: 'Use the latest vue-eslint-parser. See also https://github.com/vuejs/eslint-plugin-vue#what-is-the-use-the-latest-vue-eslint-parser-error.'
+          })
+          return
+        }
+        const tokenStore = context.parserServices.getTemplateBodyTokenStore()
+        const tokens = tokenStore.getTokens(node.templateBody, { includeComments: true })
+
+        let prevToken = tokens.shift()
+        for (const token of tokens) {
+          const spaces = token.range[0] - prevToken.range[1]
+          if (spaces > 1 && token.loc.start.line === prevToken.loc.start.line) {
+            context.report({
+              node: token,
+              loc: {
+                start: prevToken.loc.end,
+                end: token.loc.start
+              },
+              message: "Multiple spaces found before '{{displayValue}}'.",
+              fix: (fixer) => fixer.replaceTextRange([prevToken.range[1], token.range[0]], ' '),
+              data: {
+                displayValue: formatValue(token)
+              }
+            })
+          }
+          prevToken = token
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


armano2:
@mysticatea is there a way that we can have access to `VAttributeText` or something similar with whitespaces between attributes?

for now i had to parse entire `VStartTag` to remove unnecessary spaces.

mysticatea:
I think that this rule should be token-based.

```js
const ALL = {range: [0, Number.POSITIVE_INFINITY]}

return {
  Program() {
    // TODO: Check `context.parserServices.getTemplateBodyTokenStore` exists or not.
    const tokenStore = context.parserServices.getTemplateBodyTokenStore()
    const tokens = tokenStore.getTokens(ALL, {includeComments: true})

    let prevToken = tokens.shift()
    for (const token of tokens) {
      const onSameLine = (prevToken.loc.end.line === token.loc.start.line)
      const spaces = token.loc.start.column - prevToken.loc.end.column
      if (onSameLine && spaces >= 2) {
        // Report.
      }
    }
  }
}
```

mysticatea:
Spaces which don't change semantic are on outside of tokens, so we can check only extra spaces in this way.

armano2:
@mysticatea thank you for suggestion, i'm still learning how to work with eslint rules  :dango:

armano2:
@mysticatea i took your sugestion and i applied it to my code also i added basic doc for it.

i think we should change error message: `Extra whitespace detected.` seems bad :sake:

armano2:
i consider `<tag div=""/>` and `<tag div="" />` as valid

mysticatea:
@armano2 I think better if this rule focuses on only the sequences of multiple spaces. In core rule case, [no-multi-spaces](http://eslint.org/docs/rules/no-multi-spaces) is doing that, and for example, [space-before-function-parentheses](http://eslint.org/docs/rules/space-before-function-parentheses) focuses on one or zero spaces which are followed by the parenthesis of function parameters.

lib/rules/html-no-self-closing.js
```diff
@@ -28,6 +28,7 @@ function create (context) {
         return
       }
 
+      // TODO: Check `context.parserServices.getTemplateBodyTokenStore` exists or not.
```
mysticatea: This is inside of `registerTemplateBodyVisitor`, so `context.parserServices.getTemplateBodyTokenStore` should always exist.

lib/rules/no-multi-spaces.js
```diff
@@ -0,0 +1,76 @@
...
+
+    return {
+      Program (node) {
+        // TODO: Check `context.parserServices.getTemplateBodyTokenStore` exists or not.
+        const tokenStore = context.parserServices.getTemplateBodyTokenStore()
+        const tokens = tokenStore.getTokens(node.templateBody, { includeComments: true })
+
+        let prevToken = tokens.shift()
+        for (const token of tokens) {
+          if (sourceCode.isSpaceBetweenTokens(prevToken, token)) {
```
mysticatea: This method cannot be used here because this method ignores `/* */` style comments which are not comments in HTML context.

tests/lib/rules/no-multi-spaces.js
```diff
@@ -0,0 +1,136 @@
...
+    {
+      code: '<template><div class="foo "   class="    foo    " /></template>',
+      output: '<template><div class="foo " class="    foo    " /></template>',
+      errors: [
+        {
+          message: 'Extra whitespace detected.',
+          type: 'HTMLIdentifier'
+        }
+      ]
+    }
```
mysticatea: I'd like to see tests for the expression of Vue.js directives and mustaches. Those are parsed by script parser (`espree` by default), so those have valid tokens. This rule can report on multiple spaces which are in the expressions.

mysticatea:
> i think we should change error message: Extra whitespace detected. seems bad :sake:

For reference, [this](https://github.com/eslint/eslint/blob/c93a853160472918976bb5cf5914f6ec32df3d60/lib/rules/no-multi-spaces.js#L121) is the message of core `no-multi-spaces` rule.

mysticatea:
@armano2 Please use `vue-eslint-parser@v2.0.0-beta.7` since I found a bug about token's range (https://github.com/mysticatea/vue-eslint-parser/commit/d670ac99a66e8f5634cdbf991f97f89a9c626ee8). That version fixed a bug that `=` tokens include spaces which are preceded by the `=`.

lib/rules/no-multi-spaces.js
armano2: ok

lib/rules/no-multi-spaces.js
```diff
@@ -0,0 +1,76 @@
...
+                    line: prevToken.loc.end.line,
+                    column: prevToken.loc.end.column + requiredSpaces
+                  },
+                  end: {
+                    line: prevToken.loc.end.line,
+                    column: prevToken.loc.end.column + spaces
+                  }
+                },
+                message: 'Extra whitespace detected.',
+                fix: (fixer) => fixer.removeRange([prevToken.range[1] + requiredSpaces, prevToken.range[1] + spaces])
```
armano2: in this case:
```
fix: (fixer) => fixer.replaceTextRange([prevToken.range[1], prevToken.range[1] + spaces], ' '),
```

is going to be valid cus there can be \n\r symbold before space

armano2:
@mysticatea thank you for suggestions, i applied most of them except that i changed a little fixer:

```js
{
  fix: (fixer) => fixer.replaceTextRange([prevToken.range[1], prevToken.range[1] + spaces], requiredSpaces ? ' ' : '')
}
```

mysticatea:
In core, `no-multi-spaces` doesn't warn trailing spaces. It's covered by [no-trailing-spaces](http://eslint.org/docs/rules/no-trailing-spaces). I think it makes sense.

armano2:
@mysticatea than its ok now?

mysticatea:
I think better if this rule focuses on only the sequences of multiple spaces between tokens. Could you remove the following things?:

- The special handling of spaces which are before `>`. This should be covered by an independent rule with always/never options.
- Trailing spaces. This is covered by no-trailing-spaces.

armano2:
@mysticatea ok changes applied

armano2:
also idk if this is bug in parser but `HTMLSelfClosingTagClose` and `HTMLTagClose` has no value.

mysticatea:
LGTM, thank you for the great work, @armano2 !

> also idk if this is bug in parser but HTMLSelfClosingTagClose and HTMLTagClose has no value.

It's intentional. Those kinds of tokens don't have variation, so their `value` property is empty.

lib/rules/no-multi-spaces.js
```diff
@@ -0,0 +1,74 @@
...
+   * @returns {Object} AST event handlers.
+   */
+  create (context) {
+    function formatValue (token) {
+      switch (token.type) {
+        case 'HTMLSelfClosingTagClose': return '/>'
+        case 'HTMLTagClose': return '>'
+      }
+
+      return token.value
```
mysticatea: Ah, also the `value` property of `HTMLAssociation` tokens is empty.
I think `context.getSourceCode().getText(token)` is good for this purpose.

armano2: thank you i learned few new useful thinks while building this rule

armano2:
@mysticatea requested changes applied

PR #142, opened 2017-08-09
Docs: improve document's header

mysticatea:
This PR improves the header of rule documents.

`npm run update` generates the headers from rule's meta data.


mysticatea:
I added internal rules which are mostly same as [eslint/lib/internal-rules](https://github.com/eslint/eslint/tree/77bcee412bbc40b5be810f974db48e0d9964a7bc/lib/internal-rules) to be consistent of rule descriptions.

lib/rules/jsx-uses-vars.js
```diff
@@ -37,7 +37,7 @@ SOFTWARE.
 module.exports = {
   meta: {
     docs: {
-      description: 'Prevent variables used in JSX to be marked as unused',
+      description: 'prevent variables used in JSX to be marked as unused.', // eslint-disable-line consistent-docs-description
```
michalsnik: Why do you disable `consistent-docs-description` rule on this line? Can't you remove the dot?

lib/rules/no-duplicate-attributes.js
```diff
@@ -90,7 +90,7 @@ module.exports = {
   create,
   meta: {
     docs: {
-      description: 'disallow duplicate attributes.',
+      description: 'disallow duplicate attributes',
```
michalsnik: To be consistent we should call it: `disallow duplication of attributes`

lib/rules/no-reservered-keys.js
```diff
@@ -49,7 +49,7 @@ function create (context) {
 module.exports = {
   meta: {
     docs: {
-      description: 'Prevent overwrite reserved keys.',
+      description: 'disallow overwrite reserved keys',
```
michalsnik: `disallow overwriting reserved keys`

tools/update-rules.js
```diff
@@ -110,3 +110,21 @@ fs.writeFileSync(
   recommendedRulesFile,
   recommendedRulesContent
 )
+
+// Update the header of rule documents.
+const headerPattern = /^#[^\n]*\n+(?:- .+\n)*\n*/
```
michalsnik: I'd add a block comment here with example .md that matches this pattern.

lib/rules/jsx-uses-vars.js
mysticatea: Because it does not fit to descriptions which start with one of `enforce`, `require` and `disallow`. The `consistent-docs-description` rule which is cloned from `eslint` repo enforces the first word of the description.

michalsnik:
Hey @mysticatea would you mind to update this PR? Then we could push this forward, should be quick to resolve my suggestions :)

mysticatea:
Oops, I'm sorry, I have lost this from my brain. I will do.

PR #143, opened 2017-08-09
Fix small mistake in `require-valid-default-prop`

armano2:
- Improve coverage for FunctionExpression and ArrowFunctionExpression
- Fix traversing over many types (return -> continue)


armano2:
@michalsnik i tested this rule in few big projects and i found few mistakes what i made there,

- If there is more than one prop, traversing over props is stopped if type is valid
- Function and variables are throwing errors

michalsnik:
N1, good catch :)

PR #145, opened 2017-08-10
New: `vue/html-indent` rule (fixes #46)

### LINKED ISSUE #46: Rule proposition: Template Indentation
It would be great to have the ability of having only one style for the template/render function HTML, by enforcing an indentation style.

It could be configured by setting the option to `tab` for tabs or a positive number for spaces.

**Proposal:**
1. For tabs `'vue/indent': 'tab'`
```
<application>
	<component />
</application>
```

2. For 2 spaces: `'vue/indent': 2`
```
<application>
  <component />
</application>
```

2. No indentation: `'vue/indent': 0`
```
<application>
<component />
</application>
```

mysticatea:
Fixes #46.

This PR adds `vue/html-indent` rule.

The strategy of the `vue/html-indent` rule is the same as core `indent` rule. I.e., it makes the base token and offset for every token in the first traversing, then at leaving the root element, it calculate and validate expected indentation for each line.

Currently, about inside of directives and mustaches, this `vue/html-indent` rule supports only the syntax ECMAScript 2017 standard includes. It just ignores other syntaxes. The indentation rule is [one of most complex rules](https://github.com/eslint/eslint/blob/bd09cd50283760b57949802c55a6e2b5862c6d15/lib/rules/indent.js). I think this is a good start point.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-    "editor.tabSize": 2
+    "editor.tabSize": 2,
+    "javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,3 @@
 {
-    "editor.tabSize": 2,
-    "javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": true
+    "editor.tabSize": 2
 }
--- a/docs/rules/html-indent.md
+++ b/docs/rules/html-indent.md
@@ -1,0 +1,115 @@
+# enforce consistent indentation in `<template>` (vue/html-indent)
+
+## :book: Rule Details
+
+This rule enforces a consistent indentation style in `<template>`. The default style is 4 spaces as same as [the core indent rule](http://eslint.org/docs/rules/indent).
+
+- This rule checks all tags, also all expressions in directives and mustaches.
+- In the expressions, this rule supports ECMAScript 2017 syntaxes. It ignores unknown AST nodes, but it might be confused by non-standard syntaxes.
+
+:-1: Examples of **incorrect** code for this rule:
+
+```html
+<template>
+ <div class="foo">
+   Hello.
+    </div>
+</template>
+```
+
+:+1: Examples of **correct** code for this rule:
+
+```html
+<template>
+    <div class="foo">
+        Hello.
+    </div>
+</template>
+```
+
+```html
+<template>
+    <div class="foo">
+        Hello.
+    </div>
+    <div
+        id="a"
+        class="b"
+        :other-attr="{
+            aaa: 1,
+            bbb: 2
+        }"
+        @other-attr2="
+            foo();
+            bar();
+        "
+    >
+        {{
+            displayMessage
+        }}
+    </div>
+</template>
+```
+
+## :wrench: Options
+
+```json
+{
+    "vue/html-indent": ["error", type, {
+        "attribute": 1,
+        "closeBracket": 0,
+        "switchCase": 0,
+        "ignores": []
+    }]
+}
+```
+
+- `type` (`number | "tab"`) ... The type of indentation. Default is `4`. If this is a number, it's the number of spaces for one indent. If this is `"tab"`, it uses one tab for one indent.
+- `attribute` (`integer`) ... The multiplier of indentation for attributes. Default is `1`.
+- `closeBracket` (`integer`) ... The multiplier of indentation for right brackets. Default is `0`.
+- `ignores` (`string[]`) ... The selector to ignore nodes. The AST spec is [here](https://github.com/mysticatea/vue-eslint-parser/blob/master/docs/ast.md). You can use [esquery](https://github.com/estools/esquery#readme) to select nodes. Default is an empty array.
+
+:+1: Examples of **correct** code for `{attribute: 1, closeBracket: 1}`:
+
+```html
+<template>
+    <div
+        id="a"
+        class="b"
+        other-attr=
+            "{longname: longvalue}"
+        other-attr2
+            ="{longname: longvalue}"
+        >
+        Text
+    </div>
+</template>
+```
+
+:+1: Examples of **correct** code for `{attribute: 2, closeBracket: 1}`:
+
+```html
+<template>
+    <div
+            id="a"
+            class="b"
+            other-attr=
+                "{longname: longvalue}"
+            other-attr2
+                ="{longname: longvalue}"
+        >
+        Text
+    </div>
+</template>
+```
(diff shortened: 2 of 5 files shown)


mysticatea:
Thank you for your patience. I think I have done. Please review and suggest lacking tests!

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-    "editor.tabSize": 2
+    "editor.tabSize": 2,
```
Aladdin-ADD: seems not related to the `vue/html-indent`. I think it can be in another PR?

package.json
```diff
@@ -10,7 +10,8 @@
     "lint": "eslint .",
     "pretest": "npm run lint",
     "preversion": "npm run update && npm test",
-    "update": "node ./tools/update-rules.js"
+    "update": "node ./tools/update-rules.js",
+    "watch": "warun lib/rules/html-indent.js tests/lib/rules/html-indent.js -- nyc --reporter lcov -- mocha tests/lib/rules/html-indent.js"
```
Aladdin-ADD: what is the `warun`? I didn't see it in deps.

lib/rules/html-indent.js
```diff
@@ -0,0 +1,1228 @@
...
+
+  /**
+   * Get the first and last tokens of the given node.
+   * If the node is parenthesized, this gets the outermost parentheses.
+   * @param {Node} node The node to get.
+   * @param {number} [borderOffset] The least offset of the first token. Defailt is 0. This value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
+   * @returns {{firstToken:Token,lastToken:Token}} The gotten tokens.
+   */
+  function getFirstAndLastTokens (node, borderOffset) {
+    borderOffset |= 0
```
Aladdin-ADD: could you explain a little? :)

armano2: its js `magic` assign 0 if value is non numeric

Aladdin-ADD: hah.. got it! amazing! lol~

package.json
mysticatea: Oops, it was added unintentionally. I was using it to test.
Thank you.

lib/rules/html-indent.js
mysticatea: Yes, it is "casting to signed integer".

Bitwise operators (`&`, `|`, "^", `~`, `<<`, `>>`, and `>>>`) casts both operands to 32bits signed integer, then the logical OR with `0` does nothing. As the result, the `x | 0` expression casts `x` to 32bits signed integer.

armano2: @mysticatea like i said `magic`

tests/lib/rules/html-indent.js
```diff
@@ -0,0 +1,4214 @@
...
+      </template>
+    `,
+
+    // VStartTag
+    unIndent`
+      <template>
+          <div
+              aaa
+              bbb
+    `,
```
Aladdin-ADD: not a valid template, is is intentional?

and why the parser not reporting an error?

docs/rules/html-indent.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+## :wrench: Options
+
+```json
+{
+    "vue/html-indent": ["error", type, {
+        "attribute": 1,
+        "closeBracket": 0,
+        "switchCase": 0,
```
Aladdin-ADD: could you add an example for the `switchCase` option?

tests/lib/rules/html-indent.js
```diff
@@ -0,0 +1,4214 @@
...
+              :other-attr="{
+                  aaa: 1,
+                  bbb: 2
+              }"
+              @other-attr2="
+                  foo();
+                  bar();
+              "
+          >
+              {{
```
Aladdin-ADD: maybe we need some more tests for `{{}}`.

and I didn't see tests for comma, e.g:
```html
<template>
{{
  foo,
  bar,
  qux
}}
</template>
```

lib/rules/html-indent.js
```diff
@@ -0,0 +1,1228 @@
...
+  function validate (tokens, comments) {
+    // Calculate and save expected indentation.
+    const firstToken = tokens[0]
+    const actualIndent = firstToken.loc.start.column
+    const expectedIndent = getExpectedIndent(tokens)
+    if (expectedIndent === Number.MAX_SAFE_INTEGER) {
+      return
+    }
+
+    // Debug log
```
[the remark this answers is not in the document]
mysticatea: Yeah!
But I'd like to leave it to investigate for bug reports, since the debug code is long...

tests/lib/rules/html-indent.js
mysticatea: Yes, it's intentional. The HTML spec includes recovering process for invalid syntax, so `vue-eslint-parser` reports syntax errors with the `programNode.templateBody.errors` property instead of throwing the error. Then [vue/no-parsing-error](https://github.com/vuejs/eslint-plugin-vue/blob/94f387ed68149f8d12c1960d68cb5d9c376b9a89/docs/rules/no-parsing-error.md) rule would report those syntax errors.

The purpose of this test case is to check whether `vue/html-indent` rule works properly on an incomplete code. Especially, the expected indentation of closing brackets (`>`) does not apply to the last attribute even if the closing bracket is lacking.

docs/rules/html-indent.md
mysticatea: Or maybe I should remove the option from this rule since `switch` is [a prohibited keyword](https://github.com/vuejs/vue/blob/c0da43d22f8f4b9aeb4f49d4e86cd9704daaff3f/src/compiler/error-detector.js#L7-L11).

tests/lib/rules/html-indent.js
mysticatea: In technically, mustaches (`{{foo}}`) and directive values (`"foo"` of directives) are completely same for indentation. Both are `VExpressionContainer`. So I don't think we need many tests for mustaches. I added some tests.

lib/rules/html-indent.js
```diff
@@ -0,0 +1,1228 @@
...
+        const node = nodeList[i]
+        if (node == null) {
+          // Holes of an array.
+          continue
+        }
+        const elementTokens = getFirstAndLastTokens(node, lastToken != null ? lastToken.range[1] : 0)
+
+        // Collect related tokens.
+        // Commas between this and the previous, and the first token of this node.
+        if (lastToken != null) {
```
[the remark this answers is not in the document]
mysticatea: I apology that I'm inactive during a long time.

It, ensuring boolean type for conditional expression, is just my favorite style :smile:

PR #156, opened 2017-08-14
Fix 'no-shared-component-data' false positive when using arrow functions

### LINKED ISSUE #154: 'no-shared-component-data' false positive when using arrow functions

* **ESLint Version:** 3.19.0
* **eslint-plugin-vue Version:** 3.11.0
* **Node Version:** 7.8.0

```
{
	root: true,
	extends: ['plugin:vue/recommended'],
	rules: {
		'vue/html-quotes': 'error',
		'vue/html-self-closing': ['error', {
			html: {
				normal: 'never',
				void: 'any',
				component: 'never'
			},
			svg: 'never',
			math: 'never'
		}],
		'vue/no-async-in-computed-properties': 'error',
		'vue/no-dupe-keys': 'error',
		'vue/no-duplicate-attributes': 'error',
		'vue/no-shared-component-data': 'error',
		'vue/order-in-components': 'error',
		'vue/v-bind-style': 'error',
		'vue/v-on-style': 'error'
	}
}
```

```vue
<script>
	export default {
		data: () => ({
			foo: 'bar'
		})
	};
</script>
```

No linting errors

Error: `data` property in component must be a function

armano2:
fixes #154

### DIFF
--- a/tests/lib/rules/no-shared-component-data.js
+++ b/tests/lib/rules/no-shared-component-data.js
@@ -12,6 +12,12 @@ const rule = require('../../../lib/rules/no-shared-component-data')
 
 const RuleTester = require('eslint').RuleTester
 
+const parserOptions = {
+  ecmaVersion: 7,
+  sourceType: 'module',
+  ecmaFeatures: { experimentalObjectRestSpread: true }
+}
+
 // ------------------------------------------------------------------------------
 // Tests
 // ------------------------------------------------------------------------------
@@ -30,7 +36,8 @@ ruleTester.run('no-shared-component-data', rule, {
             }
           }
         })
-      `
+      `,
+      parserOptions
     },
     {
       filename: 'test.js',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-shared-component-data.js
```diff
@@ -102,7 +105,18 @@ ruleTester.run('no-shared-component-data', rule, {
...
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          data: () => {
+
+          }
+        }
+      `,
```
mysticatea: I suspect that this test case is not correct.
The filename is `.vue` and the code is in outside of `<script>` element.

michalsnik: This is sort of correct @mysticatea I think. This plugin doesn't transform `<script></script>` from a `vue` file. It's part of `eslint-vue-parser` responsibilities so here we expect that any code inside `script` tag will be accessible as usual. That's why we write this code directly without extra `script` tag. All we need to know in our rules is the extension of the file to detect if simple `export default` should be treated as component.

mysticatea: Ah, I was under the impression that `parserOptions` has `parser: "vue-eslint-parser"`. I'm sorry.

PR #158, opened 2017-08-14
⭐️New: Add rule `no-template-shadow`.

### LINKED ISSUE #101: Rule Proposal: `vue/no-shadow`

`no-shadow` should report variable definitions of `v-for` directives or `scope` attributes if those shadows the variables in parent scopes.

[ ] Enforces code style
[X] Warns about a potential error
[ ] Other (please specify:)

## Provide 2-3 code examples that this rule will warn about:

```html
<template>
    <ol v-for="i in 5">
        <ol v-for="i in 5"><!-- "i" is already declared in the upper scope. -->
            <li>item</li>
        </ol>
    </ol>
</template>
<script>
export default {
}
</script>
```

armano2:
fixes #101


tests/lib/rules/no-template-shadow.js
```diff
@@ -0,0 +1,127 @@
...
+      </script>`,
+      errors: [{
+        message: "Variable 'i' is already declared in the upper scope.",
+        type: 'Identifier'
+      }, {
+        message: "Variable 'i' is already declared in the upper scope.",
+        type: 'Identifier'
+      }]
+    },
+    {
```
Aladdin-ADD: I like to see some tests like this:
```html
<template>
<ul v-for="i in 5">
  <li>
    <span v-for="j in 10">{{i}},{{j}}</span>
  </li>
</ul>
</template>
```

docs/rules/no-template-shadow.md
```diff
@@ -0,0 +1,46 @@
...
+
+`no-shadow` should report variable definitions of v-for directives or scope attributes if those shadows the variables in parent scopes.
+
+## :book: Rule Details
+
+This rule aims to eliminate shadowed variable declarations of v-for directives or scope attributes.
+
+:-1: Examples of **incorrect** code for this rule:
+
+```html
```
michalsnik: I think we should present two or three cases here. First - a simple template example, and second with shadowed `data` or `method` property declarations. e.g.:
1.
```
<template>
   <div>
     <div v-for="i in 5">
       <div v-for="i in 10"></div>
     </div>
   </div>
 </template>
```

2.
```
<template>
   <div>
     <div v-for="i in 5"></div>
   </div>
 </template>
<script>
  export default {
    data () {
      return {
        i: 10
      }
    }
  }
</script>
```

lib/rules/no-template-shadow.js
```diff
@@ -0,0 +1,76 @@
...
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+const GROUP_NAMES = ['props', 'computed', 'data', 'methods']
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow variable declarations from shadowing variables declared in the outer scope.',
+      category: 'Possible Errors',
```
michalsnik: I think it should belong to `essential` group

lib/utils/index.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
    *
    * @param {RuleContext} context The rule context to use parser services.
    * @param {Object} templateBodyVisitor The visitor to traverse the template body.
-   * @param {Object} scriptVisitor The visitor to traverse the script.
+   * @param {Object} [scriptVisitor] The visitor to traverse the script.
```
michalsnik: Why square brackets here?

tests/lib/rules/no-template-shadow.js
```diff
@@ -0,0 +1,128 @@
...
+    '',
+    '<template><div></div></template>',
+    '<template><div v-for="i in 5"></div></template>',
+    '<template><div v-for="i in 5"><div v-for="b in 5"></div></div></template>',
+    '<template><div v-for="i in 5"></div><div v-for="i in 5"></div></template>',
+    '<template> <ul v-for="i in 5"> <li> <span v-for="j in 10">{{i}},{{j}}</span> </li> </ul> </template>',
+    {
+      filename: 'test.vue',
+      code: `<template>
+        <div v-for="i in 5"></div>
```
michalsnik: Let's add case with `i in f`, wdyt?

tests/lib/rules/no-template-shadow.js
```diff
@@ -0,0 +1,128 @@
...
+      </template>
+      <script>
+        export default {
+          data: {
+            i: 7
+          }
+        }
+      </script>`,
+      errors: [{
+        message: "Variable 'i' is already declared in the upper scope.",
```
michalsnik: Please add line on which the following error is expected

docs/rules/no-template-shadow.md
```diff
@@ -0,0 +1,46 @@
+# Disallow variable declarations from shadowing variables declared in the outer scope. (no-template-shadow)
+
+`no-shadow` should report variable definitions of v-for directives or scope attributes if those shadows the variables in parent scopes.
```
michalsnik: `no-template-shadow`

lib/utils/index.js
michalsnik: It indicates an `optional param` http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

michalsnik:
We're still in v5 beta. Let's put it in `strongly-recommended` so we can get feedback faster, and don't need to wait for the next major release to ship it by default.

PR #163, opened 2017-08-16
Rename and add options to `this-in-template`.

### LINKED ISSUE #148: Disallow `this` in template
Disallow `this` in the template

[x] Enforces code style
[ ] Warns about a potential error
[ ] Other (please specify:)

```html
<template>
     <a :href="this.link">{{this.text}}</a>
</template>
```

```html
<template>
     <div :class="this.$style.container">
          <slot />
     </div>
</template>
```

armano2:
fixes #148

### DIFF
--- a/lib/rules/this-in-template.js
+++ b/lib/rules/this-in-template.js
@@ -30,7 +30,7 @@ module.exports = {
   },
 
   /**
-   * Creates AST event handlers for no-this-in-template.
+   * Creates AST event handlers for this-in-template.
    *
    * @param {RuleContext} context - The rule context.
    * @returns {Object} AST event handlers.
--- a/lib/rules/this-in-template.js
+++ b/lib/rules/this-in-template.js
@@ -42,10 +42,34 @@ module.exports = {
       nodes: []
     }
 
-    function validateNever () {
-      return {
-        'VExpressionContainer ThisExpression' (node) {
-          if (options === 'never') {
+    utils.registerTemplateBodyVisitor(context, Object.assign({
+      VElement (node) {
+        scope = {
+          parent: scope,
+          nodes: scope.nodes.slice() // make copy
+        }
+        if (node.variables) {
+          for (const variable of node.variables) {
+            const varNode = variable.id
+            const name = varNode.name
+            if (!scope.nodes.some(node => node.name === name)) { // Prevent adding duplicates
+              scope.nodes.push(varNode)
+            }
+          }
+        }
+      },
+      'VElement:exit' (node) {
+        scope = scope.parent
+      }
+    }, options === 'never'
+      ? {
+        'VExpressionContainer MemberExpression > ThisExpression' (node) {
+          const propertyName = utils.getStaticPropertyName(node.parent.property)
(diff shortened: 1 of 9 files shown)


lib/rules/this-in-template.js
```diff
@@ -0,0 +1,99 @@
...
+    const options = context.options[0] !== 'always' ? 'never' : 'always'
+    let scope = {
+      parent: null,
+      nodes: []
+    }
+
+    function validateNever () {
+      return {
+        'VExpressionContainer ThisExpression' (node) {
+          if (options === 'never') {
```
mysticatea: I think that it should check whether variables which have the same name as the property name.

```html
<template>
    <div v-for="x of xs">
        {{this.x}} <!-- we cannot remove this `this.`. -->
    </div>
</template>
```

Also, I think that it should check whether the `this` is `MemberExpression#object`.

```html
<template>
    <child :parent="this" /> <!-- we cannot remove this `this`. -->
</template>
```

mysticatea:
Incidentally, I have improved `vue-eslint-parser` as having the relationship between variables and references: https://github.com/mysticatea/vue-eslint-parser/commit/02579df15c7eeb7658c216036895be19818a7bcb.

- `Variable#references` is an array of references which use the variable. If this is empty, it's an unused variable.
- `Reference#variable` is the variable that the reference uses. If this is `null`, it's global variable or the member of the instance.

Though, I'm not sure that it helps this rule.

armano2:
@mysticatea i i like changes with `references` it can simplify this rule, but for now there is infinite loop there: 
https://github.com/mysticatea/vue-eslint-parser/issues/16

mysticatea:
Thank you!

I'm sorry, I have realized one more thing. If property name is a reserved word (e.g. `this.class`), we cannot remove the `this.` since it causes syntax errors. Could you add the check into `"never"`?

armano2:
@mysticatea check for js reserved names added

armano2:
additionality i added support for `this['0aaaa']` and `this['foo-bar bas']` witch should not be warned

armano2:
@mysticatea with latest change to parser, reference variables are no longer there

_Sent from my Xiaomi Redmi 4A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

mysticatea:
I'm sorry for late.

> additionality i added support for this['0aaaa'] and this['foo-bar bas'] which should not be warned

Good catch!

> with latest change to parser, reference variables are no longer there

I think, in fact, `console.log()` no longer shows `Variable#references` and `Reference#variable` since those are non-enumerable properties. But those are still there. Tests exist at [vue-eslint-parser/test/variables-references.js#L152-L248](https://github.com/mysticatea/vue-eslint-parser/blob/426a9e399e7cca7f8c6a73989af40b69095ff893/test/variables-references.js#L152-L248).

michalsnik:
GJ! I'll release it at the end of this weekend @mysticatea @armano2 . I'm moving to new place, so have quite a lot on my plate lately..

PR #166, opened 2017-08-24
Fix: incorrectly flags rule valid-template-root for pug

### LINKED ISSUE #165: Beta incorrectly flags rule valid-template-root for pug templates
Ubuntu 16.04 Gnome 3

* **ESLint Version:**
3.19.0 
* **eslint-plugin-vue Version:**
beta (v3.12.0)
* **Node Version:**
6.11.0  

```
// http://eslint.org/docs/user-guide/configuring

module.exports = {
  root: true,
  parserOptions: {
    parser: 'babel-eslint',
    sourceType: 'module'
  },
  env: {
    browser: true,
  },
  // https://github.com/feross/standard/blob/master/RULES.md#javascript-standard-style
  // https://github.com/vuejs/eslint-plugin-vue
  extends: [
    'standard',
    'plugin:vue/recommended'
  ],
  plugins: [
    'html'
  ],
  'settings': {
      'html/html-extensions': ['.html'],  // don't include .vue
  },
  // add your custom rules here
  'rules': {
    // allow paren-less arrow functions
    'arrow-parens': 0,
    // allow async-await
    'generator-star-spacing': 0,
    // allow debugger during development
    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0
  }
}

```

```vue

<template lang="pug">
.example
  | blah blah blah
</template>

<script>
export default {
  data () {
    return {

    }
  }
}
</script>

```

When I had beta v3.8.0, no linter errors occurred. **After beta v3.9.0 and above**, the linter error occurs in all pug templates: valid-template-root.

I get an **valid-template-root** error for all my pug templates. It seems the latest beta updates now think that the root element is text when it is actually valid pug markup.

armano2:
fixes #165

### DIFF
--- a/lib/rules/valid-template-root.js
+++ b/lib/rules/valid-template-root.js
@@ -31,6 +31,11 @@ function create (context) {
         return
       }
 
+      const lang = utils.getAttribute(element, 'lang')
+      if (lang && lang.value.value !== 'html') {
+        return // Ignore pug and ....
+      }
+
       const hasSrc = utils.hasAttribute(element, 'src')
       const rootElements = []
       let extraText = null
--- a/tests/lib/rules/valid-template-root.js
+++ b/tests/lib/rules/valid-template-root.js
@@ -33,6 +33,10 @@ tester.run('valid-template-root', rule, {
     },
     {
       filename: 'test.vue',
+      code: '<template lang="pug"></template>'
+    },
+    {
+      filename: 'test.vue',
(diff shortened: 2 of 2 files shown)


mysticatea:
Thank you for everything, @armano2 !

However, I think this is a bug of upstream, so we should fix `vue-eslint-parser`.
I have done in #180. Please check it.

armano2:
thank you

PR #172, opened 2017-09-01
Fix: don't use internal APIs of ESLint (fixes #171)

### LINKED ISSUE #171: Use only public APIs of ESLint
Link: https://github.com/eslint/eslint/issues/9193.

`vue-eslint-parser` has been depending on the private API to traverse `<template>`'s AST, so we have to fix it to work on ESLint v4.6.0.

mysticatea:
Fixes #171.

This PR makes `eslint-plugin-vue` not using internal APIs of ESLint.
This is helpful to prevent broken by internal changes in ESLint.

For this, I changed `parserServices.registerTemplateBodyVisitor` to `parserServices.defineTemplateBodyVisitor`.
The `parserServices.registerTemplateBodyVisitor` had been registering the visitor to ESLint with internal API.
New `parserServices.defineTemplateBodyVisitor` returns a normal visitor which has a handler to call the given visitor. Rules must return the normal visitor to register it to ESLint.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,12 +4,12 @@ machine:
 
 dependencies:
   pre:
-    - nvm install 7
+    - nvm install 6
     - nvm install 8
 
 test:
   override:
-    - npm test
+    - nvm use 4 && npm test
     - nvm use 6 && npm test
-    - nvm use 7 && npm test
     - nvm use 8 && npm test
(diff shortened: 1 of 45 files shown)


michalsnik:
Looks good, but will this be backwards compatible with previous eslint versions too?

mysticatea:
Yes. I added tests for an old ESLint version to Circle CI.

PR #200, opened 2017-10-05
Fix max-attributes-per-line example json configuration

maoberlehner:
The first value of the array is expected to be the severity.
Keys must be in double quotes.


michalsnik:
Thanks @maoberlehner !

Issue #77, opened 2017-07-14
Style Guide Collaboration


chrisvfritz:
I'm developing an official Vue style guide and although not all rules there will be enforceable through this plugin, I think it makes sense to collaborate here, especially for those rules that _are_ enforceable. Right now, I'm thinking the style guide could serve as the official documentation for any recommended rules in this plugin and these two resources could be published together.

You can see the current progress [in this branch](https://github.com/vuejs/vuejs.org/blob/style-guide/src/v2/guide/style-guide.md), with the [live result here](https://vue-style-guide-dev.surge.sh/v2/guide/style-guide.html). Any feedback/contributions are welcome and rules are not set in stone, so don't panic if you see something you disagree with. 😄  

I've also not kept up-to-date with the conversation in various issues here, so I apologize if some of it is out of sync with plans you've been making. Please ping me on discussions you think I should know about, or if you'd like my input.

armano2:
👍 I like it

mysticatea:
Good idea ❤️ 
I think good if we have predefined configs about the priority A to D.
Also, the style guide is the best source of rule ideas.

chrisvfritz:
@mysticea Agreed, though inspiration can also go both ways. If you have any rules from the plugin that should be added, or if a recommended rule in the plugin conflicts, please discuss it here.

Feedback on the categories I've created are also welcome - as well as feedback on which category each rule belongs in. I tried to put a lot of research and thought into them, but this is still the first community style guide I'm creating and I'm sure there are considerations I'm not experienced enough to foresee.

chrisvfritz:
@filipalacerda Also tagging you here, as I know you've thought about this quite a lot at Gitlab and your internal style guide was actually a source of inspiration for the official one. 🙂

sdras:
This is amazing! I love it. 

A couple of notes: (these are _so_ minor)

> Complex computed properties
> Whenever possible, computed properties should be split into as many named values as possible.

I think this one could do with an example.

> Component style scoping
> Component styles should be scoped.

Clarification needed: I know a good number of people who won't scope the topmost `App.vue` component tag, but every subsequent component thereafter, are we deciding that this is an antipattern too? If not, it might be worth detailing this a bit more.

> Self-closing components
> Components with no content should be self-closing in single-file component or string templates, but never in DOM templates.

Can you please explain the inconsistency here?

This is fantastic, I'm sure it's going to be a great resource for people.

[...]

PR #205, opened 2017-10-09
Style Guide Alignment 

chrisvfritz:
I believe this is [everything that's needed](https://github.com/vuejs/eslint-plugin-vue/issues/77#issuecomment-333954255) to get the plugin to match the style guide. As I worked, I also fixed any minor bugs or inconsistencies I noticed.


chrisvfritz:
I'd rather focus on the rules we have for now. I'm mostly holding off on adding new rules to the style guide too, while we refine what we currently have. And we'll be adding references later to note what has a corresponding style guide rule here - and on the style guide, what's enforceable by ESLint.

I figure it's better to have solid set of rules available than nothing, as we wait for every possible rule to be implemented. 🙂

.gitignore
```diff
@@ -3,3 +3,4 @@
 /tests/integrations/*/node_modules
 /node_modules
 /test.*
+yarn.lock
```
michalsnik: I wouldn't ignore yarn.lock, we might want to add it to repository

README.md
```diff
@@ -75,85 +75,107 @@ Vue.component('AsyncComponent', (resolve, reject) => {
 ## :gear: Configs
 
 This plugin provides two predefined configs:
-- `plugin:vue/base` - contains necessary settings for this plugin to work properly
-- `plugin:vue/recommended` - extends base config with recommended rules (the ones with check mark :white_check_mark: in the table below)
+- `plugin:vue/base` - Settings and rules to enable correct ESLint parsing
+- `plugin:vue/essential` - Above, plus rules to prevent errors or unintended behavior
```
michalsnik: Those configs seems very nice, but I'm not sure we should group them like this in readme. Current groups are common in eslint ecosystem (take a look at original rules https://eslint.org/docs/rules/). Plus later we will probably have some rules that we don't want to enable in any config and give the user full freedom in deciding whether he needs it or not. With the proposed granulation we would also have to add another category - which may lead to unnecessary churn. Also I'm not sure whether maintaining 4 configs is a way to go. I'd rather keep it  simple stupid and provide only `base` and `recommended` config. What do you think @chrisvfritz @mysticatea ?

docs/rules/name-property-casing.md
```diff
@@ -22,18 +22,10 @@ export default {
 }
 ```
 
-:+1: Examples of **correct** code for `camelCase`:
```
michalsnik: Why did you remove this? Recommending something is one thing, but we should still have the ability to configure rule as we see fit.

lib/config/all.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
michalsnik: what do we need this config for?

michalsnik:
Great work @chrisvfritz! 🚀  I have concerns though, regarding proposed categories and number of configs ☝️

.gitignore
chrisvfritz: I thought it'd be safer to add it to `.gitignore` until an explicit decision to include it had been made, at which time it could be easily removed. Do you disagree?

README.md
chrisvfritz: Those configs seems very nice, but I'm not sure we should group them like this in readme. Current groups are common in eslint ecosystem (take a look at original rules eslint.org/docs/rules). Plus later we will probably have some rules that we don't want to enable in any config and give the user full freedom in deciding whether he needs it or not. With the proposed granulation we would also have to add another category - which may lead to unnecessary churn. Also I'm not sure whether maintaining 4 configs is a way to go. I'd rather keep it simple stupid and provide only base and recommended config. What do you think @chrisvfritz @mysticatea ?

===

> Current groups are common in eslint ecosystem (take a look at original rules eslint.org/docs/rules).

I agree it makes sense to stick with ESLint conventions wherever there's not a compelling reason to stray, but I did see some compelling reasons here.

First, I'm thinking that most people will probably discover the ESLint plugin through the style guide, so I saw a strong benefit in not forcing users to context switch after arriving here. If they want to enforce priority A rules, they'll know within seconds exactly what to do. There will be no wondering, "Wait, is Possible Errors what I want to enforce Priority A rules?"

For users who have never seen the style guide, but are familiar with ESLint conventions, I don't think the new name will cause any confusion - but perhaps there's something I'm not foreseeing. What do you think?

Second, I think there's value in the increased granularity. "Stylistic Issues" is too broad in my opinion, which is why I split stylistic rules between Priority B and C. For Priority B, there's one or more possible conventions that offer clear benefits over the alternatives. For Priority C, there are really many good options and the primary value is just in picking something and sticking to it.

Finally, I moved `jsx-uses-vars` out of "Variables" because I just don't find that a useful category, despite ESLint's recommendation. Their rules in that category include a mix of possible errors (e.g. `no-use-before-define`) and stylistic issues (e.g. `init-declarations`) and this is the only language feature (apart from ES6) that receives special treatment. I also feel that `jsx-uses-vars` only _incidentally_ has to do with variables - it's really a parsing extension that allows ESLint to work correctly, which is why I added it to `base`.

> Plus later we will probably have some rules that we don't want to enable in any config and give the user full freedom in deciding whether he needs it or not.

I'm not sure I understand this one. Don't users have the same amount of freedom either way, since they can override whatever they want? In terms of ease of use, I can't think of a case where more nested rule sets would worsen the user experience or make it more difficult for them to do something. Am I missing something though?

> With the proposed granulation we would also have to add another category - which may lead to unnecessary churn.

I might need more context to understand this one, because adding a new category wouldn't be a lot more code nor increase its complexity, if I'm understanding correctly. Can you provide a specific example of what you're thinking might happen?

> Also I'm not sure whether maintaining 4 configs is a way to go. I'd rather keep it simple stupid and provide only base and recommended config.

This is another one where I might need more context. The configs are generated with a script, so I don't believes there's any extra maintenance needed.

docs/rules/name-property-casing.md
chrisvfritz: Not necessarily. For example, the `new-cap` and `camelcase` rules don't include the ability to enforce `snake_case`. Some might argue they should be able to configure variable and constructor casing however they like, but if there's no use case, it's not included as an option.

Similarly, there are benefits for PascalCase and kebab-case in the `name` property, but I don't believe any scenario exists where you'd want to use camelCase. It would simply break convention both in JavaScript (constructors are PascalCase) and HTML (custom elements are kebab-case).

lib/config/all.js
chrisvfritz: Oops. 😅  That was something I experimented with, but decided there wasn't a use case for simply including all rules. I've deleted it now. Good catch!

chrisvfritz:
@michalsnik Thanks for the feedback! 🙂  I responded to your concerns to provide more information. Sorry for not including that context in the PR description. 😅

chrisvfritz:
@michalsnik Just pinging on this. 🙂  I tried to address your concerns above and would love to hear your thoughts.

docs/rules/attribute-hyphenation.md
```diff
@@ -15,41 +15,25 @@ Default casing is set to `always`
 :+1: Examples of **correct** code`:
 
 ```html
-<template>
-  <foo my-prop="prop">
-    <a onClick="return false"></a>
-  </foo>
-</template>
+<MyComponent my-prop="prop"/>
```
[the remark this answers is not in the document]
chrisvfritz: @Aladdin-ADD See [this rule](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended) in the style guide.

docs/rules/name-property-casing.md
michalsnik: Okay, if it's aligned with the rule now then cool :)

README.md
michalsnik: Okay, I get your point better now @chrisvfritz Thanks for adding more context :)

And regarding last three paragraphs that you quoted:
I was thinking about the case where we want to introduce a rule, but we don't want to enable it in any config by default. Let's say we'd want to give people possibility to use the given rule, but it's not aligned with the official styleguide for some reason.
Maybe it's an edge case, but we had a similar scenario in eslint-plugin-ember.

Adding another group is not a big deal though, so I think that we can move forward with the current state, and iterate over it later on 👍

michalsnik:
Yeah, you changed my mind @chrisvfritz :) Since now we have an official style guide and this is also an official plugin for Vue, it makes sense to be more opinionated and to have the rules organised as proposed. However before merging I would love to hear what @mysticatea thinks about it as well :)

README.md
chrisvfritz: @michalsnik I agree it'd be good to introduce a category like that when the need arises - and I can already think of use cases for it. For example, I like to disallow filters in many projects, because they add one more thing that developers have to learn without enabling them to do anything new. That kind of thing wouldn't fit into any (existing) categories of the style guide though.

chrisvfritz:
@michalsnik Agreed. 🙂 👍  Just added a request for his review.

michalsnik:
Thanks for the last fix @chrisvfritz, you were right :) Please rebase/merge master now, and @mysticatea can you please take a look at this PR as well?

mysticatea:
I apology that I'm inactive during a long time.

Wow, amazing work!
Looks almost good to me.

One point, I have a concern that this PR removed `recommended: true/false` property of all rules.
Because [our semver policy](https://github.com/vuejs/eslint-plugin-vue#anchor-semantic-versioning-policy) is following [ESLint semver policy](https://github.com/eslint/eslint#semantic-versioning-policy), so it's `semver-major` matter if we add a new rule into configs like `vue/recommended`. I think we need the way to add rules without config changes in order to avoid major bump rushes.

chrisvfritz:
Yes, thanks for the ping! I agree with you @mysticatea and will update this weekend. And while I think that semver policy makes sense, I also think it might not be too bad to publish major versions relatively often (e.g. as often as monthly?), especially in the beginning as there may be many new rules to add. Do you think that would be OK?

mysticatea:
Thank you, @chrisvfritz.

I agree with you. I think that one or two of new `Essential` rules is valuable to make a major release. ESLint itself has a policy they don't make two major releases within six months, but off course, plugins can make major releases more frequency. Monthly or bimonthly looks good to me.

michalsnik:
Hey @chrisvfritz would you be so kind and resolve these conflicts? :) I think we should start merging it all together now

chrisvfritz:
@michalsnik Conflicts resolved! 😄  

@mysticatea Rules no longer require a category and if they don't, they'll be added to an "Uncategorized" section after the main rules in docs, but before the "Deprecated" rules.

I also updated the default for html-indent rules from 4 spaces to 2 spaces, since that's what's used throughout all the Vue docs and just seems to be a more popular in the web community (e.g. GitHub's style guide, Google's style guide, and [in GitHub repos](https://ukupat.github.io/tabs-or-spaces/)).

mysticatea:
I'm sorry for late: `Uncategorized` looks reasonable to me.

Issue #170, opened 2017-08-29
Rule Proposal `vue/attribute-order`


mikegreiling:
## Please describe what the rule should do:

Provide a template rule to enforce placing vue-specific attributes before other properties in a template tag.

## What category of rule is this? (place an "X" next to just one item)

[X] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

#### Configuration

'vue/closing-bracket-location': ['alphabetical' | 'vue-first']

#### Examples:

```html
// bad
<a
  v-tooltip
  class="btn external-url"
  data-container="body"
  target="_blank"
  rel="noopener noreferrer nofollow"
  :title="title"
  :aria-label="title"
  :href="externalUrl">

// good
<a
  v-tooltip
  :title="title"
  :aria-label="title"
  :href="externalUrl"
  class="btn external-url"
  data-container="body"
  target="_blank"
  rel="noopener noreferrer nofollow">
```

mikegreiling:
I'd be open to a better name for this rule 😄

kaicataldo:
I wonder if more fine-grained control could be added to the config. I'm thinking something similar to to the config options for [eslint-plugin-import's order rule](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/order.md#options).

michalsnik:
I very much like this idea, plus it have already been mentioned in #77

michalsnik:
https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended is the best point of reference how the config should look like :)

erindepew:
@michalsnik @mikegreiling I just opened a PR for this rule https://github.com/vuejs/eslint-plugin-vue/pull/209 is this what you had in mind?

matt-oconnell:
Hi @michalsnik @mikegreiling, thanks for all the hard work that's already gone into this plugin. I've been working to add configuration to this rule with @erindepew. Before moving forward, we just want to ensure that we're not doing duplicate work and that we're moving in the right direction.

1) As mentioned in #77 there are recommended groupings of attributes in the style guide: https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended. I was thinking of accepting an array of group names as a configuration option to define the order. If none are given, the rule will enforce the default order specified within the style guide.

2) The could also be an alphabetical option. This would enforce alphabetical order within groups themselves.

[...]

PR #209, opened 2017-10-09
[New] Add `vue/attribute-order`

erindepew:
resolves https://github.com/vuejs/eslint-plugin-vue/issues/170

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
+|  | [vue/attributes-order](./docs/rules/attributes-order.md) | enforce order of attributes |
 | :wrench: | [vue/html-closing-bracket-newline](./docs/rules/html-closing-bracket-newline.md) | require or disallow a line break before tag's closing brackets |
 | :wrench: | [vue/html-closing-bracket-spacing](./docs/rules/html-closing-bracket-spacing.md) | require or disallow a space before tag's closing brackets |
 | :wrench: | [vue/script-indent](./docs/rules/script-indent.md) | enforce consistent indentation in `<script>` |
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -1,0 +1,113 @@
+# enforce order of attributes (vue/attributes-order)
+
+## :book: Rule Details
+
+This rule aims to enfore ordering of component attributes. The default order is specified in the [Vue styleguide](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended) and is:
+- DEFINITION
+ex: 'is'
+- LIST_RENDERING
+ex: 'v-for item in items'
+- CONDITIONALS
+ex: 'v-if', 'v-else-if', 'v-else', 'v-show', 'v-cloak'
+- RENDER_MODIFIERS
+ex: 'v-once', 'v-pre'
+- GLOBAL
+ex: 'id'
+- UNIQUE
+ex: 'ref', 'key', 'slot'
+- BINDING
+ex: 'v-model', 'v-bind'
+- OTHER_ATTR
+ex: 'customProp="foo"'
+- EVENTS
+ex: '@click="functionCall"'
+- CONTENT
+ex: 'v-text', 'v-html'
+
+:+1: Examples of **correct** code`:
+
+```html
+<div
+  is="header"
+  v-for="item in items"
+  v-if="!visible"
+  v-once id="uniqueID"
+  ref="header"
+  v-model="headerData"
+  myProp="prop"
+  @click="functionCall"
+  v-text="textContent">
+</div>
+```
+
+```html
+<div
+  v-for="item in items"
+  v-if="!visible"
+  propOne="prop"
+  propTwo="prop"
+  propThree="prop"
+  @click="functionCall"
+  v-text="textContent">
+</div>
+```
+
+```html
+<div
+  propOne="prop"
+  propTwo="prop"
+  propThree="prop">
+</div>
+```
+
+:-1: Examples of **incorrect** code`:
+
+```html
+<div
+  ref="header"
+  v-for="item in items"
+  v-once id="uniqueID"
+  v-model="headerData"
+  myProp="prop"
+  v-if="!visible"
+  is="header"
+  @click="functionCall"
+  v-text="textContent">
+</div>
+```
+
+### `order`
+
+Specify custom order of attribute groups
+
+:+1: Examples of **correct** code with custom order`:
+
+```html
+<!-- 'vue/attribute-order': [2, { order: ['LIST_RENDERING', 'CONDITIONALS', 'RENDER_MODIFIERS', 'GLOBAL', 'UNIQUE', 'BINDING', 'OTHER_ATTR', 'EVENTS', 'CONTENT', 'DEFINITION'] }] -->
+<div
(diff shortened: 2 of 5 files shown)


michalsnik:
Hey @erindepew Are you willing to work on this further on? If yes - let us know and please see last comments in #209 for more informations about possible solutions :)

erindepew:
@michalsnik sure! I was just waiting for the community to decide if they want a configuration option and if so what it should be. Are you referring to https://github.com/vuejs/eslint-plugin-vue/issues/170#issuecomment-336772966 ?

michalsnik:
Cool! Sort of, but you know what? Let's start with a simple implementation that does job well according to the official style guide, that is: let's pre-define groups of attributes and only allow people to change order of those groups in configuration, but not individual attributes.
Let's not overthink it yet.

So the default configuration could look just like this:
```
'vue/attributes-order': [2, {
  order: [
    'CONDITIONALS',
    'RENDER_MODIFIERS'
    // ...
  ],
}]
```

Groups should obviously reflect situation from style guide.

After it's done we can iterate and add custom groups support / think about possible edge scenarios etc. WDYT?

erindepew:
@michalsnik I'm going to assume that it makes sense to push off ordering them alphabetically within those groups for a future iteration?

erindepew:
@michalsnik updated, let me know what you think!

lib/rules/attribute-order.js
```diff
@@ -0,0 +1,99 @@
...
+  } else if (name === 'pre' || name === 'once') {
+    return attributeOrder.indexOf('RENDER_MODIFIERS')
+  } else if (name === 'id') {
+    return attributeOrder.indexOf('GLOBAL')
+  } else if (name === 'ref' || name === 'key' || name === 'slot') {
+    return attributeOrder.indexOf('UNIQUE')
+  } else if (name === 'model') {
+    return attributeOrder.indexOf('BINDING')
+  } else if (name === 'on') {
+    return attributeOrder.indexOf('EVENTS')
```
michalsnik: Does it catch both `v-on` and `@`?

lib/rules/attribute-order.js
```diff
@@ -0,0 +1,99 @@
...
+  } else if (name === 'if' || name === 'else-if' || name === 'else' || name === 'show' || name === 'cloak') {
+    return attributeOrder.indexOf('CONDITIONALS')
+  } else if (name === 'pre' || name === 'once') {
+    return attributeOrder.indexOf('RENDER_MODIFIERS')
+  } else if (name === 'id') {
+    return attributeOrder.indexOf('GLOBAL')
+  } else if (name === 'ref' || name === 'key' || name === 'slot') {
+    return attributeOrder.indexOf('UNIQUE')
+  } else if (name === 'model') {
+    return attributeOrder.indexOf('BINDING')
```
michalsnik: `v-bind:something` and `:something` should also be marked as binding I think, not only `v-model`.
Also what if someone adds property `model="asd"`? I think it would mark it as binding and not OTHER_ATTR. You should probably check if `attribute.directive` is true in all cases where you want to detect `v-` attribute.

README.md
```diff
@@ -161,7 +161,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 |  | [require-prop-types](./docs/rules/require-prop-types.md) | require type definitions in props |
 | :wrench: | [v-bind-style](./docs/rules/v-bind-style.md) | enforce `v-bind` directive style |
 | :wrench: | [v-on-style](./docs/rules/v-on-style.md) | enforce `v-on` directive style |
-
+| :wrench: | [attribute-order](./docs/rules/attribute-order.md) | enforce ordering of attributes |
```
michalsnik: Did you add it by hand? I see there is `fixable` set to null in rule, but I also see 🔧 here. Please run `npm run update` if you want to update readme or rule documentation header base on rule source.

lib/rules/attribute-order.js
erindepew: yup! I'm using `@click` in my tests and it seems to work fine, I'll also add `v-on` just to be sure

erindepew:
@michalsnik yup, that all makes sense. Updated!

michalsnik:
Hey @erindepew, I'll take a closer look and test run soon :) So far it's looking very promising 🚀 

A little advice for the future though: please try not to rebase during review, as people that check your code after review once again can't see actual changes you performed in order to resolve given suggestions. Rebasing is good right before requesting review or after review :)

docs/rules/attributes-order.md
```diff
@@ -0,0 +1,113 @@
...
+- CONDITIONALS
+ex: 'v-if', 'v-else-if', 'v-else', 'v-show', 'v-cloak'
+- RENDER_MODIFIERS
+ex: 'v-once', 'v-pre'
+- GLOBAL
+ex: 'id'
+- UNIQUE
+ex: 'ref', 'key', 'slot'
+- BINDING
+ex: 'v-model', 'v-bind'
```
michalsnik: Does `:something=""` count here too?

docs/rules/attributes-order.md
```diff
@@ -0,0 +1,113 @@
...
+- OTHER_ATTR
+ex: 'customProp="foo"'
+- EVENTS
+ex: '@click="functionCall"'
```
michalsnik: Does `v-on=""` count here too?

docs/rules/attributes-order.md
```diff
@@ -0,0 +1,113 @@
...
+ex: 'v-text', 'v-html'
+
+:+1: Examples of **correct** code`:
+
+```html
+<div
+  is="header"
+  v-for="item in items"
+  v-if="!visible"
+  v-once id="uniqueID"
```
michalsnik: Why there are two attributes in the same line here?

tests/lib/rules/attributes-order.js
```diff
@@ -0,0 +1,172 @@
...
+      filename: 'test.vue',
+      code: '<template><div @click="functionCall"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div myProp="prop"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div is="header" v-for="item in items" v-if="!visible" v-once id="uniqueID" ref="header" v-model="headerData" myProp="prop" @click="functionCall" v-text="textContent"></div></template>'
```
michalsnik: You should wrap those lines so it's easier to read :)

lib/rules/attributes-order.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  })
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'enforce order of attributes',
+      category: undefined,
+      recommended: false
```
chrisvfritz: This can go in the `recommended` category and we can remove `recommended: false`, since we no longer use that property.

erindepew:
Thanks @chrisvfritz and @michalsnik for the review! I made the suggested changes

wadetandy:
This looks great! Would love to get this merged so our team can start taking advantage of it. Anything still outstanding here?

mikesherov:
Congrats @erindepew!

skjnldsv:
Hey! Thanks for this awesome rule @erindepew !

There was a discussion and some comments about making [alphabetical order within groups](https://github.com/vuejs/eslint-plugin-vue/pull/209#issuecomment-355460942), is this implemented? The documentation does not state anything related :thinking: 
Thanks :)

PR #221, opened 2017-10-24
html-end-tags: don't fail for self-closing elements

### LINKED ISSUE #217: html-end-tags and html-self-closing conflict

* **ESLint Version: 4.9.0**
* **eslint-plugin-vue Version: 3.13.1**
* **Node Version: 8.2.1**

```
const OFF = 'off'
const WARN = 'warn'
const ERROR = 'error'

module.exports = {
  extends: [
    'eslint:recommended',
    'plugin:lodash/recommended',
    'plugin:you-dont-need-lodash-underscore/compatible',
    'plugin:vue/recommended',
    'standard'
  ],
  plugins: [
    'babel',
    // 'html',
    'lodash'
  ],
  parserOptions: { parser: 'babel-eslint' },
  env: {
    commonjs: true,
    jquery: true
  },
  rules: {
    'vue/html-end-tags': ERROR,
    'vue/html-self-closing': ERROR
  }
}
```

```vue
<template>
  <fa-icon class="fa-spinner fa-lg fa-pulse" title="Uploading..." />
</template>
```

No error.

'<fa-icon>' should have end tag. (vue/html-end-tags)

Setting the `html.component` option to `'never'` on the `vue/html-self-closing` rule and then using a closing tag instead of self-closing fixes the issue, but the default options cause a conflict.

mariolamacchia:
Fixes #217

### DIFF
--- a/docs/rules/html-end-tags.md
+++ b/docs/rules/html-end-tags.md
@@ -36,6 +36,7 @@ This rule reports the following elements:
         <div></div>
         <p></p>
         <p></p>
+        <div />
         <input>
         <br>
     </div>
--- a/lib/rules/html-end-tags.js
+++ b/lib/rules/html-end-tags.js
@@ -27,6 +27,7 @@ function create (context) {
       const name = node.name
       const isVoid = utils.isHtmlVoidElementName(name)
       const hasEndTag = node.endTag != null
+      const isSelfClosing = node.startTag.selfClosing
 
       if (isVoid && hasEndTag) {
         context.report({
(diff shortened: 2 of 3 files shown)


mariolamacchia:
Hey guys! A question: should `<br />` be valid or not?

mysticatea:
Thank you for contributing!

LGTM.

`<br />` is valid in HTML spec because it's a void element.
People can control preference between `<br>` and `<br/>` by [vue/html-self-closing](https://github.com/vuejs/eslint-plugin-vue/blob/0a8eb8e385e005532922fe73a01816b0af3a7952/docs/rules/html-self-closing.md) rule.

In background of this bug, actually, `<div/>` is [non-void-html-element-start-tag-with-trailing-solidus](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-non-void-html-element-start-tag-with-trailing-solidus) error in HTML spec. This rule was created on early time I didn't know Vue.js supports the trailing solidus even if it's a non-void HTML element.

PR #228, opened 2017-11-05
Update html-end-tags doc

### LINKED ISSUE #160: Question: `vue/html-end-tags`

* **ESLint Version:** 4.4.1
* **eslint-plugin-vue Version:** 3.11.0
* **Node Version:** 8.3.0

**Question**

For now rule `vue/html-end-tags` works only for non void html tags as described in doc: 
https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/html-end-tags.md#todo-brbr
```
TODO: <br></br>

parse5 does not recognize the illegal end tags of void elements.
```

We are not longer use `parse5` but issue is stil present with new parser.

michalsnik:
Fixes #160

### DIFF
--- a/docs/rules/html-end-tags.md
+++ b/docs/rules/html-end-tags.md
@@ -2,17 +2,12 @@
 
 - :wrench: The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) can automatically fix some of the problems reported by this rule.
 
-This rule enforce the way of end tags.
-
-- [Void elements] disallow end tags.
-- Other elements require end tags.
-
 ## :book: Rule Details
 
-This rule reports the following elements:
+This rule reports:
 
-- [Void elements] which have end tags.
-- Other elements which do not have end tags and are not self-closing.
+- presence of end tags on [Void elements](https://www.w3.org/TR/html51/syntax.html#void-elements)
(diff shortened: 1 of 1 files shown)


michalsnik:
@mysticatea up

docs/rules/html-end-tags.md
```diff
@@ -2,17 +2,12 @@
...
-- Other elements require end tags.
-
 ## :book: Rule Details
 
-This rule reports the following elements:
+This rule reports:
 
-- [Void elements] which have end tags.
-- Other elements which do not have end tags and are not self-closing.
+- presence of end tags on [Void elements](https://www.w3.org/TR/html51/syntax.html#void-elements)
```
chrisvfritz: Maybe better to say "an and tag"?

docs/rules/html-end-tags.md
```diff
@@ -2,17 +2,12 @@
...
+- absence of end tags or self-closed endings on other elements
```
chrisvfritz: Might be clearer to say:

> absence of both an end tag (e.g. `</div>`) and self-closing opening tag (e.g. `<div/>`) on other elements

michalsnik:
@chrisvfritz Thanks for suggestions! Updated :)

PR #232, opened 2017-11-16
Remove deprecated rules

michalsnik:


mysticatea:
Thank you!

I agree with @chrisvfritz about the conditional `Deprecated` section.

michalsnik:
Makes sense. Done @chrisvfritz @mysticatea :)

PR #269, opened 2017-12-01
Added exact as valid modifier

mokkabonna:
https://vuejs.org/v2/guide/events.html#exact-Modifier

### DIFF
--- a/lib/rules/valid-v-on.js
+++ b/lib/rules/valid-v-on.js
@@ -18,7 +18,7 @@ const utils = require('../utils')
 const VALID_MODIFIERS = new Set([
   'stop', 'prevent', 'capture', 'self', 'ctrl', 'shift', 'alt', 'meta',
   'native', 'once', 'left', 'right', 'middle', 'passive', 'esc', 'tab',
-  'enter', 'space', 'up', 'left', 'right', 'down', 'delete'
+  'enter', 'space', 'up', 'left', 'right', 'down', 'delete', 'exact'
 ])
 const VERB_MODIFIERS = new Set([
   'stop', 'prevent'


mysticatea:
Thank you for the PR!

The `exact` modifier seems to have to use with one or more system modifiers. Could you add the check the `exact` modifier is used with system modifiers?

For example:

```js
const SYSTEM_MODIFIERS = new Set([
  'ctrl', 'alt', 'shift', 'meta'
})

// ....

         if (modifier === "exact" && !node.key.modifiers.some(SYSTEM_MODIFIERS.has, SYSTEM_MODIFIERS)) {
           context.report({
             node,
             loc: node.loc,
             message: "'exact' modifier should be used with one or more system modifiers."
           })
         }
```

mokkabonna:
It's not only for system modifiers.

If I have this:

```html
<div @click.ctrl.exact="handler1" @click="handler2"></div>
```

handler 2 will be also be triggered when holding ctrl. handler 1 will only be triggered when holding ctrl.

Using exact on both works as expected:

```html
<div @click.ctrl.exact="handler1" @click.exact="handler2"></div>
```

mysticatea:
Oh, thank you for the explanation! I got it. LGTM.

mokkabonna:
Just as a follow up:

This code generates the guard: https://github.com/vuejs/vue/blob/daed1e73557d57df244ad8d46c9afff7208c9a2d/src/compiler/codegen/events.js#L81

It starts with all modifier keys and removes any that are present. So if none are present then the guard generates this:

```js
if (
  $event.ctrlKey ||
  $event.shiftKey ||
  $event.altKey ||
  $event.metaKey
) {
  return null
}
```

So if any modifier present, it does not trigger the handler.

PR #274, opened 2017-12-02
[Fix] Incorrect errors in invalid EOF cases (fixes #268)

mysticatea:
This PR fixes #268 and some similar cases:

- [Case 1](https://mysticatea.github.io/vue-eslint-demo/#eyJjb2RlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PjwhLS1cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuIiwicnVsZXMiOnsidnVlL2h0bWwtZW5kLXRhZ3MiOjIsInZ1ZS9uby1wYXJzaW5nLWVycm9yIjoyfX0=)
- [Case 2](https://mysticatea.github.io/vue-eslint-demo/#eyJjb2RlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PlxuICAgIDxzdmc+XG4gICAgICA8IVtDREFUQVt0ZXN0XG4gIDwvZGl2PlxuPC90ZW1wbGF0ZT5cbiIsInJ1bGVzIjp7InZ1ZS9odG1sLWVuZC10YWdzIjoyLCJ2dWUvbm8tcGFyc2luZy1lcnJvciI6Mn19)
- [Case 3](https://mysticatea.github.io/vue-eslint-demo/#eyJjb2RlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PlxuICAgIDwhLS1cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuIiwicnVsZXMiOnsidnVlL2h0bWwtZW5kLXRhZ3MiOjIsInZ1ZS9uby1wYXJzaW5nLWVycm9yIjoyfX0=)
- [Case 4](https://mysticatea.github.io/vue-eslint-demo/#eyJjb2RlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PlxuICAgIDwhLS1cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuIiwicnVsZXMiOnsidnVlL2h0bWwtc2VsZi1jbG9zaW5nIjoyLCJ2dWUvbm8tcGFyc2luZy1lcnJvciI6Mn19)

I.e. some rules are confused in invalid EOF. This PR fixes some rules to not report errors if the code has invalid EOF.

PR #275, opened 2017-12-02
[Update] Make `vue/html-quotes` fixable

mysticatea:
This PR makes `vue/html-quotes` rule fixable. If the quotes exist in the attribute value, the rule replaces those quotes by HTML entities.

This PR includes #274 to address invalid EOF.


michalsnik:
Nice! Thank you @mysticatea 
Let's merge this once #274 is done

PR #278, opened 2017-12-02
[Update] Make `vue/no-shared-component-data` fixable

mysticatea:
This PR makes `vue/no-shared-component-data` fixable. The autofix wraps the `data` property's value by a function expression.

At first glance, the unit tests make wrong indentations, but this is intentional. This strategy is same as core rules; one rule does one thing. A combination of rules will make correct autofix.

image

### DIFF
--- a/lib/rules/no-shared-component-data.js
+++ b/lib/rules/no-shared-component-data.js
@@ -6,10 +6,16 @@
 
 const utils = require('../utils')
 
+function isOpenParen (token) {
+  return token.type === 'Punctuator' && token.value === '('
+}
+
+function isCloseParen (token) {
+  return token.type === 'Punctuator' && token.value === ')'
+}
+
 function create (context) {
-  // ----------------------------------------------------------------------
-  // Public
-  // ----------------------------------------------------------------------
+  const sourceCode = context.getSourceCode()
 
   return utils.executeOnVueComponent(context, (obj) => {
     obj.properties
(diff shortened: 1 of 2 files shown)


lib/rules/no-shared-component-data.js
```diff
@@ -21,10 +27,33 @@ function create (context) {
...
-          node: cp.value,
-          message: '`data` property in component must be a function'
+          node: p,
+          message: '`data` property in component must be a function',
+          fix (fixer) {
+            let first = sourceCode.getFirstToken(p.value)
+            let last = sourceCode.getLastToken(p.value)
+
+            // If the value enclosed by parentheses, update the 'first' and 'last' by the parentheses.
+            let t = null
```
michalsnik: I think we should use more descriptive variables, `t` and `u` doesn't mean much.

lib/rules/no-shared-component-data.js
```diff
@@ -21,10 +27,33 @@ function create (context) {
...
+            let u = null
+            while (isOpenParen(t = sourceCode.getTokenBefore(first)) && isCloseParen(u = sourceCode.getTokenAfter(last))) {
```
michalsnik: I think we should avoid assigning values in function params, it's kind of confusing. Please simplify this part and wrap too long line.

lib/rules/no-shared-component-data.js
```diff
@@ -21,10 +27,33 @@ function create (context) {
...
+
+            // If we can upgrade requirements to `eslint@>4.1.0`, this code can be replaced by:
+            //     return [
+            //       fixer.insertTextBefore(first, 'function() {\nreturn '),
+            //       fixer.insertTextAfter(last, ';\n}')
+            //     ]
+            // See: https://eslint.org/blog/2017/06/eslint-v4.1.0-released#applying-multiple-autofixes-simultaneously
+            const range = [first.range[0], last.range[1]]
+            const valueText = sourceCode.text.slice(range[0], range[1])
+            const replacement = `function() {\nreturn ${valueText};\n}`
```
michalsnik: Can't we replace it with `data() {}` instead of a `function`?

mysticatea: I think that a function expression is better to keep the rule simple.
If we want to fix to a method shorthand, it needs extra check the `object-shorthand` rule is doing. I.e., if comments exist between the key and the value, the rule should not autofix it to avoid removing the comments.

mysticatea:
Updated about `t`/`u`.

lib/rules/no-shared-component-data.js
michalsnik: Ah, alright. There is already `object-shorthand` core rule, that is fixable, so it should take care of it :) I forgot about it.

michalsnik:
Good job @mysticatea 🙌

PR #280, opened 2017-12-04
Fix: html-indent about comments (fixes #241)

### LINKED ISSUE #241: HTML indent calculation problem with comments
If a HTML comment is the last entry in an indentation level, it expects the wrong level (always one indentation less).

For example the following example works as expected:
```
<div>
    <br>
    <!-- michalsnik rocks -->
    <br>
</div>
```
But if the comment is the last entry, we'll get a `Expected indentation of 0 spaces but found 4 spaces` error for the line of the comment:
```
<div>
    <br>
    <!-- michalsnik rocks -->
</div>
```
So it expects the comment to look like this:
```
<div>
    <br>
<!-- michalsnik rocks -->
</div>
```
I'd say, it's probably not what the author **usually** wants. I'd say it might be best to ignore indentation check for comments altogether, as we can't predict if the developer wants to comment out the last row, or write a comment for the closing tag.

mysticatea:
Fixes #241.

This PR fixes `vue/html-indent` rule about comments.

If the next token of a comment is the following tokens, the rule allows the indentation of comments at either the same as the next line or the previous line. If neither, the rule auto-fixes the indentation of comments to the same as the previous line.

- `HTMLEndTagOpen` (e.g. `</div`)
- `VExpressionEnd` (`}}`)
- `}`
- `]`
- `)`

### DIFF
--- a/lib/rules/html-indent.js
+++ b/lib/rules/html-indent.js
@@ -199,6 +199,26 @@ function isBeginningOfLine (node, index, nodes) {
 }
 
 /**
+ * Check whether a given token is a closing token which triggers unindent.
+ * @param {Token} token The token to check.
+ * @returns {boolean} `true` if the token is a closing token.
+ */
+function isClosingToken (token) {
+  return token != null && (
+    token.type === 'HTMLEndTagOpen' ||
+    token.type === 'VExpressionEnd' ||
+    (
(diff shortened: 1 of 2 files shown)


michalsnik:
Well done 🚀

PR #281, opened 2017-12-04
Fix: html-indent about binary expressions (fixes #264)

### LINKED ISSUE #264: vue/html-indent not allowing conditions with aligned indentation

* **ESLint Version:** 4.12.0
* **eslint-plugin-vue Version:** 4.0.0-beta.0
* **Node Version:** 8.9.1

```
module.exports = {
  root: true,
  parserOptions: {
    parser: 'babel-eslint',
    sourceType: 'module'
  },
  extends: ['standard', 'plugin:vue/recommended']
}
```

```vue
<template>
  <div
    :class="{
      foo: (
        a === b &&
        c === d
      )
    }"
  />
</template>
```

I expected this to be an acceptable style.

```
  ✘  vue/html-indent  Expected indentation of 10 spaces but found 8 spaces
  src/components/component.vue:6:1
          c === d
   ^
```

mysticatea:
Fixes #264.

This PR fixes `vue/html-indent` rule about the indentation of binary expressions.

Before, the second line of binary expressions always has 1 indent from the first line.
After this PR, the second line of binary expressions has the same indentation as the first line in *except* the following cases:

- The first token of the binary expression is on the same line as the previous token.
    ```js
    {
      foo: a
        + b
    }
    ```
- The first token of the binary expression is the first token of the statement.
    ```js
    a
        + b
    ```

EDIT: additionally:

- The first token of the binary expression is the first token of an argument of `CallExpression`/`NewExpression`.
    ```js
    foo(
        a
            + b,
        c
            + d
    )
    ```
- The first token of the binary expression is the first token of an element of `ArrayExpression`.
    ```js
    [
        a
            + b,
        c
            + d
    ]
    ```
- The first token of the binary expression is the first token of an expression of `SequenceExpression`.
    ```js
    (
        a
            + b,
        c
            + d
    )
    ```

### DIFF
--- a/lib/rules/html-indent.js
+++ b/lib/rules/html-indent.js
@@ -388,12 +388,19 @@ function create (context) {
   }
 
   /**
-   * Check whether a given token is the first token of a statement.
+   * Check whether a given token is the first token of:
+   *
+   * - ExpressionStatement
+   * - VExpressionContainer
+   * - A parameter of CallExpression/NewExpression
+   * - An element of ArrayExpression
+   * - An expression of SequenceExpression
+   *
    * @param {Token} token The token to check.
    * @param {Node} belongingNode The node that the token is belonging to.
-   * @returns {boolean} `true` if the token is the first token of a statement.
+   * @returns {boolean} `true` if the token is the first token of an element.
    */
-  function isFirstOfStatement (token, belongingNode) {
+  function isBeginningOfElement (token, belongingNode) {
     let node = belongingNode
 
     while (node != null) {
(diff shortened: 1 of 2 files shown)


mysticatea:
Mm, I'm considering that the following case (the first token of the binary expression is the first token of a parameter/array element) should have one indent:

```js
foo(
    a +
        b,
    c +
        d
);

[
    a +
        b,
    c +
        d,
];
```

mysticatea:
I updated this PR.

Issue #252, opened 2017-11-26
Rule proposal: `props-names-casing`


michalsnik:
**Style guide:**
https://vuejs.org/v2/style-guide/#Prop-name-casing-strongly-recommended

**Description:**
This rule would enforce proper casing of props in vue components (camelCase).
We already have a rule for checking HTML attributes casing, and this would make a good addition to [that rule](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/attribute-hyphenation.md).

We could add different casing as options and set camelCase as default, but I think we should definitely forbid two of them anyway:
- kebab-case - which doesn't make any sense and should be avoided
- PascalCase - as this is the casing recommended for components, and users should not be able to mix props with components

so in the end there may be two options:
- camelCase
- snake_case

Which makes me think if this rule shouldn't be named `props-camel-casing` and force only one style of props names.

```
// bad
props: {
  'greeting-text': String
}
```

```
// good
props: {
  greetingText: String
}
```

mysticatea:
Could we use just core [camelcase](https://eslint.org/docs/rules/camelcase) rule?

michalsnik:
Core rule won't work in case someone's using array of props @mysticatea

PR #289, opened 2017-12-10
[New] Add `prop-name-casing`

bfkbx960:
This PR implements rule proposed in #252


michalsnik:
Thanks for your work @bfkbx960 ! We really appreciate it :) We're going to merge new rules once stable v4 is out 🚀

lib/rules/prop-name-casing.js
```diff
@@ -0,0 +1,73 @@
...
+      if (item.type !== 'Property') {
+        return
+      }
+
+      const propName = item.key.type === 'Literal' ? item.key.value : item.key.name
+      const convertedName = converter(propName)
+      if (convertedName !== propName) {
+        context.report({
+          node: item,
+          message: 'Prop "{{name}}" is not {{caseType}}.',
```
michalsnik: I think it should be: `is not in`

lib/rules/prop-name-casing.js
```diff
@@ -0,0 +1,73 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'enforce specific casing for the Prop name in Vue components',
+      category: 'strongly-recommended'
```
michalsnik: Can you please change the category to be `upcoming`?
```
category: 'upcoming' // 'strongly-recommended'
```
We're not going to enable this rule by default until the next major release, so we'll stash them in special category for everyone to see what's going to be enabled in next version :)

bfkbx960:
@michalsnik suggestions applied, thank you for code review.
In my local environment the test will succeed but it seems to fail here :(

michalsnik:
Thanks @bfkbx960! If you remove `node_modules`, set node to version 4, install all dependencies once again and run `npm test` on node 4 - it's failing. It's working on master though.

michalsnik:
or it's not working even on master now o.O which is soo weird, as yeasterday I released 4.1.0 and everything passed

PR #291, opened 2017-12-16
Fix html-self-closing example json

maoberlehner:
`vue/html-self-closing` must be used as a key in the eslint config to
make it work correctly. Also this is consitent with the example json for
other rules.


michalsnik:
Ah, good catch! Thanks @maoberlehner

PR #297, opened 2017-12-20
chore(package): Swap out `requireindex`

### LINKED ISSUE #290: Issues with loading plugin via ESLint API when running Jest

* **ESLint Version:** 4.12.1
* **eslint-plugin-vue Version:** 4.0.0.beta.2
* **Node Version:** 9.2.0

```
module.exports = {
  plugins: [
    'jest'
  ],
  extends: [
    'eslint:recommended',
    'plugin:jest/recommended',
    'plugin:vue/recommended',
    'standard'
  ],
  env: {
    'jest/globals': true
  }
}
```

Tried testing my ESLint config via Jest.

```vue
const { CLIEngine } = require('eslint')

const configFile = require.resolve('../.eslintrc.js')
const engine = new CLIEngine({ configFile })

test('should include required parser options', () => {
  const config = engine.getConfigForFile(configFile)
  expect(config.parserOptions.ecmaVersion).toBeGreaterThanOrEqual(6)
  expect(config.parserOptions).toHaveProperty('sourceType', 'module')
})

test('should include required environments', () => {
  const config = engine.getConfigForFile(configFile)
  expect(config.env).toHaveProperty('browser', true)
  expect(config.env).toHaveProperty('es6', true)
  expect(config.env).toHaveProperty('node', true)
})

test('should include required plugins', () => {
  const config = engine.getConfigForFile(configFile)
  expect(config.plugins).toContain('jest')
  expect(config.plugins).toContain('vue')
})
```

Config to load correctly, tests to pass.

```
    Failed to load config "plugin:vue/recommended" to extend from.
    Referenced from: /mnt/c/Users/kwilliams/Projects/eslint-config/.eslintrc.js

      at configMissingError (node_modules/eslint/lib/config/config-file.js:187:19)
      at loadConfigFile (node_modules/eslint/lib/config/config-file.js:213:27)
      at loadFromDisk (node_modules/eslint/lib/config/config-file.js:486:18)
      at load (node_modules/eslint/lib/config/config-file.js:550:20)
      at configExtends.reduceRight (node_modules/eslint/lib/config/config-file.js:421:36)
          at Array.reduceRight (<anonymous>)
      at applyExtends (node_modules/eslint/lib/config/config-file.js:403:28)
      at loadFromDisk (node_modules/eslint/lib/config/config-file.js:514:22)
      at Object.load (node_modules/eslint/lib/config/config-file.js:550:20)
      at Config.loadSpecificConfig (node_modules/eslint/lib/config.js:139:46)
      at new Config (node_modules/eslint/lib/config.js:101:14)
      at new CLIEngine (node_modules/eslint/lib/cli-engine.js:420:23)
      at Object.<anonymous> (__tests__/eslintrc.test.js:4:16)
          at Generator.next (<anonymous>)
          at new Promise (<anonymous>)
          at Generator.next (<anonymous>)
          at <anonymous>
```

The reason this happens is that the `requireindex` dependency checks filenames against `require.extensions` (1), which is deprecated. When running Jest, `require.extensions` is empty, and thus the config files never get loaded (2). `requireindex` hasn't been updated since 2014. Can we look for a suitable replacement, or would it make sense to integrate it into the plugin's codebase as a utility function without `require.extensions`?

1. https://github.com/stephenhandley/requireindex/blob/master/index.js#L45
2. https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/index.js#L7

Edit: Looks like this issue was brought up with Jest already: https://github.com/facebook/jest/issues/2017. Given that this is still an issue, and I feel it is more proper to move away from deprecated API anyway, I still think the best solution would be to stop using `requireindex`.

scriptdaemon:
Swap out `requireindex` dependency with the more recently updated
dependency `require-all`. The old dependency uses deprecated API that
breaks when running in a Jest testing environment.

Fixes #290

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,10 +1,16 @@
 'use strict'
 
 const resolve = require('path').resolve
-const requireIndex = require('requireindex')
+const requireAll = require('require-all')
 
-const rules = requireIndex(resolve(__dirname, 'rules'))
-const configs = requireIndex(resolve(__dirname, 'config'))
+const rules = requireAll({
+  dirname: resolve(__dirname, 'rules'),
+  filter: /^([\w\-]+)\.js$/
+})
+const configs = requireAll({
+  dirname: resolve(__dirname, 'config'),
+  filter: /^([\w\-]+)\.js$/
+})
 
 module.exports = {
   rules,
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "eslint": "^3.18.0 || ^4.0.0"
   },
   "dependencies": {
-    "requireindex": "^1.1.0",
+    "require-all": "^2.2.0",
     "vue-eslint-parser": "^2.0.1-beta.1"
   },
   "devDependencies": {


michalsnik:
Thanks @scriptdaemon !

PR #301, opened 2017-12-21
[Fixes #293] Update indent rule

### LINKED ISSUE #293: [vue/html-indent] unexpected indentation level

* **ESLint Version:** 4.13.1
* **eslint-plugin-vue Version:** 4.0.0-beta.3
* **Node Version:** 9.3.0

```
'vue/html-indent': ['error', 2, {
  'attribute': 1
}],
```

```vue
<template>
  <div class="alpha">
    <div
      class="bravo"
    />
    <div v-for="n in list" :key="n"
      class="charlie"
    >
      {{ n }}
    </div>
    <div
      v-for="n in list"
      :key="n"
      class="delta"
    >
      {{ n }}
    </div>
  </div>
</template>

<script>
export default {
  data () {
    return {
      list: ['a', 'b', 'c'],
    }
  },
}
</script>
```

I expect the template to be considered valid.

```
…/src/demo.vue
  7:1  error  Expected indentation of 9 spaces but found 6 spaces  vue/html-indent
```

Note how the indentation of `class="bravo"` and `class="delta"` is fine, but `class="charlie"` is somehow trying to align with the `v-for="…"` on the previous line.

---

We tend to keep properties `v-if`, `v-else-if`, `v-else`, `v-for`, `key` and `is` on the same line as the opening element, where the remaining properties are spread onto their own lines. I understand the fix for my problem is to move all attributes off the opening line, but the current behavior remains unexpected.

michalsnik:
As pointed in #293 currently indent rule requires attributes to be aligned as the first one:
```html
<div id="foo"
     class="abc"
     style="font-color: red;"
>
</div>
```

It's not that common I think, and most people prefer to use it like this:
```html
<div id="foo"
  class="abc"
  style="font-color: red;"
>
</div>
```

This PR updates indent rule to force the second version.

Though I'm thinking whether we shouldn't add an optional setting to make it possible to align to the first token if needed.

What do you think guys? @mysticatea @chrisvfritz

### DIFF
--- a/lib/rules/html-indent.js
+++ b/lib/rules/html-indent.js
@@ -308,7 +308,7 @@ function create (context) {
    * @param {number} offset The offset to set.
    * @returns {void}
    */
-  function processNodeList (nodeList, leftToken, rightToken, offset, rootNode) {
+  function processNodeList (nodeList, leftToken, rightToken, offset, alignVertically = true) {
     let t
 
     if (nodeList.length >= 1) {
@@ -358,13 +358,14 @@ function create (context) {
           setBaseline(baseToken)
         }
 
-        if (!options.alignAttributesVertically && rootNode) {
-          // Align tokens relatively to passed root node
-          // So it's possible to force proper position for VAttributes
-          setOffset(alignTokens, offset, template.getFirstToken(rootNode))
-        } else {
+        if (alignVertically) {
           // Align the rest tokens to the first token.
           setOffset(alignTokens, 0, baseToken)
+        } else {
+          // Align tokens relatively to passed root node
+          // So it's possible to force proper position for VAttributes
+          const rootNode = nodeList && nodeList[0].parent
+          setOffset(alignTokens, offset, template.getFirstToken(rootNode))
         }
       }
     }
@@ -772,7 +773,7 @@ function create (context) {
       const openToken = template.getFirstToken(node)
       const closeToken = template.getLastToken(node)
 
-      processNodeList(node.attributes, openToken, null, options.attribute, node)
+      processNodeList(node.attributes, openToken, null, options.attribute, options.alignAttributesVertically)
       if (closeToken != null && closeToken.type.endsWith('TagClose')) {
         setOffset(closeToken, options.closeBracket, openToken)
       }
--- a/lib/rules/html-indent.js
+++ b/lib/rules/html-indent.js
@@ -34,6 +34,7 @@ function parseOptions (type, options) {
     attribute: 1,
     closeBracket: 0,
     switchCase: 0,
+    alignAttributesVertically: true,
     ignores: []
   }
 
@@ -52,6 +53,9 @@ function parseOptions (type, options) {
   }
   if (Number.isSafeInteger(options.switchCase)) {
     ret.switchCase = options.switchCase
+  }
+  if (options.alignAttributesVertically != null) {
+    ret.alignAttributesVertically = options.alignAttributesVertically
   }
   if (options.ignores != null) {
     ret.ignores = options.ignores
@@ -354,7 +358,7 @@ function create (context) {
           setBaseline(baseToken)
         }
 
-        if (rootNode) {
+        if (!options.alignAttributesVertically && rootNode) {
           // Align tokens relatively to passed root node
           // So it's possible to force proper position for VAttributes
           setOffset(alignTokens, offset, template.getFirstToken(rootNode))
@@ -1302,6 +1306,7 @@ module.exports = {
           'attribute': { type: 'integer', minimum: 0 },
           'closeBracket': { type: 'integer', minimum: 0 },
           'switchCase': { type: 'integer', minimum: 0 },
+          'alignAttributesVertically': { type: 'boolean' },
           'ignores': {
             type: 'array',
             items: {
--- a/lib/rules/html-indent.js
+++ b/lib/rules/html-indent.js
@@ -304,7 +304,7 @@ function create (context) {
    * @param {number} offset The offset to set.
    * @returns {void}
    */
-  function processNodeList (nodeList, leftToken, rightToken, offset) {
+  function processNodeList (nodeList, leftToken, rightToken, offset, rootNode) {
     let t
 
     if (nodeList.length >= 1) {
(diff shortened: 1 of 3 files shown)


mysticatea:
Hmm, I think it's enough common style that it aligns rest attributes to the first attribute; it exists in the source code of this issue page as well (though both styles are mixed).

chrisvfritz:
I personally don't see a use case for either of them. 😄  They seem just as strange as these styles in JavaScript:

``` js
const myObject = { foo: 'a',
                   bar: 'b',
                   baz: 'c',
}
```

``` js
const myObject = { foo: 'a',
  bar: 'b',
  baz: 'c',
}
```

I don't think ESLint even provides an option to format objects with either of these styles. 

That said, @mysticatea is right that these are _both_ unfortunately quite common. I believe at least one popular editor (don't remember which one) even auto-formats HTML in the first style by default. So an optional setting may be the best way to go, defaulting to the current behavior for backwards-compatibility.

michalsnik:
Well, it's not a JS, so it might differ after all :D I'm fine with an optional setting, but would argue that the second version should be a default one, and `aligned to first argument` as optional formatting preference. This rule is fixable, so it shouldn't be a big problem @chrisvfritz

Also we have another rule that doesn't allow attributes in first line in case there are other attributes in the following lines (`max-attributes-per-line`) so we don't have to worry too much about this kind of change I think, especially when the official preferred way of writing components doesn't approve above scenarios anyway :)

chrisvfritz:
I'm not strongly opinionated in which of these is the default. I think it's possible that the current (aligned) style is more popular, but I find your proposed change slightly less offensive for multi-line. 😄 I can see arguments either way.

On a technical note, since this is fixable: how do we know the specific indentation level to use if we can't tell from context?

michalsnik:
I'll ask on discord to see how people think about it :)

`less offensive alignment` is the argument I was looking for 🙌 

And about indentation level, what do you mean by context? In the first example we know the alignment of the first attribute and simply set expected alignments on the reset of attributes to be equal. And in the second case we set expected indentation on each but the first attribute to be shifted by `options.attribute` setting value relatively to their parent tag node. Is this what you asked about @chrisvfritz? :)

chrisvfritz:
What I mean is that in cases like this:

``` js
<div id="foo"
class="abc"
style="font-color: red;"
>
</div>
```

When we fix the indentation to match your proposed style (1 indentation level from the opening tag), how do we know whether to indent 2 spaces, 4 spaces, 1 tab, etc?

mysticatea:
@michalsnik I think that vertical alignment is enough common style. We have received such request frequently in core as well. Also, the change of default behavior requires mejor version up.

About option form, how about `{attribute: {multiplier: number, valign: boolean}}`?

michalsnik:
@chrisvfritz We know it by checking settings of this rule. 

@mysticatea the proposition with `valign` looks nice 👍 Though I still don't like HTML being compared with JavaScript in terms of its preferred style. But if we compare those two - the core `indent` rule in eslint by default also dosen't expect vertically aligned variables, and the following version is preferred:
```js
const a = 1,
  b = 2,
  c = 3;
```
I think it should look exactly the same here. And since we're still in beta I don't think it would be a big problem to introduce this change now. In my opinion better now than later :)

mysticatea:
In `vue/html-indent`, in the const declaration case, the rule aligns vertically. :)

michalsnik:
Ah, I missed it. Then perhaps this should also be updated 🤔

mysticatea:
Arrays and objects are the same. If the first element is on the same line as the left paren, it aligns rest elements. The rule handles the condition as the sign of vertical alignment at all. Especially in the variable declaration case, this simplifies implementation hugely.

mysticatea:
About https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/html-indent.js#L1073-L1087

michalsnik:
Variables declarations indeed look better vertically aligned, but as for objects, arrays and attributes I think we could quite easily update this rule to be less strict by default. Also cases with attributes, objects or arrays are, or should be most common to use, everything more should rather be avoided in interpolated expressions in my opinion anyway.

mysticatea:
I believe it's not "less strict". Just it fits preference or not. The number of spaces is A or B. 

I thought the order of commonly is the following. I didn't think `third` case is intentional. It's the reason that the rule makes vertical alignment if `a:1` is not at beginning of line.

```js
let first = {
    a: 1,
    b: 2,
    c: 3
}
let second = { a: 1,
               b: 2,
               c: 3 }
let third = { a: 1,
    b: 2,
    c: 3
}
```

Anyway, I can accept the change of the default behavior, however, I can't accept if we do without semver-major. We are following to ESLint semantic versioning policy. It intends that bug fixes never break user's CI if the user depends on eslint/plugins with npm's `~`. It intends that changes never break user's CI if the user depends on eslint/plugins with npm's `^`.

michalsnik:
Okay let's meet at the middle ground then :) I'll update this PR as I proposed, but under additional option and I won't change the default behaviour to not introduce unnecessary confusion @mysticatea 

About that third case - to be honest I didn't see such style in JS ever. Maybe we don't have to change anything about arrays and object in interpolated expressions and bindings at all 🤔 And that optional setting would only take care of HTML attributes indentation. Simple as that.

michalsnik:
I as promised, I added `alignAttributesVertically` setting. I thought about `attribute: { multipier: 1, valign: true }`, but I'm not a big fan of nested configurations. We would have to support both attribute passed as a number and as an object too, that would unnecessarily complicate this setting, so I kept it simple stupid instead.

lib/rules/html-indent.js
```diff
@@ -349,11 +353,19 @@ function create (context) {
...
         }
 
-        // Align the rest tokens to the first token.
+        // Set baseline
         if (nodeList.some(isBeginningOfLine)) {
           setBaseline(baseToken)
         }
-        setOffset(alignTokens, 0, baseToken)
+
+        if (!options.alignAttributesVertically && rootNode) {
```
mysticatea: Would you move `options.alignAttributesVertically` to the call site? `processNodeList` is a common logic for many expressions/statements. I'd like to change the value of `rootNode` by the option.

michalsnik:
Done, check if that's what you wanted me to do @mysticatea :)

PR #302, opened 2017-12-21
[Fixes #294] Fix no-unused-vars rule

### LINKED ISSUE #294: vue/no-unused-vars needs configuration options similar to eslint/no-unused-vars

* **ESLint Version:** v4.13.1
* **eslint-plugin-vue Version:** eslint-plugin-vue@^4.0.0-beta.2
* **Node Version:** v8.4.0

```
{
    "env": {
        "browser": true
    },
    "extends": [
        "eslint:recommended",
        "plugin:vue/recommended"
    ],
    "parserOptions": {
        "sourceType": "module"
    },
    "rules": {
        "indent": [
            "error",
            4,
            {
                "SwitchCase": 1
            }
        ],
        "linebreak-style": [
            "error",
            "windows"
        ],
        "quotes": [
            "error",
            "double",
            {
                "avoidEscape": true,
                "allowTemplateLiterals": true
            }
        ],
        "semi": [
            "error",
            "always"
        ],
        "no-console": [
            "off"
        ],
        "no-unused-vars": [
            "error",
            {
                "args": "after-used",
                "argsIgnorePattern": "^_"
            }
        ],
        "vue/html-self-closing": "off",
        "vue/html-indent": [
            "error", 
            4
        ],
        "vue/max-attributes-per-line": "off",
        "vue/order-in-components": "off"
    }
}
```

```vue
<template>
<div id="app">
      <div v-for="(thing, key, index) in things" :key="thing.id" :class="{ 'every-other': index % 2 === 0 }">
        {{ thing.id }} - {{ index }}
      </div>
</div>
</template>
<style>
.every-other {
  background-color: red;
}

div {
  width: 100px;
}
</style>
<script>
new Vue({
  el: '#app',
  data: () => ({
    things:  {
      thing1: { id: "thing 1" },
      thing2: { id: "thing 2" },
      thing3: { id: "thing 3" },
      thing4: { id: "thing 4" },
      thing5: { id: "thing 5" },
      thing6: { id: "thing 6" }
    }
  })
})
</script>
```

v-for supports optional 2nd and 3rd params, and it's not unexpected that the 2nd param would be unused.  eslint/no-unused-vars allows for configurations that take this into account, such as in my config: 
```
        "no-unused-vars": [
            "error",
            {
                "args": "after-used",
                "argsIgnorePattern": "^_"
            }
        ],
```

The 'key' variable is unused and therefore causes an error.
[eslint-plugin-vue] [vue/no-unused-vars] 'key' is defined but never used.

michalsnik:
This PR fixes `no-unused-vars` rule, so it doesn't throw a warning, when some of the properties are not used even when the last property was used.

As pointed in #294 currently code like this throws warnings:
```html
<template>
  <!-- 'v'  is defined but never used, 'i' ..., 'c' ... -->
  <div v-for="(v, i, c) in foo">{{c}}</div>
</template>
```

In this PR I'm checking if the last property was referenced anywhere. If yes then we simply ignore other properties, if not -> we check references of all properties as before.

### DIFF
--- a/tests/lib/rules/no-unused-vars.js
+++ b/tests/lib/rules/no-unused-vars.js
@@ -66,12 +66,16 @@ tester.run('no-unused-vars', rule, {
       errors: ["'f' is defined but never used."]
     },
     {
-      code: '<template><div v-for="(v, i, c) in foo"></div></template>',
-      errors: ["'v' is defined but never used.", "'i' is defined but never used.", "'c' is defined but never used."]
+      code: '<template><div v-for="(a, b, c) in foo"></div></template>',
+      errors: ["'a' is defined but never used.", "'b' is defined but never used.", "'c' is defined but never used."]
     },
     {
-      code: '<template><div v-for="(v, i, c) in foo">{{i}}</div></template>',
-      errors: ["'v' is defined but never used.", "'c' is defined but never used."]
+      code: '<template><div v-for="(a, b, c) in foo">{{a}}</div></template>',
+      errors: ["'b' is defined but never used.", "'c' is defined but never used."]
+    },
+    {
+      code: '<template><div v-for="(a, b, c) in foo">{{b}}</div></template>',
+      errors: ["'c' is defined but never used."]
     },
     {
       code: '<template><div v-for="(item, key) in items" :key="item.id">{{item.name}}</div></template>',
--- a/tests/lib/rules/no-unused-vars.js
+++ b/tests/lib/rules/no-unused-vars.js
@@ -40,28 +40,42 @@ tester.run('no-unused-vars', rule, {
     },
     {
       code: '<template scope="props"><span v-if="props"></span></template>'
+    },
+    {
+      code: '<template><div v-for="(item, key) in items" :key="key">{{item.name}}</div></template>'
+    },
+    {
+      code: '<template><div v-for="(v, i, c) in foo">{{c}}</div></template>'
     }
   ],
   invalid: [
     {
       code: '<template><ol v-for="i in 5"><li></li></ol></template>',
-      errors: ['\'i\' is defined but never used.']
+      errors: ["'i' is defined but never used."]
     },
     {
       code: '<template scope="props"></template>',
-      errors: ['\'props\' is defined but never used.']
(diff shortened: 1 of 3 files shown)


mysticatea:
As it happens, there is a discussion that core `no-unused-vars` rule should report all unused variables which are after the last used variable: https://github.com/eslint/eslint/issues/9750

I like that approach.
Would you change this PR by using a `for` loop which goes from last to first?

```js
for (let i = node.variables.length - 1; i >= 0 && node.varibles[i].references.length === 0; --i) {
  // report it.
}

**Comment by michalsnik** [comment]
I think this PR introduces exactly that @mysticatea, it works as follows:
```
<template>
  <div v-for="(v, i, c) in foo">{{v}}</div>
</template>
```
☝️ reports: `i` and `c`

```
<template>
  <div v-for="(v, i, c) in foo">{{i}}</div>
</template>
```
☝️  reports: `c`

```
<template>
  <div v-for="(v, i, c) in foo">{{c}}</div>
</template>
```
☝️  reports nothing

File: tests/lib/rules/no-unused-vars.js [hunk 1]
```diff
@@ -40,28 +40,42 @@ tester.run('no-unused-vars', rule, {
...
+      errors: ["'f' is defined but never used."]
     },
     {
       code: '<template><div v-for="(v, i, c) in foo"></div></template>',
-      errors: ['\'v\' is defined but never used.', '\'i\' is defined but never used.', '\'c\' is defined but never used.']
+      errors: ["'v' is defined but never used.", "'i' is defined but never used.", "'c' is defined but never used."]
+    },
+    {
+      code: '<template><div v-for="(v, i, c) in foo">{{i}}</div></template>',
+      errors: ["'v' is defined but never used.", "'c' is defined but never used."]
```
mysticatea: You are wrong in the 2nd case.

michalsnik: Holly molly! You're absolutely right. Thank you for being so vigilant! I'll update it tomorrow :)

michalsnik:
Should be good now @mysticatea

Issue #248, opened 2017-11-26
Rule proposal: `component-filename-casing`


michalsnik:
Similar to `name-property-casing` this rule would enforce proper filename casing of vue component.
https://vuejs.org/v2/style-guide/#Single-file-component-filename-casing-strongly-recommended

It should be quite easy to do in eslint. While parsing AST we can check for filename and report the root element if it's not ok.

chrisvfritz:
I think this would be good to have, but I'm really torn on what the default should be. 😕

michalsnik:
I think PascalCase is the only right default :) it would be common with the way you import components

chrisvfritz:
The only issue is that PascalCase can cause issues when renaming files on case-insensitive OSes and with Git, e.g. when renaming `Todolist.vue` to `TodoList.vue`. The benefit is that PascalCase works better with autocompletion (and snippets) in editors. In the style guide, we say both are acceptable and use PascalCase in most examples, because it's what most of the team uses.

When setting defaults for other users though, I feel uncomfortable with an out-of-the-box config that can cause subtle issues and even be difficult to diagnose and fix if they've never encountered the problem before. I think ideally, the default would be "either convention, as long as it's consistently one or the other." Is that possible?

If not, I think it might be best to either make kebab-case the default, so that users have to explicitly opt in to a potentially dangerous file naming scheme, or not include it in any config by default.

michalsnik:
This is a valid concern @chrisvfritz. Unfortunately we can't compare one file to another. If we use PascalCase everywhere, but would have a rule with `kebab-case` default, that we can't even put in `recommended` config, I'm not sure we should even implement it. Perhaps it's just one of these rules, that we don't need to have.

michalsnik:
If someone thinks otherwise and have strong opinion, please reopen this issue to continue discussion :) I'm going to close topics that have no conclusion after a while by default.

JonathanDn:
@michalsnik Is there any progress or any plugins / rules you know of that can enforce `kebab-case` `.vue` file names with `eslint-plugin-vue` ?

PR #305, opened 2017-12-25
[New] Add filename-case rule

corbinu:
I use unicorn in my eslint-config and its filename-case rule does not work with Vue conventions so this rule replaces it.

I made this part of the recommended set but happy to take any feedback on changes :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
+|  | [filename-case](./docs/rules/filename-case.md) | enforce file-name conventions |
 | :wrench: | [html-quotes](./docs/rules/html-quotes.md) | enforce quotes style of HTML attributes |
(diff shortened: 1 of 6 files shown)


michalsnik:
Hi @corbinu thanks for working on this. Turns out we have a proposition for a similar rule here: #248 though it's only about Vue components. And because it's eslint plugin for vue I think we should stick here with `.vue` files only and if you don't want eslint-plugin-unicorn to interfere with your Vue files. it's probably a good idea to update the original rule there with additional setting where you could black list some extensions.

corbinu:
@michalsnik Unfortunately that has been discussed before and didn't seem to go anywhere https://github.com/sindresorhus/eslint-plugin-unicorn/issues/23

I can try merging over there is a harder pr to write to allow any file to be passed. If it gets through great would just would hate to do all the extra work if they aren't interested

michalsnik:
I get your point @corbinu , what I proposed was to actually add some kind of `ignoreExtenstions` setting there so it would simply not check files with given extensions. This way other plugins could utilise their own standards for given files - in our case `.vue` files. This should be easier to explain and get approval to work on I think :)

And here I'd stick with rule just for `.vue` files casing anyway.

corbinu:
@michalsnik That is exactly what was proposed in import. Closing for now though

PR #309, opened 2017-12-30
New: vue/script-indent rule (fixes #118)

### LINKED ISSUE #118: Request to allow vue component js to optionally start at an indentation level of 1 when wrapped in <script></script>
**The version of ESLint you are using**: v4.3.0

**The version of eslint-plugin-vue you are using**: 3.8.0

**The problem you want to solve**: Javascript in a vue component expects an initial indentation level of 0 despite being wrapped in `<script></script>`. The following currently reports an error:
```
<script>
    export default {

    };
</script>
```

**Your take on the correct solution to problem**: When modifying a vue component, the linter should optionally allow for an initial indentation level of 1 when wrapped in `<script></script>`.

The [eslint-plugin-html](https://github.com/BenoitZugmeyer/eslint-plugin-html) plugin solves the indentation level issue for html files, and works nicely with eslint to only lint the javascript portion of vue components. Ideally eslint-plugin-vue could work in a similar way while also linting the template portion of the component (as it currently does).

Thanks!

mysticatea:
Fixes #118.

This PR adds `vue/script-indent` rule.
The `vue/script-indent` rule is sharing all logic with `vue/html-indent` rule. The shared logic is at `lib/utils/indent-common.js`.

Also, this PR refactored the tests of `vue/html-indent` rule. New tests load files from `tests/fixtures/html-indent/` then use it as valid cases. And it makes invalid tests from the valid tests automatically -- removing all indentation and checking warning/fixing for correct indentation.
Similarly, the tests of `vue/script-indent` rule are in `tests/fixtures/script-indent/`.


mysticatea:
I have imported `alignAttributesVertically` from #301.

michalsnik:
Please resolve conflicts in your spare time :) I'll try to review it in the next few days

tests/lib/rules/html-indent.js
```diff
@@ -9,18 +9,69 @@
...
+function loadPatterns (additionalValid, additionalInvalid) {
+  const valid = fs
+    .readdirSync(FIXTURE_ROOT)
+    .map(filename => {
+      const code0 = fs.readFileSync(path.join(FIXTURE_ROOT, filename), 'utf8')
+      const code = code0.replace(/^<!--(.+?)-->/, `<!--${filename}-->`)
+      const baseObj = JSON.parse(/^<!--(.+?)-->/.exec(code0)[1])
+      return Object.assign(baseObj, { code, filename })
+    })
+  const invalid = valid
```
michalsnik: It's hard to read and know which fixture is good, or not. You're detecting them here, but I think it would be muuuch easier to just add another comment or put valid fixtures in `valid` directory and so on. Then you wouldn't need to do any confusing magic here. Or am I'm missing something? (It's very possible :D)

mysticatea: I'll add comments about what this does.

This addresses the code in fixtures as valid.
And this creates an invalid case from every valid case -- it removes all indentation and checks whether the rule restores all indentation or not.

mysticatea: So if there are ignored lines, we can't use fixtures. And it's the reason that some tests exist on outside of fixtures.

mysticatea: I added comments.

To realize `invalid` sub-directory, it needs a DSL to specify expected indentation for each line. Indeed it would be useful, but I avoided it for my time.

michalsnik: Fine, now it's much easier to understand. Thank you for explanation and comments :)

PR #190, opened 2017-09-21
[New] Add `html-closing-bracket-newline` (fixes #169)

mysticatea:
This PR adds a new rule to enforce the location of closing brackets `>` of HTML tags.

PR #312, opened 2017-12-31
New: html-closing-bracket-spacing rule (fixes #229)

### LINKED ISSUE #229: Rule proposal: `html-tag-spacing`
As pointed in #199 `html-self-closing` rule fix end tags, but doesn't care about proper spacing before end tags. It is an intended behaviour as one rule should do just one thing.

However currently we don't have a rule that would make it possible to check and fix wrong spacing in HTML tags.

I propose to add a similar rule to: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-tag-spacing.md

Possible settings:
```
{
  "closingSlash": "never | always",
  "beforeSelfClosing": "never | always",
  "afterOpening": "never | always",
  "beforeClosingOpening": "never | always"
}
```

**Incorrect**
```
<!-- closingSlash: "never" -->
<input/ >
<div>< /div>

<!-- beforeSelfClosing: "always" -->
<input/>

<!-- afterOpening: "never" -->
< input />

<!-- beforeClosingOpening: "never" -->
<div ></div>
```

**Correct**
```
<!-- closingSlash: "never" -->
<input/>
<div></div>

<!-- beforeSelfClosing: "always" -->
<input />

<!-- afterOpening: "never" -->
<input />

<!-- beforeClosingOpening: "never" -->
<div></div>
```

mysticatea:
Fixes #229.

This PR adds `vue/html-closing-bracket-spacing` rule. This is the spacing style version of #190. It requires or disallows a space before tag's closing brackets.

### DIFF
--- a/docs/rules/html-closing-bracket-spacing.md
+++ b/docs/rules/html-closing-bracket-spacing.md
@@ -2,7 +2,7 @@
 
 - :wrench: The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) can automatically fix some of the problems reported by this rule.
 
-This rule enforces space style before `>` of tags.
+This rule enforces consistent spacing style before closing brackets `>` of tags.
 
 ```html
 <div class="foo"> or <div class="foo" >
--- a/docs/rules/html-closing-bracket-spacing.md
+++ b/docs/rules/html-closing-bracket-spacing.md
@@ -1,0 +1,83 @@
+# require or disallow a space before tag's closing brackets (html-closing-bracket-spacing)
+
+- :wrench: The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) can automatically fix some of the problems reported by this rule.
+
+This rule enforces space style before `>` of tags.
+
+```html
+<div class="foo"> or <div class="foo" >
+<div class="foo"/> or <div class="foo" />
+```
+
+## Rule Details
+
+This rule has options.
+
+```json
+{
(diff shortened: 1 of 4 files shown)


docs/rules/html-closing-bracket-spacing.md
```diff
@@ -0,0 +1,83 @@
+# require or disallow a space before tag's closing brackets (html-closing-bracket-spacing)
+
+- :wrench: The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) can automatically fix some of the problems reported by this rule.
+
+This rule enforces space style before `>` of tags.
```
michalsnik: `This rule enforces consistent spacing style before closing brackets `>` of elements.`

tests/lib/rules/html-closing-bracket-spacing.js
```diff
@@ -0,0 +1,92 @@
...
+    {
+      code: '<template><div></div',
+      options: [{ startTag: 'never', endTag: 'never' }]
+    }
+  ],
+  invalid: [
+    {
+      code: '<template>\n  <div >\n  </div >\n  <div/>\n</template>',
+      errors: [
+        { message: "Expected no space before '>', but found.", line: 2 },
```
michalsnik: It would be also good to check whether column of the `>` is properly reported.

mysticatea:
I updated this PR.

As a note, this is the range of warnings:

image

michalsnik:
Brilliant 🚀

PR #316, opened 2018-01-01
Add "upcoming" rules category

michalsnik:
This PR adds new category named `upcoming` so we can merge new rules, and show them in separate category with proper description, that those will most likely be enabled by default in next major release. It will serve us as some kind of a buffer. `uncategorized` didn't seem ok for that purpose I think, as we might have some uncategorised rules, that we wouldn't want to actually be enabled anywhere by default.

### DIFF
--- a/tools/update-rules.js
+++ b/tools/update-rules.js
@@ -37,7 +37,10 @@ const categories = [
   'base',
   'essential',
   'strongly-recommended',
-  'recommended'
+  'recommended',
+  'use-with-caution',
+  'uncategorized',
+  'upcoming'
 // Only include categories with existing rules
 ].filter(category =>
   rules.some(entry =>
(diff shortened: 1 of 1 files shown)


mysticatea:
I'm not sure if it's necessary.

- I think `uncategorized` category is enough. It means that "any preset doesn't enable those rules" and upcoming rules are so.
- We have multiple presets (`recommended`, `strongly-recommended`, `essential`, and `base`), so users can't know if the upcoming rule is related to them. 
- We can lazy to decide that which rule should be under which category until the next major release.

michalsnik:
I thought of it like a good place to distinguish which rules are to be put in any of our preset configuration in the next release from rules that simply doesn't belong anywhere. Users will be able to see what's coming soon and be curious about those rules. WDYT @chrisvfritz?

PR #318, opened 2018-01-01
[Fixes #315] Report missing prop types only for assigned arrays and objects

### LINKED ISSUE #315: False positive in rule "vue/require-prop-types"

* **ESLint Version:** 4.14.0
* **eslint-plugin-vue Version:** 4.0.0
* **Node Version:** 8.5.0

```js
// basic configuration
module.exports = {
  root: true,
  extends: [
    'airbnb-base',
    'plugin:vue/recommended',
  ],
  parserOptions: {
    parser: 'babel-eslint'
  },
  plugins: ['vue']
}
```

```js
import props from './Foo.props';

// @vue/component
export default {
  name: 'Foo',
  props,
  render: h => h('div', 'foo'),
};
```

Should not raise require-prop-types error

Setting props in a component that are defined outside of it raises the error:

```
C:\dev\foobar\src\app.js
  6:3   error  Props should at least define their types vue/require-prop-types

```

michalsnik:
#315

### DIFF
--- a/tests/lib/rules/require-prop-types.js
+++ b/tests/lib/rules/require-prop-types.js
@@ -69,6 +69,24 @@ ruleTester.run('require-prop-types', rule, {
         }
       `,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          props
+        }
+      `,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
+          props: externalProps
+        }
+      `,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' }
     }
   ],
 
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-prop-types.js
```diff
@@ -69,6 +69,24 @@ ruleTester.run('require-prop-types', rule, {
         }
       `,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        export default {
```
mysticatea: It's needed to enclose by `<script>` element.

michalsnik: Nope, it's not needed. It works good as is :) It's enough to put `export default {` and `.vue` filename :)

mysticatea: Oh, I got it. I had misunderstood that the tests are using `vue-eslint-parser`.

PR #322, opened 2018-01-02
[Fixes #319] Add Vue.extend support

### LINKED ISSUE #319: Add `Vue.extend` as valid component rule trigger
> All component-related rules are being applied to code that passes any of the following checks:
> - Vue.component() expression
> - export default {} in .vue or .jsx file

**Suggestion**

Add `Vue.extend` to the triggers for for applying the component related rules. It is a commonly used api fo creating components in a raw formats without SFC.

michalsnik:
#319

This PR adds support for detecting `Vue.extend` as components, also it adds missing information about `Vue.mixin` in readme.

Unfortunately I think it should be merged once we're ready to release new major version, as it might potentially break some CI builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ module.exports = {
 All component-related rules are being applied to code that passes any of the following checks:
 
 * `Vue.component()` expression
+* `Vue.extend()` expression
+* `Vue.mixin()` expression
 * `export default {}` in `.vue` or `.jsx` file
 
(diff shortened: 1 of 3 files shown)


blake-newman:
Not sure this is an issue, i have seen in places where `export default Vue.extend` is used. Will this capture this aswell. I think this is used with TS based projects, may not be so much of an issue now.

michalsnik:
It's not an issue, should work fine @blake-newman :) We look for CallExpressions so `export` part is entirely omitted in this check. We only check for `export default` in `.vue` and `.jsx` files to determine it as component.

PR #338, opened 2018-01-08
Fix: unintentional disabling

mysticatea:
I noticed that `<!-- eslint-disable -->` affects to the code in `<script>` element.
This PR fixes `vue/comment-directive` rule to clear state at the end of `<template>` element.

```html
<template>
  <!-- eslint-disable -->
  <div id id="a">Hello</div>
</template>
<script>
  var a // `no-unused-vars` rule didn't trigger.
</script>
```

.... Or I wonder if people expect to share the state between `<template>` and `<script>`? I thought no.

```html
<template>
  <!-- eslint-disable -->
  <div id id="a">Hello</div>
</template>
<script>
  /* eslint-enable */
  var a
</script>
```

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -43,6 +43,12 @@ module.exports = {
               state[group].disableRules.delete(rule)
             }
             break
+          case 'clear':
+            state.block.disableAll = false
+            state.block.disableRules.clear()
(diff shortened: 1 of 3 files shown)


michalsnik:
Great, thank you!

PR #341, opened 2018-01-09
Chore: add `require-meta-docs-url` internal rule

mysticatea:
This PR adds `require-meta-docs-url` internal rule to verify `meta.docs.url` property on all rule files (related to #340). Details:

1. I found [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) rules use `module.exports.create` to distinguish whether a file is an ESLint rule or not. If the `create` property is not a method (e.g. a variable reference), `eslint-plugin-eslint-plugin` rules ignore the file. As the result, most rules in this plugin are ignored.  
  6589d4a fixes the problem. It moves every `craete` function into the RHS of `module.exports`.
2. 61e3a4f adds `require-meta-docs-url` internal rule.
3. 404d433 applies the internal rule to our rules. And I improved those URL to version specific documents.
4. e7c3ec4 modifies `npm version <type>` script to update those URL by new version.

> I have a plan to send `require-meta-docs-url` rule to `eslint-plugin-eslint-plugin`. After I have done it, I will send another PR to upgrade `eslint-plugin-eslint-plugin` plugin.


michalsnik:
Please resolve conflict and feel free to merge it @mysticatea :)

PR #348, opened 2018-01-14
Rule enhancement: add option "consistent" for mustache-interpolation-spacing

g-plane:
Details in docs and test cases.


michalsnik:
Hey @g-plane, I don't think it's a good option. It is a bit misleading, as it's not very consistent actually. This option allows consistent spacing per single interpolation, but then in one file you might end up with different spacing across all interpolation, which is not very consistent. I'd rather stick with `never` and `always` as is - I think those are the only consistent options possible here :)

Issue #299, opened 2017-12-21
Make order-in-components rule fixable


ifraixedes:
**Tell us about your environment**

* **ESLint Version:** 3.19.0
* **eslint-plugin-vue Version:** 4.0.0-beta.3
* **Node Version:** 9.3.0

**Question**

I'm wondering if the rule [order-in-components](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/order-in-components.md) could not be auto fixable by `--fix`

michalsnik:
Hey @ifraixedes 👋  It probably is. Maybe not that easy but should be doable.
Though we have slightly different priorities now, so not every rule have been written with `fix` functionality unfortunately. If you'd like to help us with this one, feel free :) Pull Requests are always welcome 🙌

ifraixedes:
@michalsnik I totally understand that isn't a priority and it was never my intention, in case that my initial comment looked like that.

My question had the goal to poll if it could or not be doable rather than requesting this feature in high priority; although I could imagine that the rule could be auto fixable, I've never developed anything for ESLint, so I wanted to confirm that the rule could be auto fixable.

I don't currently have time now for this, but having the issue labeled as accepted proposition and enhancement is a way to let to the others, not just me 🙂 , that doing so, would be accepted without having to spend time doing a PR without a previous confirmation that it's a nice feature to have.

Cheers!

PR #381, opened 2018-02-07
[Update] Make `vue/order-in-components` fixable

ota-meshi:
This PR makes `vue/order-in-components` fixable. #299
In case of `The "A" property should be above the "B" property` error, autofix will move A before B

### DIFF
--- a/tests/lib/rules/order-in-components.js
+++ b/tests/lib/rules/order-in-components.js
@@ -432,14 +432,7 @@ ruleTester.run('order-in-components', rule, {
         };
       `,
       parserOptions,
-      output: `
-        export default {
-          data() {
-          },
-          test: obj.fn(),
-          name: 'burger',
-        };
-      `,
+      output: null,
       errors: [{
         message: 'The "name" property should be above the "data" property on line 3.',
         line: 6
@@ -457,14 +450,7 @@ ruleTester.run('order-in-components', rule, {
         };
       `,
       parserOptions,
-      output: `
-        export default {
-          data() {
-          },
-          test: new MyClass(),
-          name: 'burger',
-        };
-      `,
+      output: null,
       errors: [{
         message: 'The "name" property should be above the "data" property on line 3.',
         line: 6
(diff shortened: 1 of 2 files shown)


mysticatea:
Thank you for the contribution! And I make apologies that I overlooked this PR. :bow:

Reordering needs to handle carefully because each property value can have side-effects. So ESLint core rules don't support reordering in autofix basically.

In this case, I think we can support reordering if it's sure those properties don't have side-effects.
Would you add the check whether all properties don't have side-effects into the `fix()` function? As details, all computed property keys and all property values don't contain `CallExpression`, `NewExpression`, `UpdateExpression`, `AssignmentExpression`, and `TaggedTemplateExpression`.

ota-meshi:
Thank you for checking.

I changed it to run `fix()` only when it's sure that there are no side effects.

In order to prevent side effects from future ES changes, it runs `fix()` only when all and each of the properties have `node.type` that do not cause side effects as far as I know, in them.

Although implementation has become complex, I think this is the only way to fully prevent troubles. However, it has become complex, so I would appreciate your feedback on whether to actually implement this change.

tests/lib/rules/order-in-components.js
```diff
@@ -283,11 +370,444 @@ ruleTester.run('order-in-components', rule, {
...
+      code: `
+        export default {
+          data() {
+          },
+          test: obj.fn(),
+          name: 'burger',
+        };
+      `,
+      parserOptions,
+      output: `
```
mysticatea: Would you use `output: null` to specify "not fix" explicitly?

ota-meshi: I did not know that. Thank you for the information.

ota-meshi:
Thank you for review!
I changed it again, so please check it.

ota-meshi:
I'm glad you approved!
I appreciate your quick response 😸

bbugh:
This doesn't seem to work when I've customized an extra property that I want to have automatically sorted. I notice a discussion about side effects. Is there a way to tell it to automatically fix "unapproved" properties as well? We want it in a specific order on purpose, not just as a warning. If I remove the item from the rule list (`validations` in this case), then everything sorts again.

michalsnik:
Please fire a separate issue Brian with more informations :)

bbugh:
Thank you for confirming that it is an issue, I will make a new issue.

PR #393, opened 2018-02-16
Docs: update README.md

mysticatea:
This adds two things and removes one thing:

1. Clarifies the requirements for `eslint --fix` and `babel-eslint`.
2. Adds an FAQ entry for `eslint-plugin-html`.
3. Removes an FAQ entry for indentation since we have `vue/script-indent`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ You can try this plugin on the Web.
 ## :grey_exclamation: Requirements
 
 - [ESLint](http://eslint.org/) `>=3.18.0`.
+  - `>=4.7.0` to use `eslint --fix`.
+  - `>=4.14.0` to use with `babel-eslint`.
 - Node.js `>=4.0.0`
 
 ## :cd: Installation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -234,14 +236,20 @@ If you already use other parser (e.g. `"parser": "babel-eslint"`), please move i
 
 The `vue-eslint-parser` uses the parser which is set by `parserOptions.parser` to parse scripts.
 
-### Can my javascript code have increased indentation?
+### Why don't work on `.vue` file?
```
michalsnik: Perhaps this sounds a bit more natural:
`Why doesn't it work on .vue file?`

README.md
```diff
@@ -234,14 +236,20 @@ If you already use other parser (e.g. `"parser": "babel-eslint"`), please move i
...
 
-It depends on the version of eslint you're using.
+1. Make sure you don't have `eslint-plugin-html` in your config. The `eslint-plugin-html` extract the inside of `<script>` tags from the source code, but `eslint-vue-plugin` requires `<script>` tags and `<template>` tags in order to find script and template.
```
michalsnik: 1. `extracts the content from <script> tags, but ...`
2. `in order to distinguish template and script in single file components`

README.md
```diff
@@ -234,14 +236,20 @@ If you already use other parser (e.g. `"parser": "babel-eslint"`), please move i
...
+  ```diff
+    "plugins": [
+      "vue",
+  -   "html"
+    ]
+  ```
 
-You can however use [indent-legacy](https://eslint.org/docs/rules/indent-legacy) rule instead.
-More informations [here](https://eslint.org/docs/user-guide/migrating-to-4.0.0#indent-rewrite).
+2. Make sure your tool does linting for `.vue` files.
```
michalsnik: Perhaps: `Make sure your tool is set to lint .vue files.`

README.md
```diff
@@ -234,14 +236,20 @@ If you already use other parser (e.g. `"parser": "babel-eslint"`), please move i
...
+  - CLI targets only `.js` files by default. You have to specify additional extensions by `--ext` option or glob patterns. E.g. `eslint "src/**/*.{js,vue}"` or `eslint src --ext .vue`.
+  - VSCode targets only JavaScript or HTML files by default. You have to add `{"autoFix": true, "language": "vue"}` into `eslint.validate` entry.
```
michalsnik: `autoFix` is not necessary I guess :)
It's enought to just add another language to array, like so:
```
"eslint.validate": [
        "javascript",
        "javascriptreact",
        "vue",
        "vue-html"
    ],
```

mysticatea: It's necessary because `vscode-eslint` does not autofix on additional file types by default. (https://github.com/Microsoft/vscode-eslint/issues/185#issuecomment-267570350)

Issue #250, opened 2017-11-26
Rule proposal: `component-name-in-template-casing`


michalsnik:
**Style guide:**
https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended

**Description:**
This rule would ensure proper component names in templates. As we only check `.vue` files in terms of templates we can provide both options: `kebab-case` and `PascalCase`

**Good examples:**
- kebab-case
```
<the-component />
```

- PascalCase
```
<TheComponent />
```

**Bad examples:**
```
<theComponent />
<Thecomponent />
<The-component />
```
and obviously with `kebab-case` option it would warn about `PascalCase` and vice-versa. 

Default option would be `PascalCase`.

ota-meshi:
I'm sorry. I made a mistake in refs at PR 402 m(_ _)m

Mouvedia:
> Default option would be PascalCase.

It shouldn't be the default.
cf https://stackoverflow.com/questions/22545621/do-custom-elements-require-a-dash-in-their-name

BTW it could be interesting to add a rule to warn about these:
- annotation-xml
- color-profile
- font-face
- font-face-src
- font-face-uri
- font-face-format
- font-face-name
- missing-glyph

probil:
@Mouvedia, It should https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended
And vue components are not custom elements

Mouvedia:
@probil The rule is called `component-name-in-template-casing` and the style guide that you have linked just above support the kebab-case for DOM templates and "everywhere" else _but_ single file components.
It concludes with:

> **using kebab-case everywhere is also acceptable**

Why? Because using PascalCase in SFC and  kebab-case everywhere else would be inconsistent: we should recommend and support a solution that doesn't require to make that distinction.

probil:
@Mouvedia Agree with you. `PascalCase` is not the best option for everyone. But the same for `kebab-case`. 
I don't think this eslint plugin is able to parse DOM templates at all. Because currently it relies on `*.vue` extension and `// @vue/component` comment so only SFC will be checked. But DOM templates are usually hidden somewhere in the back-end templates. _Correct me please, if I'm wrong here._

> In most projects, component names should always be PascalCase in single-file components

[...]

PR #397, opened 2018-02-19
⭐️New: Add `vue/component-name-in-template-casing`

ota-meshi:
This PR adds `vue/component-name-in-template-casing` rule.
This implements rule proposed in #250


lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,88 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'enforce specific casing for the component naming style in template',
+      category: 'strongly-recommended',
```
michalsnik: Please set `category` to undefined, as it would require us to do major release. And we're not going to do so in the next 2 weeks. We're focusing on minor changes and will do a major release once per 2-3 months probably to avoid confusion.

lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,88 @@
...
+    schema: [
+      {
+        enum: allowedCaseOptions
+      }
+    ]
+  },
+
+  create (context) {
+    const options = context.options[0]
+    const caseType = allowedCaseOptions.indexOf(options) !== -1 ? options : 'PascalCase'
```
michalsnik: You could add `const defaultCase = 'PascalCase'` right below `allowedCaseOptions`, and just use it here. It would be more obv I guess :)

lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,88 @@
...
+        if (!utils.isCustomComponent(node)) {
+          return
+        }
+
+        const name = node.rawName
+        const casingName = casing.getConverter(caseType)(name)
+        if (casingName !== name) {
+          const startTag = node.startTag
+          const open = tokens.getFirstToken(startTag)
+          context.report({
```
michalsnik: Please add empty line above, for clarity

lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,88 @@
...
+            node: open,
+            loc: open.loc,
+            message: 'Component name "{{name}}" is not {{caseType}}.',
+            data: {
+              name,
+              caseType: caseType
```
michalsnik: You can also use shorthand here :)

ota-meshi:
@michalsnik
Thank you for review!
I changed it, so please check it.

lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,90 @@
...
+              const endTag = node.endTag
+              if (!endTag) {
+                return fixer.replaceText(open, `<${casingName}`)
+              }
+              // If we can upgrade requirements to `eslint@>4.1.0`, this code can be replaced by:
+              // return [
+              //   fixer.replaceText(open, `<${casingName}`),
+              //   fixer.replaceText(endTag, `</${casingName}>`)
+              // ]
+              const code = `<${casingName}${sourceCode.text.slice(open.range[1], endTag.range[0])}</${casingName}>`
```
mysticatea: Would you use the first token of `endTag` instead of the node?
I think there is no real problem, but people can put spaces and attributes between `</foo` and `>`. It's unintentional behavior if this rule deletes the spaces and attributes since this rule focuses on name casing.

ota-meshi:
I think we do not have to emphasize html because this rule targets `custom components` and `html elements` like `<div>` and `<span>` are not targets.
However, I am not good at English,  so maybe I haven't understood your intention correctly.

docs/rules/component-name-in-template-casing.md
```diff
@@ -0,0 +1,54 @@
...
+  <TheComponent />
+  <theComponent />
+  <Thecomponent />
+  <The-component />
+</template>
+```
+
+## :wrench: Options
+
+Default casing is set to `PascalCase`.
```
Mouvedia: Since #250 was still open I didn't realise that the PR was already merged.
It shouldn't have been the default for the reasons Iv listed on https://github.com/vuejs/eslint-plugin-vue/issues/250#issuecomment-381360755

lib/rules/component-name-in-template-casing.js
```diff
@@ -0,0 +1,91 @@
...
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+const casing = require('../utils/casing')
+
+const allowedCaseOptions = ['PascalCase', 'kebab-case']
+const defaultCase = 'PascalCase'
```
Mouvedia: https://github.com/vuejs/eslint-plugin-vue/pull/397/files#r184395330

Mouvedia: https://github.com/vuejs/eslint-plugin-vue/pull/397/files#r184395330

shroudedcode:
Any updates on this pull request, @ota-meshi? I would love to see this merged!

trollepierre:
I made the modification in this pull request: https://github.com/vuejs/eslint-plugin-vue/pull/485
can we merge it?

ota-meshi:
I'm sorry to reply late. I recognize the following situation.

1. Rule name
It is proposed that `html-element-name-casing` should be used, but [I think that `component-name-in-template-casing` is good](https://github.com/vuejs/eslint-plugin-vue/pull/397#issuecomment-368066556).
Originally `component-name-in-template-casing` is also the name defined in #250, so I would like to ask the opinion of @michalsnik.

2. Default casing
I personally like `kebab-case`. However, in the official [StyleGuide](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended) of the Vue community, it is recommended to unify it with `PascalCase` first. Since `eslint-plugin-vue` is the official plugin of the Vue community, I think that `PascalCase` should be most easily available.

Either way, I will respect the opinion of the main members(@michalsnik @mysticatea @chrisvfritz), so I would like to receive opinions.

trollepierre:
@michalsnik @mysticatea @chrisvfritz : do you have an opinion of this topic, so we can merge this PR or the other PR I made with the requested change (https://github.com/vuejs/eslint-plugin-vue/pull/485#pullrequestreview-124832375) ?

chrisvfritz:
- Rule name: I personally like `component-name-in-template-casing` for the name. Since in-DOM HTML can _never_ be PascalCase, I think `html-element-name-casing` would be too broad.

- Default casing: Yep, PascalCase should be the default for _components_.

**Here's the issue though**: Can this rule distinguish between Vue components and non-Vue custom elements? Those will still need to be kebab-case. For that reason, I'm not sure this rule is actually feasible to implement.

I think new, uncategorized `html-element-name-kebab-casing` rule could be a good idea for people who prefer that style though, because that would actually implementable.

docs/rules/component-name-in-template-casing.md
samit4me: From my understanding, we can only lint Single File Components and according to the [Vue Style Guide](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended) PascalCase is strongly recommended, so this seems like a sensible default. If you want to use kebab-case it's configurable so it can be either. +1 for this PR, really looking forward to it being merged 🎉

samit4me: It looks like this plugin uses the Style Guide as the source of truth, so If you believe kebab-case should be the default (a lot of people I work with agree) it might be more appropriate to raise a new issue over on the Style Guide repo itself, see https://github.com/vuejs/vuejs.org/blob/e93b8371d73e4467dd8152703ddf1db423f489a2/src/v2/style-guide/index.md#single-file-component-filename-casing-strongly-recommended

chrisvfritz:
@ota-meshi The problem is, I'm not sure it's actually a solvable problem. 😞

And yes, I recommend creating a new issue for that rule. It should be much less controversial than this one and porbably not even require any options. 🙂

ota-meshi:
@chrisvfritz
I added a new issue.#499
Please check whether the contents are consistent.

michalsnik:
Sorry guys for such a long absence..

I think that:
- `component-name-in-template-casing` is a good name
-`PascalCase` is sensible default aligned with our style guide

Also:
- this rule checks only content of SFC's `<template>` tags
- this rule looks only for custom elements, using [this helper](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/index.js#L184) thus it's exactly targeting what name suggests: `component-name`

Given that I think it's ready to be merged. Unless I missed some vital information in all above comments? It would still be in an `unassigned` category for some time, and it would give as a moment to test it in real life projects.

ota-meshi:
@michalsnik  I agree.
If few custom elements are used, I think that it can be solved with option.
I think that it is unlikely that products using Vue.js uses a lot of custom elements, so I think that this rule will be useful.

stevenadams:
@ota-meshi @michalsnik - any reason this rule is still left in undefined category? According to the style guide it should be 'strongly-recommended' 

https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/component-name-in-template-casing.js#L25

lfaoro:
This creates an issue with Vuetify, it'll transform all Vuetify components from e.g. `<v-app>` -> `<VApp>`

zifnab87:
It also creates problems with bootstrap-vue. e.g `<b-row>` --> `<BRow>`

armano2:
@lfaoro @zifnab87 you know that this `<v-app>` and `<b-row>` its going to work?

lfaoro:
as a quick workaround turn it off in the `.eslintrc.js`

```
  rules: {
    "vue/component-name-in-template-casing": "off"
  },
```

ota-meshi:
I did not know that it is must use `kebab-case` when defining components with `kebab-case`.
https://vuejs.org/v2/guide/components-registration.html

This rule reports false positives.

I think that it is necessary to limit the verification target to those registered within the component.
I will fix this rule.

adrlen:
I also have those warnings :

```
warning: Component name "router-link" is not PascalCase (vue/component-name-in-template-casing)
warning: Component name "transition-group" is not PascalCase (vue/component-name-in-template-casing)
etc
```

With Vue native components.

ota-meshi:
@adrlen Hello!
`bootstrap-vue` seems not to be able to use PascalCase, but I think that `router-link` and `transition-group` can also be used in PascalCase.

If you prefer kebab-case you can also change the setting.

```json
  "vue/component-name-in-template-casing": ["error", "kebab-case"]
```

ota-meshi:
I might not notice this thread comments so please open the new issue.

adrlen:
Hi @ota-meshi, thanks for your response !

It works indeed with `Transition`, `TransitionGroup`, but I have this error when using `RouterView` and `RouterLink` :

```
[Vue warn]: Unknown custom element: <RouterView> - did you register the component correctly? For recursive components, make sure to provide the "name" option.
```

ota-meshi:
I don't know the cause 🤔 
Are you installing and using the [vue-router](https://www.npmjs.com/package/vue-router) package?

https://www.npmjs.com/package/vue-router

adrlen:
https://github.com/vuejs/vue-router/releases/tag/v3.0.2

I was using 3.0.1 ;)

Thanks again.

ota-meshi:
@adrlen Thank you for the information!

It seems that PascalCase has become available for v3.0.2. 😄

https://github.com/vuejs/vue-router/pull/1842

https://github.com/vuejs/vue-router/blob/v3.0.2/src/install.js#L46
https://github.com/vuejs/vue-router/blob/v3.0.1/src/install.js#L46

lfaoro:
Still, this rule should be removed from the defaults

PR #402, opened 2018-02-22
[Update] Make `vue/prop-name-casing` fixable

ota-meshi:
This PR makes `vue/prop-name-casing` fixable.
This implements autofix of rules proposed in #252

### DIFF
--- a/lib/rules/prop-name-casing.js
+++ b/lib/rules/prop-name-casing.js
@@ -1,0 +1,73 @@
+/**
+ * @fileoverview Requires specific casing for the Prop name in Vue components
+ * @author Yu Kimura
+ */
+'use strict'
+
+const utils = require('../utils')
+const casing = require('../utils/casing')
+const allowedCaseOptions = ['camelCase', 'snake_case']
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+function create (context) {
+  const options = context.options[0]
+  const caseType = allowedCaseOptions.indexOf(options) !== -1 ? options : 'camelCase'
+  const converter = casing.getConverter(caseType)
+
+  // ----------------------------------------------------------------------
+  // Public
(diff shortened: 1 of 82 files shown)


lib/rules/prop-name-casing.js
```diff
@@ -46,7 +56,12 @@ function create (context) {
           data: {
             name: propName,
             caseType: caseType
-          }
+          },
+          fix: canFixPropertyName(item) ? fixer => {
+            return item.key.type === 'Literal'
+              ? fixer.replaceText(item.key, item.key.raw.replace(item.key.value, convertedName))
```
mysticatea: If the original name is `"🍻"`, what does it happen?

ota-meshi: Why!? It became "" !!!
Emoji is difficult...

ota-meshi:
Thanks for review!
I changed it. to auto-fix if only for ASCII printable characters.

tests/lib/rules/prop-name-casing.js
```diff
@@ -189,12 +217,156 @@ ruleTester.run('prop-name-casing', rule, {
...
+        export default {
+          props: {
+            [greeting_text]: String
+          }
+        }
+      `,
+      output: null,
+      parserOptions,
+      errors: [{
+        // bug ?
```
michalsnik: I guess so, it shouldn't throw error in this case I think.

michalsnik: But we can deal with it in separate task.

ota-meshi: Thank you for your view
I wrote an issue #442

F3n67u:
Make prop-name-casing fixable is very dangerous, because the fix code only fix the props name and leave the reference code same. I think the fix should leave to manual, we just show the alert and do nothing else.

Issue #403, opened 2018-02-22
Rule proposal: `no-use-v-if-with-v-for`


ota-meshi:
**Please describe what the rule should do:**

This rule warns never use v-if on the same element as v-for.

Style guide:
https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential

**What category of rule is this? (place an "X" next to just one item)**

[X] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about:**

```xml
<ul>
  <li
    v-for="user in users"
    v-if="user.isActive"
    :key="user.id"
  >
    {{ user.name }}
  </li>
</ul>
```

**Why should this rule be included?**

This rule is stated in the Style guide as ESSENTIAL.
It is similar to [`vue/no-confusing-v-for-v-if`](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/no-confusing-v-for-v-if.md), but the content of `code examples` is not warned.

I confirmed issues, but I could not find duplicate rules.
I'm sorry if it was duplicated.

mysticatea:
Thank you for this issue.

Looks good to me.
Maybe the existing `vue/no-confusing-v-for-v-if` rule should be deprecated after we add `vue/no-use-v-if-with-v-for` rule.

cc: @chrisvfritz @michalsnik

ota-meshi:
Thank you for considering the proposal !!

chrisvfritz:
I like it too. 🙂

kenneyHong:
> I like it too. 🙂
I do not like(我不喜欢)

PR #406, opened 2018-02-23
⭐️New: Add `vue/no-use-v-if-with-v-for` rule

ota-meshi:
This PR adds `vue/no-use-v-if-with-v-for` rule.
This implements rule proposed in #403

### DIFF
--- a/docs/rules/no-use-v-if-with-v-for.md
+++ b/docs/rules/no-use-v-if-with-v-for.md
@@ -80,8 +80,8 @@ computed: {
 
 ```html
 <TodoItem
-  v-if="complete"
   v-for="todo in todos"
+  v-if="todo.shown"
   :todo="todo"
 />
 ```
@@ -90,8 +90,8 @@ computed: {
 
 ```html
 <TodoItem
+  v-if="complete"
   v-for="todo in todos"
-  v-if="todo.shown"
   :todo="todo"
 />
 ```
--- a/docs/rules/no-use-v-if-with-v-for.md
+++ b/docs/rules/no-use-v-if-with-v-for.md
@@ -1,0 +1,98 @@
+# disallow use v-if on the same element as v-for. (vue/no-use-v-if-with-v-for)
+
+> Never use `v-if` on the same element as `v-for`.
+>
+> There are two common cases where this can be tempting:
+>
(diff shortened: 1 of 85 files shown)


docs/rules/no-use-v-if-with-v-for.md
```diff
@@ -0,0 +1,98 @@
...
+
+`allowUsingIterationVar` - Enables The `v-if` directive use the reference which is to the variables which are defined by the `v-for` directives.
+
+```js
+'vue/no-use-v-if-with-v-for': ['error', {
+  allowUsingIterationVar: true // default: false
+}]
+```
+
+:+1: Examples of additional **correct** code for this rule with sample `"allowUsingIterationVar": true` options:
```
michalsnik: I guess this is supposed to be `incorrect` right?

docs/rules/no-use-v-if-with-v-for.md
```diff
@@ -0,0 +1,98 @@
...
+
+```html
+<TodoItem
+  v-if="complete"
+  v-for="todo in todos"
+  :todo="todo"
+/>
+```
+
+:-1: Examples of additional **incorrect** code for this rule with sample `"allowUsingIterationVar": true` options:
```
michalsnik: And this correct

lib/rules/no-use-v-if-with-v-for.js
```diff
@@ -0,0 +1,85 @@
...
+      "VAttribute[directive=true][key.name='if']" (node) {
+        const element = node.parent.parent
+
+        if (utils.hasDirective(element, 'for')) {
+          if (isUsingIterationVar(node)) {
+            if (!allowUsingIterationVar) {
+              context.report({
+                node,
+                loc: node.loc,
+                message: "The 'v-for' list variable should be replace with a new computed property that returns your filtered list by this 'v-if' condition."
```
michalsnik: Can we make `list` to be dynamic value in this message? Also what do you think about this kind of message?
```
The 'someArray' variable inside 'v-for' directive should be replaced with a computed property that returns filtered array instead. You should not mix 'v-for' with 'v-if'.
```

ota-meshi:
@michalsnik Thank you for review!
I changed it, so please check it.

PR #412, opened 2018-02-26
[Update] Make 'vue/attributes-order' fixable

mokkabonna:
I should probably add docs etc. But I prefer to get some feedback first. I am not very familiar with the eslint api, so I might be doing something wrong.

However it worked fine for me, I had an app with 30+ files that now failed because of this added rule. Manually applying the fix in this PR did wonders for me. After running fix, I have 0 errors.


mokkabonna:
Lol, after changing the title I realized the other identically titled PR!

Closing in favor of that one, although my implementation is shorter. It might be missing some corner cases though..

PR #421, opened 2018-03-06
🐛Fix: v-bind order in `vue/attribute-order` rule

### LINKED ISSUE #410: v-bind attribute should be sorted by its real name
In new rule `vue/attributes-order`, **v-bind** has an order as same as v-model, but I think it should not be regarded as a single group. For example **v-bind:is** should be placed at first but not together with other attributes.

### LINKED ISSUE #422: vue/attributes-order: v-dom-portal errors if placed after v-if

* **ESLint Version: 4.14.0**
* **eslint-plugin-vue Version: 4.3.0**
* **Node Version: 8.9.4**

```
"vue/attributes-order": [2, {
          order: [
              "DEFINITION",
              "LIST_RENDERING",
              "CONDITIONALS",
              "RENDER_MODIFIERS",
              "GLOBAL",
              "UNIQUE",
              "CONTENT",
              "OTHER_ATTR",
              "BINDING",
              "EVENTS"
          ]
      }]

```

```vue
<div v-if="isVisible" v-dom-portal="true" :class="classes"></div>
```

v-dom-portal directive should not fail linting, as it should be OTHER_ATTR imo

Instead it seems to be recognised as either DEFINITION or LIST_RENDERING, and the following message displays:

```
Attribute "v-dom-portal" should go before "v-if". (vue/attributes-order)
```

ydfzgyj:
close #410 , close #422


docs/rules/attributes-order.md
```diff
@@ -17,10 +17,10 @@ ex: 'v-once', 'v-pre'
 ex: 'id'
 - UNIQUE
 ex: 'ref', 'key', 'slot'
-- BINDING
-ex: 'v-model', 'v-bind', ':property="foo"'
+- TWO\_WAY\_BINDING
+ex: 'v-model'
 - OTHER_ATTR
-ex: 'customProp="foo"'
+ex: 'custom-prop="foo"', 'v-bind:prop="foo"', ':prop="foo"', 'v-custom-directive'
```
michalsnik: I don't think we should mix other attributes with bindings.
Let's keep both groups: `TWO_WAY_BINDINGS` and `BINDINGS`

ydfzgyj: It's a common situation that when user needs to change a static property value to a dynamic binding value. If we separate them into two groups, the user may have to modify the order of these properties and feel confused. Maybe it's still better to let user themselves to sort these attributes.

ydfzgyj:
Only solved the conflict but did not change anything else

docs/rules/attributes-order.md
```diff
@@ -4,7 +4,7 @@
 
 ## :book: Rule Details
 
-This rule aims to enforce ordering of component attributes. The default order is specified in the [Vue styleguide](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended) and is:
+This rule aims to enfore ordering of component attributes. The default order is specified in the [Vue styleguide](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended) and is:
```
chrisvfritz: Looks like a typo here.

Issue #434, opened 2018-03-16
Rule Proposal: `no-v-html` to prevent potential XSS attacks


n-zeplo:
Please describe what the rule should do:

This rule errors or warns developers of the use of v-html and its potential to lead to XSS attacks.

What category of rule is this? (place an "X" next to just one item)

[ ] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[X] Other (please specify:) Security enhancement

Provide 2-3 code examples that this rule will warn about:

<template>
  <div v-html="userHTML"></div>
</template>

<template>
  <ul v-html="userHTML"></ul>
</template>

Why should this rule be included?

Injecting unescaped html into the view can be unsafe depending on the source of the html. If it ties into user input it has the potential for [XSS attacks](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)). This rule should be able to set as warning and optional as this directive might be needed for trusted HTML coming from the server side. 

Similar implementation in [react eslint ruleset](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-danger.md).

PR #435, opened 2018-03-18
⭐️New: Add `vue/no-v-html` rule

n-zeplo:
This PR adds vue/no-v-html rule.
This implements rule proposed in #434

### DIFF
--- a/lib/rules/no-v-html.js
+++ b/lib/rules/no-v-html.js
@@ -1,0 +1,33 @@
+/**
+ * @fileoverview Restrict or warn use of v-html to prevent XSS attack
+ * @author Nathan Zeplowitz
+ */
+'use strict'
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lib/rules/no-v-html.js
```diff
@@ -0,0 +1,33 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definitionutilu
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'disallow use of v-html to prevent XSS attack',
+      category: 'security enhancement',
```
michalsnik: Please use `undefined` as category. We're going to assign categories in major releases, as it results in automatic generation of sharable configs.

n-zeplo: I've updated the PR with this change.

n-zeplo:
Thanks @michalsnik! Made the change you have requested.

Issue #415, opened 2018-03-01
rule-proposal: add a new line after html-closing-bracket


Alendorff:
**What category of rule is this? (place an "X" next to just one item)**

[X] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something

It could be a very convenient with the `html-closing-bracket-newline` rule usage.

**Motivation:**

Currently, this piece of html a fully valid when used with vue/recommended and this rule:

```
"vue/html-closing-bracket-newline": ["error", {
      "singleline": "never",
      "multiline": "always"
    }],
```

They are all different:

```html
<li>
  <a
    href="http://router.vuejs.org/"
    target="_blank"
  >vue-router</a>
</li>
<li>
  <a
    href="http://vuex.vuejs.org/"
    target="_blank"
  >
    vuex
  </a>
</li>
<li>
  <a
    href="http://vue-loader.vuejs.org/"
    target="_blank"
  >vue-loader
  </a>
</li>
```

**Suggestion**

With suggested rule it could looks like the following:

```html
<li>
  <a
    href="http://router.vuejs.org/"
    target="_blank"
  >
    vue-router
  </a>
</li>
```

```html
<div>
  <div
    some="prop"
    another="prop"
  >
    <some-html>
      <even-more />
    </some-html>
  </div>
</li>
```

But If there is no children I suggest no newline.

```html
<div>
  <div some="prop"
       another="prop"
  ></div>
</li>
```

michalsnik:
I feel like I've heard this proposition before 🤔Perhaps we were talking about it while discussing about two original rules. I don't think that we should add another responsibility for this rule. We should rather add new one, e.g. `html-content-new-line`, which would make sure that in case the element is multiline one, then its content shouldn't be at the same line as the bracket. I think it makes sense! :)

michalsnik:
I see it like this:

**Settings:**
```
"vue/html-content-newline": ["error", {
  "singleline": "ignore",
  "multiline": "always"
}],
```

**Incorrect:**
```html
<div
  class="panel"
>content</div>  
```

**Correct:**
```html
<div class="panel">content</div>
<div
  class="panel"
>
  content
</div>
```

----------------------------

**Settings:**
```
"vue/html-content-newline": ["error", {
  "singleline": "always",
  "multiline": "never"
}],
```

**Incorrect:**
```html
<div class="panel">panel</div>

<div
  class="panel"
>
  content
</div>  
```

[...]

PR #445, opened 2018-03-29
⭐️New: Add `vue/html-content-newline`

ota-meshi:
This PR adds `vue/html-content-newline` rule.
This implements rule proposed in #415


michalsnik:
We've talked with @chrisvfritz and this rule caused a really vivid conversation. In result we came up with an updated idea. What if we name it `html-tag-attrs-content-new-line`, and force putting content on new line whenever the HTML tag has any attribute, so that we don't mix the content and attributes on the same line? The rule would be very simple to use, it would either be ON or OFF.

**Correct code:**
```
<label>Lorem ipsum</label>
<div class="panel">
  content
</div>
```

Alendorff:
> we name it html-tag-attrs-content-new-line, and force putting content on new line whenever the HTML tag has any attribute

I think naming is minor thing, because rules are typically googled by use-cases or read directly at the documentation... But about behavior, I think it shouldn't conflict with the rule `max-attributes-per-line`. 

For example I currently have such config:
```javascript
    "vue/max-attributes-per-line": [
      2,
      {
        "singleline": 2,
        "multiline": {
          "max": 1,
          "allowFirstLine": true
        }
      }
    ]
```
And this is valid examples:
```html
<div arg1 arg2>Valid</div>

<div arg1
     arg2
     arg3
>
  Valid too
</div>
```
So if I'll apply this new rule I would still expect the same behavior. But the rule should help to fight with different code which I've mentioned at the [issue](https://github.com/vuejs/eslint-plugin-vue/issues/415).
So I would expect the following code as a valid output:
```html
<div arg1>Valid</div>

<div arg1 arg2>Valid</div>

<div arg1>
  Valid
</div>

<div arg1 arg2>
  Valid
</div>

<div arg1 
     arg2
>
  Valid
</div>
```

chrisvfritz:
> I think it shouldn't conflict with the rule `max-attributes-per-line`.

I agree with this. So for example, I wouldn't expect this to be invalid:

```vue
<div
  class="panel"
>content</div>
```

Because the attribute _is_ technically on a new line from the content. _However_ @Alendorff, I would expect these to be invalid:

```vue
<div arg1>content</div>

<div arg1 arg2>content</div>
```

And autofixed to:

```vue
<div arg1>
  content
</div>

<div arg1 arg2>
  content
</div>
```

Because I don't see how it conflicts with `max-attributes-per-line` - though apologies if I've missed something. The last example you gave would also be valid:

```vue
<div arg1 
     arg2
>
  content
</div>
```

Does that make sense?

Alendorff:
Got it, thanks. Yes, make sense.

ota-meshi:
@michalsnik @chrisvfritz 

> html-tag-attrs-content-new-line

Sounds good to me.

However, I would like to enforce new line in the following html after `<tr>` and before `</ tr>`.

```html
<tr><td>cell1</td>
<td>cell2</td>
<td>cell3</td></tr>
```

I think it is necessary to have a different rule from `html-tag-attrs-content-new-line`.
That rule enforces a new line after the start tag and before the end tag if the content or children are multiple lines.

chrisvfritz:
@ota-meshi I agree would be useful! 🙂 Although with this specific proposal:

> That rule enforces a new line after the start tag and before the end tag if the content or children are multiple lines.

Then we may have an issue with autofix, because the execution order would become important. For example, with this code:

```js
<tr><td foo>cell1</td></tr>
```

We'd have to ensure that any other rules that add/remove newlines is run first. In the case of `html-tag-attrs-content-new-line`, it would _make_ the content multiline where it wasn't before:

```js
<tr><td foo>
  cell1
</td></tr>
```

Without controlling the execution order, users would have situations where they'd have to ESLint with `--fix` multiple times to actually fix everything.

@michalsnik Is controlling the execution order of rules something that's feasible or do we need to think of a slightly different rule that doesn't rely on it?

chrisvfritz:
> Since eslint verifies up to 10 times until autofix is resolved, I think the possibility of this issue remaining is very low.

@ota-meshi Oh, cool! I didn't know that. 🙂

ota-meshi:
Closing to continue working with #551 and #552.

PR #446, opened 2018-03-29
Idea to allow eslint to skip over and allow <script lang="coffee"> in *.vue files

shreeve:
I'd like to use eslint in my *.vue files, but it bombs out with coffeescript, when using <script lang="coffee">.

Instead of disabling eslint for all *.vue files, this change skips over the coffeescript.

This is probably a terrible approach, but perhaps there's a way that the preprocess function could be overridden by user config?

### DIFF
--- a/lib/processor.js
+++ b/lib/processor.js
@@ -3,8 +3,17 @@
  */
 'use strict'
 
+// detect <script lang="coffee">
+const coffee = /^<script\s+lang=["']coffee[^]*^<\/script>/m
+
 module.exports = {
   preprocess (code) {
+
+    // do not lint coffeescript
+    code = code.replace(coffee, source => {
+      source.replace(/[^\n]/g, '')
+    })
+
     return [code]
   },
 


michalsnik:
Hello @shreeve I'm not sure what would be the best solution for this issue, hence I requested review from @mysticatea that actually crated the whole parser for vue.
In terms of typescript though, I guess you can use `typescript-eslint-parser` to make eslint know what's going on there. Unfortunately there's no coffee parser.

shreeve:
@michalsnik - Cool, thanks. @mysticatea - Any suggestions on how to support coffeescript in *.vue files?

mysticatea:
Hi. I apologize for the delay. I was inactive for several months.

I understand that it's inconvenient for CoffeeScript users.

ESLint checks AST which is defined by [this page](https://eslint.org/docs/developer-guide/working-with-custom-parsers#the-ast-specification). We can use custom parsers, so we can check the source code of various kinds with ESLint.

This plugin uses [vue-eslint-parser](https://github.com/mysticatea/vue-eslint-parser) to parse `.vue` files. The `vue-eslint-parser` parses the files as HTML then parses `<script>` tags with the configured parser. People can configure the script parser with `parserOptions.parser` property of the `.eslintrc` file. If it's omtited, `vue-eslint-parser` uses [espree](https://github.com/eslint/espree), the default parser of ESLint, to parse `<script>` tags.

TypeScript users can use [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser) to parse `<script>` tags.
But I don't know if there is the custom parser for CoffeeScript.
If you want to support CoffeeScript, probably you have to make a custom parser. Documentation is here: https://eslint.org/docs/developer-guide/working-with-custom-parsers

I added the functionality to skip parsing `<script>` tags completely into `vue-eslint-parser`: https://github.com/mysticatea/vue-eslint-parser/commit/9b947b11fac73ac5baac1fee35c19b47afb21f1c
It will allow us to check only templates.

armano2:
@shreeve you can use this parser
https://www.npmjs.com/package/eslint-plugin-coffeescript

PR #458, opened 2018-04-12
docs: fix typo in plugin name

jmosul:
Believe this should be that way around.  Blindly tried to use it in an `npm install`.   Suppose I should read what I'm installing in more detail...

But so the next person isn't confused as simply as I was :joy:


michalsnik:
Thanks @jmosul it certainly wasn't intended :D

PR #461, opened 2018-04-18
Update attribute-hyphenation to allow a custom ignore list

MechJosh0:
Changes:
- By default it'll still ignore `data-` and `aria-`, added to the default ignore list is `slot-scope` as this is a Vue attribute that cannot be un-hyphenated.
- Users can now do `"vue/attribute-hyphenation": ["error", "never", {"ignore": ["data-", "aria-", "my-prop"]}]` to overwrite the existing ignore list.
- Added `/.idea` folder created by Phpstorm to the `/.gitignore` list.

### DIFF
--- a/docs/rules/attribute-hyphenation.md
+++ b/docs/rules/attribute-hyphenation.md
@@ -1,17 +1,17 @@
-# enforce attribute naming style in template (vue/attribute-hyphenation)
+# enforce attribute naming style on custom components in template (vue/attribute-hyphenation)
 
 - :gear: This rule is included in `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"`.
 - :wrench: The `--fix` option on the [command line](http://eslint.org/docs/user-guide/command-line-interface#fix) can automatically fix some of the problems reported by this rule.
 
 ## :wrench: Options
 
-Default casing is set to `always`
+Default casing is set to `always` with `['data-', 'aria-', 'slot-scope']` set to be ignored
 
 ```
-'vue/attribute-hyphenation': [2, 'always'|'never']
+'vue/attribute-hyphenation': [2, 'always'|'never', { 'ignore': ['data-', 'aria-', 'slot-scope'] }]
 ```
 
-### `"always"` - Use hyphenated name. (It errors on upper case letters.)
+### `[2, "always"]` - Use hyphenated name. (It errors on upper case letters.)
 
 :+1: Examples of **correct** code`:
 
@@ -25,7 +25,7 @@ Default casing is set to `always`
 <MyComponent myProp="prop"/>
 ```
 
-### `"never"` - Don't use hyphenated name. (It errors on hyphens except `data-` and `aria-`.)
+### `[2, "never"]` - Don't use hyphenated name. (It errors on hyphens except `data-`, `aria-` and `slot-scope-`.)
 
 :+1: Examples of **correct** code`:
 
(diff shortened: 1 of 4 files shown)


docs/rules/attribute-hyphenation.md
```diff
@@ -1,17 +1,17 @@
-# enforce attribute naming style in template (vue/attribute-hyphenation)
+# enforce attribute naming style on custom components in template (vue/attribute-hyphenation)
```
michalsnik: This part is automatically generated based on the rule's metadata informations. If you want to update it please - do so in rule file instead. Otherwise it'll be overwritten during next release.

lib/rules/attribute-hyphenation.js
```diff
@@ -22,14 +22,40 @@ module.exports = {
...
-    const useHyphenated = options !== 'never'
+    const option = context.options[0]
+    const optionsPayload = context.options[1]
+    const useHyphenated = option !== 'never'
+    let ignoredAttributes = []
+
+    if (optionsPayload && optionsPayload.ignore) {
+      ignoredAttributes = optionsPayload.ignore
+    } else {
+      ignoredAttributes = ['data-', 'aria-', 'slot-scope']
```
michalsnik: You can set ignoredAttributes above and remove the whole `else` clause.

lib/rules/attribute-hyphenation.js
```diff
@@ -48,9 +74,14 @@ module.exports = {
     }
 
     function isIgnoredAttribute (value) {
-      if (value.indexOf('data-') !== -1 || value.indexOf('aria-') !== -1) {
+      const isIgnored = !ignoredAttributes.every(function (attr) {
```
michalsnik: Perhaps you can use `.some` and remove the negation?
```
ignoredAttributes.some(function (attr) {
  return value.indexOf(attr) !== -1;
});
```

docs/rules/attribute-hyphenation.md
```diff
@@ -38,3 +38,17 @@ Default casing is set to `always`
 ```html
 <MyComponent my-prop="prop"/>
 ```
+
+### `[2, "never", { 'ignore': ['data-', 'aria-', 'slot-scope', 'custom-prop'] }]` - Don't use hyphenated name but allow custom attributes
```
michalsnik: I wonder whether it wouldn't be easier for users to just merge their settings with default ones. I can't imagine a case where you'd want to remote any of the default settings - they are sort of required for this rule to work properly. So I'd just assume these setting are always there and whatever user adds here - will be merged to the default array.
What do you think about it?

lib/rules/attribute-hyphenation.js
MechJosh0: This makes sense, I have done this change.

docs/rules/attribute-hyphenation.md
MechJosh0: Is it just this one line coming from `meta.docs.description` and needs updating? If so I've done it, let me know if there is anything else.

docs/rules/attribute-hyphenation.md
MechJosh0: I was providing more flexibility for long term use if things change, but I do agree that it's unlikely people will want to overwrite these defaults. I have updated it so it will not overwrite the default array but instead `push` to it.

lib/rules/attribute-hyphenation.js
MechJosh0: That does make more sense, thanks.

tests/lib/rules/attribute-hyphenation.js
```diff
@@ -30,18 +30,23 @@ ruleTester.run('attribute-hyphenation', rule, {
...
     {
       filename: 'test.vue',
-      code: '<template><div data-id="foo" aria-test="bar"><a onClick="" my-prop="prop"></a></div></template>',
+      code: '<template><div data-id="foo" aria-test="bar" slot-scope="{ data }"><a onClick="" my-prop="prop"></a></div></template>',
       options: ['never']
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hypen="foo"><a onClick="" my-prop="prop"></a></div></template>',
+      options: ['never', { 'ignore': ['data-', 'aria-', 'slot-scope', 'custom-hypen'] }]
```
michalsnik: Ah, one thing you missed. We don't need to pass `'data-', 'aria-', 'slot-scope'` anymore :)

PR #464, opened 2018-04-21
(Fixes #420) Fix false positive in no-side-effect-in-cp

### LINKED ISSUE #420: False positive in vue/no-side-effects-in-computed-properties

Node v9.3.0
eslint-plugin-vue v

* **ESLint Version:** 4.18.0
* **eslint-plugin-vue Version:** 4.3.0
* **Node Version:** 9.3.0

```
module.exports = {
  root: true,
  env: {
    browser: true,
    node: true
  },
  parserOptions: {
    parser: 'babel-eslint'
  },
  extends: [
    'plugin:vue/essential'
  ],
  plugins: [
    'vue'
  ],
}
```

```vue
computed: {
   categorized() {
      const categories = {}

      this.types.forEach(c => {
        categories[c.category] = categories[c.category] || []
        categories[c.category].push(c)
      })

      return categories
    },
},
```

This function takes an array of types, each of which has a category property, and returns an array of categories, each of which has an array of types within that category.

This function is appropriate as a computed function as it does not modify `this.types`. 

> 47:7  error  Unexpected side effect in "categorized" computed property  vue/no-side-effects-in-computed-properties
>✖ 1 problem (1 error, 0 warnings)

The issue is with line 7 from the snippet: `categories[c.category].push(c)`. The regex for `no-side-effects-in-computed-properties` catches ANY use of `push`, even if it's on unrelated arrays. It even catches `push` calls within comments, for instance:
```
   unacceptable() {
      return this.types.map(t => {
        // [].push('xxx')
        return t
   })
```

I'm not certain what an appropriate fix is for this. Perhaps it should try to verify that the array being modified is not within the scope of the computed function?

michalsnik:
This PR fixes #420, by introducing parser that simplifies given MemberExpression or CallExpression to version that can be properly verified in terms of potential side-effect in the chain.

So for example this kind of code:
```js
this.categories['test']
  .map(c => c.items)
  .sort()[0]
  .anotherProp
  .slice(1)
  .reduce((acc, item) => {
    // item.push(10)
    return acc + item.length;
  }, 0)
  .length
```

Is being simplified to this:
```
this.categories.test.map().sort()[].anotherProp.slice().reduce().length
```

And now we can properly check for potential side-effects in the given chain without potential false-positives hidden in inner scopes.

### DIFF
--- a/lib/rules/no-side-effects-in-computed-properties.js
+++ b/lib/rules/no-side-effects-in-computed-properties.js
@@ -41,7 +41,7 @@ module.exports = {
         },
         // this.xxx.func()
         'CallExpression' (node) {
-          const code = context.getSourceCode().getText(node)
+          const code = utils.parseMemberOrCallExpression(node)
           const MUTATION_REGEX = /(this.)((?!(concat|slice|map|filter)\().)[^\)]*((push|pop|shift|unshift|reverse|splice|sort|copyWithin|fill)\()/g
 
           if (MUTATION_REGEX.test(code)) {
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -617,5 +617,42 @@ module.exports = {
       return
     }
     return body.errors.some(error => typeof error.code === 'string' && error.code.startsWith('eof-'))
+  },
+
+  /**
+   * Parse CallExpression or MemberExpression to get simplified version without arguments
+   *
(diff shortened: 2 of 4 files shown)


michalsnik:
Let's do it in separate PR, you can fire an issue to discuss it @chrisvfritz. Although everything about this rule is a balance between being useful and not full of false positives. But we can discuss ofc :)

PR #473, opened 2018-05-07
[Docs] Single File Component Usage

deckar01:
Document the single file component configuration in the usage section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,17 @@ module.exports = {
 }
 ```
 
+### Single File Components
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,17 @@ module.exports = {
 }
 ```
 
+### Single File Components
+
+Eslint only targets `.js` files by default. You must include the `.vue` extension using [the `--ext` option](https://eslint.org/docs/user-guide/configuring#specifying-file-extensions-to-lint) or a glob pattern.
```
gaomd: Should be `ESLint`, the capital `E`.

PR #476, opened 2018-05-17
Fix: Allow props with `validator` without `type` in `vue/require-prop-types`

st-sloth:
According to [the docs](https://vuejs.org/v2/guide/components-props.html#Prop-Validation), if the `type` property is insufficient and more complex type-checking is required, one can use the `validator` property. Currently `eslint-plugin-vue` reports an error in such cases.

### DIFF
--- a/docs/rules/require-prop-types.md
+++ b/docs/rules/require-prop-types.md
@@ -17,24 +17,24 @@ props: ['status']
 :+1: Examples of **correct** code for this rule:
 
 ```js
+// Without options, just type reference
 props: {
   status: String
 }
 ```
 
 ```js
+// With options with type field
 props: {
   status: {
     type: String,
     required: true,
-    validate: function (value) {
-      return ['syncing', 'synced', 'version-conflict', 'error'].indexOf(value) !== -1
-    }
   }
 }
 ```
 
 ```js
+// With options without type field but with validator field
 props: {
   status: {
     required: true,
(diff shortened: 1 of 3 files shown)


docs/rules/require-prop-types.md
```diff
@@ -34,6 +34,20 @@ props: {
 }
 ```
 
+```js
+props: {
+  status: {
+    required: true,
+    validator: function (value) {
```
michalsnik: Or actually, perhaps we can remove this code, and use above example with `validate` property changed to proper `validator`?

st-sloth: I didn't feel right editing another example at the time. Now I looked up vue code and found no mention of `validate` as prop option at all. So it makes sense to change it to `validator`. However, I think it would be clearer if the docs contained at least three cases:
```js
// Without options, just type reference
props: {
  status: String
}

// With options with type field
props: {
  status: {
    type: String,
    required: true
  }
}

// With options without type field but with validator field
props: {
  status: {
    required: true,
    validator: function (value) {
      return (
        value === null ||
        Array.isArray(value) && value.length > 0
      )
    }
  }
}
```
so the readers understand that `type` is not required if `validator` is present. There is also the case when both `type` and `validator` are used but I don't think it's relevant here since both are covered separately.

So what if I remove `validate` from the second example, so it's just about `type`, and also add explicit comments like above?

michalsnik: Sounds good to me @st-sloth , go for it :)

st-sloth:
Done! Updated docs and rebased on master.

PR #484, opened 2018-05-30
chore: remove yarn-error.log

sinchang:


.gitignore
```diff
@@ -5,3 +5,4 @@
 /node_modules
 /test.*
 yarn.lock
+yarn-error.log
```
shroudedcode: There should be a final newline here. See the [`.editorconfig`](https://github.com/vuejs/eslint-plugin-vue/blob/master/.editorconfig#L8).
Otherwise I fully agree with you, there's no need to add this file to version control.

PR #504, opened 2018-06-24
⭐️ New: Support ESLint 5

ChristianMurphy:
https://eslint.org/blog/2018/06/eslint-v5.0.0-released
ESLint 5 drops support for Node 4, updated CI accordingly.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,7 @@ dependencies:
 
 test:
   override:
-    - nvm use 4 && npm test
+    - nvm use 4 && npm i eslint@4.19.1 --no-save && npm test
     - nvm use 6 && npm test
     - nvm use 8 && npm test
     # Test for the minimum version we are supporting.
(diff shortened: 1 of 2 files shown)


mysticatea:
Thank you for this PR.

I'm seeing a parser problem: https://github.com/mysticatea/vue-eslint-parser/issues/26
I think that we have to resolve that issue before support 5.0.0.

patrickabkarian:
How about now

michalsnik:
I think this change should be released in a new major version of eslint-plugin-vue. Currently build fails, as vue-eslint-parser requires eslint@5 as peer dependency, and we can't drop Node 4 support out of the blue. I'm going to prepare new beta soon, and will try include this change and drop support for Node 4 there. Thanks for working on it @ChristianMurphy

michalsnik:
I'm closing this one in favor of #532 where @mysticatea updated also mentioned parser :)

PR #513, opened 2018-07-07
Fix attribute-hyphenation ignore lists

LukeShu:
Commit f9e1eee3fe711b9dfedc037fe25aa5570370296e introduced the "ignore" config option for the attribute-hyphenation rule.  However, that commit has a few problems:

 - The added testcase doesn't actually test the functionality at all
 - The code only works correctly if the ignore list has exactly 1 item in it
 - There are spelling mistakes in the test

See each individual commit message for more details.

### DIFF
--- a/tests/lib/rules/attribute-hyphenation.js
+++ b/tests/lib/rules/attribute-hyphenation.js
@@ -130,27 +130,74 @@ ruleTester.run('attribute-hyphenation', rule, {
     },
     {
       filename: 'test.vue',
-      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo" third-custom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
-      output: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo" thirdCustom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
-      options: ['never', { 'ignore': ['custom-hyphen', 'second-custom'] }],
+      code: '<template><div><custom v-bind:my-prop="prop" :second-prop="test"></custom></div></template>',
+      output: '<template><div><custom v-bind:my-prop="prop" :secondProp="test"></custom></div></template>',
+      options: ['never', { ignore: ['my-prop'] }],
       errors: [{
-        message: "Attribute 'third-custom' can't be hyphenated.",
+        message: "Attribute ':second-prop' can't be hyphenated.",
         type: 'VDirectiveKey',
         line: 1
       }]
     },
     {
-      // This is the same code as the `'ignore': ['custom-hyphen']`
-      // valid test; to verify that setting ignore actually makes the
-      // difference.
       filename: 'test.vue',
-      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo"><a onClick="" my-prop="prop"></a></custom></template>',
-      output: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" customHyphen="foo"><a onClick="" my-prop="prop"></a></custom></template>',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/attribute-hyphenation.js
```diff
@@ -46,7 +46,7 @@ ruleTester.run('attribute-hyphenation', rule, {
     {
       filename: 'test.vue',
       code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hypen="foo"><a onClick="" my-prop="prop"></a></custom></template>',
-      options: ['never', { 'ignore': ['custom-hypen'] }]
+      options: ['never', { 'ignore': ['custom-hypen', 'second-custom'] }]
```
michalsnik: Can you add `second-custom` attribute in above html and also add extra failing test with same `ignore` list, but only with one attribute in HTML?

michalsnik:
Thanks for working on it @LukeShu! I added one comment :)

tests/lib/rules/attribute-hyphenation.js
LukeShu: I'm not entirely sure what you're asking for for the failing test case.  Is this what you're looking for?
```js
    {
      filename: 'test.vue',
      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hypen="foo" third-custom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
      output: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hypen="foo" thirdCustom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
      options: ['never', { 'ignore': ['custom-hypen', 'second-custom'] }],
      errors: [{
        message: "Attribute 'third-custom' cann't be hyphenated.",
        type: 'VDirectiveKey',
        line: 1
      }]
    },
```

LukeShu:
v2:
 - Make requested changes to the tests

The change from v1 to v2 is as follows:

```patch
diff --git a/tests/lib/rules/attribute-hyphenation.js b/tests/lib/rules/attribute-hyphenation.js
index 70d798f..ce659cc 100644
--- a/tests/lib/rules/attribute-hyphenation.js
+++ b/tests/lib/rules/attribute-hyphenation.js
@@ -45,7 +45,7 @@ ruleTester.run('attribute-hyphenation', rule, {
     },
     {
       filename: 'test.vue',
-      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo"><a onClick="" my-prop="prop"></a></custom></template>',
+      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo" second-custom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
       options: ['never', { 'ignore': ['custom-hyphen', 'second-custom'] }]
     }
   ],
@@ -128,6 +128,17 @@ ruleTester.run('attribute-hyphenation', rule, {
         line: 1
       }]
     },
+    {
+      filename: 'test.vue',
+      code: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo" third-custom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
+      output: '<template><custom data-id="foo" aria-test="bar" slot-scope="{ data }" custom-hyphen="foo" thirdCustom="bar"><a onClick="" my-prop="prop"></a></custom></template>',
+      options: ['never', { 'ignore': ['custom-hyphen', 'second-custom'] }],
+      errors: [{
+        message: "Attribute 'third-custom' can't be hyphenated.",
+        type: 'VDirectiveKey',
+        line: 1
+      }]
+    },
     {
       // This is the same code as the `'ignore': ['custom-hyphen']`
       // valid test; to verify that setting ignore actually makes the
```

michalsnik:
Thanks @LukeShu ! I fixed the test and also added some missing scenarios

PR #518, opened 2018-07-11
Add 'OTHER_DIRECTIVE' category to attributes-order

ElMassimo:
Follow-up to https://github.com/vuejs/eslint-plugin-vue/pull/209, fixes an issue where custom directives are required to go first, before `v-for` and `v-if`:

screen shot 2018-07-11 at 10 39 08 am

### Description 📖 
This pull request adds an `OTHER_DIRECTIVE` category for any directives that don't match the existing:
- `LIST_RENDERING`
- `CONDITIONALS`
- `RENDER_MODIFIERS`
- `BINDING`
- `EVENTS`
- `CONTENT`

allowing to order custom directives as necessary (by default, after the unique modifiers). This is specially convenient so that `v-for` and `v-if` come before custom directives.

#### Before
```html
<input v-autofocus v-if="displayInput"/>
``` 

#### After
```html
<input v-if="displayInput" v-autofocus/>
```


lib/rules/attributes-order.js
```diff
@@ -23,6 +23,8 @@ function getAttributeType (name, isDirective) {
       return 'EVENTS'
     } else if (name === 'html' || name === 'text') {
       return 'CONTENT'
+    } else {
+      return 'OTHER_DIRECTIVE'
```
ElMassimo: At the moment this category is not defined, so custom directives are always required to go first.

michalsnik:
Hey @ElMassimo ! Thanks for working on it, however we already have a PR with such change I think: https://github.com/vuejs/eslint-plugin-vue/pull/421
It is waiting for the next major release, that's coming really soon now.

Issue #222, opened 2017-10-24
Boolean props should be false by default always. (require-default-prop)


jpickwell:
**Tell us about your environment**

* **ESLint Version: 4.9.0**
* **eslint-plugin-vue Version: 3.13.1**
* **Node Version: 8.2.1**

**Please show your full configuration:**

```
```

**What did you do? Please include the actual source code causing the issue.**

```vue
<script>
export default {
  props: {
    showText: {
      type: Boolean,
      default: true
    },
    autoPlay: Boolean
  }
}
</script>
```

**What did you expect to happen?**
An error should be shown for the `showText` prop indicating that it should not have a default of `true`. The `autoPlay` prop should considered correct even with the `require-default-prop` rule turned on.

**What actually happened? Please include the actual, raw output from ESLint.**
With the `require-default-prop` rule enabled, `autoPlay` will be flagged. `showText` will be considered correct.

To make Vue components work like idiomatic HTML elements `Boolean` attributes/props should always default to `false`. If a `Boolean` prop defaults to `true` then a user who wants to turn that prop off has to bind the prop to a constant `false` (e.g., `:show-text="false"`).

The `require-default-prop` rule would need to be modified to either enforce this, or ignore `Boolean` props. If the rule ignores `Boolean` props, then a new rule could be added to enforce proper `Boolean` props.

`Boolean` props should also never be required.

`someBooleanProp: Boolean` is how all `Boolean` props should be declared. There's no need for any other prop options.

At the very least, something should be added to the rules to allow enforcing this. I also believe this rule should be added to the Style Guide.

michalsnik:
This is a really interesting proposal @jpickwell. I agree that setting default value for `Boolean` property seems daunting and it makes perfect sense to ignore this type in `require-default-prop` rule. But validating `default` property's content for this type of prop sounds like a job for a new rule `no-truthy-boolean-prop` or something like that.

What do you think about this @chrisvfritz ? It might be a good practice to add in the styleguide, as it would definitely improve consistency in components APIs.

chrisvfritz:
If we do include it, I think it should be a new rule, but probably not part of the style guide. I agree that most defaults should be false, but there are legitimate exceptions. For example, you may have an alert component that you want to be dismissible 99% of the time in your app. But sometimes, you want notifications that aren't dismissible, so you make the behavior optional with a new `dissmissible` prop that defaults to true. Does that make sense?

byCedric:
[...]

PR #529, opened 2018-07-21
Ignore Boolean props in require-default-prop rule

michalsnik:
This PR addresses #222 by not requiring `default` key for `Boolean` props.

I thought about considering the following cases:
```js
{
  a: Boolean,
  b: [Boolean],
  c: {
    type: Boolean,
  },
  d: {
    type: [Boolean],
  }
}
```

But I'm no longer sure we need to support cases with arrays that contain only `Boolean` type, and perhaps we should always require default key, when prop type is described by an `ArrayExpression` 🤔 WDYT @chrisvfritz ?


michalsnik:
Oh, now I see I didn't cover `d` actually. I'll update it @chrisvfritz

chrisvfritz:
Thanks! 🙂

PR #534, opened 2018-07-25
[WIP] Docs: add vuepress documentation

mysticatea:
This is working in progress.

I'm trying to make the documentation with [VuePress](https://vuepress.vuejs.org/).

You can review it with:

1. Checkout this branch.
2. `npm install`
3. `npm run docs:watch`
4. Open http://localhost:8080/eslint-plugin-vue/

It will be similar to https://mysticatea.github.io/eslint-plugin-es/

----

Pros:

- We get more structured pages. (E.g., the rule list in the sidebar)
- We get interactive sample code blocks. (Likely https://mysticatea.github.io/eslint-plugin-es/rules/no-async-iteration.html)
    - These code blocks are made with [vue-eslint-editor](https://mysticatea.github.io/vue-eslint-editor/) and [eslint4b](https://github.com/mysticatea/eslint4b).

Cons:

- It's hard to keep old version's content.


armano2:
@mysticatea @michalsnik what is our priority on finishing it?
https://armano2.github.io/eslint-plugin-vue/

i'm unsure if it's better to focus on rules or on documentation.

michalsnik:
@armano2 I think we should focus on documentation now - making it easier to browse through all rules. Having interactive examples in the documentation with per-rule playground would be a really nice addition and an accent on the top of `eslint-plugin-vue@5` that we plan to releasy very very soon.

michalsnik:
Having that said - all new rules are now put `on hold` and we're only merging bug fixes, so you can easily work on the documentation if you're willing to help with this too :)

armano2:
What do you guys think about adding badges to indicate changes in rules between versions

like: https://vuepress.vuejs.org/guide/using-vue.html#badge

armano2:
Issues:
- `vue/comment-directive` is not supported: https://github.com/mysticatea/vue-eslint-editor/pull/1 PR
- `vue/multiline-html-element-content-newline` has issue with example: #663 PR
- `vue/return-in-computed-property` has issue with example: #655 PR
- `vue/require-render-return` has issue with example: #655 PR

Missing rule detail / description
- [ ] `vue/attribute-hyphenation` [link](https://armano2.github.io/eslint-plugin-vue/rules/attribute-hyphenation.html)
- [ ] `vue/this-in-template` [link](https://armano2.github.io/eslint-plugin-vue/rules/this-in-template.html)

demo: https://armano2.github.io/eslint-plugin-vue/

mysticatea:
I'm sorry for disappeared. And very very thank you, @armano2 !!

armano2:
ok, small update, editor looks broken now; https://github.com/mysticatea/vue-eslint-editor/issues/3

michalsnik:
@armano2 what's left here to do? How can I help?

docs/user-guide/README.md
```diff
@@ -0,0 +1,213 @@
...
+You have to include the `.vue` extension using [the `--ext` option](https://eslint.org/docs/user-guide/configuring#specifying-file-extensions-to-lint) or a glob pattern because ESLint targets only `.js` files by default.
+
+Examples:
+
+```bash
+eslint --ext .js,.vue src
+eslint "src/**/*.{js,vue}"
+```
+
+### Editor integrations
```
michalsnik: I think this would be nice as a separate section. just as `Usage`. It will be easier to navigate to it using side nav.

docs/user-guide/README.md
```diff
@@ -0,0 +1,213 @@
...
+#### IntelliJ IDEA / JetBrains WebStorm
+
+In the **Settings/Preferences** dialog (`Ctrl+Alt+S`), choose JavaScript under **Languages and Frameworks** and then choose **ESLint** under **Code Quality Tools**.
+On the **ESLint page** that opens, select the *Enable* checkbox.
+
+If your ESLint configuration is updated (manually or from your version control), open it in the editor and choose **Apply ESLint Code Style Rules** on the context menu.
+
+read more: [JetBrains - ESLint](https://www.jetbrains.com/help/idea/eslint.html)
+
+### The code of components
```
michalsnik: What do you think about calling it `How ESLint detects components?` or `Components detection`

docs/user-guide/README.md
```diff
@@ -0,0 +1,213 @@
...
+      "vue",
+  -   "html"
+    ]
+  ```
+
+2. Make sure your tool is set to lint `.vue` files.
+  - CLI targets only `.js` files by default. You have to specify additional extensions by `--ext` option or glob patterns. E.g. `eslint "src/**/*.{js,vue}"` or `eslint src --ext .vue`. If you use `@vue/cli-plugin-eslint` and the `vue-cli-service lint` command - you don't have to worry about it.
+  - If you are having issues with configuring editor please read [editor integrations](#editor-integrations)
+
+## :traffic_light: Versioning policy
```
michalsnik: What do you think about putting this and the following sections in `Introduction` and keeping `User Guide` a bit cleaner?

tools/update-docs.js
```diff
@@ -24,46 +24,100 @@ const last = require('lodash/last')
...
+  }
+
+  write () {
+    fs.writeFileSync(this.filePath, this.content)
+  }
+
+  updateHeader () {
+    const { ruleId, meta } = this.rule
+    const categoryIndex = categories.findIndex(category => category.categoryId === meta.docs.category)
+    const title = `# ${meta.docs.description} (${ruleId})`
```
michalsnik: What do you think about preparing header to look like this?
image

michalsnik:
For the reference, we should also remember to update link to rules after merging this PR here: https://github.com/jfmengels/eslint-rule-documentation/blob/master/plugins.json

tools/update-docs.js
armano2: i like this

armano2: i did some digging, and looks like we will have to write/override css to have something like this.

armano2:
@michalsnik i formated all rules to use editor to preview, i moved some "notes" to `:::warning`

from things TODO

Missing rule detail / description
- `vue/attribute-hyphenation` [link](https://armano2.github.io/eslint-plugin-vue/rules/attribute-hyphenation.html)
- `vue/this-in-template` [link](https://armano2.github.io/eslint-plugin-vue/rules/this-in-template.html)

and also
`valid-template-root` [link](https://armano2.github.io/eslint-plugin-vue/rules/valid-template-root.html#rule-details) is not formatted, i'm unsure if we want to keep all examples or split in some way

we could apply your suggestions about cleaning up User Guide

-----

thanks to @mysticatea issue with missing tooltips in editor is gone 

-----

in next few days i'm going to have little less time, than.... feel free to work on it :)

michalsnik:
I can take it from here @armano2 :)

michalsnik:
Weird @armano2. First thing I did was to remove node_modules and clean cache, but it didn't help. Now I updated this branch on my other computer and it works 🤔

michalsnik:
I think that's it! And since it's all about documentation there's no better person to mention here, but @chrisvfritz would you give it a look? :) I think It's the final part before releasing v5

armano2:
Can you add restriction that deploy is going to be done only from master branch in circleci

armano2:
@michalsnik latter if i will find few hours i'm going to add badges as indicator when options/rules was added to plugin.

chrisvfritz:
@armano2 Thanks! Apologies for missing that comment in the above thread. 😅 I just checked out the demo you linked to and the interactive editor looks fantastic. 😻

tools/deploy.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+npm run docs:build
+cd docs/.vuepress/dist
+
+git init
+git add -A
+git commit -m 'deploy'
+git push -f git@github.com:vuejs/eslint-plugin-vue.git master:gh-pages
+
```
armano2: What do you think about using something like this instead:

https://github.com/eldarlabs/ghpages-deploy-script/blob/master/scripts/deploy-ghpages.sh

michalsnik: What are the benefits of using it instead? I copied deploy script from the official Vuepress documentation :)

armano2: instead of making one commit and force pushing it over and over again, its adding changes only

PR #544, opened 2018-08-02
(fixes #494) Ignore indent rules on files other than .vue

### LINKED ISSUE #494: `vue/script-indent` applies to js files but should not

* **ESLint Version:** 4.19.1
* **eslint-plugin-vue Version:** 4.5.0
* **Node Version:** 10.4.0

```json
{
  "extends": [
    "nc/vue",
    "plugin:vue/recommended"
  ],
  "plugins": [
    "vue"
  ],
  "env": {
    "browser": true
  },
  "settings": {
    "import/resolver": {
      "webpack": {
        "config": "vuejsapp/build/webpack.base.conf.js"
      }
    },
    "import/ignore": ["node_modules"]
  },
  "rules": {
    "import/extensions": ["error", "always", {
      "js": "never",
      "vue": "never"
    }],
    "vue/html-self-closing": ["error", {
      "html": {
        "void": "any",
        "normal": "any",
        "component": "always"
      },
      "svg": "any",
      "math": "any"
    }],
    "vue/attributes-order": ["off"],
    "vue/order-in-components": ["off"],
    "vue/html-closing-bracket-newline": ["error", {
      "singleline": "never",
      "multiline": "always"
    }],
    "vue/html-closing-bracket-spacing": ["error", {
      "startTag": "never",
      "endTag": "never",
      "selfClosingTag": "always"
    }],
    "vue/prop-name-casing": ["error", "camelCase"],
    "vue/script-indent": ["error", 2, {
      "baseIndent": 1
    }]
  }
}
```

I've added this rule:
```js
"vue/script-indent": ["error", 2, {
  "baseIndent": 1
}]
```
Vue store module: 

Before `eslint --fix`:
```js
# src/store/modules/siteCustomerNavigation.js
export default {
  namespaced: true,
  state: {
    recentHistory: {
      items: [],
    },
  },
  mutations: {
    SOCKET_USER_NAVIGATION_UPDATED(state, navigation) {
      //...
      if (currentItemIndex > -1) {
        const updatedItem = Object.assign({},
          state.recentHistory.items[currentItemIndex],
          navigation
        );
      }
    },
  }
};
```

Indent rule applied to `*.js` files.

After `eslint --fix`:
```js
export default {
  namespaced: true,
  state: {
    recentHistory: {
      items: [],
    },
  },
  mutations: {
    SOCKET_USER_NAVIGATION_UPDATED(state, navigation) {
      //...
      if (currentItemIndex > -1) {
        const updatedItem = Object.assign({},
                                          state.recentHistory.items[currentItemIndex],
                                          navigation
        );
      }
    },
  }
};
```

Due to documentation it should enforce indend in `<script>` of vue component only. But actually it changes js files that are not vue components.

michalsnik:
As pointed in #494 indent rules affects all files, thus `script-indent` lints not only `.vue` files, but also `.js` files.

I think that we should encourage usage of native eslint `indent` rule and `script-indent` only for single file components,  where people might want the extra base indentation.


mysticatea:
LGTM, but this is a breaking change we should write in the changelog.

michalsnik:
We're still in beta phase @mysticatea and actually this change won't introduce more issues being reported, rather less - so it shouldn't break anything.

mysticatea:
@michalsnik It's a surprising change to people which is using the rule for `.js` files as well. They will lose linting about indentation. It's the reason I said, "we should write in the changelog."

chrisvfritz:
@mysticatea I agree it should definitely be included in the changelog! 🙂 But is reducing the scope of a rule typically considered a "breaking change" in ESLint? In this context, I was thinking of a breaking change as anything that could produce potentially undesired errors. I'm quite inexperienced in the ESLint community though, so maybe I've been thinking about it wrong. 😅

Issue #323, opened 2018-01-03
Rule Proposal: Ensure prop.type is an Object


richardtallent:
**Please describe what the rule should do:**

Ensure that the `type` key for a property is passed as an Object, not as a string.

**What category of rule is this? (place an "X" next to just one item)**

[ ] Enforces code style
[x] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about:**

```js
props: {
	myFieldWithBadType: {
		type: "Object",
		default: function() {
			return {}
		},
	},
	myOtherFieldWithBadType: {
		type: "Number",
		default: 1,
	},
}
```

**Why should this rule be included in ESLint (instead of a plugin)?**

Passing a string to `type` rather than the name of the type results in a rather opaque error message from deep in the bowels of Vue, and it's an easy mistake to make.

michalsnik:
This is an interesting proposal @richardtallent It might indeed be useful for newcomers that learn Vue 👍

chrisvfritz:
I like this! It would go in `essentials` and a more complete list of situations to match might be:

``` js
import Foo from './foo'
const Bar = ''

// ...

props: {
  a: 'Object',
  b: { type: 'Number' },
  c: { type: `String` },
  d: { type: Foo + '' },
  e: { type: Bar },
}
```

PR #546, opened 2018-08-03
(implements #323) Add `require-prop-type-constructor` rule

michalsnik:
This PR implements rule proposed in #323.

Unfortunately we can't detect all possible scenarios, hence I focused on those that are trivial and should catch most commonly made mistakes. We still allow CallExpressions and MemberExpressions, as we cannot assume those will not represent or return constructor implicitly.


lib/rules/require-prop-type-constructor.js
```diff
@@ -0,0 +1,98 @@
...
+      url: 'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.1/docs/rules/require-prop-type-constructor.md'
+    },
+    fixable: 'code',  // or "code" or "whitespace"
+    schema: []
+  },
+
+  create (context) {
+    const fix = node => fixer => {
+      if (node.type === 'Literal') {
+        return fixer.replaceText(node, node.value)
```
ota-meshi: It is need to check if the value type is string.
Hexadecimal literals such as `0xF` are converted to numeric literals like `15`.

michalsnik: I didn't consider this case, as it seems extremely rare and nearly impossible to get unless you really want to break it.

ota-meshi: > extremely rare and nearly impossible to get unless you really want to break it.

I think that you are right. I do not think it is necessary to take time and fix it.

PR #547, opened 2018-08-05
[New] Add `vue/html-tag-attrs-content-newline`

ota-meshi:
This PR adds `vue/html-tag-attrs-content-newline` rule.
This implements rule proposed in https://github.com/vuejs/eslint-plugin-vue/pull/445#issuecomment-408873646

It is related to #415.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 |    | Rule ID | Description |
 |:---|:--------|:------------|
 | :wrench: | [vue/component-name-in-template-casing](./docs/rules/component-name-in-template-casing.md) | enforce specific casing for the component naming style in template |
+| :wrench: | [vue/html-tag-attrs-content-newline](./docs/rules/html-tag-attrs-content-newline.md) | require a line break before and after contents whenever the HTML tag has any attribute |
 | :wrench: | [vue/script-indent](./docs/rules/script-indent.md) | enforce consistent indentation in `<script>` |
 
 ### Deprecated
--- a/docs/rules/html-tag-attrs-content-newline.md
+++ b/docs/rules/html-tag-attrs-content-newline.md
@@ -36,7 +36,7 @@ This rule enforces a line break before and after html contents whenever the HTML
 
 ```json
 {
-    "vue/html-content-newline": ["error", {
+    "vue/html-tag-attrs-content-newline": ["error", {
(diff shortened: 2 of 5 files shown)


michalsnik:
Thank you @ota-meshi ! We really appreciate your engagement 🥇 

I’ve been testing both of these rules now in a normal project:
- https://github.com/vuejs/eslint-plugin-vue/pull/445
- https://github.com/vuejs/eslint-plugin-vue/pull/547

Now that these two are ready to play a bit, I wanted to To get a decent comparison of how they both perform in template-heavy conditions.

And I must say - the old one, that we decided is too problematic actually works better in practice in my opinion :sweat_smile:
But I think the `multiline` option should not be possible to override - it doesn’t seem to serve any purpose. After setting it to `never` - code starts to look totally unreadable. Additionally we might however add an extra value to the `singleline` option, so we have: `ignore`, `never`, `always` and `if-has-attribute` (or something like this).

The reasons why I consider it better after all are:
- It's not too strict by default, and allows very common (and not necessarily unreadable) cases like:
```
<span class="item-badge">{{ badge }}</span>
```
- But it warns about:
```
<tr><td>
  {{ col }}
</td></tr>
```
that there should be a new line after `<tr>` and before `</tr>` 👍
- with an additional option for `singleline` (e.g. `if-has-attr`) it might also report the first case, but in the end I think it should be a team's decision to make, not ours.

I would love to hear what is your opinion on this topic @ota-meshi 
I know we've been going back-and-forth with this, and I'm really sorry for the confusion..

The more I think about it I have more ideas.. The newest one is about splitting this rule into two:
- `multiline-html-element-content-newline` - on/off - if the tag is multiline - force putting content in newline
- `singleline-html-element-content-newline` - on/off - with turned on - content should always go to new line, we might however introduce an extra `strict` flag, that could be set to false or true, and depending on this flag - we might allow having content in one line if there are no attributes on the element.

@chrisvfritz I think I have too many ideas regarding this topic...

michalsnik:
That's what I exactly imagined @ota-meshi :) I think it is plain and simple, less things to think about.

One case I won't bother about though:
```
<div>multiline end tag</div
>
```
I mean we'd treat it as multiline and force newline, because of the overall location that covers 2 lines, but not because the end tag itself is multiline.

Multiline end tags should rather be reported separately, but it's not a big priority and we probably won't create a rule for that.

ota-meshi:
Closing to continue working with #551 and #552.

PR #552, opened 2018-08-11
[New] Add `vue/singleline-html-element-content-newline` rule

ota-meshi:
This PR adds `vue/singleline-html-element-content-newline` rule.
This implements rule proposed in https://github.com/vuejs/eslint-plugin-vue/pull/547#issuecomment-410517455

It is related to #415.

### DIFF
--- a/docs/rules/singleline-html-element-content-newline.md
+++ b/docs/rules/singleline-html-element-content-newline.md
@@ -1,0 +1,102 @@
+# require a line break before and after the contents of a singleline element (vue/singleline-html-element-content-newline)
+
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule enforces a line break before and after the contents of a singleline element.
+
+
+:-1: Examples of **incorrect** code:
+
+```html
+<div attr>content</div>
+
+<tr attr><td>{{ data1 }}</td><td>{{ data2 }}</td></tr>
+
+<div attr><!-- comment --></div>
+```
+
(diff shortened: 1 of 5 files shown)


docs/rules/singleline-html-element-content-newline.md
```diff
@@ -0,0 +1,102 @@
...
+```json
+{
+  "vue/singleline-html-element-content-newline": ["error", {
+    "strict": false,
+    "ignoreNames": ["pre", "textarea"]
+  }]
+}
+```
+
+- `strict` ... if false, if there are no attributes on the element, this rule allows having contents in one line.
```
michalsnik: What do you think about renaming `strict` option that I proposed earlier to something a bit more descriptive, like `ignoreWhenNoAttribures`?
```
- `ignoreWhenNoAttributes` ... allows having contents in one line, when given element has no attributes. default `true`

File: tests/lib/rules/singleline-html-element-content-newline.js [hunk 2]
```diff
@@ -0,0 +1,439 @@
...
+      `,
+      output: `
+        <template>
+          <div attr>
+<!-- singleline comment -->
+</div>
+        </template>
+      `,
+      errors: [
+        'Expected 1 line break after closing bracket of the "div" element, but no line breaks found.',
```
michalsnik: Same question as in the other rule, what do you think about using `opening tag` and `closing tag` phrases?

docs/rules/singleline-html-element-content-newline.md
ota-meshi: I think that the name `ignoreWhenNoAttribures` is easy to understand and very good!

ota-meshi:
@michalsnik Thank you for the review!
I changed it, so please check again.

PR #581, opened 2018-09-22
⭐️New: Add `vue/require-direct-export` rule

hirokiosame:
Vue SFC only allows default export where the default export must be a component object. A lot of the linting rules in this package only work when the component is directly exported like so (this is why I mark is as "essential"):
```
export default {
     name: 'ComponentName'
};
```

My team follows the Airbnb JS style, and sometimes the components are [assigned to a variable before getting exported](https://github.com/airbnb/javascript#naming--filename-matches-export).

This rule is to deter components from being assigned to a variable before being exported so that all linting rules can be correctly applied.


lib/rules/require-direct-export.js
```diff
@@ -0,0 +1,47 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'require the component to be directly exported',
+      category: 'essential',
```
michalsnik: We add every new rule as `uncategorized` to have a nice period for testing it and introducing breaking changes if necessary, without having to wait for another major release. So I think we should change the category to `undefined` for the time being :) We're still in v5 beta though, so it might land in v5 after all - we'll see.

In theory it's perfectly fine to have the component assigned to a variable and then exported. You can use a special comment for determining that this is a Vue component and it will be linted properly:
```js
// @vue/component
const ComponentA = {
  ...
}

export default ComponentA;
```

The way I see it - essentials are here to prevent errors. And this is technically not an error-situation so I think it would better fit in `strongly-recommended` :)

docs/rules/require-direct-export.md
```diff
@@ -0,0 +1,35 @@
...
+const ComponentA = {
+	name: 'ComponentA',
+	data() {
+		return {
+			state: 1
+		}
+	}
+}
+
+export default Component A
```
michalsnik: Unwanted space in `Component A`

docs/rules/require-direct-export.md
```diff
@@ -0,0 +1,35 @@
+# require the component to be directly exported (require-direct-export)
+
+- :gear: This rule is included in `"plugin:vue/essentials"`.
+
+## Rule Details
+
+This rule aims to...
```
michalsnik: Can you please provide the short description?

lib/rules/require-direct-export.js
hirokiosame: Ahh I didn't know about the comment. Good to know, thank you!

hirokiosame:
Hey @michalsnik Apologies for the delayed changes, but made the fixes now. Let me know if there's anything else you'd like me to do!

michalsnik:
The category @hirokiosame is still not `undefined` as pointed in the comment.

Issue #265, opened 2017-11-30
Rule proposal: `require-component-name`


danielbastos11:
**Please describe what the rule should do:**

This rule warns you if the component's `name` property is undefined

**What category of rule is this?**

[x] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about:**

```jsx
// bad
export default {
  render(){
    return <div />
  }
}
```

```jsx
// good
export default {
  name: 'div-component',
  render(){
    return <div />
  }
}
```

**Why should this rule be included**

Naming components is a good practice. It makes it easier to debug your code with Vue Devtools, and helps you to keep your code organized. So much so that at least 2 rules on the Style Guide address the issue of how to name your components - one of them is a essential rule.

michalsnik:
Hey, thanks for posting this proposal. I think this might even be something worth to be mentioned in Styleguide itself / cc @chrisvfritz

chrisvfritz:
Hmm, I would probably leave this out of the style guide and keep it in the uncategorized section here, because a `name` option is not actually useful in the vast majority of cases. For example:

- When registering a component locally, as the Vue devtools will use the component's local name.
- When registering a component globally, as providing a name is part of that process.
- When importing a `.vue` component to be used in a render function, as a name will be derived from the `__file` property that's added to the default export.

So now that I think about it actually, and especially since functional components can now be written in `.vue` files, I can't think of a real use case where a component needs the `name` option to be specified. Did I miss any common (or even uncommon) scenarios?

LinusBorg:
The name is necessary when rendering a component recursively.

chrisvfritz:
Yes, that's true! I think a rule to warn about recursive components without a name would be useful. I'd put that in strongly-recommended, but probably leave it out of the style guide still because it's a bit niche.

danielbastos11:
My use case is slightly different.
I'm using [babel-plugin-transform-vue-jsx](https://github.com/vuejs/babel-plugin-transform-vue-jsx), which makes my life extremely easier by allowing me not to register components locally and providing me with a Vue.js syntax that looks almost exactly like React (this is really important in my use case)

This is a extremely simplified version of one my files:

```jsx
import Table from '../components/Table'

export default {
  name: 'SpList',
  props: {
    visits: { type: Array, default: [] }
  },
  render() {
    let { visits } = this
    return (
[...]

PR #589, opened 2018-09-27
⭐️New: Add rule `require-component-name`

hirokiosame:
This rule requires components be named

### DIFF
--- a/docs/rules/require-component-name.md
+++ b/docs/rules/require-component-name.md
@@ -1,0 +1,30 @@
+# Require components to have names (vue/require-component-name)
+
+- :gear: This rule is included in `"plugin:vue/recommended"`.
+
+## :book: Rule Details
+
+This rule requires components to have names.
+
+:-1: Examples of **incorrect** code for this rule:
+
+```js
+export default {
+    data() {
+        return {};
+    },
+    created() {},
+}
+```
+
+:+1: Examples of **correct** code for this rule:
+
+```js
+export default {
+    name: 'my-component',
+    data() {
+        return {};
+    },
+    created() {},
+}
+```
--- a/lib/configs/recommended.js
+++ b/lib/configs/recommended.js
@@ -10,6 +10,7 @@ module.exports = {
     'vue/html-quotes': 'error',
     'vue/no-v-html': 'error',
     'vue/order-in-components': 'error',
+    'vue/require-component-name': 'error',
     'vue/this-in-template': 'error'
   }
 }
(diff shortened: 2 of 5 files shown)


armano2:
Additionally we should think what should happens with mixins because for now we have 
```
const componentComments = sourceCode.getAllComments().filter(comment => /@vue\/component/g.test(comment.value))
```
and we have no way of telling if something is mixin or component

michalsnik:
Hey, have you looked at the discussion here https://github.com/vuejs/eslint-plugin-vue/issues/265 @hirokiosame ? The PR @chrisvfritz mentioned has been merged and we didn't yet agree whether we want to introduce this rule or not. I'll put on-hold for now and let's continue discussion there, as I have very mixed feelings about it.

hirokiosame:
Ah great thanks guys. Let me know what you guys decide on and I can work on it further.

rodrigopedra:
Can we add an option force the component name to match the file name?

e.g. 

This is **correct:**:

~~~vue
// src/components/MyComponent.vue

<script>
export default {
    name: 'MyComponent',
}
</script>

<template />
~~~

This is **incorrect:**:

~~~vue
// src/components/MyComponent.vue

<script>
export default {
    name: 'MyCompnent', // missed the o
}
</script>

<template />
~~~

It would help to catch cases where someone mistyped the component name.

chrisvfritz:
> If possible I wanted to understand the rationale for not specifying a name property for any component. I always set them for consistency but perhaps it is a bit of over engineering on my side.

@rodrigopedra Great question! 🙂 My reasoning is actually the same as yours: consistency. 😄 When we always use `.vue` files and don't specify a name, we guarantee that the component's name always matches the file name. When you specify the name _again_ in a `name` property, you're not only writing more code unnecessarily, but also forcing yourself to either manually make sure the two names matche (to avoid confusing behavior) or write some kind of tooling to help you avoid mistakes, like you're doing now. 🙂 

> If you can, please review it and I can send a PR

Sounds great. This would be a new rule (maybe `match-component-file-name`), so a new PR would be best. Writing unit tests for the rule will also help us review it, because they'll not only describe the behavior you're trying to create, but if the tests are robust and they pass, we can focus on higher level aspects of the code.

rodrigopedra:
@chrisvfritz thank you very much for the feedback, I will try to send a PR later this week.

rodrigopedra:
Hi @hirokiosame , check the work that was done #668 , it was initially based in your code in this PR, and we defaulted to `jsx` and added options to other file extensions.

chrisvfritz:
@michalsnik I think this can be closed unless someone objects.

Issue #535, opened 2018-07-25
exceptions option for vue/no-multi-spaces


rjgotten:
**Tell us about your environment**

* **ESLint Version:** 5.0.1
* **eslint-plugin-vue Version:** 4.5.0
* **Node Version:** 8.11.3 (LTS)

**Please show your full configuration:**

```
module.exports = {
  root: true,
  extends: "plugin:vue/recommended",
  parser: "vue-eslint-parser",
  parserOptions: {
    parser: "babel-eslint",
    sourceType: "module",
    ecmaVersion: 8
  },
  env: {
    browser: true,
    es6: true
  },
  plugins: ["vue"]
};
```

**What did you do? Please include the actual source code causing the issue.**

```vue
<i
  v-if="useIconArrow"
  :class="{
    'fa-angle-up'   : isExpanded,
    'fa-angle-down' : !isExpanded,
  }"
  class="fal"
></i>
```

**What did you expect to happen?**

No validation error from `vue/no-multi-spaces`, which is meant to handle unnecessary spaces on the template side and should not bother checking inside JS expressions. (That should be the domain of 'normal' ESLint rules.)

**What actually happened? Please include the actual, raw output from ESLint.**

`vue/no-multi-spaces` gives an error on an object literal inside a JS expression, which aligns property assignments on the colon:

```
error  Multiple spaces found before ':'  vue/no-multi-spaces
```

mysticatea:
Thank you for this issue.

This is intentional behavior rather than false positive.
I think that we can consider `exceptions` option like core [no-multi-spaces](https://eslint.org/docs/rules/no-multi-spaces) rule.

PR #591, opened 2018-09-28
(Implements #535) Add "ignoreProperties" option to "no-multi-spaces" rule

michalsnik:
This PR implements proposition from #535

### DIFF
--- a/docs/rules/no-multi-spaces.md
+++ b/docs/rules/no-multi-spaces.md
@@ -16,6 +16,15 @@ Examples of **incorrect** code for this rule:
       :style="bar"         />
 ```
 
+```html
+<i
+  :class="{
+    'fa-angle-up'   : isExpanded,
+    'fa-angle-down' : !isExpanded,
+  }"
+/>
+```
+
 Examples of **correct** code for this rule:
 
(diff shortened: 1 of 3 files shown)


docs/rules/no-multi-spaces.md
```diff
@@ -25,6 +34,36 @@ Examples of **correct** code for this rule:
...
+}]
+```
+
+:+1: Examples of **correct** code for this rule:
+
+```html
+<i
+  :class="{
+    'fa-angle-up'   : isExpanded,
+    'fa-angle-down' : !isExpanded,
```
ota-meshi: @michalsnik Hello!
Is the spaces after the colon allowed?

```js

    'fa-angle-up':   isExpanded,
    'fa-angle-down': !isExpanded,
```

It seems that this case is allowed by option `{ "Property": true }` of core [no-multi-spaces](https://eslint.org/docs/rules/no-multi-spaces) rule,
so I think that it will be the same behavior is good.

michalsnik: Good catch! I updated the cases and handle this one as well.

PR #599, opened 2018-10-07
add directive-interpolation-spacing rule

milewski:
Hi, i dont know what is the criteria used to determine which a rule should go so let me know if it should be somewhere else than "strongly-recommended" ,

This PR gives the ability to format the expression within properties binding:

```html
<template>
    <div>
        <!-- transform this: -->
        <div :class="{test:132}" />
        <!-- to: -->
        <div :class="{ test: 132 }" />
    </div>
</template>
```

this is a problem that extremely annoys me many times i give up using eslint with vue just because of this.. so i took the initiative to fix it myself

🤞

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 |    | Rule ID | Description |
 |:---|:--------|:------------|
 | :wrench: | [vue/attribute-hyphenation](./docs/rules/attribute-hyphenation.md) | enforce attribute naming style on custom components in template |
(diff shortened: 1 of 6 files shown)


milewski:
I felt that would be a way of doing that with the existing rules.. Do you have any example of how to use existing rules to format specific section? 

Because that two rules did not catch the expressions inside the directives, so maybe if I could indicate that part should be processed by rule X or Y then I could eliminate the need of adding the spacing on curly using this PR

mysticatea:
Thank you for this PR.

I prefer to re-implement of the existing rules for Vue.js template. 
`vue/object-curly-spacing` and something like will be useful because the core rules don't work in Vue.js template.

Issue #598, opened 2018-10-05
no-unused-components for file template.html


mcmc4519:
Hi!

My env: **windows 10 x64** , **node v8.11.2**
**.eslintrc**
```
{
    "parser": "vue-eslint-parser",
    "parserOptions": {
        "parser": "babel-eslint",
        "root": true,
        "ecmaVersion": 2018,
        "sourceType": "module",
        "ecmaFeatures": {
            "globalReturn": false,
            "impliedStrict": false,
            "jsx": false
        }
    },
    "extends": [
        "eslint:recommended",
        "plugin:vue/essential",
        "prettier",
        "prettier/standard",
        "plugin:vue/strongly-recommended",
        "plugin:vue/recommended",
        "plugin:you-dont-need-lodash-underscore/compatible-warn"
    ],
    "rules": {
        "prefer-arrow-callback": "error",
        "prefer-numeric-literals": "error",
        "prefer-rest-params": "error",
        "prefer-spread": "error",
        "prefer-template": "error",
        "quote-props": ["error", "as-needed"],
        "you-dont-need-lodash-underscore/all": "error",
        "vue/no-unused-vars": "error",
        "prettier/prettier": "error",
        // "linebreak-style": ["error", "windows"],
        "curly": ["error", "multi"],
        "no-debugger": "off",
        "no-var": "error",
        "no-console": "off",
        "no-const-assign": "error",
        "no-unused-vars": "warn",
        "no-await-in-loop": "error",
        "default-case": "error",
        "no-else-return": "warn",
        "no-multi-spaces": "warn",
        "no-useless-return": "warn",
        "no-case-declarations": "error",
        "camelcase": "error",
        "no-mixed-spaces-and-tabs": "error",
        "no-duplicate-imports": "error",
        "prefer-const": "warn",
        "vue/html-closing-bracket-spacing": "warn",
        "vue/html-closing-bracket-newline": [
            "warn",
            {
                "singleline": "never",
                "multiline": "always"
            }
        ],
        "vue/max-attributes-per-line": [
            "error",
            {
                "singleline": 1,
                "multiline": 1
            }
        ],
        "vue/require-prop-types": "warn",
         "vue/no-unused-components": "error"
    },
    "plugins": ["vue", "prettier", "babel"]
}
```

**devDependencies**

```
{"devDependencies": {
    "babel-eslint": "^8.2.6",
    "babel-plugin-syntax-dynamic-import": "^6.18.0",
[...]

PR #607, opened 2018-10-13
#598 Ignore templates with src tag in no-unused-components

armano2:
fix issue: #598


armano2:
i'm unsure if this is valid way to handle this, we could add reading "src" in parser, but its risky (can cause alot of issues if files are generated dynamicly, or processed/generated by scripts)

PR #612, opened 2018-10-14
⭐️New: Add rule `no-boolean-default`

hirokiosame:
_I haven't added documentation yet but I can add one if you guys like the rule._

The rule prevents Boolean props from having a default value. I put this under `essential` as it follows HTML standards of always defaulting boolean attributes to false. Otherwise, users would have to do `:prop="false"` to turn it off.

The rule has the following options:
`'constructor'` (default) enforces that Boolean props must use the constructor.
`'no-default'` allows a prop definition object, but enforces that the `default` property not be defined.
`'default-fase'` enforces that the default can be set but must be set to `false`.


hirokiosame:
Thank you @mysticatea I made the updates! Let me know if there's anything else I can do.

tests/lib/rules/no-boolean-default.js
```diff
@@ -0,0 +1,125 @@
...
+      code: `
+        const props = {};
+        export default {
+          props: {
+            ...props,
+            enabled: Boolean
+          }
+        }
+      `
+    }
```
armano2: Can you add this test?

```suggestion
    },
    {
      filename: 'test.vue',
      code: `
        export default {
          props: {
            enabled: {
              type: Boolean,
              ...data
            }
          }
        }
      `
    }
```

hirokiosame:
@armano2 Oops I misread his comment to add the spread outside of the prop schema. I updated the PR, thanks for catching

armano2:
@hirokiosame thank you 👍

docs/rules/no-boolean-default.md
```diff
@@ -0,0 +1,39 @@
...
+export default {
+  props: {
+    foo: Boolean
+  }
+}
+```
+
+
+## :wrench: Options
+- `'constructor'` (default) enforces that Boolean props must use the constructor.
```
michalsnik: There is a separate rule that checks whether prop type is a constructor, so I don't think we need it here. Perhaps the only additional setting we need here is `allowFalse`?

lib/rules/no-boolean-default.js
```diff
@@ -0,0 +1,113 @@
...
+        p.type === 'Property' &&
+            p.key.type === 'Identifier' &&
+            p.key.name === 'props' &&
+            p.value.type === 'ObjectExpression'
+      )
+    }
+
+    function getBooleanProps (propsDef) {
+      return propsDef.value.properties
+        .filter(p => {
```
michalsnik: You can chain `.filter` to improve readability

lib/rules/no-boolean-default.js
```diff
@@ -0,0 +1,113 @@
...
+          return (p.type === 'Property') && (p.value.type === 'ObjectExpression') && p.value.properties.find(p =>
```
michalsnik: Please remember to unwrap types, look at `require-default-prop` rule for more informations :)

docs/rules/no-boolean-default.md
```diff
@@ -0,0 +1,39 @@
...
+- `'no-default'` allows a prop definition object, but enforces that the `default` property not be defined.
+- `'default-fase'` enforces that the default can be set but must be set to `false`.
```
ota-meshi: `default-fase` is a typo.
Could you  rewrite it to `default-false`?
There is similar typo in line below.

lib/rules/no-boolean-default.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    function getDefaultNode (propDef) {
+      return propDef.value.properties.find(p => {
+        return (p.type === 'Property' && p.key.type === 'Identifier' && p.key.name === 'default')
+      })
+    }
+
+    return utils.executeOnVueComponent(context, (obj) => {
+      const propsDef = getPropsDef(obj)
```
armano2: If you decide to merge from master, you can simplify this code

```suggestion
      const propsDef = utils.getComponentProps(obj)
        .filter(prop => prop.value && prop.value.type === 'ObjectExpression')
```

lib/rules/no-boolean-default.js
```diff
@@ -0,0 +1,113 @@
...
+      return vueComp.properties.find(p =>
+        p.type === 'Property' &&
+            p.key.type === 'Identifier' &&
+            p.key.name === 'props' &&
+            p.value.type === 'ObjectExpression'
+      )
+    }
+
+    function getBooleanProps (propsDef) {
+      return propsDef.value.properties
```
armano2: If you decide to merge from master, you can simplify this code
```suggestion
      return propsDef
```

lib/rules/no-boolean-default.js
armano2: If you decide to merge from master, you can simplify this code
```suggestion
          return p.value.properties.some(p =>
```

lib/rules/no-boolean-default.js
armano2: p is used 2 times in this function and it can be confusing to read, please change this one to prop.

README.md
```diff
@@ -148,6 +148,7 @@ Enforce all the rules in this category, as well as all higher priority rules, wi
 |    | Rule ID | Description |
 |:---|:--------|:------------|
 |  | [vue/no-async-in-computed-properties](./docs/rules/no-async-in-computed-properties.md) | disallow asynchronous actions in computed properties |
+|  | [vue/no-boolean-default](./docs/rules/no-boolean-default.md) | Prevents boolean defaults from being set |
```
armano2: please do not update readme -> it will be done during next release

lib/configs/essential.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
   extends: require.resolve('./base'),
   rules: {
     'vue/no-async-in-computed-properties': 'error',
+    'vue/no-boolean-default': 'error',
```
armano2: please do not update configs -> it will be done during next release

michalsnik:
@armano2 I updated this PR, please take a look one more time :)

hirokiosame:
Looks great, thanks for the updates @michalsnik

Issue #564, opened 2018-08-29
vue/require-default-prop gives a false error with " as PropOptions<T>" with TypeScript


wuservices:
**Tell us about your environment**

* **ESLint Version:** 4.19.1
* **eslint-plugin-vue Version:** 4.7.1
* **Node Version:** 10.9.0

**Please show your full configuration:**

```
module.exports = {
  root: true,
  env: {
    node: true
  },
  extends: ["plugin:vue/recommended", "@vue/prettier", "@vue/typescript"],
  rules: {
    "no-console": process.env.NODE_ENV === "production" ? "error" : "off",
    "no-debugger": process.env.NODE_ENV === "production" ? "error" : "off"
  },
  parserOptions: {
    parser: "typescript-eslint-parser"
  }
};
```

**What did you do? Please include the actual source code causing the issue.**

1. Create new TypeScript project with ESLint + Prettier with Vue CLI 3.0.1
2. In `.eslintrc.js`, change `plugin:vue/essential` to `plugin:vue/recommended` to get all rules
3. Add `PropOptions` types to `HelloWorld.vue` as seen below. In this example I use `PropOptions<string>`, which is redundant (but still shows the problem). In practice, this would be a more complex type like an interface. https://github.com/vuejs/vue/issues/7640#issuecomment-365217858 is where I got inspiration for this typing syntax.

```vue
<script lang="ts">
import Vue, { PropOptions } from "vue";

export default Vue.extend({
  props: {
    msg: {
      type: String,
      required: true
    } as PropOptions<string>
  }
});
</script>
```

**What did you expect to happen?**
No errors that prop `msg` requires a default value to be set.

```
$ yarn lint
yarn run v1.9.4
$ vue-cli-service lint
 DONE  No lint errors found!
```

**What actually happened? Please include the actual, raw output from ESLint.**

It looks like adding ` as PropOptions<string>` trips up the parser such that it doesn't recognize that the `msg` prop is required and therefore doesn't need a default value

```
$ yarn lint
yarn run v1.9.4
$ vue-cli-service lint
error: Prop 'msg' requires default value to be set (vue/require-default-prop) at src/components/HelloWorld.vue:89:5:
  87 | export default Vue.extend({
  88 |   props: {
> 89 |     msg: {
     |     ^
  90 |       type: String,
  91 |       required: true
  92 |     } as PropOptions<string>
```

michalsnik:
[...]

Issue #575, opened 2018-09-16
vue/order-in-components ceases to work when typecasting Vue component


ffxsam:
**Tell us about your environment**

* **ESLint Version:** 4.19.1
* **eslint-plugin-vue Version:** 4.7.1
* **Node Version:** 8.10.0

**Please show your full configuration:**

```js
module.exports = {
  root: true,
  env: {
    node: true,
  },
  globals: {
    describe: true,
    it: true,
    cy: true,
  },
  extends: ['plugin:vue/recommended', '@vue/standard', '@vue/typescript'],
  rules: {
    // Plain JS rules
    // allow debugger during development
    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
    'comma-dangle': ['error', 'always-multiline'],
    semi: ['error', 'always'],
    'space-before-function-paren': ['error', 'never'],
    'no-var': 'error',
    'no-const-assign': 'error',
    // 'no-unused-vars': 'warn',
    'no-await-in-loop': 'error',
    'default-case': 'error',
    eqeqeq: 'warn',
    'no-alert': 'warn',
    'no-console': 'off', //process.env.NODE_ENV === 'production' ? 'error' : 'off',
    'no-unreachable': 'warn',
    'no-else-return': 'warn',
    'no-multi-spaces': 'warn',
    'no-useless-return': 'warn',
    'no-case-declarations': 'error',
    yoda: 'error',
    'no-use-before-define': 'warn',
    camelcase: ['error', { properties: 'never' }],
    'no-mixed-spaces-and-tabs': 'error',
    'no-duplicate-imports': 'error',
    'prefer-const': 'warn',
    indent: 'off',
    'space-before-function-paren': 'off',
    'standard/computed-property-even-spacing': 'off',

    // Extra Vue rules
    'vue/html-closing-bracket-spacing': 'warn',
    'vue/html-closing-bracket-newline': [
      'warn',
      {
        singleline: 'never',
        multiline: 'always',
      },
    ],
    'vue/require-prop-types': 'never', // temporary, till Array junk is fixed
  },
  parserOptions: {
    parser: 'typescript-eslint-parser',
  },
};
```

**What did you do? Please include the actual source code causing the issue.**

```ts
export default (Vue as VueConstructor<
  Vue &
    AudioState & {
      $refs: {
        audio: HTMLAudioElement;
      };
    }
>).extend({
  data() { /* ... */ },
  props: { /* ... */ },
[...]

PR #613, opened 2018-10-17
#564 #575 Fix issues with Typescript

armano2:
Add typescript specific changes:
* Add "as" helper
* Fix `vue/require-default-prop`
* Fix `vue/require-prop-types`
* Fix `vue/require-prop-type-constructor`
* Fix `vue/require-valid-default-prop`
* Allow to use `export default (Vue as VueConstructor<Vue>).extend({` syntax

Fix issues: #564 #575

### DIFF
--- a/lib/rules/require-default-prop.js
+++ b/lib/rules/require-default-prop.js
@@ -32,7 +32,7 @@ module.exports = {
      * @return {boolean}
      */
     function propIsRequired (prop) {
-      const propRequiredNode = prop.value.properties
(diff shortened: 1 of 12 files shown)


armano2:
i know that this is not solving all issues, but i think its good start

armano2:
@michalsnik do you think thats enough for first phase?

i started changing rules to support https://github.com/vuejs/vue-class-component, but i will preffer to merge this first

Issue #626, opened 2018-11-01
[FR] `inheritAttrs: false` when using `v-bind="$attrs"`


privatenumber:
Components that inherit `$attrs` into a nested element in the component typically wouldn't want to get duplicated attributes applied to the top-level element. Might be a good linting rule to warn to apply `inheritAttrs: false` when it detects `v-bind="$attrs"` being used.

michalsnik:
I'm not sure about this rule. It might lead to false-positives, the `inheritAttrs` might not be explicitly set on the component if it uses a mixin. And having it reporting in one scenario and not in the other seems not very consistent and might actually be misleading in my opinion. WDYT @chrisvfritz ?

chrisvfritz:
As long as we can avoid false positives (e.g. by just turning the rule off if we detect a mixin or `Object.assign`/spread on the component object), I think it could be useful. From what I've seen, it's very rare for `inheritAttrs: false` to come from an external source, so this would still catch 99% of cases.

PR #627, opened 2018-11-02
⭐️New: Add rule `no-duplicate-attr-inheritance`

privatenumber:
This rule makes sure you're setting `inheritAttrs` to `false` when it detects `v-bind="$attrs"`.

This rule is to prevent duplicated attribute inheritance: http://jsfiddle.net/hirokiosame/h0j3qrbm/

Can update the docs if there's consensus on the rule proposal.


armano2:
ticket: #626

lib/rules/no-duplicate-attr-inheritance.js
```diff
@@ -0,0 +1,50 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'enforce inheritAttrs: false when using v-bind="$attrs"',
+      category: 'recommended',
```
armano2: Please change category to undefined
```suggestion
      category: undefined,
```

hirokiosame:
@armano2 
What do you think of the following behavior?
- If it detects a mixin being pulled in, the rule gives the user the benefit of the doubt that `inheritAttrs: false` is set via the mixin. We can either emit a warning instead of an error, or emit no error at all.
- If the component is extended upon, then the base should be responsible for having `inheritAttrs: false` IMO. It shouldn't be the extending component's responsibility to patch a bug in the base component.

michalsnik:
Thanks for working on it  @hirokiosame , however I'm going to close this PR for the moment, until we decide whether it's something we want to consider. Let's continue discussion in the relevant issue: #626

Issue #596, opened 2018-10-03
vue/require-default-prop gives error when using a variable.


jpickwell:
**Tell us about your environment**

* **ESLint Version:** 5.6.1
* **eslint-plugin-vue Version:** 5.0.0-beta.3
* **Node Version:** 10.3.0

**Please show your full configuration:**

```
const standardConfig = require('eslint-config-standard/eslintrc.json')

const OFF = 'off'
const WARN = 'warn'
const ERROR = 'error'

module.exports = {
  extends: [
    'eslint:recommended',
    'plugin:flowtype/recommended',
    'plugin:lodash/recommended',
    'plugin:you-dont-need-lodash-underscore/compatible-warn',
    'plugin:vue/recommended',
    'standard'
  ],
  plugins: ['babel'],
  env: { commonjs: true },
  rules: {
    'array-bracket-newline': ERROR,
    'array-bracket-spacing': ERROR,
    'array-callback-return': ERROR,
    'arrow-body-style': ERROR,
    'arrow-parens': ERROR,
    'babel/new-cap': standardConfig.rules['new-cap'],
    'babel/no-invalid-this': ERROR,
    'babel/no-unused-expressions': standardConfig.rules['no-unused-expressions'],
    'babel/object-curly-spacing': [ERROR, 'always'],
    'babel/quotes': standardConfig.rules.quotes,
    'babel/semi': standardConfig.rules.semi,
    'callback-return': ERROR,
    'computed-property-spacing': ERROR,
    'consistent-return': ERROR,
    'consistent-this': [ERROR, 'self', 'that'],
    'default-case': ERROR,
    'dot-notation': ERROR,
    'for-direction': ERROR,
    'func-names': [ERROR, 'as-needed'],
    'getter-return': ERROR,
    'global-require': ERROR,
    'guard-for-in': ERROR,
    'init-declarations': [ERROR, 'never'],
    'jsx-quotes': ERROR,
    'linebreak-style': ERROR,
    'lines-between-class-members': ERROR,
    'lodash/prefer-lodash-method': [
      ERROR,
      {
        ignoreMethods: ['isArray', 'replace', 'toLower', 'toUpper', 'trim']
      }
    ],
    'max-len': [ERROR, {
      code: 120,
      ignoreTrailingComments: true,
      ignoreUrls: true
    }],
    'max-statements-per-line': ERROR,
    'new-cap': OFF, // disabled because of babel plugin
    'newline-per-chained-call': ERROR,
[...]

PR #632, opened 2018-11-06
Fix issues with props

armano2:
With those changes, its going to be easier to add support for classes in feature 

* `vue/prop-name-casing`: is working now with array props [literals]
* `vue/prop-name-casing`: reports all errors if there are non Literal keys in it
* `vue/require-default-prop`: allow to use shorthand, properties and function calls
* `vue/require-prop-types`: reports names for types different than literals
* `vue/require-prop-types`: fix false positive error in when is set to empty array
* `vue/require-prop-types`: will return now errors about each prop from ArrayExpression
* add new `getComponentProps` helper to easily deal with props 
* fix spread operator issues across all `prop` related rules

fix issue: #596 
fix issue #595


.gitignore
```diff
@@ -1,4 +1,5 @@
 .idea
+*.iml
```
michalsnik: What is this file?

armano2: https://www.jetbrains.org/intellij/sdk/docs/basics/project_structure.html
> Module Library: the library classes are visible only in this module and the library information is recorded in the module’s .iml file.

armano2:
@michalsnik all requested changes are applied, 
additionally i made interface of `getComponentProps` more like `getComputedProperties`

lib/rules/require-prop-type-constructor.js
```diff
@@ -70,31 +70,23 @@ module.exports = {
...
 
     return utils.executeOnVueComponent(context, (obj) => {
-      const node = obj.properties.find(p =>
-        p.type === 'Property' &&
-        p.key.type === 'Identifier' &&
-        p.key.name === 'props' &&
-        p.value.type === 'ObjectExpression'
-      )
+      const props = utils.getComponentProps(obj)
+        .filter(cp => cp.key && cp.value)
```
michalsnik: `cp` should rather be named `prop` here

lib/rules/require-prop-type-constructor.js
```diff
@@ -70,31 +70,23 @@ module.exports = {
...
 
-      if (!node) return
+      for (const p of props) {
```
michalsnik: Can we also use `prop` here instead of `p`?

lib/rules/require-prop-type-constructor.js
```diff
@@ -70,31 +70,23 @@ module.exports = {
...
+        if (isForbiddenType(p.value) || p.value.type === 'ArrayExpression') {
+          checkPropertyNode(p.key, p.value)
+        } else if (p.value.type === 'ObjectExpression') {
+          const typeProperty = p.value.properties.find(prop =>
```
michalsnik: And here we can use `p` or `property` then

lib/rules/require-valid-default-prop.js
```diff
@@ -88,20 +88,11 @@ module.exports = {
...
-        p.value.type === 'ObjectExpression'
-      )
-      if (!props) return
-
-      const properties = props.value.properties.filter(p =>
-        isPropertyIdentifier(p) &&
-        utils.unwrapTypes(p.value).type === 'ObjectExpression'
-      )
+      const props = utils.getComponentProps(obj)
+        .filter(cp => cp.key && cp.value && cp.value.type === 'ObjectExpression')
```
michalsnik: `cp` -> `prop`

lib/utils/index.js
```diff
@@ -365,9 +365,46 @@ module.exports = {
...
+
+    if (!propsNode) {
+      return []
+    }
+
+    let props
+
+    if (propsNode.value.type === 'ObjectExpression') {
+      props = propsNode.value.properties
+        .filter(cp => cp.type === 'Property')
```
michalsnik: `cp` -> `prop`, and everywhere else too, expect places where it stands for `computed property` :)

tests/lib/rules/require-prop-types.js
```diff
@@ -149,25 +167,43 @@ ruleTester.run('require-prop-types', rule, {
...
+        message: 'Prop "foo" should define at least its type.',
+        line: 3
+      }, {
+        message: 'Prop "bar" should define at least its type.',
+        line: 3
+      }, {
+        message: 'Prop "baz" should define at least its type.',
+        line: 3
+      }, {
+        message: 'Prop "***" should define at least its type.',
```
michalsnik: Perhaps we should call it `Unknown prop`? I won't be ideal though. I think we can do better - perhaps in another PR and only report this once if the props is an Array. Now we'll always get as many errors as we have props in the array anyway.

armano2:
@michalsnik requested changes applied

PR #642, opened 2018-11-11
Fix #614 - Improve errors in valid-v-on, detect forbidden keywords

### LINKED ISSUE #614: False invalid v-on rule, eslint
I checked the docs and i am using it correctly.
This is error:
screen shot 2018-10-18 at 04 06 00
This is warning:
screen shot 2018-10-18 at 04 07 20

VSCode: 1.28.1
Vetur: 0.13.0
Mac Mojave

michalsnik:
This PR improves error message pointed in #614. By default if there is no value or one of the two verb modifiers we throw an error, but if someone attempts to use forbidden JS keyword the parser stripes it from AST and we throw the wrong message. Now I'm making sure that the value is really empty, both in the AST and in the code itself, otherwise the rule will throw a message about forbidden value.

### DIFF
--- a/lib/rules/valid-v-on.js
+++ b/lib/rules/valid-v-on.js
@@ -167,7 +167,7 @@ module.exports = {
             context.report({
               node,
               loc: node.loc,
-              message: "'v-on' directives require a value or verb modifiers (like 'stop' or 'prevent')."
+              message: "'v-on' directives require a value or verb modifier (like 'stop' or 'prevent')."
             })
           }
         }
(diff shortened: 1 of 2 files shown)


lib/rules/valid-v-on.js
```diff
@@ -149,12 +150,26 @@ module.exports = {
...
+              node,
+              loc: node.loc,
+              message: 'Avoid using JavaScript keyword as "v-on" value: {{value}}.',
+              data: { value }
+            })
+          } else {
+            context.report({
+              node,
+              loc: node.loc,
+              message: "'v-on' directives require a value or verb modifiers (like 'stop' or 'prevent')."
```
armano2: we should not use plural word here, besides of that it looks fine 👍  
```suggestion
              message: "'v-on' directives require a value or verb modifier (like 'stop' or 'prevent')."
```

michalsnik:
@armano2 updated :)

PR #643, opened 2018-11-11
Fix #616 - update max attrs per line error message, remove white spaces

### LINKED ISSUE #616: `max-attributes-per-line` error message is not specific enough

* **ESLint Version:** 5.7.0
* **eslint-plugin-vue Version:** 5.0.0-beta.3
* **Node Version:** 9.2.0

```js
module.exports = {
  root: true,

  parserOptions: {
    parser: "babel-eslint",
    sourceType: "module",
  },

  env: {
    browser: true,
    node: true,
  },

  plugins: ["vue"],

  extends: [
    "airbnb-base",
    "plugin:vue/recommended",
  ],

  rules: {
    "consistent-return": 0,
    "max-len": 0,
    "func-names": 0,
    "global-require": 0,
    "no-underscore-dangle": 0,
    "no-param-reassign": 0,
    "no-plusplus": [
      2,
      {
        allowForLoopAfterthoughts: true,
      },
    ],
    "no-unused-vars": 1,
    "no-mixed-operators": [
      2,
      {
        allowSamePrecedence: true,
      },
    ],
    "import/no-webpack-loader-syntax": 0,
    "import/first": [2, "disable-absolute-first"],
    "import/extensions": [
      "error",
      "always",
      {
        js: "never",
        vue: "never",
      },
    ],
    "import/no-extraneous-dependencies": [
      "error",
      { devDependencies: true, optionalDependencies: false, peerDependencies: false },
    ],

    // Bug when used with webpack aliases
    // - https://github.com/benmosher/eslint-plugin-import/issues/764
    "import/extensions": 0,
    "import/no-unresolved": 0,

    // == All Vue-specific rules ==
    // Essential


    // Recommended
    "vue/component-name-in-template-casing": [
      "error",
      "kebab-case",
    ]
  },

  settings: {
    "import/resolver": {
      webpack: {
        config: "config/webpack/webpack.config.base.js",
      },
    },
  },
}
```

```vue
 <button :disabled="disabled" :aria-disabled="loading" class="vy-button" :aria-label="ariaLabel"
  @click="handleClick"
  :autofocus="autofocus"
  :type="nativeType"
  :class="[
    type ? 'vy-button--' + type : '',
    size ? 'vy-button--' + size : '',
    {
      'is-disabled': disabled,
      'is-loading': loading,
      'is-plain': plain,
    }
  ]"
>
```

The rule `vue/max-attributes-per-line` would tell me which attribute is at fault when I am using the `v-bind` shorthand.

eg. `2:11  error  Attribute "disabled" should be on a new line`

The output just shows me that it is a `v-bind` that is violating the linting rule:
`2:11  error  Attribute "bind" should be on a new line`.

I have several attributes here violating this rule, so my eslint output looks like this:
```
    2:11  error  Attribute "bind" should be on a new line              vue/max-attributes-per-line
    2:32  error  Attribute "bind" should be on a new line              vue/max-attributes-per-line
    2:57  error  Attribute "class" should be on a new line             vue/max-attributes-per-line
    2:75  error  Attribute "bind" should be on a new line              vue/max-attributes-per-line
```

It makes it more difficult to grok without specifying the actual property name.

michalsnik:
This PR addresses issue #616

[x] Replaces white spaces with new line instead of just adding new line
[x] Improves `binding` error message
[x] Improves `event` error message
[x] Improves `directive` error message

### DIFF
--- a/lib/rules/max-attributes-per-line.js
+++ b/lib/rules/max-attributes-per-line.js
@@ -70,6 +70,7 @@ module.exports = {
     const multilineMaximum = configuration.multiline
     const singlelinemMaximum = configuration.singleline
     const canHaveFirstLine = configuration.allowFirstLine
+    const template = context.parserServices.getTemplateBodyTokenStore && context.parserServices.getTemplateBodyTokenStore()
 
     return utils.defineTemplateBodyVisitor(context, {
       'VStartTag' (node) {
@@ -78,17 +79,17 @@ module.exports = {
         if (!numberOfAttributes) return
 
         if (utils.isSingleLine(node) && numberOfAttributes > singlelinemMaximum) {
-          showErrors(node.attributes.slice(singlelinemMaximum), node)
+          showErrors(node.attributes.slice(singlelinemMaximum))
         }
 
         if (!utils.isSingleLine(node)) {
           if (!canHaveFirstLine && node.attributes[0].loc.start.line === node.loc.start.line) {
-            showErrors([node.attributes[0]], node)
+            showErrors([node.attributes[0]])
           }
 
           groupAttrsByLine(node.attributes)
             .filter(attrs => attrs.length > multilineMaximum)
-            .forEach(attrs => showErrors(attrs.splice(multilineMaximum), node))
+            .forEach(attrs => showErrors(attrs.splice(multilineMaximum)))
         }
       }
     })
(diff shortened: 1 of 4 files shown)


lib/rules/max-attributes-per-line.js
```diff
@@ -128,16 +129,45 @@ module.exports = {
...
     }
 
-    function showErrors (attributes, node) {
+    function getPropData (prop) {
+      let propType = 'Attribute'
+      let propName = prop.key.name
+
+      if (utils.isBindingAttribute(prop)) {
+        propType = 'Binding'
+        propName = prop.key.raw.argument
```
armano2: argument can be null when `v-bind="$props"` and error message is going to look:
```
Binding "null" should be on a new line.
```

```suggestion
        propName = prop.key.raw.argument || 'object as props'
```

see. https://vuejs.org/v2/guide/components-props.html#Passing-the-Properties-of-an-Object

michalsnik: Good catch! I forgot about this case. I think it should actually report `Directive "bind" should be on a new line`

armano2: thank you 👍

Issue #251, opened 2017-11-26
Rule proposal: `component-name-casing`


michalsnik:
**Style guide:**
https://vuejs.org/v2/style-guide/#Component-name-casing-in-JS-JSX-strongly-recommended

**Description:**
We already have a rule called [name-property-casing](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/name-property-casing.md) but it only checks a property `name`, and dosen't care about `Vue.component('TheComponent', ...)` case.

This rule could either extend the `name-property-casing` with extra checks or simply fill the gap and check only the second case, but if someone would want to change default casing he would need to do it in two places, and that's rather not what we want, so I'm leaning more towards one rule for both cases.

mysticatea:
I agree with you.

But I think that the name `component-name-casing` causes confusing between definition's and reference's.

How about `component-definition-name-casing` and `component-reference-name-casing`?

armano2:
i can take this one.

i like one rule approach more to. but i'm not sure about name?

`component-definition-name-casing` witch checks name property and `Vue.component('***'` seems good.

armano2:
fixed in #646

PR #646, opened 2018-11-11
⭐️New: Add `component-definition-name-casing` rule.

armano2:
this rule replaces `vue/name-property-casing`
fix issue: #251

### DIFF
--- a/docs/rules/component-definition-name-casing.md
+++ b/docs/rules/component-definition-name-casing.md
@@ -1,10 +1,8 @@
-# enforce specific casing for the name property in Vue components (vue/name-property-casing)
+# enforce specific casing for component definition name (vue/component-definition-name-casing)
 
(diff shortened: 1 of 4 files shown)


docs/rules/component-definition-name-casing.md
```diff
@@ -0,0 +1,47 @@
...
+export default {
+  name: 'MyComponent'
+}
+```
+```js
+Vue.component('MyComponent', {
+  
+})
+```
+
```
ota-meshi: @armano2 Hello!
Can this rule be verified if the user defines only the name of the component as follows?

```js
const component = {} // ex. import external components
Vue.component('MyComponent', component)
```

armano2: Its working like that, but i will add additional test for it

armano2: @ota-meshi i did a little more testing, and there was few uncovered cases, now it should be ok :>

PR #647, opened 2018-11-12
⭐️New: Add `vue/valid-v-bind-sync` rule

### LINKED ISSUE #639: Rule Proposal: `valid-v-bind-sync`

This rule checks `.sync` modifier with `v-bind`.
Most checking of this rule will do the same check as `vue/valid-v-model`.

[Custom Events `.sync` Modifier](https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier)

[ ] Enforces code style
[X] Warns about a potential error
[ ] Other (please specify:)

```vue
<TheComponent v-bind:title.sync="foo + bar" />
<input v-bind:title.sync="foo"/>
<div v-for="todo in todos">
  <TheComponent v-bind:title.sync="todo" />
</div>
```

**Why should this rule be included?**

The `.sync` modifier behaves similarly to v-model, so the same problem may get into it.

I would like to ask your opinion whether this check should be a separate rule or included in `valid-v-bind`.

ota-meshi:
This PR add `vue/valid-v-bind-sync` rule. (#639)

`vue/valid-v-bind-sync` rule reports `.sync` modifier on `v-bind` directives in the following cases:

- The `.sync` modifier does not have the attribute value which is valid as LHS. E.g. `<MyComponent v-bind:aaa.sync="foo() + bar()" />`
- The `.sync` modifier is on non Vue-components. E.g. `<input v-bind:aaa.sync="foo"></div>`
- The `.sync` modifier's reference is iteration variables. E.g. `<div v-for="x in list"><MyComponent v-bind:aaa.sync="x" /></div>`

Close #639


ota-meshi:
@armano2 Thank you for your review!
I changed it. so please check agein.

lib/rules/valid-v-bind-sync.js
```diff
@@ -0,0 +1,136 @@
...
+            })
+          }
+
+          for (const reference of node.value.references) {
+            const id = reference.id
+            if (id.parent.type !== 'VExpressionContainer') {
+              continue
+            }
+
+            const variable = getVariable(id.name, element)
```
mysticatea: Maybe you can use `reference.variable` property. The `reference.variable` property is the resolved variable if the variable is defined in the template (by `v-for` or scope).

lib/rules/valid-v-bind-sync.js
```diff
@@ -0,0 +1,136 @@
...
+  return true
+}
+
+/**
+ * Check whether the given node can be LHS.
+ * @param {ASTNode} node The node to check.
+ * @returns {boolean} `true` if the node can be LHS.
+ */
+function isLhs (node) {
+  return node != null && (
```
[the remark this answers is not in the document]
mysticatea: Well, I love `!= null` in this case. It's the best way to verify whether the `node` is neither null nor undefined. `Boolean(node)` converts falsy values such as `0`/`""` to `false` (though I know it's no problem in this case).

lib/rules/valid-v-bind-sync.js
ota-meshi: @mysticatea 
Thank you for the information!
That went well!

PR #654, opened 2018-11-12
chore: Report errors only for base and essential rules

michalsnik:
Currently we report errors for all rules, but it's not a very semantic way of telling user something's not exactly ok. We should rather report errors only for things that are crucial - that is `base` and `essential` rules, and for recommendations throw warnings instead. It will also give users the ability to set treshold value. cc @chrisvfritz

### DIFF
--- a/tools/update-lib-configs.js
+++ b/tools/update-lib-configs.js
@@ -14,9 +14,11 @@ const path = require('path')
 const eslint = require('eslint')
 const categories = require('./lib/categories')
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I like it. 🙂

armano2:
@michalsnik should we add this to documentation?

chrisvfritz:
@armano2 Yes, good idea. 🙂

PR #503, opened 2018-06-23
🐛Fix: incorrect indentation in `vue/script-indent` rule

ota-meshi:
fixed #441,#443

PR #659, opened 2018-11-13
Fix: ignore preformatted tokens in `html-indent`

### LINKED ISSUE #653: Indent inside pre tag incorrectly applied
* **ESLint version:**	5.9
* **eslint-plugin-vue version:** 5 beta 4	

https://github.com/nuxt/nuxt.js/blob/dev/.eslintrc.js

**What did you do?**	
https://github.com/nuxt/nuxt.js/blob/dev/examples/routes-meta/pages/about.vue

No error for indent inside pre tag, because it will change the result in browser

**What actually happened?**
https://circleci.com/gh/nuxt/nuxt.js/15089?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ota-meshi:
This PR is for fix the issue #653.

I stored tokens in the `<pre>` element, and excluded those tokens from validation.

--
Fixed #653

The cause of this issue was due to side effects of #503.


ota-meshi:
I thought about fixing with few changes, but because I could not solve the problem of become invalid with other patterns, I solved it with logic of this PR.

lib/utils/indent-common.js
```diff
@@ -301,6 +302,29 @@ module.exports.defineVisitor = function create (context, tokenStore, defaultOpti
     }
   }
 
+  /**
+   * Sets preformatted tokens to the given element node.
+   * @param {Node} node The node to set.
+   * @returns {void}
+   */
+  function setPreformattedTokens (node) {
+    const endToken = node.endTag && tokenStore.getFirstToken(node.endTag) || tokenStore.getTokenAfter(node)
```
michalsnik: Can we use extra parentheses to make this line more clear?
```suggestion
    const endToken = (node.endTag && tokenStore.getFirstToken(node.endTag)) || tokenStore.getTokenAfter(node)
```

tests/lib/rules/html-indent.js
```diff
@@ -281,6 +281,72 @@ tester.run('html-indent', rule, loadPatterns(
...
+      `
+    },
+    {
+      filename: 'test.vue',
+      code: unIndent`
+        <template>
+          <pre>aaa
+        bbb ccc
+        ddd
+        fff</pre>
```
michalsnik: Shouldn't we also indent end tag properly?

lib/utils/indent-common.js
mysticatea: Enabling [no-mixed-operators](https://eslint.org/docs/rules/no-mixed-operators) rule will be worthful in order to automatically check the kind of suggestion. :wink:

lib/utils/indent-common.js
```diff
@@ -821,6 +850,8 @@ module.exports.defineVisitor = function create (context, tokenStore, defaultOpti
       if (node.name !== 'pre') {
         const childTokens = node.children.map(n => tokenStore.getFirstToken(n))
         setOffset(childTokens, 1, startTagToken)
+      } else {
+        setPreformattedTokens(node)
```
mysticatea: My memory is uncertain, but I'd like to try a suggestion.
There is `ignore(node)` function to ignore tokens in a given node. Can we use it here?

tests/lib/rules/html-indent.js
ota-meshi: I think it is not good to indent end tags. Because indenting the end tag affects the contents.

lib/utils/indent-common.js
ota-meshi: `ignore(node)` ignores the indentation of Token, but it can not ignore line indentation.
So, I think that it can not be used to ignore indentation of `<pre>` contents.

lib/utils/indent-common.js
ota-meshi: @michalsnik 
I merged #672 and fixed it.

michalsnik:
@mysticatea can you re-check pls? :)

mysticatea:
I will take a look. Thank you for the ping.

lib/utils/indent-common.js
mysticatea: How about this handler?

```js
    // Ignore inside of <pre>.
    "VElement[name='pre']:exit" (node) {
      const { endTag, children } = node
      for (const child of children) {
        ignore(child)
      }
      if (endTag) {
        ignore(endTag)
      }
    },
```

mysticatea: I got it. Thank you for the confirming!

mysticatea:
Just I have found, `<textarea>` has the same matter as `<pre>`.

ota-meshi:
Do I need to change the processing of `<textarea>` with this PR?

mysticatea:
Probably, separated PR is better.
I will merge this PR.

ota-meshi:
@mysticatea Thank you for merged this PR!
I will begin the work related to `<textarea>` after #676 has been merged.

Issue #667, opened 2018-11-17
Match component file name


rodrigopedra:
**Please describe what the rule should do:**

This rule check if a component's name attribute matches its file name. By default it checks only `.jsx` files, but an additional option is provided to check both `.jsx` and `.vue` files

**What category should the rule belong to?**

- [X] Enforces code style
- [ ] Warns about a potential error
- [ ] Suggests an alternate way of doing something
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**
```jsx
// src/MyComponent.jsx
export default {
  name: 'MComponent', // this rule would warn of the mismatch between component's name
  render() { return <div /> }
}
```

**Additional context**

As discussed with @chrisvfritz in the #589 PR comments

PR #668, opened 2018-11-17
match-component-file-name rule

rodrigopedra:
Described in issue #667 

Discussed in the comments of PR #589 with @chrisvfritz


rodrigopedra:
Hi @armano2 I fixed my commit to pass the lint and I think as I force pushed to have just one commit github cannot show your review. If it is not related to fixing the linting problems, I would as you to please repost your review. Sorry for force pushing I wasn't expecting this to reviwed so quickly.

I think we could add the "run `npm run lint`" to the Contributing section on the `README.md` file.

lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,57 @@
...
+    },
+    fixable: null,
+    schema: [
+      { enum: ['jsx', 'both'] }
+    ]
+  },
+
+  create (context) {
+    return utils.executeOnVueComponent(context, (object) => {
+      const nameProperty = object.properties.find((prop) => prop.key.name === 'name')
```
armano2: spread operator is going to crash here

```suggestion
      const nameProperty = object.properties
        .find(p =>
          p.type === 'Property' &&
          p.key.type === 'Identifier' &&
          p.key.name === 'name' &&
          (
            p.value.type === 'Literal' || (
              p.value.type === 'TemplateLiteral' &&
              p.value.expressions.length === 0 &&
              p.value.quasis.length === 1
            )
          )
        )
```

tests/lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,155 @@
...
+        <script>
+          export default {
+            name: 'MyComponent',
+            template: '<div />'
+          }
+        </script>
+      `,
+      options: ['both']
+    }
+  ],
```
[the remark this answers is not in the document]
rodrigopedra: I added the tests when the component has no `name` aproperty. But I don't understand what you mean with *no literal type*.

lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,57 @@
...
+
+      if (!nameProperty) {
+        return
+      }
+
+      const allowedExtensions = context.options[0] || 'jsx'
+      const name = nameProperty.value.value
```
armano2:
```suggestion
      const name = nameProperty.value.type === 'TemplateLiteral'
          ? nameProperty.quasis[0].value.cooked
          : nameProperty.value.value
```

tests/lib/rules/match-component-file-name.js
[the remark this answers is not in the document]
armano2: i added example for you here
https://github.com/vuejs/eslint-plugin-vue/pull/668#discussion_r234390234
https://github.com/vuejs/eslint-plugin-vue/pull/668#discussion_r234390400

lib/rules/match-component-file-name.js
armano2: i didn't tested it, i just wrote it ~ how i remember AST

lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,58 @@
...
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Check if the component name matches its file name',
+      category: 'recommended',
+      url: 'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.4/docs/rules/match-component-file-name.md'
+    },
+    fixable: null,
+    schema: [
+      { enum: ['jsx', 'both'] }
```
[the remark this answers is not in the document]
rodrigopedra: working on it

armano2:
Missing valid tests:
```js
export default {
  name: componentName,
  template: '<div />'
}
```
```js
// test.js(x)
export default {
  name: `test`,
  template: '<div />'
}
```
```js
// foo.js(x)
export default {
  name: `test${foo}`,
  template: '<div />'
}
```
```js
// foo.js(x)
export default {
  name,
  template: '<div />'
}
```

docs/rules/match-component-file-name.md
```diff
@@ -0,0 +1,41 @@
...
+export default {
+  name: 'MyComponent',
+  render: () {
+    return <h1>Hello world</h1>
+  }
+}
+```
+
+## :wrench: Options
+
```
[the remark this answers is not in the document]
rodrigopedra: will do it, right now refactor to handle those use cases you pointed out on the last comment. First time contribution to Vue. Learning a lot with your help. Thanks

lib/rules/match-component-file-name.js
rodrigopedra: I used and adapted version from your code in [component-definition-name-casing](https://github.com/vuejs/eslint-plugin-vue/pull/646/files#diff-14393fb5384e4c5a97aa80b9c06923b2R74) rule

tests/lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,155 @@
...
+        export default {
+          name: 'MComponent',
+          render() { return <div /> }
+        }
+      `,
+      options: ['both'],
+      errors: [{
+        message: 'Component name should match file name (MyComponent / MComponent).'
+      }]
+    }
```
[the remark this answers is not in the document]
rodrigopedra: Can you verify the new tests if they are ok?

tests/lib/rules/match-component-file-name.js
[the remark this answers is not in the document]
rodrigopedra: Can you verify the new tests if they are ok?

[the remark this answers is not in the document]
armano2: looks fine 👍

rodrigopedra:
Still missing the test of multiple components in the same file... Working (struggling) on it

armano2:
I don't think that's possible without changing logic in 
https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/index.js#L545
`executeOnVue`, `executeOnVueInstance` and `executeOnVueComponent`

armano2:
what's about making array of errors and on `Program:end` checking `array.length === 1`

rodrigopedra:
Hi  @mysticatea , good question. The implementation only checks if the component's name property is equal to its file name. It doesn't do any case conversion. So if your file is names `my-component.jsx`, the rule will check if the component's name property is `name: 'my-component'`.

In the examples and testing I used only PascalCase examples for shortness but if you look into the rule's code you will see that nothing fancy is done when comparing the component's name property and its file name.

mysticatea:
Thank you for the explanation.

Does allowing kebab-case file names make sense?
`utils/casing.js` has utilities about name conversion.
It may be `name !== filename && casing.kebabCase(name) !== filename`.

rodrigopedra:
should it be 

```js
casing.pascalCase(name) !== filename && casing.kebabCase(name) !== filename
```

so it can accept any combination of cases?

- `name: 'my-component'` with `my-component.jsx`
- `name: 'my-component'` with `MyComponent.jsx`
- `name: 'MyComponent'` with `my-component.jsx`
- `name: 'MyComponent'` with `MyComponent.jsx`

mysticatea:
Sounds good to me.

lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,125 @@
...
+        additionalProperties: false
+      }
+    ]
+  },
+
+  create (context) {
+    const options = context.options[0]
+    const extensionsArray = options && options.extensions
+    const allowedExtensions = Array.isArray(extensionsArray) ? extensionsArray : ['jsx']
+    const [, filename, extension] = /^(.+?)\.(.+)$/g.exec(path.basename(context.getFilename()))
```
mysticatea: Looks like crashing when the filename doesn't have that file extension.

(we can use `stdin` to lint (E.g., `curl a_url | eslint --stdin`), in such a case, the filename is `"<INPUT>"`).

How about?:

```suggestion
    const extension = path.extname(context.getFilename())
    const filename = path.basename(context.getFilename(), extension)
```

rodrigopedra: The `extname` method keeps the leading dot in the extension. Should I trim the dot before comparing, or do you think it is a better idea to ask the users to specify the extension with a dot in the options?

mysticatea: Good question. I don't have a strong opinion here.

For example, an ESLint CLI option need the beginning dot: `eslint src --ext .js,.vue`.
But I know another case that doesn't need the beginning dot.

I'm OK with either way.

rodrigopedra: I kept as it was, without the leading dot. As it only compares once per file I don't think it is a performance hit.

docs/rules/match-component-file-name.md
```diff
@@ -0,0 +1,153 @@
...
+- `{extensions: ["jsx"]}` (default) ... verify components in files with `.jsx` extension.
+- `{extensions: ["vue"]}` (default) ... verify components in files with `.vue` extension.
+- `{extensions: ["js"]}` (default) ... verify components in files with `.js` extension.
+- `{extensions: ["jsx", "vue"]}` ... verify components in files with `.jsx` or `.vue` extensions.
+- `{extensions: ["jsx", "js"]}` ... verify components in files with `.jsx` or `.js` extensions.
+- `{extensions: ["vue", "js"]}` ... verify components in files with `.vue` or `.js` extensions.
+- `{extensions: ["jsx", "vue", "js"]}` ... verify components in files with any of the
```
armano2: what do you think about changing it to:

> ## :wrench: Options
> 
> ```json
> {
>   "vue/match-component-file-name": ["error", {
>     "extensions": ["jsx"]
>   }]
> }
> ```
> - `"extensions": []` ... array of file extensions to be verified. Default is set to `["jsx"]`.

rodrigopedra: Looks much better

rodrigopedra:
Hi @armano2 thanks for approving the changes. Let me know if there’s anything else that is needed on this issue. And thanks a lot to you and to @mysticatea for the guidance.

docs/rules/match-component-file-name.md
```diff
@@ -0,0 +1,199 @@
...
+  "vue/match-component-file-name": ["error", {
+    "extensions": ["jsx"],
+    "shouldMatchCase": false
+  }]
+}
+```
+
+- `"extensions": []` ... array of file extensions to be verified. Default is set to `["jsx"]`.
+- `"shouldMatchCase": false` ... boolean indicating if component's name
+  should also match its file name case. Default is set to `false`.
```
armano2:
```suggestion
  should also match its file name case. Default is set to `false`.

## :books: Further reading

 - [Guide - Scoped Slots](https://vuejs.org/v2/style-guide/#Single-file-component-filename-casing-strongly-recommended)
 
```

do you mind adding this?

lib/rules/match-component-file-name.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      return casing.pascalCase(name) === filename || casing.kebabCase(name) === filename
+    }
+
+    function verifyName (node) {
+      let name
+      if (node.type === 'TemplateLiteral') {
+        const quasis = node.quasis[0]
+        name = quasis.value.cooked
```
mysticatea: The `quasis.value.cooked` can be null since ES2018, so it may cause crashing.
Would you add a test case for such a case?

- To ignore malformed template literals: E.g., `` name: `\unicode` ``

mysticatea: Never mind. It's only in TaggedTemplateLiteral.

michalsnik:
Sorry guys, but since we're still in v5 beta and all new rules are on hold - to get everything in place for official release - I reverted this PR and will merge it back in the next few days (after 5.0.0 release). I hope this is fine

michalsnik:
Reverted back to master.

PR #670, opened 2018-11-17
⭐️New: Add `vue/valid-slot-scope` rule

### LINKED ISSUE #661: Rule Proposal: `valid-slot-scope`

This rule checks whether every `slot-scope` (or `scope`) attributes is valid.
This rule reports the following cases:

- The `slot-scope` attribute does not have that attribute value. E.g. `<div slot-scope></div>`
- The `slot-scope` attribute have the attribute value which is extra access to slot data. E.g. `<div slot-scope="prop, extra"></div>`
- The `slot-scope` attribute have the attribute value which is rest parameter. E.g. `<div slot-scope="...props"></div>`

maybe "extra access" and "rest parameter" are mistakes of `destructuring`.

- [ ] Enforces code style
- [X] Warns about a potential error
- [ ] Other (please specify:)

```vue
<template>
  <TheComponent>
    <template slot-scope>
      ...
    </template>
  </TheComponent>
  <TheComponent>
    <template slot-scope="">
      ...
    </template>
  </TheComponent>
  <TheComponent>
    <template slot-scope="...props">
      ...
    </template>
  </TheComponent>
  <TheComponent>
    <template slot-scope="a, b, c">
      <!-- `b` and `c` are extra access. -->
      ...
    </template>
  </TheComponent>
</template>
```

**Additional context**

[slot-scope](https://vuejs.org/v2/api/index.html#slot-scope)

ota-meshi:
This PR add `vue/valid-slot-scope` rule. (#661)

`vue/valid-slot-scope` rule checks whether every `slot-scope` (or `scope`) attributes is valid.

Close #661


docs/rules/valid-slot-scope.md
```diff
@@ -0,0 +1,74 @@
...
+
+## :wrench: Options
+
+Nothing.
+
+## :couple: Related rules
+
+- [no-parsing-error]
+
+
```
armano2:
```suggestion
## :books: Further reading

- [Guide - Scoped Slots](https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots)

```

ota-meshi: Thank you for suggestion!

I added a link.

Since there are many titles `Related links` in other documents, I added them with the title `Related links`.

tests/lib/rules/valid-slot-scope.js
```diff
@@ -0,0 +1,393 @@
...
+        column: 30,
+        endColumn: 37
+      }]
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        <template>
+          <MyComponent>
+            <div slot-scope="a, b  ">
```
mysticatea: I'm surprised in this is not a syntax error.
Does not `no-parsing-error` catch this? If so, it looks like a bug on `vue-eslint-parser`.

armano2:
@mysticatea as far as I know scoped slots allow only one parameter, and documentation is not giving any examples for more than one

tests/lib/rules/valid-slot-scope.js
mysticatea: It was that. Now I remember that I made the parser's behavior same as the original compiler.

But it's odd if orphan tokens exist without syntax errors, so I will add the syntax error into `vue-eslint-parser`.

ota-meshi: After updating `vue-eslint-parser`, I would like to update this rule.

lib/rules/valid-slot-scope.js
```diff
@@ -66,7 +66,7 @@ module.exports = {
     docs: {
       description: 'enforce valid `slot-scope` attributes',
       category: undefined,
-      url: 'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.4/docs/rules/valid-slot-scope.md'
+      url: 'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.5/docs/rules/valid-slot-scope.md'
```
armano2: @michalsnik what do you think about moving it to config than we don't have to modify it in during every release :)

instead
`'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.5/docs/rules/valid-slot-scope.md'`
we will have
`${config.ruleUrl}/valid-slot-scope.md'
or something like that

tests/lib/rules/valid-slot-scope.js
mysticatea: I have released `vue-eslint-parser` [v3.3.0](https://github.com/mysticatea/vue-eslint-parser/releases/tag/v3.3.0).

tangjinzhou:
Why not support multiple parameters?   Our [component](https://github.com/vueComponent/ant-design-vue) uses multiple parameters.
And, I think empty value also should not be reported incorrectly.
 ref https://github.com/vueComponent/ant-design-vue/issues/301

tangjinzhou:
@ota-meshi 

> I think that blank should be reported, because it makes no sense.

The slot-scope is like a function. For example `()=> <span/>`.
I expect to get a slot that can be reused. Its role is to provide  a component with a reusable slot. 
 So should not be limited by the number of parameters.

tangjinzhou:
@ota-meshi 
Maybe it is the best practice.
But should not be so strictly prohibited.
It is best to provide a separate rule instead of no-parsing-error.

ota-meshi:
https://github.com/vuejs/eslint-plugin-vue/issues/722#issuecomment-448079140

Issue #247, opened 2017-11-26
Rule proposal: `one-component-per-file`


michalsnik:
This rule would ensure that a single file describes only one component.
https://vuejs.org/v2/style-guide/#Component-files-strongly-recommended

armano2:
i will take this one should be fast

PR #671, opened 2018-11-18
⭐️New: Add `vue/one-component-per-file` rule

armano2:
ticket: #247


armano2:
i'm unsure what we should do with mixins,

lib/rules/one-component-per-file.js
```diff
@@ -0,0 +1,44 @@
...
+  meta: {
+    docs: {
+      description: 'enforce that each component should be in its own file',
+      category: undefined, // strongly-recommended
+      url: 'https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.4/docs/rules/one-component-per-file.md'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      toManyComponents: 'There is to many components in this file.'
```
seuros: I think we should write : There is more than one component in this file.

casey6: Also this should be `tooManyComponents`.

PR #673, opened 2018-11-18
chore: add type to all rules

armano2:
Make sure that `--fix-type` cli parameter works with our plugin

https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

### DIFF
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -132,7 +132,7 @@ function isNotSideEffectsNode (node, visitorKeys) {
 
 module.exports = {
   meta: {
-    type: 'layout',
+    type: 'suggestion',
     docs: {
       description: 'enforce order of properties in components',
       category: 'recommended',
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -132,6 +132,7 @@ function isNotSideEffectsNode (node, visitorKeys) {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'enforce order of properties in components',
       category: 'recommended',
--- a/lib/rules/prop-name-casing.js
+++ b/lib/rules/prop-name-casing.js
@@ -71,6 +71,7 @@ function create (context) {
 
 module.exports = {
   meta: {
+    type: 'layout',
     docs: {
       description: 'enforce specific casing for the Prop name in Vue components',
       category: 'strongly-recommended',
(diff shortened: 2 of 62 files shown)


lib/rules/require-v-for-key.js
```diff
@@ -17,6 +17,7 @@ const utils = require('../utils')
 
 module.exports = {
   meta: {
+    type: 'suggestion',
```
michalsnik: All rules in `essential` category report rather problems.

armano2: ohh i missed that 👍

lib/rules/order-in-components.js
```diff
@@ -132,6 +132,7 @@ function isNotSideEffectsNode (node, visitorKeys) {
 
 module.exports = {
   meta: {
+    type: 'layout',
```
mysticatea: I believe this is `suggestion` because this rule changes AST hugely.

lib/rules/prop-name-casing.js
```diff
@@ -71,6 +71,7 @@ function create (context) {
 
 module.exports = {
   meta: {
+    type: 'layout',
```
mysticatea: I feel this is `suggestion` because this is about neither whitespaces nor punctuators.

lib/rules/v-bind-style.js
```diff
@@ -17,6 +17,7 @@ const utils = require('../utils')
 
 module.exports = {
   meta: {
+    type: 'layout',
```
mysticatea: I feel this is `suggestion` because this is about neither whitespaces nor punctuators.

lib/rules/v-on-style.js
```diff
@@ -17,6 +17,7 @@ const utils = require('../utils')
 
 module.exports = {
   meta: {
+    type: 'layout',
```
mysticatea: I feel this is `suggestion` because this is about neither whitespaces nor punctuators.

armano2:
`v-on-style` and `v-bind-style` determines only how code looks not how its executed

lib/rules/v-bind-style.js
armano2: `all the parts of the program that determine how the code looks rather than how it executes. These rules work on parts of the code that aren’t specified in the AST`

lib/rules/v-on-style.js
armano2: `all the parts of the program that determine how the code looks rather than how it executes. These rules work on parts of the code that aren’t specified in the AST`

mysticatea:
Those two appear on `vue-eslint-parser`'s AST as `shorthand` property, but indeed, those don't change semantic.

Hm, I'm OK.

PR #676, opened 2018-11-24
Fix: improve comment indentation (fixes #514)

### LINKED ISSUE #514: html-indent: Wants to indent comments too deep after multiline tags

* **ESLint Version:** 4.19.1
* **eslint-plugin-vue Version:** 4.5.0
* **Node Version:** 10.3.0

```
module.exports = {
    root: true,
    plugins: ["vue"],
    extends: [
	"eslint:recommended",
	"plugin:vue/recommended",
    ],
    rules: {
	"vue/max-attributes-per-line": ["warn", {
	    "multiline": {"allowFirstLine": true}
	}],
    }
}
```

```vue
<template>
  <div>
    <input type="text"
           value="foo">
    <!-- comment -->
  </div>
</template>
```

There to be no problem with the file; no warnings/or errors.

```
$ ./node_modules/.bin/eslint test.vue                                                                                    

/home/lukeshu/testcase/test.vue
  5:1  error  Expected indentation of 11 spaces but found 4 spaces  vue/html-indent

  1 error, 0 warnings potentially fixable with the `--fix` option.

$ ./node_modules/.bin/eslint --fix test.vue                                                                                                    
$ cat test.vue                                                                                                                                 
<template>
  <div>
    <input type="text"
           value="foo">
           <!-- comment -->
  </div>
</template>
```

It wants to indent the comment too deep--as if it were inside of the `<input>` tag.  This is because it is aligning to the previous line, not the previous... I'm not sure what word to use... "statement"?

mysticatea:
This PR fixes #514.

This PR modifies `processNodeList()` behavior.
It handles the tokens which are between nodes (e.g., comma tokens) as the same way as the nodes. After this PR, it handles the comments which are between nodes as well.

Then, `validate()` aligns comments that `processNodeList()` handled to the expected column. And it allows comments which are on the expected indent of the previous/next line in the same manner as before.

This fix will improve the comment indentation in templates, blocks, object literals, array literals, and class bodies.


lib/utils/indent-common.js
```diff
@@ -332,14 +342,18 @@ module.exports.defineVisitor = function create (context, tokenStore, defaultOpti
    * @param {Node|null} leftToken The left parenthesis token.
```
ota-meshi: The param name of jsdoc has not been updated.

PR #678, opened 2018-11-24
Update: add baseIndent option to vue/html-indent (fixes #292)

### LINKED ISSUE #292: Provide an option of vue/html-indent to disable the indentation level of 1

* **ESLint Version:4.13.1**
* **eslint-plugin-vue Version:4.0.0-beta.2**
* **Node Version:8.9.1**

``` js
{
    extends: 'plugin:vue/recommended',
    parser: 'vue-eslint-parser',
    plugins: ['vue'],
    rules: {
        'vue/html-indent': [2, 4, {
            attribute: 1,
            closeBracket: 0,
        }],
    }
}
```

```vue
<template>
    <div class="article">
        <div v-if="foo"></div>
        <slot></slot>
    </div>
</template>

<script>
export default {
    name: 'article',
};
</script>
```

```
<template>
<div class="article">
    <div v-if="foo"></div>
    <slot></slot>
</div>
</template>

<script>
export default {
    name: 'article',
};
</script>
```

I don't want to indent first level node of `<template>` because it's inconsistent with that in `<script>`. And sometimes I have to concat no-indentation *.html, *.js, *.css into *.vue. That means that I want they are in the same format.

mysticatea:
This PR adds `baseIndent` option to `vue/html-indent` rule. It's very similar to `vue/script-indent`'s one, but the default value is `1` as same as the current behavior.


michalsnik:
One confict to resolve @mysticatea and we're good to go :)

mysticatea:
I have resolved the conflict.

PR #679, opened 2018-11-25
chore: Fix typo in comments

armano2:


armano2:
@michalsnik done

lib/rules/no-v-html.js
```diff
@@ -6,7 +6,7 @@
 const utils = require('../utils')
 
 // ------------------------------------------------------------------------------
-// Rule Definitionutilu
```
michalsnik: Whaat happened here 🙈 😂

armano2: no clue but editor was screaming at me on this file xd

Issue #682, opened 2018-11-26
empty multiline-html-element-content-newline & singleline-html-element-content-newline


armano2:
### Tell us about your environment
* **ESLint version:**	5.9.0
* **eslint-plugin-vue version:**	5.0.0-beta.5
* **Node version:**	8.12.0

### Please show your full configuration:
```json
{
  "extends": {
    "plugin:vue/strongly-recommended"
  }
}
```

### What did you do?
**multiline-html-element-content-newline**
```vue
<div
  class="a"
  v-if="b"
  @click.stop.prevent
></div>
```
**singleline-html-element-content-newline**
```vue
<slot name="header"></slot>
```

### What did you expect to happen?
When component has no content this rule should not enforce putting new line.

### What actually happened?
```
 Expected 1 line break after opening tag (`<div>`), but no line breaks found
 Expected 1 line break after opening tag (`<div>`), but no line breaks found
```

armano2:
I'm unsure if this should be option or defaul behavior, but its common to use `<slot></slot>` without content ;>

michalsnik:
I agree, this makes sense and it shouldn't report if there is no content inside.

PR #684, opened 2018-11-26
Add `ignoreWhenEmpty` options to singleline/multiline-html-element-content-newline

armano2:
#682


docs/rules/multiline-html-element-content-newline.md
```diff
@@ -65,11 +65,14 @@ This rule enforces a line break before and after the contents of a multiline ele
 ```json
 {
     "vue/multiline-html-element-content-newline": ["error", {
+        "ignoreWhenEmpty": true,
         "ignores": ["pre", "textarea"]
     }]
 }
 ```
 
+- `ignoreWhenEmpty` ... allows having contents in one line, when given element has no content.
```
armano2: this message is wierd

michalsnik: Perhaps `disables reporting when element has no content`?

lib/rules/multiline-html-element-content-newline.js
```diff
@@ -108,6 +114,12 @@ module.exports = {
         }
 
         const getTokenOption = { includeComments: true, filter: (token) => token.type !== 'HTMLWhitespace' }
+        if (ignoreWhenEmpty && node.children.length === 0) {
```
michalsnik: What do you think about using single `if` statement?
```js
if (
  ignoreWhenEmpty &&
  node.children.length === 0 &&
  template.getFirstTokensBetween(node.startTag, node.endTag, getTokenOption).length === 0
) {
  return
}

armano2: i wanted to make it clean but with this wrapping its looks ok 👍

Issue #688, opened 2018-11-29
UTF-8 characters break PascalCase checking


mariszin:
When using UTF8 characters in component name, the plugin fails to recognize PascalCase name.

* **ESLint version:** v4.19.1	
* **eslint-plugin-vue version:** v4.7.1
* **Node version:**	v8.11.3

**Please show your full configuration:**
```js
module.exports = {
  root: true,
  env: {
    node: true
  },
  'extends': [
    'plugin:vue/recommended',
    '@vue/airbnb'
  ],
  rules: {
    'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off'
  },
  parserOptions: {
    parser: 'babel-eslint'
  }
};
```

**What did you do?**	
```vue
export default {
  name: 'DarbībaShow',
};
```

**What did you expect to happen?**
linter should recognize this as a valid PascalCase

**What actually happened?**
```
...\show.vue
  13:9  error  Property name "DarbībaShow" is not PascalCase  vue/name-property-casing

✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

PR #694, opened 2018-12-02
Fix: casing of unicode characters

armano2:
#688


tests/lib/utils/casing.js
```diff
@@ -30,30 +34,56 @@ describe('getConverter()', () => {
...
+    assert.equal(converter('Foo1BAZ'), 'foo1_b_a_z')
     assert.equal(converter('foo1b_a_z'), 'foo1b_a_z')
+    assert.equal(converter('darИībaÊÊw'), 'dar_иība_ê_êw')
+    assert.equal(converter('    foo       Bar   '), 'foo_bar')
+  })
+
+  it('unicode tests', () => {
+    const words = [
+      ['ÊtreSîne', 'être_sîne'],
+      ['darbībaÊÊw', 'darbība_ê_êw'],
```
ota-meshi: This is correct with a popular casing. However, I think there are times when there is a difference with the solution method of Vue.
https://github.com/vuejs/vue/blob/master/src/shared/util.js#L149

I'm worried about being broken by some autofix.

armano2: hmm, how it should work than xd

armano2: vue is not checking for utf characters, they will have to polyfill.

armano2: @ota-meshi maybe we should just ignore non a-z utf characters?

ota-meshi: It's difficult to decide.
@michalsnik what do you think?

ota-meshi: In most of Japan, it is forbidden to use a source other than a-z to avoid problems.

ota-meshi: @armano2 
> what do you think about making new rule which forbids usage of non ascii characters in code?

I think that rule is very nice!

mysticatea: I think that we should follow the original compiler's behavior.

> what do you think about making new rule which forbids usage of non ascii characters in code?

Sounds good to me.

michalsnik: Perhaps if we introduce this rule and put it in the `base` category - we wouldn't need this fix at all?

armano2:
What i should do with this than? Remove support for utf-8 and add rule to ban non a-zA-Z characters instead?

i feel like its base on preference, some ppl may want to use non ascii characters like #688

mysticatea:
I think that:

- the utilities have to behave as same as Vue.js's. 
- the rule which disallows non-ascii *upper case* characters in names would be nice because it can cause unexpected behavior.

michalsnik:
@mysticatea or @ota-meshi can you also take a look at the code here? I would appreciate second pair of eyes before merging it :)

mysticatea:
My opinion is not changed since https://github.com/vuejs/eslint-plugin-vue/pull/694#issuecomment-443990334. I meant that our utilities should behave as same as https://github.com/vuejs/vue/blob/dev/src/shared/util.js. That doesn't have special handling for non-ASCII characters.

armano2:
@mysticatea what should happen with non `A-Za-z:` characters than?

mysticatea:
@armano2 Why does it remove those?

armano2:
no clue, it was copied from vue :)

armano2:
i will replace it with `.replace(/[\-!#$%^&*()_+~`\s\r\n\v\t]+/gu, ' ')`

armano2:
and now it's ignoring unknown (non ascii) characters, and we will have to add new rule to warn about usage of non ascii characters.

mysticatea:
> no clue, it was copied from vue :)

In that case, I think we should keep it as is. My concern is that static analysis gets different behavior to Vue.js compiler and overlooks errors.

armano2:
@mysticatea but in vue its used in different context https://github.com/vuejs/vue/blob/a7658e03a16dc507f0abeba41aee705f773727d0/src/core/util/options.js#L260
and from what i see it got removed anyway

mysticatea:
@mysticatea My question in https://github.com/vuejs/eslint-plugin-vue/pull/694#issuecomment-450610078 was "why remove?". But https://github.com/vuejs/vue/blob/a7658e03a16dc507f0abeba41aee705f773727d0/src/core/util/options.js#L260 is not removal. So I should ask again, why does it remove those? If the removal is different behavior to Vue.js compiler, we should not remove those characters.

armano2:
that's why i changed it now, we are removing now only invalid characters 
```ts
\-!#$%^&*()_+~`"'<>,./?\[\]{}\s\r\n\v\t
```
those characters can't be used as attribute names and property names

mysticatea:
Why does remove invalid characters? I think that we should handle invalid names as invalid names because this is an error checker.

armano2:
we have to somehow determine when word ends, and those are characters on which we split text, vue compile is never removing invalid characters, it's going to be runtime error..

PR #698, opened 2018-12-03
Chore: Add prettier config

michalsnik:
This PR adds automatic generation for `prettier` config, that disables all `layout` rules. We're going to use this config in `@vue/eslint-config-prettier` as well as probably update information in prettier docs to point to this config, instead of a not-synced version that is currently provided.


armano2:
your gitconfig has wrong username/email :>

armano2:
we should also propose update to https://github.com/prettier/eslint-config-prettier/blob/master/vue.js

mysticatea:
Thank you for this PR.

The name `vue/prettier` looks enabling `prettier` format to me. How do you think `vue/no-layout-rules` or something like?

Also we need the document about the preset.

michalsnik:
Good idea @mysticatea I realised after I started creating PR to `eslint-config-prettier`:D

Damn, good catch @armano2 - I ran docs deploy and it must've changed my config.

michalsnik:
Theoretically information about config could stay in prettier documentation. If you're going to configure prettier yourself, this is where this information should be present. And if you'll decide to use @vue/config-prettier - you'll be all set, without ever having to worry about turning any rules off @mysticatea . What do you think?

armano2:
@mysticatea do you mind enabling `require-meta-type` ule in config, cus its used here.

michalsnik:
This looks fine @alexschomb. Can you show full config, how do you run eslint, does the problem exist if you run eslint through terminal? Also what versions of these packages are you using?

PR #699, opened 2018-12-04
Fix: filters validation in `valid-v-model`

armano2:
fixes: #700


mysticatea:
[The document](https://vuejs.org/v2/guide/filters.html) says:

> Filters are usable in two places: mustache interpolations and v-bind expressions (the latter supported in 2.1.0+).

But is it supported on the third place?

armano2:
it is http://jsfiddle.net/zaf7ghey/

Issue #691, opened 2018-12-01
Beta 5 brakes linter


Piterden:
Beta 5 shows many errors. Looks like it not read the config file. Beta 4 is fine.

mysticatea:
Hi. Please open an issue with our issue template. We cannot investigate problems if there are no reproduce.

Piterden:
https://asciinema.org/a/7v6elbIiu2XyqMHgzMwIOAlAe

Piterden:
https://asciinema.org/a/214596

mysticatea:
Would you fill our issue template?
That video shows actual errors but it's not enough to reproduce it because it doesn't show configs and source codes.

Piterden:
https://github.com/Piterden/vue-crossword

https://codesandbox.io/s/github/Piterden/vue-crossword/tree/master/

armano2:
in beta.5 we enabled additional rules

- 8fd7ab841a42b25f1ce0b025063fb16bd41a9337 Update uncategorized rules' categories
  - strongly-recommended: 
    - [component-name-in-template-casing]
    - [multiline-html-element-content-newline]
    - [no-spaces-around-equal-signs-in-attribute]
    - [singleline-html-element-content-newline]
  - essential:
    - [use-v-on-exact]

Piterden:
It broke our staging build

mysticatea:
Thank you.

You have `plugin:vue/recommended` in your config, so this is expected behavior.
We have updated the recommended config in that version.

We don't change the recommended config without major version up, but we did in beta.

Piterden:
Thank you for the explanation. Now it is clear enough.

armano2:
> ✖ 16 problems (2 errors, 14 warnings)
>   0 errors and 14 warnings potentially fixable with the `--fix` option.

Piterden:
`vue/component-name-in-template-casing` provides React style of component names. It is strongly recommended to use kebab-cased names.

mysticatea:
I agree with you. I like kebab-case than PascalCase strongly because HTML is case-insensitive. However, the team seems to have a different preference. :(

PR #703, opened 2018-12-04
Chore: move component-name-in-template-casing rule to uncategorized

mysticatea:
This PR moves `component-name-in-template-casing` to uncategorized from `strongly-recommended` category.

There are some reasons:

1. This rule can break user code by autofix. If a user uses `x-element` Web Component, the rule breaks it. To avoid this broken, the user needs to configure `ignores` option. Therefore, this rule doesn't fit `vue/strongly-recommended` preset.
1. There is a bikeshedding. https://github.com/vuejs/eslint-plugin-vue/pull/397#issuecomment-443663855, https://github.com/vuejs/eslint-plugin-vue/pull/397#issuecomment-443886771, https://github.com/vuejs/eslint-plugin-vue/issues/691#issuecomment-443400173, and https://github.com/vuejs/eslint-plugin-vue/issues/691#issuecomment-443400290. Users which use kebab-casing is much and our style guide allows it.

### DIFF
--- a/lib/rules/component-name-in-template-casing.js
+++ b/lib/rules/component-name-in-template-casing.js
@@ -23,7 +23,7 @@ module.exports = {
     type: 'suggestion',
     docs: {
       description: 'enforce specific casing for the component naming style in template',
-      category: 'strongly-recommended',
+      category: undefined,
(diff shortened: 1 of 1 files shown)


armano2:
@mysticatea i'm unsure if its not better to go for #714 with `registeredComponentsOnly` set to true by default

mysticatea:
I still want to move it to uncategorized because it needs additional configuration for people which use popular libraries.

PR #732, opened 2018-12-18
Update name-property-casing.md

Alex-Sokolov:
Fix code blocks preview


armano2:
there is already change with that: https://github.com/vuejs/eslint-plugin-vue/pull/709

PR #745, opened 2019-01-02
fix(no-async-in-computed-properties): Allow async in nested scopes

### LINKED ISSUE #725: no-async-in-computed-properties fails on objects containing async functions
This can be seen on the demo page for the rule [here](https://vuejs.github.io/eslint-plugin-vue/rules/no-async-in-computed-properties.html). Adding the following computed property causes an ESLint error.

```vue
test () {
  return {
    callback: async (e) => e
  }
}
```

No error since computing the result of the computed property is not asynchronous.

**What actually happened?**
There is an error on the property.

michalsnik:
Fixes #725

### DIFF
--- a/lib/rules/no-async-in-computed-properties.js
+++ b/lib/rules/no-async-in-computed-properties.js
@@ -73,6 +73,7 @@ module.exports = {
 
   create (context) {
     const forbiddenNodes = []
+    const allowedScopes = []
 
     const expressionTypes = {
       promise: 'asynchronous action',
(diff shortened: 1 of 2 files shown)


armano2:
This is not going to be perfect is still going to warn about:

```js
foo() {
 const foo = {
   async bar() => await baz()
 }
 return foo
}
```

@michalsnik what do you think about using [Code Path Analysis](https://eslint.org/docs/developer-guide/code-path-analysis) to determine if function is reachable?

on other hand there if we implement in this way we are not going to catch...

```js
foo() {
 const foo = {
   async bar() => await baz()
 }
 return foo.bar()
}
```

PR #752, opened 2019-01-04
Clarify meaning of the 'wrench' emoji on the Rules page

NathanKleekamp:
The meaning of the wrench emoji on the rules page is a bit ambiguous. These proposed changes spell it out more plainly.


michalsnik:
Thanks for the PR @NathanKleekamp ! I already updated docs, so I'm just going to close it :)

PR #756, opened 2019-01-07
fix(html-indent): Don't nest closeBracket of closing tag

### LINKED ISSUE #728: closeBracket affects end tag
* **ESLint version:**	5.10.0
* **eslint-plugin-vue version:**	5.0.0
* **Node version:**	10.10.0

```js
module.exports = {
	root: true,
	plugins: ['vue'],
	extends: [
		'plugin:vue/essential',
	],
	rules: {
		'vue/html-indent': [
			'error',
			'tab',
			{
				closeBracket: 1,
			},
		],
	},
}
```

**What did you do?**	

```vue
<template>
	<div>
		<a
			>alksjdklaj dklajsd klaj dklajs kldjaksljdkals jdklas
			jdkljaklsdjkalsjdklajskldjaskl jdklasjdklajslkdja</a
		>
	</div>
</template>
```

`yarn eslint --fix test.vue && cat test.vue`

The example deliberately has no whitespace between `<a>`/`</a>` and inner text (this is how it happens in real life - links don't typically have side spaces).

```vue
<template>
	<div>
		<a
			>alksjdklaj dklajsd klaj dklajs kldjaksljdkals jdklas
			jdkljaklsdjkalsjdklajskldjaskl jdklasjdklajslkdja</a
		>
	</div>
</template>
```

**What actually happened?**

```
yarn run v1.10.1
$ /Users/semenov/xxx/node_modules/.bin/eslint --fix test.vue
✨  Done in 0.75s.
```

```vue
<template>
	<div>
		<a
			>alksjdklaj dklajsd klaj dklajs kldjaksljdkals jdklas
			jdkljaklsdjkalsjdklajskldjaskl jdklasjdklajslkdja</a
			>
	</div>
</template>
```

---

I would expect `closeBracket` not to affect the end tag's indentation as it doesn't make sense (the whole `<a>...</a>` element becomes unbalanced).

Alternatively, there should be an option to disable `closeBracket` for end tags separately.

michalsnik:
This PR ignores `closeBracket` option for closing tags. Fixes #728

### DIFF
--- a/tests/lib/rules/html-indent.js
+++ b/tests/lib/rules/html-indent.js
@@ -381,33 +381,6 @@ tester.run('html-indent', rule, loadPatterns(
         text <span /> <!-- comment --></pre>
         </template>
       `
-    },
-    // Close bracket
-    {
-      filename: 'test.vue',
-      code: unIndent`
-        <template>
-          <a
-            href="#"
-            >Content</a
-          >
-        </template>
-      `,
-      options: [2, {
-        closeBracket: 1
-      }]
-    },
-    {
-      filename: 'test.vue',
-      code: unIndent`
-        <template>
-          <div>Content</div
-          >
-        </template>
-      `,
-      options: [2, {
-        closeBracket: 2
-      }]
     }
   ],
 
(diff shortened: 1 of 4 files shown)


tests/lib/rules/html-indent.js
```diff
@@ -381,6 +381,33 @@ tester.run('html-indent', rule, loadPatterns(
         text <span /> <!-- comment --></pre>
         </template>
       `
+    },
+    // Close bracket
+    {
+      filename: 'test.vue',
+      code: unIndent`
+        <template>
```
ota-meshi: If the source code where all lines are indented, You can be moved to fixtures.

michalsnik: I moved these examples to fixtures. And turns out there is one more issue.

The following code is fine:
```html
<!--{"options":[2, { "closeBracket": 1 }]}-->
<template>
  <a
    href="#"
    >Link 1</a
  >
</template>
```

But when there is any element below the link, like here:
```html
<!--{"options":[2, { "closeBracket": 1 }]}-->
<template>
  <a
    href="#"
    >Link 1</a
  >
  <span>text</span>
</template>
```
it wants to additionally indent the close bracket of </a> like this:
```html
<template>
  <a
    href="#"
    >Link 1</a
    >
  <span>text</span>
</template>
```

Do you have an idea what might be causing it @mysticatea ?

mysticatea:
Honestly, I think that `closeBracket` option should apply to all tags regardless of start tags or end tags.

Issue #486, opened 2018-05-31
Rule proposal: no-reserved-component-names


Mouvedia:
see https://stackoverflow.com/a/22545622/248058

Those names should produce an error:

- annotation-xml
- color-profile
- font-face
- font-face-src
- font-face-uri
- font-face-format
- font-face-name
- missing-glyph

chrisvfritz:
This makes sense, but I think it should be focused on avoiding reserved names in component definitions, e.g. `no-reserved-component-names` and it would be in `essentials`.

chrisvfritz:
Some examples of code it should catch:

```js
Vue.component('font-face', {})
```

```js
const name = 'font-face'

Vue.component(name, {})
```

```js
// filename: font-face.vue

export default {
  name: 'font-face'
}
```

```js
// filename: font-face.vue

module.exports = {
  name: 'font-face'
}
```

michalsnik:
I think we should also warn about:
```
export default {
  components: {
    'font-face': ...,
  }
}
```

This returns same warning in Vue app, as former examples. @chrisvfritz @shadskii

shadskii:
I've added warnings for locally registered components in #757. Thank you for the suggestion!

PR #757, opened 2019-01-08
 ⭐️ New: Add rule `no-reserved-component-names`

shadskii:
This rule addresses #486 and prevents name collisions with vue component and standard html elements. Both PascalCase names and kebab-case names of standard html elements are invalid. For example, `'div'` and `'Div'` are both treated as reserved names.


lib/rules/no-reserved-component-names.js
```diff
@@ -0,0 +1,113 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow the use of reserved names in component definitions',
+      category: 'essential',
```
ota-meshi: Please set undefined.

```suggestion
      category: undefined, // 'essential'
```

Categories are set at the timing of the major version up.

lib/rules/no-reserved-component-names.js
```diff
@@ -0,0 +1,113 @@
...
+  'color-profile',
+  'font-face',
+  'font-face-src',
+  'font-face-uri',
+  'font-face-format',
+  'font-face-name',
+  'missing-glyph'
+]
+
+const isLowercase = (word) => (/[a-z]/.test(word))
```
ota-meshi: `/^[a-z]*$/.test(word)` ?

docs/rules/no-reserved-component-names.md
```diff
@@ -0,0 +1,27 @@
...
+</script>
+```
+
+</eslint-code-block>
+
+## :books: Further reading
+
+- [List of html elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)
+- [List of SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element)
+- [Kebab case elements](https://stackoverflow.com/questions/22545621/do-custom-elements-require-a-dash-in-their-name/22545622#22545622)
```
ota-meshi: I think that the next link is the source, so can you add it as well?

https://w3c.github.io/webcomponents/spec/custom/#valid-custom-element-name

lib/rules/no-reserved-component-names.js
shadskii: Yes, that is a more accurate regex!

lib/rules/no-reserved-component-names.js
shadskii: My mistake, I have corrected this with your suggestion!

shadskii:
I've added warnings for locally registered components as @michalsnik suggested in #486.
For example:
```javascript
export default {
  components: {
    'font-face': {},
  }
}
```
Thanks for the suggestion!

tests/lib/rules/no-reserved-component-names.js
```diff
@@ -0,0 +1,324 @@
...
+  'textarea', 'Textarea',
+  'details', 'Details',
+  'dialog', 'Dialog',
+  'menu', 'Menu',
+  'menuitem', 'menuitem',
+  'summary', 'Summary',
+  'content', 'Content',
+  'element', 'Element',
+  'shadow', 'Shadow',
+  'template', 'Template',
```
Mouvedia: [`<template>`](https://vuejs.org/v2/guide/list.html#v-for-on-a-lt-template-gt) should be allowed right?

cf https://html.spec.whatwg.org/multipage/scripting.html#the-template-element

Mouvedia:
The `nobr`, `strike` and `acronym` tags are missing from [this list](https://github.com/vuejs/eslint-plugin-vue/pull/757/files#diff-05040e918e9890028624bfff94c813d4R18).

tests/lib/rules/no-reserved-component-names.js
```diff
@@ -0,0 +1,324 @@
...
+  'slot', 'Slot',
```
Mouvedia: Shouldn't this conflict with [slots](https://vuejs.org/v2/guide/components-slots.html)?

shadskii:
> The `nobr`, `strike` and `acronym` tags are missing from [this list](https://github.com/vuejs/eslint-plugin-vue/pull/757/files#diff-05040e918e9890028624bfff94c813d4R18).

Originally, I chose to only check for non-obsolete elements. `nobr`, `strike` and `acronym` are all obsolete (see their MDN reference [`nobr`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr), [`strike`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike), and [`acronym`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/acronym) ). You bring up a good question: _Should we check for obsolete or deprecated tags?_ I'm inclined to believe the answer is _yes_. Browsers still render deprecated tags and this will result in a name collision if a vue component shares the same name. 

MDN lists 31 obsolete and deprecated elements (see this [list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#Obsolete_and_deprecated_elements)). I have added `deprecated-html-elements.json` with these elements and updated this rule to search for them. 

Thank you for reviewing this pull request and please let me know if you agree with the updates I've made!

tests/lib/rules/no-reserved-component-names.js
shadskii: Yes, I've added a test to validate this.

tests/lib/rules/no-reserved-component-names.js
shadskii: Yes, same for `slot`, I've added a test to validate.

Mouvedia:
> I'm inclined to believe the answer is _yes_.

I agree as well, but I wanted to illustrate the need with elements that are obsolete but still heavily exploited nowadays.

> Thank you for reviewing this pull request and please let me know if you agree with the updates I've made!

These were merely comments. I didn't review your code nor do I have the time to.
Thanks _you_ for the PR.

PR #759, opened 2019-01-08
Update: Add `allowEmptyLines` option to `vue/multiline-html-element-content-newline` rule

### LINKED ISSUE #683: Allow multiple lines after opening an element (multiline-html-element-content-newline)
Hi there.
This is neither a rule proposal nor a bug. I would just like to know if there is a possibility to tailor [the rule](https://github.com/vuejs/eslint-plugin-vue/blob/v5.0.0-beta.3/docs/rules/multiline-html-element-content-newline.md) `multiline-html-element-content-newline` to my needs.

I really like to insert empty lines in longer markup files to group some of the code for better readability. Now my eslint warns me that it found 2 lines instead of 1.
I think the rule itself is useful, but I would like it to not warn me if there are more than 1 lines found.

https://github.com/vuejs/eslint-plugin-vue/blob/07a7d9f6bc7dafe6db0bf8644f35109f3c4efd3c/tests/lib/rules/multiline-html-element-content-newline.js#L425

Thanks for your input on this issue.
Cheers

J

ota-meshi:
Close #683

This PR adds `allowEmptyLines` option to `vue/multiline-html-element-content-newline` rule.

- `allowEmptyLines` ... if `true`, it allows empty lines around content. If you want to disallow multiple empty lines, use [no-multiple-empty-lines] in combination.  
    default `false`

[no-multiple-empty-lines]: https://eslint.org/docs/rules/no-multiple-empty-lines

### DIFF
--- a/docs/rules/multiline-html-element-content-newline.md
+++ b/docs/rules/multiline-html-element-content-newline.md
@@ -79,7 +79,7 @@ This rule enforces a line break before and after the contents of a multiline ele
     "vue/multiline-html-element-content-newline": ["error", {
         "ignoreWhenEmpty": true,
         "ignores": ["pre", "textarea"],
-        "arrowEmptyLine": false
+        "allowEmptyLine": false
     }]
 }
 ```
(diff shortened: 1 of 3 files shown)


ota-meshi:
@armano2 Thank you for pointing that out :confounded:

I fixed it.

docs/rules/multiline-html-element-content-newline.md
```diff
@@ -87,6 +88,8 @@ This rule enforces a line break before and after the contents of a multiline ele
     default `true`
 - `ignores` ... the configuration for element names to ignore line breaks style.  
     default `["pre", "textarea"]`
+- `allowEmptyLine` ... if `true`, allow empty line. If disallow multiple empty lines, use [no-multiple-empty-lines] in combination.
```
michalsnik: Implementation-wise it looks like `allowEmptyLines` might be a bit more correct option 🤔 

As for copy, what do you think about: "... if `true`, it allows empty lines around content. If you want to disallow multiple empty lines, use [...](...)"?

Also, does the core `no-multiple-empty-lines` rule work with Vue template by default?

ota-meshi:
@michalsnik Thank you for your review.
I fixed option name and copy. so please check again.

PR #769, opened 2019-01-12
⭐️New: Add `vue/comma-spacing` rule

ota-meshi:
This PR adds the wrapper rule of the `comma-spacing` core rule to apply to the expressions in `<template>`.


mysticatea:
Thank you for this PR.
Would you use `skipDynamicArguments` option because of a spacing rule?

ota-meshi:
@mysticatea Thank you for your comment.
Since `comma-spacing` uses `sourceCode.tokensAndComments`, it seems that `skipDynamicArguments` will not be effective.

I need to rethink, so I will close this PR.

PR #783, opened 2019-01-19
⭐️New: Add `vue/no-require-false` rule

sizer:
This PR add new rule as `no-require-false`.
`false` for require on props is default value.
Should code `require: true` expressly.


ota-meshi:
I think this rule is the same as another rule `"vue/no-boolean-default": ["error ", "no-default "]`.  
So I will close this PR.

PR #786, opened 2019-01-24
feat: add nuxt component functions to order rule

manniL:


manniL:
@michalsnik We are currently working on a few rules for Nuxt [here](https://github.com/nuxt/eslint-plugin-nuxt) ☺️ 
But of course, duplicating logic, or rules, isn't the goal, so I thought it'd fit better in the Vue plugin as it'll be part of the recommended Nuxt eslint config anyway 🙊

rchl:
I really don't like the order picked for `asyncData` (same for `fetch` probably although I'm not using it).

These are called before component is initialized and before `data` is created so I would expect the order to be:
...
props
asyncData
fetch
data
...
That would match there order of things getting merged together (first asyncData/fetch, then data).

manniL:
@rchl I think the preferred order is always debatable 🙈 

> These are called before component is initialized and before `data` is created

For now yes. With `ssrPrefetch` from v2.6 this will change. You *will* have `this` access in `asyncData`. We just have to get the changes from our side done.

>  That would match the order of things getting merged together (first asyncData/fetch, then data
`fetch` is not related to `data`, only `asyncData` will finally merge with `data`. But I see the point there.

Feel free to submit a PR and we can talk about it there in-depth ☺️

rchl:
Created #823

PR #795, opened 2019-02-01
⭐️New: Add `vue/keyword-spacing` rule

ota-meshi:
This PR adds the wrapper rule of the [keyword-spacing](https://eslint.org/docs/rules/keyword-spacing) core rule to apply to the expressions in `<template>`.


mysticatea:
Thank you for this PR.
Would you use `skipDynamicArguments` option because of a spacing rule?

ota-meshi:
@mysticatea Thank you for your review.
I changed to use `skipDynamicArguments`, so please check again.

PR #816, opened 2019-02-14
Fix no async with empty returns

dougalg:
no-async-in-computed-properties errors out when there is no return value in a function.


mysticatea:
Thank you for the PR!

dougalg:
My pleasure, thank you for the project :)

axit-joost:
Heh, I just opened issue #817, only to find this merge request like 5 seconds later :-) 👍 

Next time I'll check here first 😺

PR #819, opened 2019-02-15
⭐️attributes-order: Allow customize order for specific attribute

gzzhanghao:
This PR allow customize order for specific attributes.

e.g. place `class` before other attributes:

```javascript
{
  "vue/attributes-order": ["error", {
    "order": [
      ["class", ":class"],
      // ...DEFAULT_ORDERS
    ]
  }]
}
```

### DIFF
--- a/tests/lib/rules/attributes-order.js
+++ b/tests/lib/rules/attributes-order.js
@@ -282,6 +282,7 @@ tester.run('attributes-order', rule, {
       code:
         `<template>
           <div
+            v-important
             class="content"
             :class="className"
             v-if="!visible"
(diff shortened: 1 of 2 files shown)


tests/lib/rules/attributes-order.js
```diff
@@ -276,6 +276,94 @@ tester.run('attributes-order', rule, {
             'GLOBAL'
           ]
         }]
+    },
+    {
```
armano2: can you add test case with custom attribute not placed on first position?

gzzhanghao: @armano2 added in 439906d, 663dd4f

ota-meshi:
This PR seems to be out of date, so I close this PR.
You can update the docs, rebase and reopen this PR.

PR #823, opened 2019-02-18
chore: Keep Nuxt's 'asyncData' and 'fetch' with 'data'

rchl:
Change the order of asyncData and fetch properties to be next
to data property. All those are function-wise more or less
equivalent so IMO it makes sense to keep them together. asyncData
and fetch are primarily for setting up component's data.

@manniL


manniL:
I'd love to get feedback about this from @pi0 and @clark ☺️

lib/rules/order-in-components.js
```diff
@@ -19,11 +19,9 @@ const defaultOrder = [
   'inheritAttrs',
   'model',
   ['props', 'propsData'],
-  'data',
+  ['asyncData', 'fetch', 'data'],
```
pi0: I suggest moving fetch to the top of `asyncData` because it is more context less

rchl:
@pi0 but as I understand, having asyncData and fetch in same array means the order can be freely chosen between those. Then it shouldn't be need for moving them around?

lib/rules/order-in-components.js
rchl: @pi0 but as I understand, having asyncData and fetch in same array means the order can be freely chosen between those. Then it shouldn't be need for moving them around?

rchl: OK, I have actually updated to have them in explicit order.

rchl:
Ping

manniL:
@rchl I forgot: could we had `head` somewhere too?

rchl:
Added after props and before data related props. Scream if you disagree.

manniL:
@rchl For me it's ok. Anyway, would it make sense at the end, as you can use methods & stuff in there too?

rchl:
True. How about:
```
  'methods',
  'head',
  ['template', 'render'],
```
At the very end would work for me too but this possibly makes a bit more sense.

manniL:
@rchl Agree with the proposed positioning ☺️

rchl:
Seems ready to merge?

rchl:
Merge please. :)

manniL:
cc @michalsnik @mysticatea :relieved:

vinayakkulkarni:
OK, this messed up my code. lol

```bash
/Users/vinayak/Development/Work/frontend/pages/signup.vue
  101:3  warning  The "methods" property should be above the "head" property on line 90  vue/order-in-components

✖ 145 problems (0 errors, 145 warnings)
```

rchl:
@vinayakkulkarni it's unfortunate that this fix was lingering for so long. You have an option to either override the rule in your eslint configuration or fix automatically with eslint --fix

PR #825, opened 2019-02-21
Use correct 'v-on-function-call' rule name in options example

romansp:
Seems like it was missed in the original #481 PR.


mysticatea:
Thank you!

PR #828, opened 2019-02-22
Rule documentation should match change in #606

andrew-humu:
In #606, the rule `no-unused-component` was updated to detect component usage when registered using camelCase as well as PascalCase.

There's a line of text in the documentation that still states PascalCase as required.  It'd be nice to update that.


mysticatea:
Thank you!

Issue #800, opened 2019-02-05
Rule proposal: slot-syntax rule


ota-meshi:
**Please describe what the rule should do:**

This rule reports deprecated or removed (in Vue.js 2.6.0) slot syntax.

This rule has the following options.

```json
{
  "vue/slot-syntax": ["error", {
    "v-slot": true,
    "slot": true,
    "slot-scope": true,
    "scope": false
  }]
}
```

- `v-slot` ... If `false`, disallow `v-slot`. default `true`
- `slot` ... If `false`, disallow `slot`. default `true`
- `slot-scope` ... If `false`, disallow `slot-scope`. default `true`
- `scope` ... If `false`, disallow `scope`. default `false`

I think we need to discuss the default options.
The listed defaults are designed not to warn v2.5 - 2.6 users.

**What category should the rule belong to?**

- [X] Enforces code style
- [ ] Warns about a potential error
- [X] Suggests an alternate way of doing something
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<template>
  <ListComponent>
    <!-- ✓ GOOD -->
    <template v-slot:name="props">
      {{ props.title }}
    </template>
    <template slot="name" slot-scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
  <ListComponent>
    <!-- ✗ BAD -->
    <template slot="name" scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
</template>
```

- option `{"v-slot": true, "slot": false, "slot-scope": false, "scope": false}`

> You are using Vue.js v2.6.0+ and you want to use only `v-slot`

```vue
<template>
  <ListComponent>
    <!-- ✓ GOOD -->
    <template v-slot:name="props">
      {{ props.title }}
    </template>
  </ListComponent>
  <ListComponent>
    <!-- ✗ BAD -->
    <template slot="name" slot-scope="props">
      {{ props.title }}
    </template>
    <template slot="name" scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
</template>
```

- option `{"v-slot": false, "slot": true, "slot-scope": true, "scope": false}`

> You are using Vue.js v2.5.x

```vue
<template>
  <ListComponent>
    <!-- ✓ GOOD -->
    <template slot="name" slot-scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
  <ListComponent>
    <!-- ✗ BAD -->
    <template v-slot:name="props">
[...]

PR #838, opened 2019-02-28
⭐️New: Add vue/no-deprecated-scope-attribute rule

ota-meshi:
`vue/no-deprecated-scope-attribute` rule reports deprecated `scope` attribute in Vue.js v2.5.0+.

```vue
<template>
  <ListComponent>
    <!-- ✓ GOOD -->
    <template v-slot:name="props">
      {{ props.title }}
    </template>
    <template slot="name" slot-scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
  <ListComponent>
    <!-- ✗ BAD -->
    <template slot="name" scope="props">
      {{ props.title }}
    </template>
  </ListComponent>
</template>
```

---

refs https://github.com/vuejs/eslint-plugin-vue/issues/800#issuecomment-460646676

### DIFF
--- a/lib/rules/syntaxes/scope-attribute.js
+++ b/lib/rules/syntaxes/scope-attribute.js
@@ -1,0 +1,27 @@
+/**
+ * @author Yosuke Ota
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+module.exports = {
+  deprecated: '2.5.0',
+  createTemplateBodyVisitor (context) {
+    /**
+     * Reports `scope` node
+     * @param {VDirectiveKey} scopeKey node of `scope`
+     * @returns {void}
+     */
+    function reportScope (scopeKey) {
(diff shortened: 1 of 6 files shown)


lib/rules/syntaxes/scope-attribute.js
```diff
@@ -0,0 +1,27 @@
...
+      context.report({
+        node: scopeKey,
+        messageId: 'forbiddenScopeAttribute',
+        // fix to use `slot-scope`
+        fix: fixer => fixer.replaceText(scopeKey, 'slot-scope')
+      })
+    }
+
+    return {
+      "VAttribute[directive=true] > VDirectiveKey[name.name='scope']": reportScope
```
lovetingyuan, line 24: Why judge the "scope" as directive ? It is not a directive, it is an attribute

PR #839, opened 2019-02-28
⭐️New: Add vue/no-deprecated-slot-attribute rule

ota-meshi:
`vue/no-deprecated-slot-attribute` rule reports deprecated `slot` attribute in Vue.js v2.6.0+.

```vue
<template>
  <ListComponent>
    <!-- ✓ GOOD -->
    <template v-slot:name>
      {{ props.title }}
    </template>
  </ListComponent>
  <ListComponent>
    <!-- ✗ BAD -->
    <template slot="name">
      {{ props.title }}
    </template>
  </ListComponent>
</template>
```

---

refs https://github.com/vuejs/eslint-plugin-vue/issues/800#issuecomment-460646676


lib/rules/syntaxes/slot-attribute.js
```diff
@@ -0,0 +1,60 @@
...
+              attr.key.name.name === 'slot-scope' ||
+              attr.key.name.name === 'scope'
+            ))
+          return fixSlotToVSlot(fixer, slotAttr, scopeAttr)
+        }
+      })
+    }
+
+    return {
+      "VAttribute[directive=false][key.name='slot']": reportSlot
```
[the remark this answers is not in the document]
ota-meshi: Thank you for your suggestion.
I did not know that `:slot="slotName"` is valid.

lib/rules/syntaxes/slot-attribute.js
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Convert to `v-slot`.
+     * @param {object} fixer fixer
+     * @param {VAttribute | null} slotAttr node of `slot`
+     * @param {VAttribute | null} scopeAttr node of `scope`
+     * @returns {*} fix data
+     */
+    function fixSlotToVSlot (fixer, slotAttr, scopeAttr) {
+      const nameArgument = slotAttr && slotAttr.value && slotAttr && slotAttr.value.value
+        ? `:${slotAttr.value.value}`
```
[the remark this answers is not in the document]
ota-meshi: I changed it so that it will not be fixed in that cases.

ota-meshi:
Thank you for your review!
I changed this. please check again.

tests/lib/rules/no-deprecated-slot-attribute.js
```diff
@@ -0,0 +1,287 @@
...
+      code: `
+      <template>
+        <LinkList>
+          <a :slot="slot.name" />
+        </LinkList>
+      </template>`,
+      output: `
+      <template>
+        <LinkList>
+          <a v-slot:[slot.name] />
```
mysticatea: Actually, this case is invalid because dot in dynamic argument causes syntax error.

```
{
    "directive": true,
    "name": { "name": "slot" },
    "argument": { "name": "[slot" },
    "modifiers": [{ "name": "name]" }]
}
```

mysticatea: Hm, sorry, it might be fixed...

mysticatea: - `v-bind` ... https://jsfiddle.net/mysticatea/zuyjom9m/73/
- `v-slot` ... https://jsfiddle.net/mysticatea/zuyjom9m/76/

Looks like `v-slot:[slot.name]` works, but `v-bind` doesn't work. So this is OK.

ota-meshi: I searched for this issue. I think that it will be fixed by the PR https://github.com/vuejs/vue/pull/9585 :smile:

lib/rules/syntaxes/slot-attribute.js
```diff
@@ -0,0 +1,120 @@
...
+        return true
+      }
+
+      if (!slotAttr.value.expression) {
+        // parse error or empty expression
+        return false
+      }
+      const slotName = sourceCode.getText(slotAttr.value.expression).trim()
+      // If space is included it can not be converted.
+      return !/\s/.test(slotName)
```
mysticatea: Would you make this check stricter a bit? It's good if the same check as `canConvertFromSlotToVSlot` function. For example, it should reject `a>b?c:d` or something like.

```suggestion
      return !/[^a-z]/i.test(slotName)
```

ota-meshi: Thank you for telling me that issue!
I fixed that.

tests/lib/rules/no-deprecated-slot-attribute.js
```diff
@@ -0,0 +1,297 @@
...
+      code: `
+      <template>
+        <LinkList>
+          <a slot="name" />
+        </LinkList>
+      </template>`,
+      output: `
+      <template>
+        <LinkList>
+          <a v-slot:name />
```
mysticatea: Hmm, sorry, I had overlooked this in the previous review.

In some cases, `slot` and `v-slot` should be in different places.

- `v-slot` must be on a `<template>` element which is a direct child of a custom element.
- `v-slot` for the default slot can be on a custom element directly if there are no named slots.

So in this case, the `<a>` element must be wrapped by a `<template>` element.

In [the example of nested default slots](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0001-new-slot-syntax.md#nested-default-slots), it will be moved to the parent element.

Therefore, autofix may be tough to implement... 🤔

ota-meshi: Sorry for the late reply.
Thank you for pointing this out.

Sorry, I did not understand `v-slot` correctly.
I will change it so that it will only autofix if `slot` is attached to `<template>`.

PR #807, opened 2019-02-07
Breaking: support new syntax in Vue.js 2.6

mysticatea:
This PR updates `vue-eslint-parser` to 6.0.0, contains Vue.js 2.6 support.

> https://github.com/mysticatea/vue-eslint-parser/releases/tag/v6.0.0

This release contains drastic change about `VDirectiveKey` node to support dynamic argument syntax. Please see the release note.

This PR contains:

- [X] update package.json for `vue-eslint-parser@^6.0.0`.
- [X] update rules to pass tests.
- [X] add some tests about new syntax to check not to crash.

PR #843, opened 2019-02-28
⭐️New: Add vue/no-static-inline-styles rule

### LINKED ISSUE #826: Rule proposal: vue/no-inline-styles

This rule detect inline styles in components.
It's (subjectively) a good practice to separate styles from the view layout, when possible. This rule detects static inline style objects when they contain literal values. `v-bind:style` and `:style` are acceptable because it's sometimes necessary to set styles based on data, computed property or methods result.

- [x] Enforces code style
- [ ] Other (please specify:)

The following pattern is considered a warning:
```vue
<span style="background-color: #FFF">Hello {{ name }}</span>;
```

The following patterns are not considered a warning::
```vue
<span :style="background-color: color }">Hello {{ name }}</span>;
<span v-bind:style="background-color: color }">Hello {{ name }}</span>;
```

**Additional context**
The rule enforce team members to use SFC `<style scope>` instead of inline style attributes.

ota-meshi:
This PR adds the `vue/no-static-inline-styles` rule.
`vue/no-static-inline-styles` rule reports the static inline style given to the element.

We recommend moving the reported inline styles to the `<style>` element.

---

Close #826


cde06:
Thank you so much for this PR!! 
I hope it will be release soon.

lib/rules/no-static-inline-styles.js
```diff
@@ -0,0 +1,138 @@
...
+      "VAttribute[directive=false][key.name='style']" (node) {
+        context.report({
+          node,
+          messageId: 'forbiddenStyleAttr'
+        })
+      }
+    }
+    if (!context.options[0] || !context.options[0].allowBinding) {
+      visitor[
+        "VAttribute[directive=true][key.name.name='bind'][key.argument.name='style']"
```
hirokiosame: Should this be `[key.name='bind'][key.argument='style']` ?

mysticatea: After #807, AST has been changed a bit to support dynamic argument.

hirokiosame: Ah of course. Thank you for the explanation.

cde06:
Hello,
Can we release this fix or are we waiting something else?
Thanks

PR #860, opened 2019-03-25
docs: add algolia search

posva:
I asked for an index and added the algolia search. Creating the PR to test out the deploy is working correctly

This should help users when searching for rules


posva:
Locally, I keep getting `getHeader` is not a function. Definitely a problem with one of the loaders installed. I think this could be prevented by using a yarn.lock

posva:
Seems to be working 🎉 
Screen Shot 2019-03-25 at 10 01 43

Docsearch is such an amazing product 😍

mysticatea:
Wow, really nice ❤️ 

Thank you very much!

PR #881, opened 2019-04-19
Add vue-eslint-parser version to "Tell us about your environment"

Hyzual:
Since my last two requests were entirely dependent on `vue-eslint-parser`'s new major version breaking rules, It could help maintainers to know that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -12,7 +12,8 @@ about: Create a report to help us improve
 
 **Tell us about your environment**	
 * **ESLint version:**	
-* **eslint-plugin-vue version:**	
+* **eslint-plugin-vue version:** 
+* **vue-eslint-parser version:** 
 * **Node version:**	
 
 **Please show your full configuration:**


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -12,7 +12,8 @@ about: Create a report to help us improve
 
 **Tell us about your environment**	
 * **ESLint version:**	
-* **eslint-plugin-vue version:**	
+* **eslint-plugin-vue version:** 
+* **vue-eslint-parser version:**
```
kazupon: I like this approach!
But, I think that we should be supported to provide easily with reporting tool like `vue info`.
Maybe we just need to write as the notes.

https://github.com/vuejs/eslint-plugin-vue/commit/5d26e943bcba5544d820e9152a796582cc9a4beb#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R49

ota-meshi:
Thank you for this PR.

Currently, there is no way to correctly know the version of the parser that ESLint recognizes, so I think it is difficult for users to recognize the version of the parser. So this PR will be closed.

Thank you for your understanding.

Hyzual:
No problem, thanks!

PR #910, opened 2019-06-22
refactor: remove eslint internal `traverser`

fisker:
eslin@6 has moved `traverser.js` to `lib/shared`

https://github.com/eslint/eslint/pull/11555

PR #912, opened 2019-06-24
pref: improve traverse

fisker:
this PR should speed up the node traverse, improves #910


fisker:
cc @mysticatea

ota-meshi:
I don't like this way of throwing an error because it looks like a trick.
I think that it is good to add API to break to traverseNodes.

fisker:
@ota-meshi `traverseNodes` is `require("vue-eslint-parser").AST.traverse`,  there is not api to break for now, I think throw is the best I can do. 

If this is not OK, I can still add `result` and `parent` check to avoid nest loop, should be better than the original implementation.

PR #923, opened 2019-07-05
Update eslint peer dependency to 6

McInkay:
The plugin has been updated to eslint 6, but the peer dependency still requires 5.


McInkay:
Nevermind, didn't check to see if there was one already.

PR #943, opened 2019-08-14
chore(utils): minor refactor

jamesgeorge007:
Refactor code to be more concise.

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -266,7 +266,7 @@ module.exports = {
       prev != null &&
       prev.startTag.attributes.some(a =>
         a.directive &&
-        (['if', 'else-if'].indexOf(a.key.name.name) !== -1)
+        (['if', 'else-if'].includes(a.key.name.name))
       )
     )
   },
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -266,7 +266,7 @@ module.exports = {
       prev != null &&
       prev.startTag.attributes.some(a =>
         a.directive &&
-        (a.key.name.name === 'if' || a.key.name.name === 'else-if')
+        (['if', 'else-if'].indexOf(a.key.name.name) !== -1)
       )
     )
   },


jamesgeorge007:
cc @kazupon

lib/utils/index.js
```diff
@@ -266,7 +266,7 @@ module.exports = {
       prev != null &&
       prev.startTag.attributes.some(a =>
         a.directive &&
-        (a.key.name.name === 'if' || a.key.name.name === 'else-if')
+        (['if', 'else-if'].indexOf(a.key.name.name) !== -1)
```
kazupon: Thank you for your PR!

IMO, it look good for me, due to  readable. :)
but for large scale projects, it may affect performance. 

@ota-meshi 
What do you think about it?

jamesgeorge007: I guess `includes` suit here better :thinking:

ota-meshi: There are lines that do not use Array in the same file, and other files.
If we rewrite all of these, I'm worried about the performance degradation.

- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/index.js#L436
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/index.js#L794
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/attributes-order.js#L21
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/require-render-return.js#L39
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/require-v-for-key.js#L36
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/require-valid-default-prop.js#L100
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/valid-template-root.js#L92
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/valid-v-model.js#L28
- https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/utils/indent-common.js#L254

jamesgeorge007: IMO the respective change proves to be helpful if more values are to be matched in the future.
Instead of chaining further, we're gonna add them to an array which seems to be much concise.

jamesgeorge007: @ota-meshi shall I update all these files :thinking:

ota-meshi: I thought it was good to follow the [ESLint] habits, so I checked the [ESLint] source code.

Both `||` and Array were used, but `||` is often used.

So, I think it is good not to make this change.

[ESLint]: https://github.com/eslint/eslint

jamesgeorge007: IMO it would be good having the respective change as within this PR 
cc @kazupon

ota-meshi:
I think this change is not good.
If you don't understand, you can resume the discussion.

PR #945, opened 2019-08-15
New: Add `vue/name-property-required` rule

LukeeeeBennett:
Require that a `name` property is set.

Originally wanted to handle name value being an `Identifier` but I'm not sure its worth the hassle as I've only ever written `Literal` name values and I can't think of any critical case that would require an `Identifier` value type.  Please let me know if this is a bad assumption.

Also let me know if you'd prefer just to check that the property key exists and not that the value is also a truthy `Literal`, understandable.

### DIFF
--- a/lib/rules/name-property-required.js
+++ b/lib/rules/name-property-required.js
@@ -1,0 +1,44 @@
+/**
+ * @fileoverview Require a name property in Vue components
+ * @author LukeeeeBennett
+ */
+'use strict'
+
+const utils = require('../utils')
+
+function isNameProperty (node) {
+  return node.type === 'Property' && node.key.name === 'name'
+}
+
+function hasTruthyLiteralValue (node) {
+  return node.value.type === 'Literal' && node.value.value
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'require a name property in Vue components',
+      category: undefined,
+      url: 'https://eslint.vuejs.org/rules/name-property-required.html'
+    },
+    fixable: null,
+    schema: []
+  },
+
+  create (context) {
+    return utils.executeOnVue(context, component => {
+      const isValid = component.properties.some(property => {
+        return isNameProperty(property) &&
+          hasTruthyLiteralValue(property)
+      })
+
+      if (isValid) return
+
+      context.report({
+        node: component,
+        message: 'Required name property is undefined.'
+      })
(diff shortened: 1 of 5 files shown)


LukeeeeBennett:
I'm not sure why the netlify builds failed because there is no error message but non-zero exit code :man_shrugging:

LukeeeeBennett:
*mic tap*

lib/rules/name-property-required.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * @fileoverview Require a name property in Vue components
+ * @author LukeeeeBennett
+ */
+'use strict'
+
+const utils = require('../utils')
+
+function isNameProperty (node) {
+  return node.type === 'Property' && node.key.name === 'name'
```
ota-meshi: Please check if it is computed property.

tests/lib/rules/name-property-required.js
```diff
@@ -0,0 +1,96 @@
...
+  valid: [
+    {
+      filename: 'ValidComponent.vue',
+      code: `
+        export default {
+          name: 'IssaName'
+        }
+      `,
+      parserOptions
+    }
```
ota-meshi: Could you add the following testcase.

```js
export default {
    [name]: 'Name'
}
```

lib/rules/name-property-required.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function hasTruthyLiteralValue (node) {
+  return node.value.type === 'Literal' && node.value.value
```
ota-meshi: I think that another rule is necessary to verify whether it is Literal.
What do you think?

lib/rules/name-property-required.js
LukeeeeBennett: It now ignores computed properties, is this what you wanted?

tests/lib/rules/name-property-required.js
LukeeeeBennett: Thanks :)

lib/rules/name-property-required.js
LukeeeeBennett: Sounds like a good idea yes :) I will open a different PR for it and link here

ota-meshi:
@LukeeeeBennett Hello!
`npm install` failed because the data cached in circleci is broken. It's not your fault.

PR #960, opened 2019-09-17
Fixed jsx parsing error

ota-meshi:
refs eslint/eslint#11018


ota-meshi:
Since there is a problem with the circleci cache, opened a new PR(#961) and close this PR.

PR #961, opened 2019-09-17
Fixed jsx parsing error

ota-meshi:
refs eslint/eslint#11018


ota-meshi:
@kazupon Hello! I think that the installation of `acorn` failed because of the netlify cache.
Can you clear the cache?

Issue #920, opened 2019-07-03
ESLint 6 compatibility


jtsom:
**Tell us about your environment**	
* **ESLint version:**	6.0.1
* **eslint-plugin-vue version:**	5.2.3
* **Node version:** 12.5.0

Upgrading to ESLint v6 shows warnings when installing:
`npm WARN eslint-plugin-vue@5.2.3 requires a peer of eslint@^5.0.0 but none is installed. You must install peer dependencies yourself.`
`npm WARN vue-eslint-parser@5.0.0 requires a peer of eslint@^5.0.0 but none is installed. You must install peer dependencies yourself`

719media:
I hit this same warning. I looked to see if the package.json for `vue-eslint-parser@6`was using eslint 6 yet (to see if whenever eslint-plugin-vue gets around to updating the parser it may alleviate the problem), but it's also still peerDependency on eslint 5, so I'm not sure if there are compat reasons that the authors know about which makes this difficult to just change.

davidpelayo:
Any updates on this?

jpickwell:
@davidpelayo, the v5.2.3 tag was created on June 24, 2019. The commit you refer to was created on July 8, 2019. They haven’t published a new version containing that commit.

jtsom:
Almost a month later now, any chance this will be updated to support ESLint 6 sometime soon??

kaechele:
Given that it seems to be only a matter of tagging a release and releases usually are made periodically, statistically the chance for seeing a release that contains the fix for this issue is increasingly likely.
Stochastically, given that we have seen a release of this project about once every month in the past, it is extremely likely that this will resolve itself within the next 4 weeks.

alex-ut:
This has become a bit more urgent because `eslint-utils`, a dependency of `eslint`, has a known vulnerability in it which is fixed in the `eslint` version 6 suite.  Please make a release of `eslint-plugin-vue` soon(er).

nysos3:
@mysticatea Any hope of a release soon? ❤️  (Especially seeing as you're the one to report the aforementioned vulnerability 👀)

juhasev:
This is a show stopper for many packages out there. I may have to hold on upgrading because of this. Hopefully eslint-6 compatibility gets out soon!

limeandcoconut:
Any movement on this? It's an issue

thefoxie:
Been watching this issue for a long time and still nothing.
[...]

PR #979, opened 2019-11-06
Chore: Update packages to support ESlint v6

michalsnik:
This PR addresses: https://github.com/vuejs/eslint-plugin-vue/issues/920

It:
- bumps few packages to ensure Eslint v6 compatibility
- updates required Node.js version, as in ESLint v6 support for Node.js@6 has been dropped
- removes ESLint v5 compatibility build from our CircleCI config

It looks however we might still need to get `vue-eslint-parser` updated, as its current peerDependencies list include only ESLint `^5.0.0` - cc @mysticatea


package.json
```diff
@@ -40,7 +40,7 @@
     "url": "https://github.com/vuejs/eslint-plugin-vue/issues"
   },
   "engines": {
-    "node": ">=6.5"
+    "node": ">=8.10"
   },
   "peerDependencies": {
     "eslint": "^5.0.0 || ^6.0.0"
```
joey91133: can chanege `^6.0.0` ?

PR #982, opened 2019-11-07
fix attributes-order rule when some types are ommited

shrpne:
Omitting type from order list should allow using attributes of such type everywhere regardless of their order.

But for now it has a bug, which I have fixed.

E.g. omitting `OTHER_ATTR` from the order list should allow `class` attr to be used everywhere.

### DIFF
--- a/lib/rules/attributes-order.js
+++ b/lib/rules/attributes-order.js
@@ -97,8 +97,12 @@ function create (context) {
       previousNode = null
     },
     'VAttribute' (node) {
-      if ((currentPosition === -1) || (currentPosition <= getPosition(node, attributePosition, sourceCode))) {
-        currentPosition = getPosition(node, attributePosition, sourceCode)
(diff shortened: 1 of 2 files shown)


ota-meshi:
Thank you for this PR. Sorry for late reply.

I don't know the benefits of your proposed ordering style. Can you tell me the benefits?

shrpne:
This PR should make `attributes-order` more flexible.

I want it to be more flexible because I prefer HTML attributes to be first and Vue attributes to be second.
E.g. 
```html
<a href="#" v-if="condition" :my-prop="myProp">
<!-- OR -->
<my-component class="foo" v-if="condition" :my-prop="myProp"/>
```

Some other developers may prefer another style and flexibility will allow them to achieve it.

ota-meshi:
Hmm... I don't think you can keep your ordering style by making this rule flexible.
Probably merging your PR is the same as turning off this rule.

I think it's good that you make an ESLint plugin and make rules to keep your favorite style.

PR #985, opened 2019-11-10
Upgrade vue-eslint-parser@^7.0.0

ota-meshi:


g-plane:
What's the progress of this PR? The old version `vue-eslint-parser` depends on `espree@6`, which doesn't support ES2020.

Issue #601, opened 2018-10-09
Rule: vue/sort-props


jpickwell:
**Please describe what the rule should do:**

Allow enforcing a sort order for props.

**What category of rule is this? (place an "X" next to just one item)**

[X] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about:**

```js
/* eslint "vue/sort-props": [2. "asc"] */
{
  props: {
    required: Boolean,
    disabled: Boolean // error
  }
}
```

```js
/* eslint "vue/sort-props": [2. "desc"] */
{
  props: {
    disabled: Boolean,
    required: Boolean // error
  }
}
```

armano2:
duplicate #541

jpickwell:
@armano2, that one is talking about sorting the keys inside a prop; i'm talking about just sorting the props themselves

armano2:
@jpickwell ahh well i got confused a little sorry 👍

jpickwell:
Understandable.

loren138:
Closed by #997

ota-meshi:
Thank you @loren138!

Issue #286, opened 2017-12-08
Question: Configure `vue/order-in-components` to play nice with `sort-keys`?


Xotic750:
I have [`vue/order-in-components` ](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/order-in-components.md) set, but I also have [`sort-keys`](https://eslint.org/docs/rules/sort-keys) enabled for natural sort order.

I wanted to know if there is some configuration that I can use to make the two rules play nicely together? 

Currently when I sort `vue` components in order, then `sort-keys` errors about natural sort order. I know that I can litter my components with `eslint` disable comments, but I would prefer no to do that if it is at all possible?

michalsnik:
I'm afraid we can't do anything about that. One rule does not know anything about any other, and we can't interfere with what other rules should or shouldn't lint. You have to either disable `sort-keys` per component or disable it completely.

Xotic750:
That's a pity. Thanks for the reply.

Xotic750:
Thanks, I am also able to do this with eslint comments but that is something that we wish to avoid.

cantoute:
or disable the rule for *.vue

.eslintrc.js

    overrides: [
        {
            files: ['*.vue'],
            rules: {
                'sort-keys': 'off',
            },
        },

Xotic750:
Thanks, but not the solution we were looking for as then the rule is off for standard objects.

Emoko-Uy:
I haven't found a bettor option... most of my code is in imported from .js files where the filter works

cantoute:
I haven't found a bettor option... most of my code is in imported from .js files where the filter works.
And as the rule sort-keys don't --fix it's not such a use full rule.

Xotic750:
We disabled the rule a while back after, and a colleague wrote a sorter using jscodeshift.

Issue #541, opened 2018-07-31
Add props properties order rule


Soullivaneuh:
As it's done in `order-in-components` rule, it would be nice to have a another one (or improve the existing one?) to force a particular order of each props.

For example, this code part might be OK:

```javascript
props: {
  domain: {
    type: String,
    required: true,
  },
}
```

And not this one, because `required` is placed before `type`:

```javascript
props: {
  domain: {
    required: true,
    type: String,
  },
}
```

What do you think?

PR #997, opened 2019-12-05
New Rule vue/sort-keys

loren138:
This rule creates a version of sort-keys which is compatible with `order-in-components`.

This addresses the issue raised in #286 as well fulfilling the request in #601.


.eslintignore
```diff
@@ -2,7 +2,7 @@
 /coverage
 /node_modules
 /tests/fixtures
-/tests/integrations/*/node_modules
+/tests/integrations/eslint-plugin-import
```
loren138: EsLint 6.7.0 changes eslintignore parsing.  The previous one seems to have ignored the whole folder which I'm now explicitly doing.  Otherwise, you get errors about eslint-plugin-ignore not being installed when running `npm run lint` after a clean install.

ota-meshi:
Thank you for this PR.

Could you change the to ignore the following Vue component options?

- props.xxx  
  e.g.

  ```js
  props: {
    foo: {
      type: String,
      default: 'foo'
    }
  }
  ```

- computed.xxx  
  e.g.

  ```js
  computed: {
    foo: {
      get() {},
      set(foo) {}
    }
  }
  ```

- watch.xxx  
  e.g.

  ```js
  watch: {
    foo: {
      handler() {},
      deep: true,
    }
  }
  ```

- directives.xxx  
  e.g.

  ```js
  directives: {
    focus: {
      bind() {},
      inserted() {}
    }
  }
  ```

- components.xxx
- mixins[*].xxx
- extends
- inject.xxx

- model  
  e.g.

  ```js
  model: {
    prop: 'checked',
    event: 'change'
  }
  ```

loren138:
Hi,
I'm not sure I completely understand your request.

My goal was to make sure all the props, computed, data, methods, etc are in alphabetical order while allowing the component order to be set by `vue/order-in-components`.

That would make this example invalid
```
props: {
  foo: {
    type: String,
    default: 'foo'
  }
}
```

it would have to have default before type
```
props: {
  foo: {
    default: 'foo',
    type: String
  }
}
```

The main goal, however, was to make this invalid:
```
props: {
  foo: {
    default: 'foo',
    type: String
  },
  bar: {
    default: 'foo',
    type: String
  },
}
```

and require it to have `bar` before `foo`:
```
props: {
  bar: {
    default: 'foo',
    type: String
  },
  foo: {
    default: 'foo',
    type: String
  }
}
```

ota-meshi:
Hi! @loren138 .

It doesn't exist now, but we may add rules to enforce the order of props definition in the future. This already has an issue open #541.

I think you need to ignore the some options in the Vue component so that they don't conflict when adding new rules.

loren138:
In that case, what things would be enforced by this rule?  Can you give an example of something that you think would be invalid (while also ignoring all the things you mentioned above)?

ota-meshi:
I think rule #541 will report an error with the following code:

```js
props: {
  foo: {
    default: 'foo',
    type: String // The `type` property should come before the `default` property.
  },
}
```

The following code is valid:

```js
props: {
  foo: {
    type: String,
    default: 'foo'
  },
}
```

The following code does not report an error with rule #541.
However, the `vue/sort-keys` rule reports an error.

```js
props: {
  foo: { },
  bar: { },
}
```

loren138:
Got it, I'll work on that.

loren138:
@ota-meshi I think the latest commit addresses your comments.  I made what to ignore configurable in case people don't want to use any of the new rules or want to force alphabetical order until those are ready.

lib/rules/sort-keys.js
```diff
@@ -0,0 +1,239 @@
...
+
+    // The stack to save the previous property's name for each object literals.
+    let stack = null
+
+    let errors = []
+
+    const reportErrors = (isVue) => {
+      if (isVue) {
+        errors = errors.filter((error) => {
+          let filter = error.hasUpper
```
ota-meshi: Vue component objects may not be top-level objects. Excluding it with hasUpper can cause false positives.

```js
const obj = {
  foo() {
    Vue.component('my-component', {
      name: 'app',
      data() {}
    })
  }
}
```

lib/rules/sort-keys.js
```diff
@@ -0,0 +1,239 @@
...
+
+        if (!isValidOrder(prevName, thisName)) {
+          errors.push({
+            node,
+            hasUpper: !!stack.upper,
+            loc: node.key.loc,
+            message: "Expected object keys to be in {{natual}}{{insensitive}}{{order}}ending order. '{{thisName}}' should be before '{{prevName}}'.",
+            parentName: stack.upper && stack.upper.prevName,
+            grandparentName: stack.upper && stack.upper.upper && stack.upper.upper.prevName,
+            greatGrandparentName: stack.upper && stack.upper.upper && stack.upper.upper.upper && stack.upper.upper.upper.prevName,
```
ota-meshi: It may false negatives if the objects are not chained.

```js
export default {
  computed: {
    foo () {
      return {
        b,
        a
      }
    }
  }
}
```

docs/rules/sort-keys.md
```diff
@@ -0,0 +1,103 @@
...
+# vue/sort-keys
+> enforce sort-keys within components after the top level details
+
+This rule is almost the same rule as core [sorts-keys] rule but it will not error on top component properties allowing that order to be enforced with `order-in-components`.
+
+## Options
+
+```json
+{
+    "sort-keys": ["error", "asc", {"caseSensitive": true, "natural": false, "minKeys": 2}]
```
ota-meshi: Could you include the properties added to the example?

loren138:
@ota-meshi Thanks for catching those issues.  They are fixed now.  Let me know if you see anymore.

loren138:
@ota-meshi Any further comments on this PR?

ota-meshi:
@loren138  sorry, but I haven't checked it yet.
I'm a little busy now, so please wait.

Issue #1035, opened 2020-01-15
Supports for Vue.js 3.x


ota-meshi:
In this issue, I will list the Vue.js 3.x changes that need to be supported by `eslint-plugin-vue`.

---

- [x] Support for Fragments. [Plans for the Next Iteration of Vue.js](https://medium.com/the-vue-point/plans-for-the-next-iteration-of-vue-js-777ffea6fabf)
  - [x] #1038 Change `vue/valid-template-root` to allow multiple root nodes.
  - [x] #1038 Add `vue/no-multiple-template-root` rule that disallows multiple root nodes (for Vue 2).
- [x] Removing `v-bind`'s `.sync` modifier and replacing it with an argument on `v-model`. [RFC0005](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0005-replace-v-bind-sync-with-v-model-argument.md)
  - [x] #1039 Change `vue/valid-v-model` to allows argument (component only).
  - [x] #1039 Add `vue/no-v-model-argument` rule that disallows argument (for Vue 2).
  - [x] #1039 Add `vue/no-deprecated-v-bind-sync` rule  that disallows `v-bind.sync`.
- [x] Functional API Change. [RFC0007](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0007-functional-async-api-change.md)
  - [x] #1119 Add `vue/no-deprecated-functional-template` rule  that disallows `<template functional>`.
- [x] Global API changes. [RFC0009](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0009-global-api-change.md)
  - [x] #1073 Add the CallExpression of `app.component` and `app.mixin` to `utils/index.isVueComponent`.
- [x] `v-model` API changes [RFC0011](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0011-v-model-api-change.md)
  - [x] #1039 Modify `vue/valid-v-model` rule to provide support for custom modifiers. https://github.com/vuejs/eslint-plugin-vue/issues/1035#issuecomment-578446775
  - [x] #1039 Add `vue/no-custom-modifiers-on-v-model` rule to checks if v-model does not contain custom modifiers (Vue 2 backward compatibility) https://github.com/vuejs/eslint-plugin-vue/issues/1035#issuecomment-578446775
- [ ] Composition API. [RFC0013](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0013-composition-api.md)
[...]

PR #1038, opened 2020-01-21
Feature/support for fragments

przemkow:
Hi! After @sodatea 's tweet, I decided to try my first contribution to Vuejs project :) This PR is based on the previous implementation of `vue/valid-template-root` if you think that something is missing or should be changed I will try to do it as soon as possible!

---
### Add support for Vue 3 fragments.    

**Relates to:**
 #1035 

**Scope:**
* `vue/valid-template-root`: Remove a single root element check. New implementation checks two different cases:
  * When `<template>` does not have `src` attribute, it needs to contain at least one element
  * When `<template>` have `src` attribute it needs to be empty

* `vue/no-multiple-template-root`: New rule-based on previous `vue/valid-template-root` implementation used to check if template contains single root element (implemented for Vue 2 backward compatibility)

**Clarification needed:**
How can we differentiate "essential" rules for Vue 2 from the one for Vue 3? Currently `vue/no-multiple-template-root` is added to `configs/essential.js` but I'm not sure how to do it correctly to do not affect build for Vue 3.


ota-meshi:
I think creating categories for Vue3 is a separate task.

PR #1039, opened 2020-01-25
Vue3 v-model API changes

przemkow:
### Add suppor for `v-model` API changes
 * support for `v-model:argument` syntax (replacement for `v-bind.sync`)
 *  support for `v-model.customModifier` syntax

**Relates to:**
 #1035 
[RFC0005](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0005-replace-v-bind-sync-with-v-model-argument.md)
[RFC0011](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0011-v-model-api-change.md)

**Scope:**
* `vue/valid-v-model` - modify rule implementation to do not report an error when:
  * `v-model` is used with an argument on Vue Component. (`v-model:argument` syntax)
  * `v-model` is used with custom modifiers on Vue Component. (`v-model. customModifier` syntax)

* `vue/no-v-model-argument`- new rule which reports an error when `v-model` is used with an argument on Vue Component (Vue 2 backward compatibility rule, previously handled by `vue/valid-v-model`)

* `vue/no-deprecated-v-bind-sync` - new rule which reports an error when deprecated `.sync` modifier is used on `v-bind` directive. 

* `vue/no-custom-modifiers-on-v-model` - new rule which reports an error when v-model is used with custom modifiers on Vue Component (Vue 2 backward compatibility rule, previously handled by vue/valid-v-model)

**Clarification needed:**
- [ ] [Object spread behaviour](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0005-replace-v-bind-sync-with-v-model-argument.md#duplicating-the-object-spread-behavior-of-v-bindsyncxxx) (previously handled by `v-bind.sync="xxx"`). 
According to RFC it's not clear which option was selected for the final implementation. Testing that with `Vue 3.0.0-alpha.3` it seems that spread on `v-model` is currently not supported, nevertheless, it requires additional confirmation before merge.

### DIFF
--- a/lib/rules/no-deprecated-v-bind-sync.js
+++ b/lib/rules/no-deprecated-v-bind-sync.js
@@ -22,7 +22,7 @@ module.exports = {
       category: undefined,
       url: 'https://eslint.vuejs.org/rules/no-deprecated-v-bind-sync.html'
     },
-    fixable: null,
+    fixable: 'code',
     schema: [],
     messages: {
       syncModifierIsDeprecated: "'.sync' modifier on 'v-bind' directive is deprecated. Use 'v-model:propName' instead."
@@ -35,7 +35,17 @@ module.exports = {
           context.report({
             node,
             loc: node.loc,
-            messageId: 'syncModifierIsDeprecated'
+            messageId: 'syncModifierIsDeprecated',
+            fix: (fixer) => {
+              const isUsingSpreadSyntax = node.key.argument == null
+              const hasMultipleModifiers = node.key.modifiers.length > 1
+              if (isUsingSpreadSyntax || hasMultipleModifiers) {
+                return
+              }
+
+              const bindArgument = context.getSourceCode().getText(node.key.argument)
+              return fixer.replaceText(node.key, `v-model:${bindArgument}`)
+            }
           })
         }
       }
--- a/lib/rules/no-deprecated-v-bind-sync.js
+++ b/lib/rules/no-deprecated-v-bind-sync.js
@@ -1,0 +1,44 @@
+/**
+ * @author Przemyslaw Falowski (@przemkow)
+ * @fileoverview Disallow use of deprecated `.sync` modifier on `v-bind` directive (in Vue.js 3.0.0+)
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow use of deprecated `.sync` modifier on `v-bind` directive (in Vue.js 3.0.0+)',
+      category: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-deprecated-v-bind-sync.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      syncModifierIsDeprecated: "'.sync' modifier on 'v-bind' directive is deprecated. Use 'v-model:propName' instead."
+    }
+  },
+  create (context) {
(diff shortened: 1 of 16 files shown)


lib/rules/no-deprecated-v-bind-sync.js
```diff
@@ -0,0 +1,51 @@
...
+  create (context) {
+    return utils.defineTemplateBodyVisitor(context, {
+      "VAttribute[directive=true][key.name.name='bind']" (node) {
+        if (node.key.modifiers.map(mod => mod.name).includes('sync')) {
+          context.report({
+            node,
+            loc: node.loc,
+            messageId: 'syncModifierIsDeprecated',
+            fix: (fixer) => {
+              const isUsingSpreadSyntax = node.key.argument == null
```
ota-meshi: If there are modifiers other than `.sync`, I think it is necessary to disabled autofix. e.g. `.prop`

tests/lib/rules/no-deprecated-v-bind-sync.js
```diff
@@ -0,0 +1,113 @@
...
+    {
+      filename: 'test.vue',
+      code: "<template><MyComponent :foo='bar'/></template>"
+    }
+  ],
+
+  invalid: [
+    {
+      filename: 'test.vue',
+      code: "<template><MyComponent v-bind:foo.sync='bar'/></template>",
```
ota-meshi: Could you please add an autofix test by adding a `output` property.

https://eslint.org/docs/developer-guide/nodejs-api#ruletester

tests/lib/rules/no-deprecated-v-bind-sync.js
```diff
@@ -0,0 +1,113 @@
...
+    {
+      filename: 'test.vue',
+      code: '<template><div><div v-for="x in list"><MyComponent :foo.sync="foo[typeof x]" /></div></div></template>',
+      errors: ["'.sync' modifier on 'v-bind' directive is deprecated. Use 'v-model:propName' instead."]
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div><div v-for="x in list"><MyComponent :foo.sync="foo[tag`${x}`]" /></div></div></template>',
+      errors: ["'.sync' modifier on 'v-bind' directive is deprecated. Use 'v-model:propName' instead."]
+    }
```
ota-meshi: Could you add a test for dynamic arguments.

lib/rules/no-deprecated-v-bind-sync.js
```diff
@@ -0,0 +1,51 @@
...
+              if (isUsingSpreadSyntax) {
+                return
+              }
+              return fixer.replaceText(node.key, `v-model:${node.key.argument.rawName}`)
```
ota-meshi: Maybe this doesn't work for dynamic argument.
Building the fixed source code using `context.getSourceCode(). getText(node.key.argument)` may work.

https://github.com/mysticatea/vue-eslint-parser/blob/master/docs/ast.md#vdirectivekey

lib/rules/no-deprecated-v-bind-sync.js
przemkow: good catch, I disabled fix when there are multiple modifiers

tests/lib/rules/no-deprecated-v-bind-sync.js
przemkow: Sure! Test for dynamic arguments were added in lines [57-68](https://github.com/vuejs/eslint-plugin-vue/pull/1039/files#diff-18440a7c387083fb7072fe9611287ecaR58-R68)

lib/rules/no-deprecated-v-bind-sync.js
przemkow: You are completely right, I totally missed the fact that `VDirectiveKey` supports not only `VIdentifier` but also `VExpressionContainer`. PR updated, dynamic argument should work correctly now.

PR #1042, opened 2020-01-30
Turn off require-v-for-key by default.

Timmmm:
This lint is too restrictive, and isn't able to detect the cases when it is wrong.

See https://github.com/vuejs/vetur/issues/261

### DIFF
--- a/lib/configs/essential.js
+++ b/lib/configs/essential.js
@@ -21,7 +21,7 @@ module.exports = {
     'vue/require-component-is': 'error',
     'vue/require-prop-type-constructor': 'error',
     'vue/require-render-return': 'error',
-    'vue/require-v-for-key': 'error',
+    'vue/require-v-for-key': 'off',
     'vue/require-valid-default-prop': 'error',
     'vue/return-in-computed-property': 'error',
     'vue/use-v-on-exact': 'error',


przemkow:
@Timmmm I recently created PR to this repository and it seems that file you edited is [generated automatically](https://github.com/vuejs/eslint-plugin-vue/blob/a807a1aaa27678ae98bec0a6d41ee64cfff4e0b7/lib/configs/essential.js#L3) during `npm run update` command (it's generated based on the configuration in `lib/rules/*.js`). 

PS . You can also find in `eslint-plugin-vue` docs developers guide which might be pretty helpful ([link](https://eslint.vuejs.org/developer-guide/#proposing-a-new-rule-or-a-rule-change)) 🙂

ota-meshi:
Hi @Timmmm. This category is based on the style guide.

https://vuejs.org/v2/style-guide/#Keyed-v-for-essential

You can change the settings freely using the ".eslintrc. *" Files.

https://eslint.vuejs.org/user-guide/#usage

Timmmm:
Unfortunately disabling this in `.eslintrc` has no effect.

Timmmm:
Also the style guide says:

> Then in the rare, performance-critical scenarios where object constancy isn’t necessary, you can make a conscious exception.

But you can't do that with this rule enabled.

PR #1043, opened 2020-02-02
feat(RFC0015): add no-deprecated-filter rule

przemkow:
### Add `no-deprecated-filter` rule

**Relates to:**
 #1035 
[RFC0015](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0015-remove-filters.md)

**Scope:**
* `vue/no-deprecated-filter` - This rule reports usage of `filters` syntax removed in Vue.js 3.0.0+

**Clarification needed:**
- [ ] This rule is based on `VFilterSequenceExpression`. As Vue.js 3 is going to drop support for filters, are they any plans to drop support for that node in AST generated by `vue-eslint-parser`?  
It seems like a reasonable move as `{{ 0b0001 | 0b0010 }}` in Vue.js 3 AST should be probably interpreted as `BinaryExpression`


ota-meshi:
Thank you for this PR.

I think people rarely do bitwise operations on templates. So I don't think there is a problem, but if you strictly parse it, you need to work on adding options to "vue-eslint-parser".

PR #1064, opened 2020-03-05
Fix iterateProperties to support arrow functions

frenchrabbit:
current iterateProperties doesn't support arrow function properties. However it supports functions and iterate their returned objects. 

So we can't treat properties from arrow functions same way:

default way works:
```
data: function () {
  return {
   property: 'value'
  }
},
```

But following examples won't work:
```
data: () => {
  return {
   property: 'value'
  }
},
```

or like this:

```
data: () => ({
  property: 'value'
}),
```

Pull request adds support for `ArrowFunctionExpression` in `iterateProperties`.

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -693,6 +693,8 @@ module.exports = {
         yield * this.iterateObjectExpression(item.value, name)
       } else if (item.value.type === 'FunctionExpression') {
         yield * this.iterateFunctionExpression(item.value, name)
+      } else if (item.value.type === 'ArrowFunctionExpression') {
+        yield * this.iterateArrowFunctionExpression(item.value, name)
       }
     }
   },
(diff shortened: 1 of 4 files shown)


frenchrabbit:
@ota-meshi , added same cases but with 2 types of arrow functions(`() => {return...}`, `()=>({})`), where was tests with `function () {}`

frenchrabbit:
Thank you, for approving. 

So what's next? We need some other review?
When it could be in npm approximately? 

We a going to release project soon where we need this changes, we could use fork there, but this changes seems quite useful for everyone, that's why we made PR.

ota-meshi:
Hi @frenchrabbit. 
I don't know when the release will be, but I will probably release this change with a beta release (7.x-beta) that supports Vue.js 3.
I would normally release this change in a minor release, but I don't have that time.

frenchrabbit:
Ok, thank you. Looking forward to see it in release.

PR #1068, opened 2020-03-07
New: Add `vue/no-watch-after-await` rule

ota-meshi:
This PR adds the `vue/no-watch-after-await` rule.

The `vue/no-watch-after-await` rule reports the `watch()` after `await` expression. 

---

ref #1035


ota-meshi:
I was unable to reproduce this asynchronous issue with Vue3 "setup".
I keep the status "WIP" until I see the actual problem.

ota-meshi:
I have found that registering a watch asynchronously has the problem that the watch remains after the instance of the Vue component is destroyed.

ota-meshi:
Hmm... but, the remaining watches can be deleted with StopHandle.

Issue #1058, opened 2020-02-24
Allow underscore prefix to disable no-unused-vars


fiesh:
If one wants to repeat a block `n` times, a `v-for="i in n"` comes in handy.

Now `vue/no-unused-vars` will rightfully complain that `i` is unused.

In TypeScript, there's a similar issue with function arguments that are not used on purpose.  These are ignored as unused if prefixed with a `_`.

How about allowing this for `no-unused-vars` as well?  So that `v-for="_i in n"` would not warn, while `v-for="i in n"` would.

IWANABETHATGUY:
I think `_` is better, which is same as other lint rule, i can provide a pr for this

IWANABETHATGUY:
okey, I no-unused-var is not only contains `v-for` stuff, so your proposal is better,    👍

PR #1070, opened 2020-03-08
feat: add fixable, no more error when unused variable with prefix _

IWANABETHATGUY:
implement proposal https://github.com/vuejs/eslint-plugin-vue/issues/1058

feat https://github.com/vuejs/eslint-plugin-vue/issues/1058

### DIFF
--- a/docs/rules/no-unused-vars.md
+++ b/docs/rules/no-unused-vars.md
@@ -34,8 +34,16 @@ This rule report variable definitions of v-for directives or scope attributes if
 
 ## :wrench: Options
 
-Nothing.
-
+```js
+{
+    "vue/no-unsed-vars": [{
+        "ignorePattern": '^_',
+    }]
+}
+```
+- `ignorePattern` ... disables reporting when your definitions of v-for directives or scope attributes match your ignorePattern Regular expression. default `null`, will ignore nothing
+## :rocket: Suggestion
+- When your ignorePattern set to `^_`, we could provide a suggestion which add a prefix`_` to your variable and no more eslint error
 ## :mag: Implementation
 
 - [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-unused-vars.js)
--- a/docs/rules/no-unused-vars.md
+++ b/docs/rules/no-unused-vars.md
@@ -8,13 +8,12 @@ description: disallow unused variable definitions of v-for directives or scope a
 > disallow unused variable definitions of v-for directives or scope attributes
 
 - :gear: This rule is included in all of `"plugin:vue/essential"`, `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"`.
-- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 ## :book: Rule Details
 
 This rule report variable definitions of v-for directives or scope attributes if those are not used.
 
-<eslint-code-block fix :rules="{'vue/no-unused-vars': ['error']}">
+<eslint-code-block :rules="{'vue/no-unused-vars': ['error']}">
 
 ```vue
 <template>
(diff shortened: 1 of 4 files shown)


IWANABETHATGUY:
okay, i will do that.

IWANABETHATGUY:
> Thank you for this PR.
> 
> I like to allow users to ignore variables with specific names.
> 
> But, I think it should be an option like `varsIgnorePattern`, just like ESLint's [`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars).
> 
> https://eslint.org/docs/rules/no-unused-vars
> 
> The `vue/no-unused-vars` rule cannot distinguish between `varsIgnorePattern` and `argsIgnorePattern`, so I think it's better to add `ignorePattern` option.
> 
> Also, if we change the variable name by auto fix, it may conflict with other variables. So, could you change it so that it does not autofix it.

Do you mean only add the `varIgnorePattern` , don't provide fixable option?

lib/rules/no-unused-vars.js
```diff
@@ -18,26 +18,49 @@ module.exports = {
...
+      }
+    ]
   },
 
   create (context) {
+    const option = context.options[0] || { varIgnorePattern: '^_' }
+    const pattern = option['varIgnorePattern'] || '^_'
+    // only use for construct a regularExpression
+    // eslint-disable-next-line no-eval
+    const regExp = eval(`/${pattern}/`)
```
ota-meshi: Do not use `eval()`.

You can handle it in the same way as ESLint's no-unused-vars.

https://github.com/eslint/eslint/blob/master/lib/rules/no-unused-vars.js#L106

lib/rules/no-unused-vars.js
```diff
@@ -18,26 +18,49 @@ module.exports = {
...
       url: 'https://eslint.vuejs.org/rules/no-unused-vars.html'
     },
-    fixable: null,
-    schema: []
+    fixable: 'code',
+    schema: [
+      {
+        'type': 'object',
+        'properties': {
+          'varIgnorePattern': {
```
ota-meshi: Could you change the name to `ignorePattern`?

lib/rules/no-unused-vars.js
```diff
@@ -18,26 +18,49 @@ module.exports = {
...
+            'type': 'string'
+          }
+        },
+        'additionalProperties': false
+      }
+    ]
   },
 
   create (context) {
+    const option = context.options[0] || { varIgnorePattern: '^_' }
```
ota-meshi: If user don't set option, could you change the default behavior to not ignore everything?

lib/rules/no-unused-vars.js
```diff
@@ -18,26 +18,49 @@ module.exports = {
...
             node: variable.id,
             loc: variable.id.loc,
             message: `'{{name}}' is defined but never used.`,
-            data: variable.id
+            data: variable.id,
+            suggest: [
+              {
+                desc: 'Replace the unused-var with default ignore pattern',
+                fix: function (fixer) {
+                  return fixer.replaceText(variable.id, `_${variable.id.name}`)
```
ota-meshi: I think it is good to use SuggestionAPI.

However, this suggestion is only valid if `'^_'` is set.

tests/lib/rules/no-unused-vars.js
```diff
@@ -98,6 +115,21 @@ tester.run('no-unused-vars', rule, {
...
+    },
+    {
+      code: '<template><span slot-scope="props"></span></template>',
+      errors: ["'props' is defined but never used."],
+      options: [{ varIgnorePattern: '^ignore' }]
+    },
+    {
+      code: '<template><span><template scope="props"></template></span></template>',
+      errors: ["'props' is defined but never used."],
+      options: [{ varIgnorePattern: '^ignore' }]
```
ota-meshi: Could you add test cases for suggestions?

https://eslint.org/docs/developer-guide/nodejs-api#testing-suggestions

docs/rules/no-unused-vars.md
```diff
@@ -8,12 +8,13 @@ description: disallow unused variable definitions of v-for directives or scope a
 > disallow unused variable definitions of v-for directives or scope attributes
 
 - :gear: This rule is included in all of `"plugin:vue/essential"`, `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"`.
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 ## :book: Rule Details
 
 This rule report variable definitions of v-for directives or scope attributes if those are not used.
```
ota-meshi: Could you add options document.

IWANABETHATGUY:
> LGTM! Thank you very match!
😀

Issue #922, opened 2019-07-04
Rule proposal: no `method` as component option


manniL:
**Please describe what the rule should do:**
To avoid typos for the `methods` declaration, like
```js
export default {
  method: {
    myMethod(){ /*...*/ }
  }
}
```

I suggest to add a rule which prohibits `method` as a key in the default component export ☺️ 	

**What category should the rule belong to?**

- [ ] Enforces code style
- [x] Warns about a potential error
- [ ] Suggests an alternate way of doing something
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**
```vue
export default {
  name: 'app',
  props: {
    propA: Number
  },
  method: {
    myMethod() { return true }
  }
}
```

ota-meshi:
I don't know how best to implement this rule, so please give your opinions.

We have the following ideas:

- `vue/no-potential-property-typo`
    This is the rule proposed by @IWANABETHATGUY. #1072
    This rule reports option names that similar but unmatched the standard option names or those set by the user.
    Whether the property name is reported depends on the edit distance.
    :+1:: If the option name provided by the user's newly introduced Vue plugin or the user's own option name is not the similar as the existing option name, it will not be reported, so there is no need to change the ESLint configuration.
    :-1:: It may not be possible to report invalid option names because they depend on the edit distance.

    e.g. 

    ```js
    export default {
      // ok
      testtt: {},
    }
    ```

- `vue/no-unknown-component-options`
    This rule reports option names that are not in the whitelist.
    Whitelists are listed by standard option names or set by the user.
    :+1:: Since it reports option names other than whitelist, you can report all incorrect option names.
    :-1:: It is difficult for users to fully build a whitelist. It is necessary to list all the option names provided by the Vue.js plugin used by the user and the option names available in the user's own framework.

- `vue/no-restricted-component-options`
    This rule reports option names that are blacklisted.
    If the user sets the `method` in the blacklist, it can be reported `method`.
    :-1:: With this rule, you must list all possible typos in order to prevent typos.

I like the `vue/no-unknown-component-options` rule because I know the framework I use.
However, I know that people may not be aware of the options provided by the framework they are using. In this case, users may find it easier to use the `vue/no-potential-property-typo` rule.

IWANABETHATGUY:
I think `vue/no-unknown-component-options` is better, for solving the downside of the `vue/no-unknown-component-options`, we could set default presets as ```['vue']```

[...]

PR #1072, opened 2020-03-09
New Add `vue/no-potential-property-typo` rule

IWANABETHATGUY:
https://github.com/vuejs/eslint-plugin-vue/issues/922


IWANABETHATGUY:
i make a new rule , `no-potential-typo` rule to warn user a potential typo error, when user define a custom component option is similar to `vue` built-in option, here is an example;
```javascript
export default {
  // ok 
  data() {

  },
  // ok , this maybe a custom property
  testtt: {},
  // bad, this is more likely to be  a typo
  method: {}
}
```

ota-meshi:
Hi @IWANABETHATGUY.
I think it is difficult to judge typos automatically.
I'm worried that this rule may require users to use more suppression comments.

I like a rule to have users set a whitelist and report non-whitelist properties. e.g. `no-unknown-component-options` rule.
Users can also add properties such as `beforeRouteEnter`.

https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

I like to list similar names in the whitelist as a suggestion to fix the wrong property.

What do you think?

IWANABETHATGUY:
`beforeRouterLeave` is absolute alright for this rule, usually when two string editdistance is. greater than 0 and lessEqual than 1 we report that warning, 
here is the built-in  options  `lib/utils/vue-component-options.json`, we could default this rule for warning not a error.
i don't think people will  define a component option like this
```javascript
export default {
  dat() {},
  method: {},
  comput: {},
  filter: {}
}
```
we default the threshold is 1, only the option in code is very similar but not a vue built-in option will report. we don't check the inner property, so i think this is safe for user;
what do you think?

ota-meshi:
I think people want to detect invalid properties. (The cause of the invalid property may be a typo.)
I think it makes sense to be able to check that properties other than those on the whitelist are not set as a result of the CI.
However, even if you can check that there are no typos (depends on edit distance) as a result of the CI, it doesn't make sense.

I agree that define a component option whitelist is difficult.
We may wish to provide some whitelist presets.
e.g. Vue.js standard, router, vuex, nuxt and more.
Users can whitelist presets and custom property names.

IWANABETHATGUY:
bacuase the indent, the test case is pretty ugly, do you have some good idea for this? and do i need more test case for this rule?

ota-meshi:
Hi @IWANABETHATGUY!
I'm not sure if this rule is best, so I'd like to start with an issue #922.

IWANABETHATGUY:
@ota-meshi could you please help for review

lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,104 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow a potential typo in your component property',
+      categories: ['essential'],
```
ota-meshi: Could you change the categories to be set to undefined?
New rules set categories at major release.

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+
+      filename: 'test.vue',
+      code: `
+      <script>
+      export default {
+        data() {}
+      }
+      </script>
+      `.trim(),
+      options: [{ presets: ['vue'] }]
```
ota-meshi: Could you add test cases using a preset other than vue and test cases using multiple presets.

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+        dat: {},
+        methods: {}
+      }
+      </script>
+              `.trim()
+            }
+          ]
+        }
+      ],
+      options: [{ custom: ['data', 'methods'] }]
```
ota-meshi: Could you use custom option test cases that is not available in the presets?

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+      filename: 'test.vue',
+      code: `
+      <script>
+      export default {
+        data() {}
+      }
+      </script>
+      `.trim(),
+      options: [{ presets: ['vue'] }]
+    }
```
ota-meshi: Could you use test cases that are valid using custom and threshold?

lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,104 @@
...
+          }
+        }
+      }
+    ]
+  },
+
+  create: function (context) {
+    const option = context.options[0] || {}
+    const custom = option['custom'] || []
+    const presets = option['presets'] || []
```
ota-meshi: Could you change Vue.js standard options to validate by default.

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+    {
+      filename: 'test.vue',
+      code: `
+      <script>
+      export default {
+        dat: {},
+        method: {}
+      }
+      </script>
+      `.trim(),
```
ota-meshi: Could you remove `trim()`?

ota-meshi: Please change other same points.

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+      errors: [
+        {
+          message: "'dat' may be a typo, which is similar to option [data].",
```
ota-meshi: Could you test the location of the error using the `line` and `column` properties?

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+      options: [{ presets: ['vue'] }]
+    },
+    // test if give preset and the potentialTypoList length is zero, just for 100% test cover
+    {
+
+      filename: 'test.vue',
+      code: `
+      <script>
+      export default {
+        data() {}
```
ota-meshi: Could you add a test case that tests all the names of the presets?

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,245 @@
...
+      }
+      </script>
+      `.trim(),
+      options: [{ presets: ['vue'] }]
+    }
+    // give me some code that won't trigger a warning
```
ota-meshi: Could you add test case that set `custom: ['abcde', 'abcd']` as option and vue with options for `'abcde'` and `'abcd'`?

lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,104 @@
...
+    const threshold = option['threshold'] || 1
+    let candidateOptions = []
```
ota-meshi: The assigned value does not seem to be used. Remove the assignment or change it to use `push(...arg)`.

lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,104 @@
...
+    if (presets.includes('all')) {
+      candidateOptions = Object.keys(vueComponentOptions).reduce((pre, cur) => {
+        return [...pre, ...vueComponentOptions[cur]]
+      }, [])
+    } else {
+      candidateOptions = presets.reduce((pre, cur) => {
+        return [...pre, ...vueComponentOptions[cur]]
+      }, [])
+    }
+    candidateOptions = [...new Set(candidateOptions.concat(custom))]
```
ota-meshi: Destructuring can be used.

```suggestion
    candidateOptions = [...new Set([...candidateOptions,...custom])]
```

tests/lib/rules/no-potential-component-option-typo.js
IWANABETHATGUY: could you please make it more clear, i was wonder if you mean all the options in preset.json, or add a test case, that presets option is ['vue', 'vue-router', 'nuxt']

IWANABETHATGUY:
i still have a change request, but i have some question, could please give me some tips?

tests/lib/rules/no-potential-component-option-typo.js
ota-meshi: I want you to add a test case that can verify that all of the names defined in `vue-component-options.json` are valid.

ota-meshi:
I want you to add a test case that can verify that all of the names defined in `vue-component-options.json` are valid.

tests/lib/rules/no-potential-component-option-typo.js
IWANABETHATGUY: finish that😊

tests/lib/rules/no-potential-component-option-typo.js
```diff
@@ -0,0 +1,406 @@
...
+        fooooo: {}
+      }
+      </script>
+      `,
+      options: [{ custom: ['custom', 'foo'], threshold: 2 }]
+    },
+    // test all valid vue options
+    {
+      filename: 'test.vue',
+      code: VueComponentJsonCheckVue,
```
ota-meshi, line 125: Hmm.. I don't think it makes sense to put it in a separate file.
Could you put the code inline here?

Issue #958, opened 2019-09-12
Rule Proposal: no-computed-in-data


scyclops:
**Please describe what the rule should do:**

Ensure computed properties are not used in the data() function because data() runs before computed properties are calculated.  So computed properties are undefined when used in data().

**What category should the rule belong to?**

- [ ] Enforces code style
- [X] Warns about a potential error
- [ ] Suggests an alternate way of doing something
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**
```vue

<script>
export default {
    data(){
        return {
            value: 'hello ' + this.world
        }
    },
    computed: {
        world() {
            return 'world'
        }
    }
}
</script>


```

ota-meshi:
The rule has been implemented #1653.

PR #1075, opened 2020-03-11
New Add `vue/no-computed-in-data` Rule

IWANABETHATGUY:
Fixes: #958


IWANABETHATGUY:
finish https://github.com/vuejs/eslint-plugin-vue/issues/958

lib/utils/index.js
```diff
@@ -464,7 +464,7 @@ module.exports = {
     return computedPropertiesNode.value.properties
       .filter(cp => cp.type === 'Property')
       .map(cp => {
-        const key = cp.key.name
+        const key = cp.key.type === 'Identifier' ? cp.key.name : cp.key.type === 'Literal' ? cp.key.value : undefined
         let value
```
IWANABETHATGUY: a property key can be a string literal, so i add this condition

IWANABETHATGUY:
@ota-meshi could please help for review?

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,81 @@
...
+        if (node === dataAstNode) {
+          dataAstNode = undefined
+        }
+      },
+      MemberExpression (node) {
+        if (dataAstNode) {
+          // a memberExpression `like this.a.c.d` -> when traverse to c.d we can got the the full name -> this.a.c.d
+          const fullName = utils.parseMemberExpression(node).join('.')
+          if (memberExpressionMap[fullName]) {
+            memberExpressionMap[fullName] = [...memberExpressionMap[fullName], node]
```
ota-meshi: Check the scope and change to trace correctly. `no-side-effects-in-computed-properties.js` is helpful.
Also add test cases to check if this works correctly.

```js
{
  data () {
    return {
      fn() { return this.foo }
    }
  },
  computed: {
    foo() {}
  }
}
```

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,81 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+const utils = require('../utils')
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'require computed properties are not used in the data()',
+      categories: ['strongly-recommended'],
```
ota-meshi: Could you change the categories to be set to undefined

lib/utils/index.js
ota-meshi: Maybe you can use `getStaticPropertyName(cp)`.

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      'Property:exit' (node) {
+        if (node === dataAstNode) {
+          dataAstNode = undefined
+        }
+      },
+      MemberExpression (node) {
+        if (dataAstNode) {
+          // a memberExpression `like this.a.c.d` -> when traverse to c.d we can got the the full name -> this.a.c.d
+          const fullName = utils.parseMemberExpression(node).join('.')
```
ota-meshi: I think it's enough to check the first two.

```suggestion
          const fullName = utils.parseMemberExpression(node).slice(0,2).join('.')
```

I think this change eliminates the need to use startsWith. at 61.

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,81 @@
...
+          } else {
+            memberExpressionMap[fullName] = [node]
+          }
+        }
+      },
+      ...utils.executeOnVue(context, obj => {
+        const computedPropertyNameList = utils.getComputedProperties(obj).map(item => `this.${item.key}`)
+        Object.keys(memberExpressionMap).forEach(fullName => {
+          let index = -1
+          for (let i = 0, len = computedPropertyNameList.length; i < len; i++) {
```
ota-meshi: Could you use findIndex?

docs/rules/no-computed-in-data.md
```diff
@@ -0,0 +1,81 @@
...
+sidebarDepth: 0
+title: vue/no-computed-in-data
+description: require computed properties are not used in the data()
+---
+# vue/no-computed-in-data
+> require computed properties are not used in the data()
+
+- :gear: This rule is included in `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"`.
+
+> require computed properties are not used in the data()
```
ota-meshi: There is a duplicate description.

docs/rules/no-computed-in-data.md
```diff
@@ -0,0 +1,81 @@
...
+
+- :gear: This rule is included in `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"`.
+
+Please describe the origin of the rule here.
+
+## :book: Rule Details
+
+This rule report computed properties are used in data property
+
+Examples of **incorrect** code for this rule:
```
ota-meshi: Could you rewrite a document like any other document?

`Examples of ** incorrect ** code for this rule:`
Is no longer used.

tests/lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,271 @@
...
+          }
+        },
+        computed: {
+          foo() {}
+        }
+      }
+      </script>
+      `
+    },
+    // should not warn when objectExpression is not a vue component
```
IWANABETHATGUY, line 112: @ota-meshi test for https://github.com/vuejs/eslint-plugin-vue/pull/1075/files#r400643684

lib/rules/no-computed-in-data.js
IWANABETHATGUY: image
if i set the categories to be undefined, the eslint will report this error

IWANABETHATGUY: i change the categories same as `vue/no-side-effects-in-computed-properties`, is that ok?

IWANABETHATGUY: i misunderstand your meaning before, i set it to undefined already😂

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+      },
+      ...utils.executeOnVue(context, obj => {
+        // check if targetObjectExpression is Vue component
+        if (targetObjectExpression !== obj) {
+          return
+        }
+        const computedPropertyNameList = utils.getComputedProperties(obj).map(item => `this.${item.key}`)
+        Object.keys(memberExpressionMap).forEach(fullName => {
+          const index = computedPropertyNameList.findIndex(name => fullName.startsWith(name))
```
ota-meshi, line 89: There is no need to use startsWith.

```suggestion
          const index = computedPropertyNameList.findIndex(name => fullName === name)
```

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,105 @@
...
+      ':function:exit' (node) {
+        scopeStack.pop()
+      },
+      'Property:exit' (node) {
+        if (node === dataAstNode) {
+          dataAstNode = undefined
+        }
+      },
+      MemberExpression (node) {
+        if (dataAstNode && dataPropertyScope === topOfStack(scopeStack)) {
```
ota-meshi, line 68: Vue components may not be at the top level.

```js
export default {
  components: {
    // @vue/component
    MyComponent: {
      data() {
        return {
          bar: this.baz
        }
      },
      computed: {
        baz() {}
      }
    }
  },
}
```

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,105 @@
...
+          // a memberExpression `like this.a.c.d` -> when traverse to c.d we can got the the full name -> this.a.c.d
+          const fullName = utils.parseMemberExpression(node).slice(0, 2).join('.')
+          if (memberExpressionMap[fullName]) {
+            // check if parent node in this array, if true ignore this node, such as `{a: this.a.c.d}`, this traverse function visit order is `this.a.c.d` -> `a.c.d` -> `c.d`
+            const hasParentNodeInArray = memberExpressionMap[fullName].some(nodeInMap => node.parent === nodeInMap)
```
ota-meshi, line 73: I've found an issue with duplicate reports when the chain is deep. Perhaps checking the `parent` is a problem.

```vue
      <script>
      export default {
        data() {
          return {
            a: this.test.foo.bar.baz,
          }
        },
        computed: {
          test() {
            return {}
          },
        }
      }
      </script>
```

lib/rules/no-computed-in-data.js
```diff
@@ -0,0 +1,105 @@
...
+ * @author IWANABETHATGUY
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+const utils = require('../utils')
+function topOfStack (stack) {
+  console.assert(Array.isArray(stack))
```
g-plane, line 12: Why is there a `console.assert` call here?

ota-meshi:
This PR seems to be out of date, so I will close it.

PR #1076, opened 2020-03-12
Fix guide link on valid-v-bind-sync rule docs

Lucidiot:
Copy-paste error on the [valid-v-bind-sync docs](https://eslint.vuejs.org/rules/valid-v-bind-sync.html): The link to the Vue.js guide for `.sync` pointed to a Markdown link syntax.


ota-meshi:
Thank you!!

Issue #932, opened 2019-07-22
vue/prop-name-casing false positive reaction


yekver:
**Tell us about your environment**	
* **ESLint version: 6.0.1**	
* **eslint-plugin-vue version: 5.2.3**	
* **Node version: 12.6.0**	

**What did you do?**	

```vue
export default {
  props: {
    $css: { type: Function, default: () => '' },
  }
}
```

**What did you expect to happen?**
No warning should be emitted

**What actually happened?**
`Prop "$css" is not in camelCase.eslint(vue/prop-name-casing)`

sirlancelot:
The same is happening for props prefixed with `_` as well:

![image](https://user-images.githubusercontent.com/71916/62729940-d8ea9100-b9d3-11e9-85f9-ab83a64f09a9.png)

Even worse, the linter is automatically "fixing" this which actually breaks code that would depend on the underscore being there. I'm using the underscore as a way to represent "internal" props which should never be accessed from templates.

sirlancelot:
Related: #862

PR #1078, opened 2020-03-16
Fix: vue/prop-name-casing false positive reaction #932

IWANABETHATGUY:
fix https://github.com/vuejs/eslint-plugin-vue/issues/932

### DIFF
--- a/lib/utils/casing.js
+++ b/lib/utils/casing.js
@@ -1,6 +1,7 @@
 const assert = require('assert')
 
 const invalidChars = /[^a-zA-Z0-9:]+/g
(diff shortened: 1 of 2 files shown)


IWANABETHATGUY:
@ota-meshi please help for review

ota-meshi:
Hmm. I think the problem is not with the `camelCase()`, but with the rule not having any settings to exclude specify names.

This PR changes may not report non-CamelCase names. e.g. `_item`

ota-meshi:
I close this PR because I don't think it should be merged. If you disagree, please reopen this PR.

PR #1083, opened 2020-03-23
New: Add `vue/no-deprecated-data-object-declaration` rule

yoyo930021:
This PR adds the vue/no-deprecated-data-object-declaration rule.

This rule reports use of deprecated object declaration on `data` property.

[RFC0019](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0019-remove-data-object-declaration.md)
ref #1035

### DIFF
--- a/docs/rules/no-deprecated-data-object-declaration.md
+++ b/docs/rules/no-deprecated-data-object-declaration.md
@@ -1,0 +1,65 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-deprecated-data-object-declaration
+description: disallow using deprecated object declaration on data
+---
+# vue/no-deprecated-data-object-declaration
+> disallow using deprecated object declaration on data
+
+- :gear: This rule is included in all of `"plugin:vue/vue3-essential"`, `"plugin:vue/vue3-strongly-recommended"` and `"plugin:vue/vue3-recommended"`.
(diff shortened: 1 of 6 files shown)


ota-meshi:
Thank you for this PR.

However, I think it is the same as the [no-shared-component-data](https://eslint.vuejs.org/rules/no-shared-component-data.html) rule.

yoyo930021:
The parent instance is slow disallow object declaration.
I think can change it and rename?

ota-meshi:
Thank you for information. You're right.

docs/rules/no-deprecated-data-object-declaration.md
```diff
@@ -0,0 +1,52 @@
...
+export default {
+  data () {
+    return {
+      foo: null
+    }
+  }
+}
+</script>
+```
+
```
ota-meshi: Could you add a sample of `createApp({}).mount()`?

tests/lib/rules/no-deprecated-data-object-declaration.js
```diff
@@ -0,0 +1,262 @@
...
+      code: `
+        new Vue({
+          data: {
+            foo: 'bar'
+          }
+        })
+      `,
+      output: `
+        new Vue({
+          data: function() {
```
ota-meshi: For Vue.js 2 this seems to be an invalid declaration. Vue.js 3 doesn't seem to have `new Vue ()`.
Can you change it to exclude?

https://vuejs.org/v2/api/#data

yoyo930021:
I finished.

PR #1090, opened 2020-04-01
Support module.exports in addition to export default

catrope:
Recognize exported objects in .vue files as Vue components, even if
they're exported using module.exports rather than export default.

Also recognize module.exports for the purposes of require-direct-export:
let A = { /* stuff */ }; module.exports = A;
is now a violation of this rule.

### DIFF
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -88,7 +88,7 @@ All component-related rules are applied to code that passes any of the following
 * `app.component()` expression
 * `app.mixin()` expression
 * `createApp()` expression
-* `export default {}` in `.vue` or `.jsx` file
+* `export default {}` or `module.exports = {}` in `.vue` or `.jsx` file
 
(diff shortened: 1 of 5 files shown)


ota-meshi:
Hi.

You can use "// @vue/component".

https://eslint.vuejs.org/user-guide/#how-does-eslint-detect-components

catrope:
I know, and as a short-term workaround we will do that, but since `module.exports` and `export default` are conceptually the same thing (just in ES5 vs ES6), I thought it would make sense for eslint to automatically detect it as well. Otherwise we'll have to manually add `// @vue/component` above every `module.exports` statement in every `.vue` file in our ES5 repos.

ota-meshi:
Hmm.. As far as I know, `module.exports` is not an ES specification, so I'm wondering if this change needs to be included.

ota-meshi:
I've considered but I don't merge this PR. `module.exports` is not ES specification, so it depends on bundler.
If you want to verify `module.exports`, you use the comments in `//@vue/component`.

Thank you for your understanding.

PR #1099, opened 2020-04-16
New: Add `vue/require-v-if-inside-transition` rule

ota-meshi:
This PR adds the `vue/require-v-if-inside-transition` rule.

This rule reports elements inside `<transition>` that do not control the display.

ref #1035


sodatea:
Hmmm, now I feel a little bit regretful on the rule naming…
Now that `v-show` is also allowed, is `require-v-if-` the right name for it?

ota-meshi:
Hi @sodatea.

I also considered the name, but `v-if` is a very short name and it is easy to understand the meaning of "control the presence of the content". I couldn't think of a better name than this.
Do you have any good name ideas?

sodatea:
What about `require-toggle-inside-transition`? I'm really bad at naming things 😂

ota-meshi:
I think that's good naming! I like that name! I would like to change the rule name in the next release.
Thank you:smile:

Issue #775, opened 2019-01-14
Rule Proposal: Boolean prop shorthand (`prop` instead of `:prop="true"`)


manniL:
**Please describe what the rule should do:**

I've seen code like `<div :boolProp="true" />` often where you could use `<div boolProp />` instead. I'd love to see a rule for that (plus an auto-fix option which should be fairly possible)

**What category should the rule belong to?**
- [x] Enforces code style
- [ ] Warns about a potential error
- [x] Suggests an alternate way of doing something
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**
```vue
<template>
  <div>
    <button :disabled="true">
  </div>
</template>
```

armano2:
i think there should be an option to enable/disable this feature.
- `always`
```vue
// valid
<button :disabled="true">
// invalid
<button disabled>
  ```
- `never`
```vue
// valid
<button disabled>
// invalid
<button :disabled="true">
```

manniL:
@armano2 Fully agree. `always`/`never` should be configurable.

Like the name :relaxed:

PR #1102, opened 2020-04-19
New: Add `vue/require-shorthand-boolean-prop` rule

inker:
This PR adds the `vue/require-shorthand-boolean-prop` rule.

ref https://github.com/vuejs/eslint-plugin-vue/issues/775


ota-meshi:
Thank you for this PR.

Due to the props definition of the component used, `:isValid="true"` and `isValid` are not equivalent.
I don't think this rule can report errors correctly.

What do you think?

inker:
I see what you mean. Currently, it only works for boolean props. Will investigate if it's possible to do prop type checking.

inker:
> Due to the props definition of the used component, `:isValid="true"` and `isValid` are not equivalent.

I see what you mean. Currently, it only works for boolean props. Will investigate if it's possible to do prop type checking.

ota-meshi:
I don't think there is a good way to do type checking right now, so I'll close this PR. If you know a good way, please let me know.

PR #1137, opened 2020-05-18
repro: arrows function as emits validators fails

cexbrayat:
Using an arrow function as an emit validator fails:

```
error: Expected to return a boolean value in "selected" emits validator (vue/return-in-emits-validator) at src/chapters/tests/User.vue:17:19:
  15 |
  16 |   emits: {
> 17 |     selected: () => true
     |                   ^
  18 |   },
```

I was keen to fix it myself, but I don't know what is missing. So here is a simple repro.


ota-meshi:
I will fix this bug. Thank you.

PR #1206, opened 2020-06-11
Chores: Add JSDoc type checking with TypeScript

ota-meshi:
This PR adds JSDoc and TypeScript types to enable type checking.

I also did some refactoring to make the types work better.


lib/rules/no-unused-components.js
```diff
@@ -79,17 +83,23 @@ module.exports = {
             ignoreReporting = true
           }
         },
+        /** @param {VAttribute} node */
         "VAttribute[directive=false][key.name='is']"(node) {
+          if (!node.value) {
```
ota-meshi, line 88: MEMO: This change fixes the bug reported in #1221.

PR #1216, opened 2020-06-23
chore(tools): use ESLint class

### LINKED ISSUE #1215: [tools] use ESLint class instead of CLIEngine
Since version 7.0.0, ESLint supports `ESLint` class instead of `CLIEngine` class.
https://github.com/eslint/eslint/pull/12939
https://github.com/eslint/rfcs/pull/40

So, if it's possible, let's use `ESLint` class!

tyankatsu0105:
fix #1215

### DIFF
--- a/tools/update-lib-configs.js
+++ b/tools/update-lib-configs.js
@@ -88,6 +88,10 @@ categories.forEach((category) => {
 })
 
 // Format files.
-const linter = new eslint.CLIEngine({ fix: true })
-const report = linter.executeOnFiles([ROOT])
-eslint.CLIEngine.outputFixes(report)
+async function main() {
+  const linter = new eslint.ESLint({ fix: true })
+  const report = await linter.lintFiles([ROOT])
+  eslint.ESLint.outputFixes(report)
+}
+
+main()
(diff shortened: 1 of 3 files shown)


tyankatsu0105:
All tests failed... why??🤔

ota-meshi:
Probably using ESLint v7.3 seems to fail some test cases. I think this is because there are some test cases with empty errors in the indented test cases.

I will fix this.

ota-meshi:
@tyankatsu0105 I fixed test cases. Could you rebase this PR?

tools/update-lib-configs.js
```diff
@@ -88,6 +88,10 @@ categories.forEach((category) => {
 })
 
 // Format files.
-const linter = new eslint.CLIEngine({ fix: true })
-const report = linter.executeOnFiles([ROOT])
-eslint.CLIEngine.outputFixes(report)
+async function main() {
```
ota-meshi, line 91: This process is not the "main" of this script, so can you rename it to "format" etc.?

tyankatsu0105: done

PR #1222, opened 2020-06-26
feat: Add no-empty-component-block rule

### LINKED ISSUE #952: New rule: no-empty-component-block
The rule should avoid empty `<style scoped></style>`, script or template tags. If there are no style rules, a style tag is not necessary.

Edge case: Tags that have a src attribute (should be valid).

- [x] Enforces code style
- [ ] Other (please specify:)

```vue
<template>
  <h1>Hey</h1>
</template>

<script></script>

<style scoped></style>
```

```vue
<template>
  <h1>Hey</h1>
</template>

<style scoped></style>
```

```vue
<template>
</template>

<script>
export default { 
  render (h) { return h('div','Hi') }  // Or so
}
</script>

```

tyankatsu0105:
close #952 

This PR adds vue/no-empty-component-block rule.

### DIFF
--- a/lib/rules/no-empty-component-block.js
+++ b/lib/rules/no-empty-component-block.js
@@ -11,6 +11,7 @@
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
+
 /**
  * check whether has attribute `src`
  */
@@ -24,6 +25,17 @@ function hasAttributeSrc(componentBlock) {
     ).length > 0
 
   return hasAttribute && hasSrc
+}
+
+/**
+ * check whether value under the component block is only whitespaces or break lines
+ */
+function isValueOnlyWhiteSpacesOrLineBreaks(componentBlock) {
+  return (
+    componentBlock.children.length === 1 &&
+    componentBlock.children[0].type === 'VText' &&
+    /^(\s|\n)+$/.test(componentBlock.children[0].value)
+  )
 }
 
 module.exports = {
@@ -67,7 +79,10 @@ module.exports = {
           // https://vue-loader.vuejs.org/spec.html#src-imports
           if (hasAttributeSrc(componentBlock)) return
 
-          if (componentBlock.children.length === 0) {
+          if (
+            isValueOnlyWhiteSpacesOrLineBreaks(componentBlock) ||
+            componentBlock.children.length === 0
+          ) {
             context.report({
               node: componentBlock,
               loc: componentBlock.loc,
--- a/lib/rules/no-empty-component-block.js
+++ b/lib/rules/no-empty-component-block.js
@@ -1,0 +1,79 @@
+/**
+ * @author tyankatsu <https://github.com/tyankatsu0105>
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+/**
+ * check whether has attribute `src`
+ */
+function hasAttributeSrc(componentBlock) {
+  const hasAttribute = componentBlock.startTag.attributes.length > 0
+
+  const hasSrc =
+    componentBlock.startTag.attributes.filter(
+      (attribute) =>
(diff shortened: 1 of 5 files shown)


lib/rules/no-empty-component-block.js
```diff
@@ -0,0 +1,79 @@
...
+  },
+
+  /**
+   * @param {RuleContext} context - The rule context.
+   * @returns {RuleListener} AST event handlers.
+   */
+  create(context) {
+    return {
+      Program(node) {
+        const componentBlocks = node.templateBody.parent.children
```
ota-meshi, line 56: Since `vue-eslint-parser` v7.0.0+ you can use `parserServices.getDocumentFragment()`.
This allows you to get the root element without the `<template>` block.

https://github.com/mysticatea/vue-eslint-parser/releases/tag/v7.0.0

I'm using this API in the `padding-line-between-blocks` rule.

https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/padding-line-between-blocks.js#L146

tyankatsu0105: Cool feature👍
Thanks for sharing.

tests/lib/rules/no-empty-component-block.js
```diff
@@ -0,0 +1,130 @@
...
+        {
+          message: '`<template>` is empty. Empty block is not allowed.'
+        },
+        {
+          message: '`<script>` is empty. Empty block is not allowed.'
+        }
+      ]
+    },
+    {
+      code: '<template></template><script></script><style></style>',
```
ota-meshi, line 88: Could you add testcases that contains whitespaces such as line breaks? It probably won't report because it contains VText.

tyankatsu0105: Wow. Good point!
I'll fix it.

lib/rules/no-empty-component-block.js
tyankatsu0105: done

lib/rules/no-empty-component-block.js
```diff
@@ -26,6 +27,17 @@ function hasAttributeSrc(componentBlock) {
...
 }
 
+/**
+ * check whether value under the component block is only whitespaces or break lines
+ */
+function isValueOnlyWhiteSpacesOrLineBreaks(componentBlock) {
+  return (
+    componentBlock.children.length === 1 &&
+    componentBlock.children[0].type === 'VText' &&
+    /^(\s|\n)+$/.test(componentBlock.children[0].value)
```
tyankatsu0105, line 37: I'm not sure if this regex is cover all cases...

ota-meshi: Maybe using `/^\s*$/` will also work.
However, I'm not good at using regular expressions, so I often do the following.

```js
!componentBlock.children[0].value.trim()
```

lib/rules/no-empty-component-block.js
```diff
@@ -0,0 +1,99 @@
...
+
+    return {
+      Program(node) {
+        for (const componentBlock of componentBlocks) {
+          if (
+            componentBlock.name !== 'template' &&
+            componentBlock.name !== 'script' &&
+            componentBlock.name !== 'style'
+          )
+            return
```
ota-meshi, line 77: I just noticed. Use `continue` here.

lib/rules/no-empty-component-block.js
```diff
@@ -0,0 +1,99 @@
...
+
+          // https://vue-loader.vuejs.org/spec.html#src-imports
+          if (hasAttributeSrc(componentBlock)) return
```
ota-meshi, line 80: Use `continue` here as well.

tyankatsu0105: done

lib/rules/no-empty-component-block.js
tyankatsu0105: done

lib/rules/no-empty-component-block.js
tyankatsu0105: It seems nice.
I reflected this code.

tests/lib/rules/no-empty-component-block.js
tyankatsu0105: I added some invalid cases.

PR #1226, opened 2020-06-30
Fixed no-side-effects-in-computed-properties when el is object

ktmouk:
Hi, thanks for great plugin!

I'm using object property to set `el` of Vue.
However, I get a `TypeError` when I update eslint-plugin-vue to `v7.0.0-alpha.7` and run eslint.

for example:

```javascript
const dummy = { app: '#app' }

new Vue({
  el: dummy.app,
  template: '<p>test</p>',
});
```

```
TypeError: Cannot read property 'body' of undefined
```

I think the cause of this issue is that the `scopeStack` has no default value.
but, I’m not sure if this PR is the best solution.

Could you please review this PR or use another best solution to fix this issue?
Thanks.

### DIFF
--- a/lib/rules/no-side-effects-in-computed-properties.js
+++ b/lib/rules/no-side-effects-in-computed-properties.js
@@ -31,15 +31,9 @@ module.exports = {
   create(context) {
     /** @type {Map<ObjectExpression, ComponentComputedProperty[]>} */
     const computedPropertiesMap = new Map()
-    /**
-     * @typedef {object} ScopeStack
-     * @property {ScopeStack} upper
-     * @property {BlockStatement | Expression} body
-     */
-    /**
-     * @type {ScopeStack}
-     */
-    let scopeStack
+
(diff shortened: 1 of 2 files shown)


lib/rules/no-side-effects-in-computed-properties.js
```diff
@@ -31,15 +31,9 @@ module.exports = {
...
-     * @typedef {object} ScopeStack
-     * @property {ScopeStack} upper
-     * @property {BlockStatement | Expression} body
-     */
-    /**
-     * @type {ScopeStack}
-     */
-    let scopeStack
+
+    /** @type { { upper: any, body: null | BlockStatement | Expression } } */
```
ktmouk, line 35: I referred to
https://github.com/vuejs/eslint-plugin-vue/blob/eb20820df997cbd3afef4ea2a3589f7a04ffb620/lib/rules/require-valid-default-prop.js#L109

ota-meshi:
Thank you!

PR #1259, opened 2020-07-21
Add `allowProps` option to `vue/require-explicit-emits` rule.

ota-meshi:
Even if you declare it in props, a warning message and fallthrough can be stopped, so I add an option to allow this.
https://github.com/vuejs/vue-next/blob/00ab9e2e8506d108958895cda4e977dfb16b53f9/packages/runtime-core/src/componentEmits.ts#L50

By default this option remains disabled.

### DIFF
--- a/docs/rules/require-explicit-emits.md
+++ b/docs/rules/require-explicit-emits.md
@@ -74,7 +74,37 @@ export default {
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+  "vue/require-explicit-emits": ["error", {
+    "allowProps": false
+  }]
+}
+```
+
+- `"allowProps"` ... If `true`, allow event names defined in `props`. default `false`
+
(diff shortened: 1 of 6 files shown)


Robo-Rin:
@ota-meshi This doesn't seem to account for custom events in kebab-case declared as a prop, do you think we should add this to support the workaround mentioned here:
https://github.com/vuejs/core/issues/5220#issuecomment-1007488240

For example currently this does work but is ugly:
```js
props: ['onCustom-event'],
// ...
this.$emit('custom-event')
```

So it would be great if we can do:
```js
props: ['onCustomEvent'],
// ...
this.$emit('custom-event')
```

Robo-Rin:
Nvm, I went ahead and made a PR for this: https://github.com/vuejs/eslint-plugin-vue/pull/2248 😄

PR #1267, opened 2020-07-29
Add `v-for-delimiter-style` rule

### LINKED ISSUE #1228: Rule proposal: `v-for-delimiter-style`
You can use both `v-for="x in xs"` and `v-for="x of xs"`. There should be a rule to enforce one of the styles (probably configurable).

- [X] Enforces code style
- [ ] Other (please specify:)

```vue
<!-- "vue/v-for-delimiter-style": ["warn", "in"] -->
<div v-for="x in xs" /><!-- GOOD -->
<div v-for="x of xs" /><!-- BAD -->

<!-- "vue/v-for-delimiter-style": ["warn", "of"] -->
<div v-for="x of xs" /><!-- GOOD -->
<div v-for="x in xs" /><!-- BAD -->
```

**Additional context**
See https://vuejs.org/v2/guide/list.html#Mapping-an-Array-to-Elements-with-v-for:

> You can also use `of` as the delimiter instead of `in`, so that it is closer to JavaScript’s syntax for iterators.

FloEdelmann:
Closes #1228.

### DIFF
--- a/lib/rules/v-for-delimiter-style.js
+++ b/lib/rules/v-for-delimiter-style.js
@@ -18,7 +18,7 @@ const utils = require('../utils')
 
 module.exports = {
   meta: {
-    type: 'suggestion',
+    type: 'layout',
     docs: {
       description: "enforce `v-for` directive's delimiter style",
       categories: undefined,
--- a/lib/rules/v-for-delimiter-style.js
+++ b/lib/rules/v-for-delimiter-style.js
@@ -1,0 +1,69 @@
+/**
+ * @fileoverview enforce `v-for` directive's delimiter style
+ * @author Flo Edelmann
+ * @copyright 2020 Flo Edelmann. All rights reserved.
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


lib/rules/v-for-delimiter-style.js
```diff
@@ -0,0 +1,69 @@
...
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
```
ota-meshi, line 21: Could you change it to "layout".

tests/lib/rules/v-for-delimiter-style.js
```diff
@@ -0,0 +1,98 @@
...
+      code: '<template><div v-for="x of xs"></div></template>',
+      options: ['of']
+    }
+  ],
+  invalid: [
+    {
+      filename: 'test.vue',
+      code: '<template><div v-for="x of xs"></div></template>',
+      output: '<template><div v-for="x in xs"></div></template>',
+      errors: ["Expected 'in' instead of 'of' in 'v-for'."]
```
ota-meshi, line 63: Could you use an object expression to check the location of the error?

https://eslint.org/docs/developer-guide/nodejs-api#ruletester

```suggestion
      errors: [{
        message: "Expected 'in' instead of 'of' in 'v-for'.",
        column: ?,
      }]
```

PR #1107, opened 2020-04-22
Add Nuxt and Vue Router properties in `order-in-components`

FloEdelmann:
Closes #980.

PR #1268, opened 2020-07-31
Reorder Nuxt.js's `fetch` method in `vue/order-in-components`

FloEdelmann:
Nuxt.js [version 2.12.0](https://github.com/nuxt/nuxt.js/releases/tag/v2.12.0) introduced a [new fetch syntax](https://github.com/nuxt/nuxt.js/pull/6880) (see the related [docs page](https://nuxtjs.org/guides/components-glossary/pages-fetch)). Since `fetch` now has access to the Vue component's `this` context, I propose to move it below `data` in `vue/order-in-components`'s default order.

This updated order also doesn't conflict with https://github.com/vuejs/eslint-plugin-vue/pull/1107#discussion_r436599013:

> I believe that Nuxt properties should be at least before methods, I don't want to scroll til the end to see what special properties my page has

As this change would probably be a breaking change, I suggest to make this decision before releasing stable v7.0.0.


ota-meshi:
I'm not familiar with nuxt APIs so please let me ask.

Why do we need to change the order suggested in the comments below?
https://github.com/vuejs/eslint-plugin-vue/pull/1107#discussion_r436693385

ota-meshi:
Does that mean that by the time I merged the previous PR, users couldn't use `this` in `fetch`, it changed?

FloEdelmann:
Using `this` in `fetch` is possible since v2.12.0, i.e. since 2020-03-17.

The previous PR (where the order of the other Nuxt properties was changed) was merged on 2020-06-09.

However, `fetch`'s position didn't change since #823 (merged 2019-08-17).

ota-meshi:
I misunderstood. As you say, the order of `fetch` was not changed in #1107.
This change is sounds good to me.

PR #1276, opened 2020-08-12
New: Add `vue/no-mustache` rule

james2doyle:
This rule disallows the usage of mustache interpolation.

There are a couple reasons to do this:

- you prefer directives (`v-{text,html,t}`)
- you want to make sure each piece of text is attached to a DOM node
- you already use a template language in your project that uses braces

### DIFF
--- a/docs/rules/no-mustache.md
+++ b/docs/rules/no-mustache.md
@@ -1,0 +1,42 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-mustache
+description: disallow usage of mustache interpolations
+---
+# vue/no-mustache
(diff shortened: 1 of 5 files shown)


james2doyle:
Any update on this? Is there someone to review?

ota-meshi:
Sorry for late reply.

I've never developed with a configuration that requires this rule, so I don't understand the need for this rule.
Can you keep this PR open to solicit the opinions of many?

g-plane:
I don't prefer adding this rule, since this looks a very rare use case.

ota-meshi:
I couldn't understand why we needed `no-mustache` rule. So I close this PR.

PR #1288, opened 2020-08-27
Add `ignoreComponents` to `attribute-hyphenation` rule

meowtec:


ota-meshi:
I'm not sure why `ignoreComponents` option is needed.
Can you please explain when `ignoreComponents` option is useful?

meowtec:
@ota-meshi 
I prefer `myProps` then `my-props` when I am using custom components, because I can simply copy the prop name from the component file and paste to where I use the component. 
So I would use `'vue/attribute-hyphenation': ['warn', 'never']`.

While the most popular component libraries such as `Vueify` and `ElementUI` provide docs and examples using `attribute-hyphenation`.

So I want to use `attribute-hyphenation` in those component libraries (which has prefix like `el-` and `v-`), or I have to convert it to camelCase...

```html
<!-- copy `myProps` from MyComponent source code -->
<MyComponent
  myProps="xxx"
/>
<!-- copy `first-day-of-week` from the element-ui docs -->
<el-date-picker
  :first-day-of-week="1"
>
```

ota-meshi:
Why don't you use camelCase in Vueify and ElementUI component?
I don't see your source code unifying the hyphenation styles of attributes, so I think you need to turn this rule off.

meowtec:
> 为什么不在Vueify和ElementUI组件中使用camelCase？
> 我看不到您的源代码统一属性的连字符样式，因此我认为您需要关闭此规则。

Because those document 

> Why don't you use camelCase in Vueify and ElementUI components?
> I don't see your source code unifying the hyphenation styles of attributes, so I think you need to turn this rule off.

Because their documents are all `attribute-hyphenation`.
For example: a long prop: `modal-append-to-body` in https://element.eleme.io/#/zh-CN/component/dialog#attributes

If I need to use it, I have to modify `modal-append-to-body` to `modalAppendToBody`, it wastes much time.

meowtec:
Maybe it is better to submit MR to docs of ElementUI / Vueify, adding a checkbox for toggling between attribute-hyphenation and camelCase.

ota-meshi:
> If I need to use it, I have to modify modal-append-to-body to modalAppendToBody, it wastes much time.

You can modify all your source code with a single command using the `--fix` option.
Also, the `vue/attribute-hyphenation` rule supports `--fix`.

https://eslint.org/docs/user-guide/command-line-interface#fixing-problems
https://eslint.vuejs.org/rules/attribute-hyphenation.html#vue-attribute-hyphenation

I don't think anyone will use `ignoreComponents` option, so I'll close this PR.

PR #1305, opened 2020-09-21
Make rule vue/no-unregistered-components ignore recursive components

### LINKED ISSUE #1304: Optionally ignore vue/no-unregistered-components warnings against recursive components
vue/no-unregistered-components

fewer warnings

My idea was to add the argument `ignoreRecursive` to the rule. If true, it would check the component's name against the name of the component it's inserted into (for this, the parent component _needs_ to have its `name` property set), and, if they're the same, do not warn.

```vue
<template>
    <span>{{ note.text }}</span>
    <Note
        v-for="reply in note.replies"
        :note="reply"
    />
</template>

<script>
    export default {
        name: 'Note',
        props: {
            note: {
                type: Object,
                required: true,
            },
        },
    };
</script>
```

It marks the recursive `Note` component inserted in the template with a warning.

The warning would not appear in recursive components, as the reference is in the `name` property, so it is self-registered.

arrudaje:
Right now when we use recursive components (when we want to use the current component inside of itself - with the example of a `Note` component that can have a number of replies, that look and behave exactly like the `Note` component, and they stack inside of one another), eslint-plugin-vue complains against it, making us disable this rule every time we want to use a recursive component. 

The suggestion is to include a new argument to the rule, `ignoreRecursive`, which, if `true`, checks whether the name of the parent component is the same as the child component (making it thus a recursive component), not warning about this rule on it.

Resolves #1304

### DIFF
--- a/lib/rules/no-unregistered-components.js
+++ b/lib/rules/no-unregistered-components.js
@@ -62,6 +62,9 @@ module.exports = {
         properties: {
           ignorePatterns: {
             type: 'array'
+          },
+          ignoreRecursive: {
+            type: 'boolean'
           }
         },
         additionalProperties: false
@@ -73,14 +76,33 @@ module.exports = {
     const options = context.options[0] || {}
     /** @type {string[]} */
     const ignorePatterns = options.ignorePatterns || []
+    /** @type {boolean} */
+    const ignoreRecursive = options.ignoreRecursive || false
     /** @type { { node: VElement | VDirective | VAttribute, name: string }[] } */
     const usedComponentNodes = []
     /** @type { { node: Property, name: string }[] } */
     const registeredComponents = []
+    /** @type {string} */
+    let componentName = ''
 
-    return utils.defineTemplateBodyVisitor(
-      context,
-      {
+    return utils.compositingVisitors(
+      utils.defineVueVisitor(context, {
+        /** @param {ObjectExpression} obj */
+        onVueObjectEnter(obj) {
+          const nameProperty = obj.properties.find(
+            /**
+             * @param {ESNode} p
+             * @returns {p is (Property & { key: Identifier & {name: 'name'}, value: ObjectExpression })}
+             */
+            (p) => p.key.name === 'name'
+          )
+
+          if (nameProperty) {
+            componentName = nameProperty.value.value
+          }
+        }
+      }),
+      utils.defineTemplateBodyVisitor(context, {
         /** @param {VElement} node */
         VElement(node) {
           if (
@@ -156,6 +178,11 @@ module.exports = {
               )
                 return false
 
+              // Check recursive components if they are ignored
+              if (ignoreRecursive) {
+                return kebabCaseName !== casing.kebabCase(componentName)
+              }
+
               // Component registered as `foo-bar` cannot be used as `FooBar`
               if (
                 casing.isPascalCase(name) &&
@@ -178,7 +205,7 @@ module.exports = {
               })
             )
         }
-      },
+      }),
       utils.executeOnVue(context, (obj) => {
         registeredComponents.push(...utils.getRegisteredComponents(obj))
       })
--- a/lib/rules/no-unregistered-components.js
+++ b/lib/rules/no-unregistered-components.js
@@ -62,9 +62,6 @@ module.exports = {
         properties: {
           ignorePatterns: {
             type: 'array'
-          },
-          ignoreRecursive: {
-            type: 'boolean'
           }
         },
         additionalProperties: false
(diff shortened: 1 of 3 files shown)


lib/rules/no-unregistered-components.js
```diff
@@ -73,14 +76,33 @@ module.exports = {
...
+    let componentName = ''
 
-    return utils.defineTemplateBodyVisitor(
-      context,
-      {
+    return utils.compositingVisitors(
+      utils.defineVueVisitor(context, {
+        /** @param {ObjectExpression} obj */
+        onVueObjectEnter(obj) {
+          const nameProperty = obj.properties.find(
```
ota-meshi, line 92: Could you move the process here to around line 210?

lib/rules/no-unregistered-components.js
```diff
@@ -73,14 +76,33 @@ module.exports = {
...
+            /**
+             * @param {ESNode} p
+             * @returns {p is (Property & { key: Identifier & {name: 'name'}, value: ObjectExpression })}
+             */
+            (p) => p.key.name === 'name'
```
ota-meshi, line 97: You can use `utils.findProperty`.

lib/rules/no-unregistered-components.js
```diff
@@ -62,6 +62,9 @@ module.exports = {
         properties: {
           ignorePatterns: {
             type: 'array'
+          },
+          ignoreRecursive: {
```
ota-meshi, line 66: I didn't know that I could use self component in `<template>` using the same name as `name` option.
I don't need the this option because I think it is false positives and a bug.

ota-meshi (changes requested):
Thank you for this PR and report! Sorry for the late reply.

I didn't know that I could use self component in `<template>` using the same name as `name` option.
I think it's a bug, so I don't think it's necessary to add option.

lib/rules/no-unregistered-components.js
```diff
@@ -73,14 +76,33 @@ module.exports = {
...
+          const nameProperty = obj.properties.find(
+            /**
+             * @param {ESNode} p
+             * @returns {p is (Property & { key: Identifier & {name: 'name'}, value: ObjectExpression })}
+             */
+            (p) => p.key.name === 'name'
+          )
+
+          if (nameProperty) {
+            componentName = nameProperty.value.value
```
ota-meshi, line 101: I think it's probably better to add it to `registeredComponents`.

arrudaje:
> Thank you for this PR and report! Sorry for the late reply.
> 
> I didn't know that I could use self component in `<template>` using the same name as `name` option.
> I think it's a bug, so I don't think it's necessary to add option.

Thank you for your reply!
Actually it's a feature in Vue. I don't know why they decided to remove the [guide (v2)](https://vuejs.org/v2/guide/components-edge-cases.html#Recursive-Components) for recursive components in v3, but I found a mention in this page [https://v3.vuejs.org/api/options-misc.html#name](https://v3.vuejs.org/api/options-misc.html#name).
Please tell me what you think, and this probably can clear out some of the comments you made here also, although I will try to work on them.

lib/rules/no-unregistered-components.js
```diff
@@ -178,7 +184,20 @@ module.exports = {
...
+            registeredComponents.push({
+              node: nameProperty,
+              name: nameProperty.value.value
+            })
+          }
+        }
+      }),
       utils.executeOnVue(context, (obj) => {
         registeredComponents.push(...utils.getRegisteredComponents(obj))
       })
```
ota-meshi, line 203: Could you please change it as follow?
Also, remove `utils.compositingVisitors`.

```js
      utils.executeOnVue(context, (obj) => {
        registeredComponents.push(...utils.getRegisteredComponents(obj))
        
          const nameProperty = utils.findProperty(obj, 'name')

          if (nameProperty) {
            registeredComponents.push({
              node: nameProperty,
              name: nameProperty.value.value
            })
          }
      })
```

lib/rules/no-unregistered-components.js
```diff
@@ -73,14 +76,17 @@ module.exports = {
...
     /** @type {string[]} */
     const ignorePatterns = options.ignorePatterns || []
+    /** @type {boolean} */
+    const ignoreRecursive = options.ignoreRecursive || false
     /** @type { { node: VElement | VDirective | VAttribute, name: string }[] } */
     const usedComponentNodes = []
     /** @type { { node: Property, name: string }[] } */
     const registeredComponents = []
+    /** @type {string} */
+    let componentName = ''
```
ota-meshi, line 86: Remove unused variable.

ota-meshi (changes requested):

lib/rules/no-unregistered-components.js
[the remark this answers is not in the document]
arrudaje: Also sorry for not reading you well, I have applied your comments and it makes much more sense now, thanks :)

ota-meshi (approved):
LGTM! Thank you!

PR #1317, opened 2020-10-04
Avoid redundant global self-require of the plugin

constgen:
In some really specific cases when the project patches `require()` to alias some modules (e.g. to fix an issue with ESLint sharable configs) this plugin may crash. This happens because it requires itself globally using its name which is redundant. 

Other plugins that are testes on the same project have no such issue.

This PR converts such paths to relative paths.


ota-meshi (approved):
Thank you!

constgen:
Thank you too. When can I expect it published to NPM? Really waiting for it.

Issue #1260, opened 2020-07-21
feat: custom-event-name-casing custom casing


privatenumber:
Extension of this bug report: https://github.com/vuejs/eslint-plugin-vue/issues/1250

## Request
For the ability to customize what kind of casing should be enforced on event names.

#### Name spacing
In my codebase we namespace events using `:` with the component name for frictionless composability. That way, when composing components, there would be no collision in emitted events. (eg. `input:update`)

For example, a custom Input component would emit `input:update`, which triggers the new v-model value (not to takeover native event names with different values). There could be another component, for example  Search, that uses this Input component and emit another "update" event with more data via `search:update`.

#### camelCase in Vue 3 docs
With Vue v3, `v-model` replaces `.sync` and their docs seem to contain [a lot of examples using camelCase in event names](https://v3.vuejs.org/guide/component-custom-events.html#multiple-v-model-bindings).

(This is pretty confusing considering they still recommend event names be case insensitive (lower case) at the top.)

If this isn't a mistake, then surely, some people would want to use camelCase in their event names. (https://github.com/vuejs/docs-next/issues/308)

Edit:
Changing to `v-model.prop-name` seems to be necessary in a non-compiled environment (haven't tested in a compiled one). Both `update:propName` and `update:prop-name` seem to work. [Codepen](https://codepen.io/hirokiosame/pen/KKVbraX)

### Proposed API

Just some ideas I had:
- `ignoreCharacters: [':']`
- `allowCharacters: [':']`
- `allowNamespace: true`
- `allowDelimiters: ':'`

ota-meshi:
Thank you for your proposal.

However, it seems to me that the options you have proposed are a bit limited.

The following is my idea.

Add the `ignorePatterns` option. You can specify an array of regular expressions with the `ignorePatterns` option.
Ignores event names whose names match the specified pattern from this rule's report.

e.g. `"ignorePatterns": ["^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$"]`

What do you think? Does the `ignores` option solve your needs?

privatenumber:
Yeah that works! Thank you @ota-meshi 🙏

jacekkarczmarczyk:
Vuetify also uses `:` syntax for events, for example `click:row`, so that would be a great addition

bitencode:
👍 Vue itself uses examples of `update:my-prop` to distinguish which prop is being updated.

[...]

PR #1321, opened 2020-10-08
added options for events that ignore custom-event-name-casing

devTeaa:
Hi, currently there are some components from library that still uses pascalCase when passing the function
`<customComponent @maskClick="handleMaskClick"/>` this changes added options so it can ignore some event naming just like in `component-name-in-template-casing`.

What do you think?

### DIFF
--- a/docs/rules/custom-event-name-casing.md
+++ b/docs/rules/custom-event-name-casing.md
@@ -59,6 +59,37 @@ export default {
 ```
 - `ignores` (`string[]`) ... The event names to ignore. Sets the event name to allow. For example, custom event names, Vue components event with special name, or Vue library component event name. You can set the regexp by writing it like `"/^name/"` or `click:row` or `fooBar`.
 
+### `"ignores": ["fooBar", "/^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$/u"]`
+
+<eslint-code-block :rules="{'vue/custom-event-name-casing': ['error', {ignores:'/^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$/u'}]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <button @click="$emit('click:row')" />
+  <button @click="$emit('fooBar')" />
+
+  <!-- ✗ BAD -->
+  <button @click="$emit('myEvent')" />
+</template>
+<script>
+export default {
+  methods: {
+    onClick () {
+      /* ✓ GOOD */
+      this.$emit('click:row')
+      this.$emit('fooBar')
+
+      /* ✗ BAD */
+      this.$emit('myEvent')
+    }
+  }
+}
+</script>
+```
+
+</eslint-code-block>
+
 
 ## :books: Further Reading
 
--- a/docs/rules/custom-event-name-casing.md
+++ b/docs/rules/custom-event-name-casing.md
@@ -57,7 +57,7 @@ export default {
   }]
 }
 ```
-- `ignores` (`string[]`) ... The event names to ignore. Sets the event name to allow. For example, custom event names, Vue components event with special name, or Vue library component event name. You can set the regexp by writing it like `"/^name/"` or `update:input` or `fooBar`.
+- `ignores` (`string[]`) ... The event names to ignore. Sets the event name to allow. For example, custom event names, Vue components event with special name, or Vue library component event name. You can set the regexp by writing it like `"/^name/"` or `click:row` or `fooBar`.
 
 
 ## :books: Further Reading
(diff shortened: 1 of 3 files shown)


ota-meshi (changes requested):
Thank you for this PR! 

Perhaps this PR solves https://github.com/vuejs/eslint-plugin-vue/issues/1260#issuecomment-662104360.
Could you add test cases to check ignores `input:update`, `search:update` and `click:row`?

Could you also add description of the new option to the documentation?
https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/custom-event-name-casing.md

devTeaa:
Updated the docs

image

tests/lib/rules/custom-event-name-casing.js
```diff
@@ -166,6 +166,66 @@ tester.run('custom-event-name-casing', rule, {
...
+          onClick() {
+            this.$emit('input:update')
+            this.$emit('search:update')
+            this.$emit('click:row')
+          }
+        }
+      }
+      </script>
+      `,
+      options: [{ ignores: ['input:update', 'search:update', 'click:row'] }]
```
ota-meshi, line 228: Could you replace it with a testcase that uses regular expressions?

```suggestion
      options: [{ ignores: ['/^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$/u'] }]
```

docs/rules/custom-event-name-casing.md
```diff
@@ -49,7 +49,16 @@ export default {
...
-Nothing.
+
+```json
+{
+  "vue/custom-event-name-casing": ["error", {
+    "ignores": []
+  }]
+}
+```
+- `ignores` (`string[]`) ... The event names to ignore. Sets the event name to allow. For example, custom event names, Vue components event with special name, or Vue library component event name. You can set the regexp by writing it like `"/^name/"` or `update:input` or `fooBar`.
```
ota-meshi, line 60: `update:*` is already ignored without using the `ignores` option, so I think it's easier for people to understand by using the `click:row` in Vuetify.

```suggestion
- `ignores` (`string[]`) ... The event names to ignore. Sets the event name to allow. For example, custom event names, Vue components event with special name, or Vue library component event name. You can set the regexp by writing it like `"/^name/"` or `click:row` or `fooBar`.
```

docs/rules/custom-event-name-casing.md
```diff
@@ -49,7 +49,16 @@ export default {
...
+
```
ota-meshi, line 61: Could you add an example to the documentation that actually uses the option?



````md
### `"ignores": ["fooBar", "/^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$/u"]`

<eslint-code-block :rules="{'vue/custom-event-name-casing': ['error', {ignores:'/^[a-z]+(?:-[a-z]+)*:[a-z]+(?:-[a-z]+)*$/u'}]}">

```vue
<template>
  <!-- ✓ GOOD -->
  <button @click="$emit('click:row')" />
  <button @click="$emit('fooBar')" />

  <!-- ✗ BAD -->
  <button @click="$emit('myEvent')" />
</template>
<script>
export default {
  methods: {
    onClick () {
      /* ✓ GOOD */
      this.$emit('click:row')
      this.$emit('fooBar')

      /* ✗ BAD */
      this.$emit('myEvent')
    }
  }
}
</script>
```

</eslint-code-block>
````


Other example: https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/no-useless-v-bind.md#ignoreincludescomment-true

File: tests/lib/rules/custom-event-name-casing.js (hunk 1 above)
Comment (ota-meshi) [inline tests/lib/rules/custom-event-name-casing.js:228]: Could you also add testcases that reports a violation that doesn't match the regular expressions.

**Review by ota-meshi — CHANGES_REQUESTED** [review]

File: docs/rules/custom-event-name-casing.md (hunk 2 above)
Comment (devTeaa) [reply docs/rules/custom-event-name-casing.md:60]: This is accidentally the same I swear 😂

**Comment by devTeaa** [comment]
Added an example for `ignores` options

image

File: package.json [hunk 4]
```diff
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "start": "npm run test:base -- --watch --growl",
-    "test:base": "mocha \"tests/lib/**/*.js\" --reporter dot",
+    "test:base": "mocha \"tests/lib/rules/custom-event-name-casing.js\" --reporter dot",
```
ota-meshi, line 8: Would you please revert this change?

devTeaa: sorry, I will

devTeaa (commented):
Done

ota-meshi (approved):
LGTM! Thanks for your contribution!
I will include this change in the next release.

devTeaa:
Thanks for helping the review!

PR #1335, opened 2020-10-17
fix(eslint-plugin-vue): [valid-v-bind-sync] Added Case: when valid is then valid .sync

### LINKED ISSUE #1298: "vue/valid-v-bind-sync" is not accounting for "is" attribute
`vue/valid-v-bind-sync` 

Fewer

```
<table>
    <tr is="my-row" 
        :some-prop.sync="somePropValue"
        :some-other-prop.sync="someOtherPropValue">
        <td></td>
    </tr>
</table>
```

Displays the error/warning, `'.sync' modifiers aren't supported on <tr> non Vue-components`, for every prop containing the `.sync` modifier.

Evaluate any `.sync` modifier implementation as valid if the target element has a valid `is` attribute.

**Additional context**
The `is` attribute is used for [dynamic components](https://vuejs.org/v2/guide/components.html#Dynamic-Components) and to work around [limitations of in-DOM templates](https://vuejs.org/v2/guide/components.html#DOM-Template-Parsing-Caveats) (see Vue [documentation](https://vuejs.org/v2/api/#is)):
```
<!-- necessary because `<my-row>` would be invalid inside -->
<!-- a `<table>` element and so would be hoisted out      -->
<table>
  <tr is="my-row"></tr>
</table>
```

aggmoulik:
Fixes #1298 
Signed-off-by: Moulik Aggarwal <qwertymoulik@gmail.com>

### DIFF
--- a/lib/rules/valid-v-bind-sync.js
+++ b/lib/rules/valid-v-bind-sync.js
@@ -82,6 +82,17 @@ function maybeNullObjectMemberExpression(node) {
   return false
 }
 
+function isValidIs(node){
+  let { attributes }  = node
+  let isAttribute = attributes.filter((attr) => {
+    if(attr.type === "VAttribute" && attr.key.name === "is") return attr
+  })
+  if(isAttribute.length > 0 && ['tr', 'li', 'option'].includes(node.parent.name)){
+    return true
+  }
+  return false
+}
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
@@ -120,7 +131,7 @@ module.exports = {
         const element = node.parent.parent
         const name = element.name
 
-        if (!isValidElement(element)) {
+        if (!isValidElement(element) && !isValidIs(node.parent)) {
           context.report({
             node,
             loc: node.loc,
--- a/lib/rules/valid-v-bind-sync.js
+++ b/lib/rules/valid-v-bind-sync.js
@@ -82,12 +82,15 @@ function maybeNullObjectMemberExpression(node) {
   return false
 }
 
-function isValidIs(node){
-  let { attributes }  = node
-  let isAttribute = attributes.filter((attr) => {
-    if(attr.type === "VAttribute" && attr.key.name === "is") return attr
+function isValidIs(node) {
+  const { attributes } = node
+  const isAttribute = attributes.filter((attr) => {
+    if (attr.type === 'VAttribute' && attr.key.name === 'is') return attr
   })
-  if(isAttribute.length > 0 && ['tr', 'li', 'option'].includes(node.parent.name)){
+  if (
+    isAttribute.length > 0 &&
+    ['tr', 'li', 'option'].includes(node.parent.name)
+  ) {
     return true
   }
   return false
--- a/lib/rules/valid-v-bind-sync.js
+++ b/lib/rules/valid-v-bind-sync.js
@@ -84,16 +84,18 @@ function maybeNullObjectMemberExpression(node) {
 
 function isValidIs(node) {
   const { attributes } = node
-  const isAttribute = attributes.filter((attr) => {
-    if (attr.type === 'VAttribute' && attr.key.name === 'is') return attr
+  const isAttribute = attributes.some((attr) => {
+    // check for `VAttribute`
+    if (attr.type === 'VAttribute') {
+      // check for `is` attribute
+      if (attr.key.type === 'VIdentifier' && attr.key.name === 'is') return true
+
+      // check for `:is` `bind` attribute
+      if (attr.key.type === 'VDirectiveKey' && attr.key.argument.name === 'is')
+        return true
+    }
   })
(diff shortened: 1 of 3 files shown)


lib/rules/valid-v-bind-sync.js
```diff
@@ -82,6 +82,20 @@ function maybeNullObjectMemberExpression(node) {
...
 }
 
+function isValidIs(node) {
+  const { attributes } = node
+  const isAttribute = attributes.filter((attr) => {
+    if (attr.type === 'VAttribute' && attr.key.name === 'is') return attr
+  })
+  if (
+    isAttribute.length > 0 &&
+    ['tr', 'li', 'option'].includes(node.parent.name)
```
ota-meshi, line 92: Could you remove the element name check?
The `is` attribute can be used on all elements. There is no need to check the element name.

lib/rules/valid-v-bind-sync.js
```diff
@@ -82,6 +82,20 @@ function maybeNullObjectMemberExpression(node) {
   return false
 }
 
+function isValidIs(node) {
+  const { attributes } = node
+  const isAttribute = attributes.filter((attr) => {
+    if (attr.type === 'VAttribute' && attr.key.name === 'is') return attr
```
ota-meshi, line 88: We should also check `v-bind:is`.

lib/rules/valid-v-bind-sync.js
```diff
@@ -82,6 +82,20 @@ function maybeNullObjectMemberExpression(node) {
   return false
 }
 
+function isValidIs(node) {
+  const { attributes } = node
+  const isAttribute = attributes.filter((attr) => {
```
ota-meshi, line 87: Could you use `some` instead of `filter`.

ota-meshi (changes requested):
Thank you for this PR!
I have some change requests.

aggmoulik:
Hy @ota-meshi I have updated your changes and added a cli utility also, as it helps us to check individual rules and makes our development faster.

aggmoulik:
Also, Add `hacktoberfest` label in the PR or in the repository tags.

ota-meshi:
I'm currently maintaining this repository almost alone, so I can't afford to accept all PR during the `hacktoberfest`, so I won't add `hacktoberfest` label.
Also, I'm sorry, this PR can't promise to complete all reviews and merge in October.

aggmoulik:
@ota-meshi I think I can help you in that as I am working to be a active-contributor not for just hacktoberfest. I can help in code review as I already do in my company.

lib/rules/valid-v-bind-sync.js
```diff
@@ -82,6 +82,22 @@ function maybeNullObjectMemberExpression(node) {
...
+function isValidIs(node) {
+  const { attributes } = node
+  const isAttribute = attributes.some((attr) => {
+    // check for `VAttribute`
+    if (attr.type === 'VAttribute') {
+      // check for `is` attribute
+      if (attr.key.type === 'VIdentifier' && attr.key.name === 'is') return true
+
+      // check for `:is` `bind` attribute
+      if (attr.key.type === 'VDirectiveKey' && attr.key.argument.name === 'is')
```
ota-meshi, line 94: You need to check if it is v-bind. Also, the argument can be null.

https://github.com/vuejs/vue-eslint-parser/blob/master/docs/ast.md#vdirectivekey

In addition, could you add a test to make sure it is not judged as `is`?

```vue
<tr v-on:is="myRow" :some-prop.sync="somePropValue">
<tr v-bind="myRow" :some-prop.sync="somePropValue">
```

package.json
```diff
@@ -19,7 +19,8 @@
     "version": "npm run lint -- --fix && git add .",
     "update": "node ./tools/update.js",
     "docs:watch": "vuepress dev docs",
-    "docs:build": "vuepress build docs"
+    "docs:build": "vuepress build docs",
+    "tests": "mocha"
```
ota-meshi, line 23: Do not change the script. You may be able to use `npx mocha`.

tests/lib/rules/valid-v-bind-sync.js
```diff
@@ -145,6 +145,48 @@ tester.run('valid-v-bind-sync', rule, {
...
+        </table>
+      </template>
+      `
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+        <table>
+          <tr v-bind:is="my-row"
```
ota-meshi, line 182:
```suggestion
          <tr v-bind:is="myRow"
```

ota-meshi (changes requested):
Sorry for the late reply. I posted some comments.

package.json
ota-meshi, line 23:
```suggestion
    "docs:build": "vuepress build docs"
```

ota-meshi:
I would like to share the fix related to this issue with another fix. So I'll fix this PR and merge it.

Issue #1362, opened 2020-12-03
vue/custom-event-name-casing forcing camelCase on vue2 code


cain:
**What rule do you want to change?**
Removing rule from vue2 `vue/custom-event-name-casing`.

Official vue documentation recommends using kebab case for naming custom event names. https://vuejs.org/v2/guide/components-custom-events.html#Event-Names
"For these reasons, we recommend you always use kebab-case for event names."

**Does this change cause the rule to produce more or fewer warnings?**
Fewer

**How will the change be implemented? (New option, new default behavior, etc.)?**
Removing rule from `plugin:vue/recommended` and `plugin:vue/essential`

**Please provide some example code that this change will affect:**

```vue
$emit('somethingCool') // throws linting error
```

**What does the rule currently do for this code?**

**What will the rule do after it's changed?**

**Additional context**
version: 7.1

cain:
See sandbox link: https://codesandbox.io/s/silly-framework-zb0uj?file=/src/App.vue:162-184

ota-meshi:
If you are using vue2, we recommend always using kebab-case.
https://vuejs.org/v2/guide/components-custom-events.html#Event-Names

However, for another reason, we plan to remove `vue/custom-event-name-casing` rule from the configuration.

- #1347
- https://github.com/vuejs/docs-next/issues/656

cain:
> If you are using vue2, we recommend always using kebab-case.
> https://vuejs.org/v2/guide/components-custom-events.html#Event-Names

That's the whole point of my issue @ota-meshi 

Looking forward to the next release.

PR #1363, opened 2020-12-03
change: remove vue/custom-event-name-casing from essential config

cain:
#1362


ota-meshi:
This file is automatically generated so you can edit it directly and it will be restored. Check the comments at the top of the file.
Also, I will close this PR as I plan to remove this rule from the configuration in the next release.

PR #1391, opened 2020-12-26
Fixed false positives for `v-bind="object"` syntax in `vue/attributes-order` rule

ota-meshi:
This PR fixes false positives for `v-bind="object"` syntax in `vue/attributes-order` rule.

For example, there were the following false positives:

```vue
<template>
  <div
    v-bind="object"
    id="x"> <!-- Attribute "id" should go before "v-bind". -->
  </div>
</template>
```

However, in Vue3, changing this order will change the behavior.

`v-bind="object"` changes the rule to report an error only if it's okay to change the order.

### DIFF
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -91,7 +91,33 @@ This rule aims to enforce ordering of component attributes. The default order is
 
 </eslint-code-block>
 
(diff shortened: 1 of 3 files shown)


danyadev:
I didn't understand a little how it works...

in the config, I did not define the attributes-order property, so the default settings are used.

image

PR #1396, opened 2020-12-29
Make no-ref-as-operand fixable close #1394

### LINKED ISSUE #1394: Make vue/no-ref-as-operand fixable

- **ESLint version:** 7.16.0
- **eslint-plugin-vue version:** 7.4.0
- **Node version:** 14.15.3

Make vue/no-ref-as-operand fixable.

**Additional context**

Since this autofix will change not just the format but also the meaning of the code, maybe it should not be auto-fixed.

sapphi-red:
This PR makes `vue/no-ref-as-operand fixable` fixable. close #1394 
The auto fix adds `.value` after the `Ref` value where it is needed.

### DIFF
--- a/lib/rules/no-ref-as-operand.js
+++ b/lib/rules/no-ref-as-operand.js
@@ -15,7 +15,7 @@ module.exports = {
       categories: ['vue3-essential'],
       url: 'https://eslint.vuejs.org/rules/no-ref-as-operand.html'
     },
-    fixable: null,
+    fixable: 'code',
     schema: [],
     messages: {
       requireDotValue:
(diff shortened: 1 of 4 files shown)


lib/rules/no-ref-as-operand.js
```diff
@@ -46,6 +46,9 @@ module.exports = {
         messageId: 'requireDotValue',
         data: {
           method: data.method
+        },
+        fix(fixer) {
+          return [fixer.insertTextAfter(node, '.value')]
```
ota-meshi, line 51: If you want to return a single fix data, you don't need to use an array.

```suggestion
          return fixer.insertTextAfter(node, '.value')
```

ota-meshi (changes requested):
Thank you for this PR.
I have one request.

sapphi-red:
Thank you for reviewing! I fixed it.

ota-meshi (approved):
LGTM! Thank you!

Issue #1393, opened 2020-12-29
vue/no-side-effects-in-computed-properties with composition api's computed


sapphi-red:
**What rule do you want to change?**
vue/no-side-effects-in-computed-properties

**Does this change cause the rule to produce more or fewer warnings?**
more warnings

**How will the change be implemented? (New option, new default behavior, etc.)?**
new default behavior

**Please provide some example code that this change will affect:**

```vue
<script>
import { defineComponent, ref, computed } from 'vue'

export default defineComponent({
  setup() {
    const arr = ref(['aaa', 'b', 'c'])
    const sorted = computed(() => arr.value.sort((a,b) => a.length - b.length)) // here arr is being mutated
    return { sorted }
  }
})
</script>
```

**What does the rule currently do for this code?**
no warning

**What will the rule do after it's changed?**
show warning like `computed` properties

**Additional context**

Actually, this new feature does not affect computed **properties**, so it might have to be a new rule.

sapphi-red:
I think `vue/return-in-computed-property` and `vue/no-async-in-computed-properties` has the same thing with this.
Should I create another issue?

PR #1398, opened 2020-12-30
Add composition api's computed function support to vue/no-async-in-computed-properties refs #1393

sapphi-red:
This PR adds support for composition api's computed function to `vue/no-async-in-computed-properties`.
refs #1393

### DIFF
--- a/docs/rules/no-async-in-computed-properties.md
+++ b/docs/rules/no-async-in-computed-properties.md
@@ -2,21 +2,21 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-async-in-computed-properties
-description: disallow asynchronous actions in computed properties
+description: disallow asynchronous actions in computed properties and functions
 since: v3.8.0
 ---
 # vue/no-async-in-computed-properties
 
-> disallow asynchronous actions in computed properties
+> disallow asynchronous actions in computed properties and functions
 
 - :gear: This rule is included in all of `"plugin:vue/vue3-essential"`, `"plugin:vue/essential"`, `"plugin:vue/vue3-strongly-recommended"`, `"plugin:vue/strongly-recommended"`, `"plugin:vue/vue3-recommended"` and `"plugin:vue/recommended"`.
 
-Computed properties should be synchronous. Asynchronous actions inside them may not work as expected and can lead to an unexpected behaviour, that's why you should avoid them.
+Computed properties and functions should be synchronous. Asynchronous actions inside them may not work as expected and can lead to an unexpected behaviour, that's why you should avoid them.
 If you need async computed properties you might want to consider using additional plugin [vue-async-computed]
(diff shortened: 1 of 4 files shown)


docs/rules/no-async-in-computed-properties.md
```diff
@@ -2,21 +2,21 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-async-in-computed-properties
-description: disallow asynchronous actions in computed properties
+description: disallow asynchronous actions in computed properties and functions
```
sapphi-red, line 5: `computed values` maybe better?
I'm not sure how I should call it.

lib/utils/index.js
```diff
@@ -860,6 +860,44 @@ module.exports = {
       })
   },
 
+  /**
+   * Get getter body from computed function
+   * @param {CallExpression} callExpression call of computed function
+   * @return {FunctionExpression | ArrowFunctionExpression | null} getter function
+   */
+  getComputedGetterBody(callExpression) {
```
ota-meshi, line 868: Can you rename it to include the meaning that it is a function that only accepts `computed (() => {})`?
Or change the argument to be acceptable and use this function in `getComputedProperties` as well.

```js
/**
  * @param {Expression} expr
  */
getComputedGetterBody(expr) {
}

// ...
// use
const getter = utils.getComputedGetterBody(node.arguments[0])
```

tests/lib/rules/no-async-in-computed-properties.js
```diff
@@ -639,6 +711,191 @@ ruleTester.run('no-async-in-computed-properties', rule, {
...
+        },
+        {
+          message: 'Unexpected timed function in computed function.',
+          line: 12
+        },
+        {
+          message: 'Unexpected timed function in computed function.',
+          line: 13
+        }
+      ]
```
ota-meshi, line 898: Could you add test cases for the arrow function that omits return?

lib/rules/no-async-in-computed-properties.js
```diff
@@ -77,13 +77,16 @@ module.exports = {
   },
   /** @param {RuleContext} context */
   create(context) {
+    /** @type {Map<ObjectExpression, ComponentComputedProperty[]>} */
+    const computedPropertiesMap = new Map()
+    /** @type {Set<FunctionExpression | ArrowFunctionExpression>} */
+    const computedFunctionNodes = new Set()
```
ota-meshi, line 83: You can probably do the same with an array. Since the array is simpler, can you change it to an array?

ota-meshi (changes requested):

sapphi-red:
Thank you for reviewing! I fixed it.

ota-meshi (approved):
Looks good to me! Thank you!

Issue #1399, opened 2020-12-30
Rule suggestion: `vue/valid-next-tick`


FloEdelmann:
**Please describe what the rule should do:**
Calling `Vue.nextTick` or `vm.$nextTick` without passing a callback and without awaiting the returned Promise is likely a mistake (probably a missing `await`).

**What category should the rule belong to?**
[ ] Enforces code style (layout)
[X] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
import { nextTick as nt } from 'vue';

// BAD:
nt();
Vue.nextTick();
this.$nextTick();
nt;
Vue.nextTick;
this.$nextTick;

// GOOD:
await nt();
await Vue.nextTick();
await this.$nextTick();
nt().then(callback);
Vue.nextTick().then(callback);
this.$nextTick().then(callback);
nt(callback);
Vue.nextTick(callback);
this.$nextTick(callback);
```

**Additional context**

- [`Vue.nextTick` API in Vue 2](https://vuejs.org/v2/api/#Vue-nextTick)
- [`vm.$nextTick` API in Vue 2](https://vuejs.org/v2/api/#vm-nextTick)
- [Global API Treeshaking](https://v3.vuejs.org/guide/migration/global-api-treeshaking.html)
- [Global `nextTick` API in Vue 3](https://v3.vuejs.org/api/global-api.html#nexttick)
- [Instance `$nextTick` API in Vue 3](https://v3.vuejs.org/api/instance-methods.html#nexttick)

Should #1318 be part of this rule (with option `enforceStyle: 'promise'` or `enforceStyle: 'callback'`)?

ota-meshi:
Thank you for suggesting this rule. I like this rule!
I think it makes sense that it's a different rule than `vue/next-tick-style` rule.

PR #1400, opened 2020-12-30
Add `vue/next-tick-style` rule

### LINKED ISSUE #1318: Rule suggestion: `vue/next-tick-style`
Since Vue v2.1.0, [`Vue.nextTick`](https://vuejs.org/v2/api/#Vue-nextTick) and [`vm.$nextTick`](https://vuejs.org/v2/api/#vm-nextTick) return a Promise if no callback is provided. So both of the following are valid:

* `Vue.nextTick(callback)`
* `Vue.nextTick().then()` / `await Vue.nextTick()`

Or inside components:

* `this.$nextTick(callback)`
* `this.$nextTick().then()` / `await this.$nextTick()`

There should be a rule that enforces one of the styles to make the codebase more consistent. I suggest that the Promise-style is the default of this rule, because it makes the code more concise and is generally well supported as of now.

However, enforcing the `await` or `.then()` style is out of scope for this plugin, there is already the [`eslint-plugin-promise` plugin](https://github.com/xjamundx/eslint-plugin-promise) with the [`promise/prefer-await-to-then` rule](https://github.com/xjamundx/eslint-plugin-promise/blob/v4.0.1/docs/rules/prefer-await-to-then.md).

[ &nbsp;] Warns about a potential error (problem)
[ &nbsp;] Suggests an alternate way of doing something (suggestion)
[ &nbsp;] Other (please specify:)

```js
// "vue/next-tick-style": "warn"
// "vue/next-tick-style": ["warn", "promise"]

// BAD:
Vue.nextTick(() => callback())
Vue.nextTick(callback)
this.$nextTick(() => callback())
this.$nextTick(callback)

// GOOD:
Vue.nextTick().then(() => callback())
await Vue.nextTick(); callback()
this.$nextTick().then(() => callback())
await this.$nextTick(); callback()
```

```js
// "vue/next-tick-style": ["warn", "callback"]

// BAD:
Vue.nextTick().then(() => callback())
await Vue.nextTick(); callback()
this.$nextTick().then(() => callback())
await this.$nextTick(); callback()

// GOOD:
Vue.nextTick(() => callback())
Vue.nextTick(callback)
this.$nextTick(() => callback())
this.$nextTick(callback)
```

FloEdelmann:
Closes #1318.

Before reviewing, please have a look at #1399. Does it make sense to merge these two rules into one? If so, I'll try to implement that.

### DIFF
--- a/lib/rules/next-tick-style.js
+++ b/lib/rules/next-tick-style.js
@@ -90,7 +90,7 @@ function isAwaitedPromise(callExpression) {
 
 module.exports = {
   meta: {
-    type: 'layout',
+    type: 'suggestion',
     docs: {
       description: 'enforce Promise or callback style in `nextTick`',
       categories: undefined,
--- a/lib/rules/next-tick-style.js
+++ b/lib/rules/next-tick-style.js
@@ -131,8 +131,8 @@ module.exports = {
             node,
             message:
               'Use the Promise returned by `nextTick` instead of passing a callback function.',
-            *fix(fixer) {
-              yield fixer.insertTextAfter(node, '().then')
+            fix(fixer) {
+              return fixer.insertTextAfter(node, '().then')
             }
           })
         }
--- a/lib/rules/next-tick-style.js
+++ b/lib/rules/next-tick-style.js
@@ -66,6 +66,19 @@ function getVueNextTickCallExpression(identifier, context) {
   return undefined
 }
 
+/**
+ * @param {CallExpression} callExpression
+ * @returns {boolean}
+ */
+function isAwaitedPromise(callExpression) {
+  return (
+    callExpression.parent.type === 'AwaitExpression' ||
+    (callExpression.parent.type === 'MemberExpression' &&
+      callExpression.parent.property.type === 'Identifier' &&
+      callExpression.parent.property.name === 'then')
+  )
+}
+
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


lib/rules/next-tick-style.js
```diff
@@ -0,0 +1,142 @@
...
+        if (
+          callExpression.arguments.length !== 0 ||
+          !isAwaitedPromise(callExpression)
+        ) {
+          context.report({
+            node,
+            message:
+              'Use the Promise returned by `nextTick` instead of passing a callback function.',
+            *fix(fixer) {
+              yield fixer.insertTextAfter(node, '().then')
```
ota-meshi, line 135: If there is only one fix, apply as follows.

```suggestion
            fix(fixer) {
              return fixer.insertTextAfter(node, '().then')
```

ota-meshi (changes requested):
Thank you for this PR!
I have some requests.

lib/rules/next-tick-style.js
```diff
@@ -0,0 +1,141 @@
...
+  )
+}
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'layout',
```
ota-meshi, line 88: Sorry. I forgot to send a comment.
I think it's better to categorize this rule as "suggestion". The changes caused by warning are little different in behavior.

lib/rules/next-tick-style.js
```diff
@@ -0,0 +1,141 @@
...
+    identifier.parent.type === 'MemberExpression' &&
+    identifier.parent.object.type === 'Identifier' &&
+    identifier.parent.object.name === 'Vue' &&
+    identifier.parent.parent.type === 'CallExpression'
+  ) {
+    return identifier.parent.parent
+  }
+
+  // Vue 3 Global API: import { nextTick as nt } from 'vue'; nt()
+  if (identifier.parent.type === 'CallExpression') {
```
ota-meshi, line 48: Sorry. I have noticed now.
You need to make sure that the arguments do not include nextTick.
You can pass it as a callback.

```js
foo()
   .then(nextTick)
   .then(() => {})
```

ota-meshi (approved):
Looks good to me! Thank you!

PR #1401, opened 2020-12-30
Add `vue/no-constant-condition` rule

### LINKED ISSUE #1358: Rule Proposal: vue/no-constant-condition
like [no-constant-condition](https://eslint.org/docs/rules/no-constant-condition) in eslint, but works with `v-if`/`v-show`.

[ ] Other (please specify:)

```vue
<div v-if="true"></div>
<div v-if="true || prop1"></div>
```

**Additional context**

FloEdelmann:
Closes #1358.

### DIFF
--- a/docs/rules/no-constant-condition.md
+++ b/docs/rules/no-constant-condition.md
@@ -1,0 +1,29 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-constant-condition
+description: disallow constant expressions in conditions
+since: v7.5.0
+---
+# vue/no-constant-condition
+
+> disallow constant expressions in conditions
+
+This rule is the same rule as core [no-constant-condition] rule but it applies to the expressions in `<template>`.
+
+## :books: Further Reading
+
+- [no-constant-condition]
+
+[no-constant-condition]: https://eslint.org/docs/rules/no-constant-condition
+
+## :rocket: Version
+
+This rule was introduced in eslint-plugin-vue v7.5.0
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-constant-condition.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-constant-condition.js)
+
+<sup>Taken with ❤️ [from ESLint core](https://eslint.org/docs/rules/no-constant-condition)</sup>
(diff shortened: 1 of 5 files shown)


FloEdelmann:
The test is failing in Node.js v8, because template literals are not supported there.

The original ESLint rule added support (and tests) for template literal only in ESLint v7.0.0 (see the [original PR](https://github.com/eslint/eslint/pull/12816)), where support for Node.js < v10 was dropped.

Should I just drop those tests?

tests/lib/rules/no-constant-condition.js
```diff
@@ -0,0 +1,111 @@
...
+        {
+          messageId: 'unexpected',
+          type: 'TemplateLiteral',
+          column: 31,
+          endColumn: 36
+        }
+      ]
+    },
+    {
+      code: '<template><CustomButton v-if="``" /></template>',
```
FloEdelmann, line 100: The test is failing in Node.js v8, because template literals are not supported there.

The original ESLint rule added support (and tests) for template literal only in ESLint v7.0.0 (see the [original PR](https://github.com/eslint/eslint/pull/12816)), where support for Node.js < v10 was dropped.

Should I just drop those tests?

ota-meshi: Comment out this test case for now.

lib/rules/no-constant-condition.js
```diff
@@ -0,0 +1,27 @@
...
+
+const { wrapCoreRule } = require('../utils')
+
+// eslint-disable-next-line no-invalid-meta, no-invalid-meta-docs-categories
+module.exports = wrapCoreRule('no-constant-condition', {
+  create(_context, { coreHandlers }) {
+    return {
+      VDirectiveKey(node) {
+        if (
+          node.name.name === 'if' &&
```
ota-meshi, line 14: Could you include `v-else-if` and `v-show` as well?

docs/rules/no-constant-condition.md
```diff
@@ -0,0 +1,29 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-constant-condition
+description: disallow constant expressions in conditions
+since: v7.5.0
```
ota-meshi, line 6: Could you remove the version as I will add it when I release it?

ota-meshi (changes requested):
Thank you for this PR! I have some requests.

ota-meshi (approved):

PR #1404, opened 2021-01-02
Add `vue/valid-next-tick` rule

### LINKED ISSUE #1399: Rule suggestion: `vue/valid-next-tick`

FloEdelmann:
Closes #1399.

### DIFF
--- a/lib/rules/valid-next-tick.js
+++ b/lib/rules/valid-next-tick.js
@@ -75,22 +75,6 @@ function isAwaitedPromise(callExpression) {
   )
 }
 
-/**
- * @param {ASTNode} node
- * @returns {FunctionExpression|undefined}
- */
-function getClosestFunction(node) {
-  while (node.parent !== null) {
-    node = /** @type {ASTNode} */ (node.parent)
-
-    if (node.type === 'FunctionExpression') {
-      return node
-    }
-  }
-
-  return undefined
-}
-
 // ------------------------------------------------------------------------------
 // Rule Definition
 // ------------------------------------------------------------------------------
@@ -137,18 +121,18 @@ module.exports = {
 
         if (parentNode.arguments.length === 0) {
           if (!isAwaitedPromise(parentNode)) {
-            const closestFunction = getClosestFunction(parentNode)
-            const isAsyncFunction = closestFunction
-              ? closestFunction.async
-              : false
-
             context.report({
               node,
               message:
                 'Await the Promise returned by `nextTick` or pass a callback function.',
-              fix: isAsyncFunction
-                ? (fixer) => fixer.insertTextBefore(parentNode, 'await ')
-                : undefined
+              suggest: [
+                {
+                  desc: 'Add missing `await` statement.',
+                  fix(fixer) {
+                    return fixer.insertTextBefore(parentNode, 'await ')
+                  }
+                }
+              ]
             })
           }
           return
--- a/lib/rules/valid-next-tick.js
+++ b/lib/rules/valid-next-tick.js
@@ -67,12 +67,46 @@ function getVueNextTickNode(identifier, context) {
  * @returns {boolean}
  */
 function isAwaitedPromise(callExpression) {
-  return (
-    callExpression.parent.type === 'AwaitExpression' ||
-    (callExpression.parent.type === 'MemberExpression' &&
-      callExpression.parent.property.type === 'Identifier' &&
-      callExpression.parent.property.name === 'then')
-  )
+  if (callExpression.parent.type === 'AwaitExpression') {
+    // cases like `await nextTick()`
+    return true
+  }
+
+  if (callExpression.parent.type === 'ReturnStatement') {
+    // cases like `return nextTick()`
+    return true
+  }
+
+  if (
+    callExpression.parent.type === 'MemberExpression' &&
+    callExpression.parent.property.type === 'Identifier' &&
+    callExpression.parent.property.name === 'then'
+  ) {
+    // cases like `nextTick().then()`
+    return true
+  }
+
+  if (
+    callExpression.parent.type === 'VariableDeclarator' ||
+    callExpression.parent.type === 'AssignmentExpression'
+  ) {
+    // cases like `let foo = nextTick()` or `foo = nextTick()`
+    return true
+  }
+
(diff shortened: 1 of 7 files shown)


lib/rules/valid-next-tick.js
```diff
@@ -0,0 +1,167 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce valid `nextTick` function calls',
+      categories: ['vue3-essential', 'essential'],
```
ota-meshi, line 103: Adding categories is a breaking change and will be done in the major version. Could you comment out?

lib/rules/valid-next-tick.js
```diff
@@ -0,0 +1,167 @@
...
+    return utils.defineVueVisitor(context, {
+      /** @param {Identifier} node */
+      Identifier(node) {
+        const parentNode = getVueNextTickParentNode(node, context)
+        if (!parentNode) {
+          return
+        }
+
+        if (parentNode.type !== 'CallExpression') {
+          context.report({
```
ota-meshi, line 120: I don't think we need to report this.
The user may store nextTick in a variable.

```js
const nt = Vue.nextTick
```

lib/rules/valid-next-tick.js
```diff
@@ -0,0 +1,167 @@
...
+
+  return undefined
+}
+
+/**
+ * @param {CallExpression} callExpression
+ * @returns {boolean}
+ */
+function isAwaitedPromise(callExpression) {
+  return (
```
ota-meshi, line 70: Change it to check for `Promise.xxx()` calls such as `Promise.all()`.
You also probably need to consider `foo = nextTick()` to be asynchronous.

lib/rules/valid-next-tick.js
```diff
@@ -0,0 +1,167 @@
...
+            const isAsyncFunction = closestFunction
+              ? closestFunction.async
+              : false
+
+            context.report({
+              node,
+              message:
+                'Await the Promise returned by `nextTick` or pass a callback function.',
+              fix: isAsyncFunction
+                ? (fixer) => fixer.insertTextBefore(parentNode, 'await ')
```
ota-meshi, line 142: There are some ways to fix it, so I think it's best not to fix it automatically.
You can implement the suggestion instead.
https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

lib/rules/valid-next-tick.js
```diff
@@ -0,0 +1,167 @@
...
+            node,
+            message: '`nextTick` is a function.',
+            fix(fixer) {
+              return fixer.insertTextAfter(node, '()')
+            }
+          })
+          return
+        }
+
+        if (parentNode.arguments.length === 0) {
```
ota-meshi, line 130: You need to make sure that the arguments do not include nextTick.
You can pass it as a callback.

```js
foo()
   .then(nextTick)
   .then(() => {})
```

ota-meshi (changes requested):
Thank you for your contribution! I have some requests.

lib/rules/valid-next-tick.js
FloEdelmann: I think it would be good to catch cases like `await this.$nextTick` instead of `await this.$nextTick()`, as it can be quite hard to find. Also, do we really need to consider assignment expressions? I don't think `const nt = Vue.nextTick` is a common pattern.

lib/rules/valid-next-tick.js
FloEdelmann: Using `nextTick` only makes sense sequentially (e.g. after a value change, to wait for a DOM re-render). So using `Promise.all([Vue.nextTick, someOtherPromise])` is not really a common pattern and might indeed be a misuse of `nextTick`, right?

lib/rules/valid-next-tick.js
ota-meshi: It's not as common as you say. But `const nt = Vue.nextTick` is clearly looks intentional. I think it needs to be excluded.
I think we should check `await this.$nextTick` and` this.$nextTick.then()`.

lib/rules/valid-next-tick.js
ota-meshi: Users may compare it to promises that are shorter than `nextTick()`. For example, `import(...)` is probably faster after the second time.
I don't think the following usage is misused.

```js
const [ , myUtility] = await Promise.all([nextTick(), import('./my-utility')])
myUtility.foo(vm)
```

Users may also use promises like queues.

```js
let queue =  nextTick()

if (foo) {
  queue = queue.then(() => procFoo())
}
if (bar) {
  queue = queue.then(() => procBar())
}

return queue
```

ota-meshi (approved):
Sorry for late reply. Thank you for changing. LGTM!

PR #1406, opened 2021-01-03
Add composition api's computed function support to vue/return-in-computed-property refs #1393

sapphi-red:
This PR adds support for composition api's computed function to `vue/return-in-computed-property `.
refs #1393


ota-meshi (approved):
LGTM! Thank you!

PR #1425, opened 2021-01-26
feat(no-bare-strings-in-template): ✨ allow to ignore specific elements

eyalch:
This rule is awesome, but there are some elements which should always be ignored. A good example for this is the [VIcon](https://vuetifyjs.com/en/components/icons/) element from [Vuetify](https://vuetifyjs.com/en/), which is used like this:

```vue
<VIcon>mdi-alarm</VIcon>
```

With the changes I introduced, this element can be ignored like this:

```json
{
  "vue/no-bare-strings-in-template": [
    "error",
    {
      "ignores": ["VIcon", "v-icon"]
    }
  ]
}
```

### DIFF
--- a/docs/rules/no-bare-strings-in-template.md
+++ b/docs/rules/no-bare-strings-in-template.md
@@ -12,7 +12,7 @@ since: v7.0.0
 ## :book: Rule Details
 
 This rule disallows the use of bare strings in `<template>`.  
-In order to be able to internationalize your application, you will need to avoid using plain strings in your templates. Instead, you would need to use a template helper specializing in translation.
+In order to be able to internationalize your application, you will need to avoid using plain strings in your templates. Instead, you would need to use a template helper specializing in translation.  
 
 This rule was inspired by [no-bare-strings rule in ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md).
 
--- a/docs/rules/no-bare-strings-in-template.md
+++ b/docs/rules/no-bare-strings-in-template.md
@@ -5,7 +5,6 @@ title: vue/no-bare-strings-in-template
 description: disallow the use of bare strings in `<template>`
 since: v7.0.0
 ---
-
 # vue/no-bare-strings-in-template
 
 > disallow the use of bare strings in `<template>`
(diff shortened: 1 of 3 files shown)


ota-meshi:
Thank you for this PR.

I think the name of that option very confusing. I don't know that "ignores" is the element name. It's also confusing if we need to add options to ignore by class name in the future.
Also, if you don't want to ignore the attribute check, consider the name of the option that will tell you.

Also, please do not make any document changes that are not relevant to this option.

eyalch:
@ota-meshi Fixed!

Please let me know if there's anything else that should be changed or can be improved.

ota-meshi:
With the following settings:

```json
{
    "ignoreComponents": ["v-icon"]
}
```

Why is the error reported in the following source code?

```xml
<v-icon v-text="foo" />
```

eyalch:
I'm not sure why. I've just added a test to check the `v-text` prop and it passes. How can I reproduce this?

ota-meshi:
I made a mistake, forgot to write the single quote.

```js
    {
      options: [{ ignoreComponents: ['v-icon'] }],
      code: `
      <template>
        <v-icon v-text="'foo'" />
      </template>
      `
    }
```

ota-meshi:
We also need to consider how to handle the following aria-label.

```vue
<v-icon aria-label="My Account" role="img" aria-hidden="false">
  mdiAccount
</v-icon>
```

https://vuetifyjs.com/en/components/icons/#semantic-svg-icons

eyalch:
@ota-meshi What would you expect to happen? IMO, if we set:

```json
{
  "vue/no-bare-strings-in-template": [
    "error",
    {
      "ignoreComponents": ["VIcon", "v-icon"]
    }
  ]
}
```

then this:

```vue
<VIcon aria-label="Alarm">mdi-alarm</VIcon>
```

should yield _only_ the following error:

```Unexpected non-translated string used in `aria-label`.```

What do you think?

eyalch:
@ota-meshi What do you say?

ota-meshi:
I think users will actually expect what you mean, but I think the option should be ignored as it is an option that ignores the component.
I think the name of the option is still confusing.

eyalch:
@ota-meshi Just so I understand correctly, you think that `ignoreComponents` is confusing, yes?

What about `ignoredComponents`? Or `componentsToIgnore`?

ota-meshi:
I think that if you name an option `ignoreComponents` (or `componentsToIgnore`), the user will think that the component should be ignored if the user specifies that option. This means that any attributes given to the component should also be ignored. In fact, it does not ignore the attributes.
So I think you need to rename the option or change the configuration of the option.

ota-meshi:

Issue #1110, opened 2020-04-28
Mixing max-attributes-per-line and a minimum number


jwillp:
I would like to do a mix of `vue/max-attributes-per-line` and a minimum, where I would prefer to have the attributes on one line instead of being multiline, but only up to a certain maximum number.
So somewhat of a mixture between min and max.

I haven't found a way to achieve this. Is there any way to do that?

Or at least, if it is not possible to force attributes to be on one line.

**Tell us about your environment**	
* **ESLint version:**
`6.8.0`
* **eslint-plugin-vue version:**	
`6.2.2`
* **Node version:**	
`v12.16.1` + `yarn: 1.21.1`

**Please show your full configuration:**

```typescript
module.exports = {
    root: true,
    env: {
        node: true
    },
    extends: [
        'plugin:vue/essential',
        'plugin:vue/recommended',
        '@vue/standard',
        '@vue/typescript',
        '@vue/typescript/recommended'
    ],
    rules: {
        'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
        'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
        indent: ['error', 4],

        '@typescript-eslint/no-explicit-any': 'off',
        '@typescript-eslint/ban-ts-ignore': 'off',

        'vue/script-indent': ['warn', 4, { baseIndent: 0 }],
        'vue/html-indent': ['warn', 4, { baseIndent: 0, alignVertically: true }],
        'vue/max-attributes-per-line': ['error', {
            singleline: 4,
            multiline: {
                max: 4,
                allowFirstLine: true
            }
        }]
    },
    parserOptions: {
        parser: '@typescript-eslint/parser',
        ecmaVersion: 2020
    }
}
```

**What did you do?**	

```vue
<div class="general-info">
    <figure
        class="avatar avatar-primary"
        data-initials="HW"
    />
</div>
```

**What did you expect to happen?**
```vue
<div class="general-info">
    <figure  class="avatar avatar-primary" data-initials="HW"  />
</div>
```

**What actually happened?**
Nothing, it stays unchanged

mapleby:
Me too.This is a headache for me, and I don't want them to use multiple lines of attributes or multiple attributes because it makes the code hard to understand.

mapleby:
I just found a solution, vue/max-len 😁

ota-meshi:
Thank you for posting this.

Vue team recommends one attribute per line.

[...]

PR #1431, opened 2021-02-04
New: `Add vue/single-attribute-single-line` rule

jacksonGross:
This PR adds the `vue/single-attribute-single-line` rule.

Idea based on https://github.com/vuejs/eslint-plugin-vue/issues/1110

First time contributing so hopefully I've followed the rules here 😄

### DIFF
--- a/docs/rules/single-attribute-single-line.md
+++ b/docs/rules/single-attribute-single-line.md
@@ -1,0 +1,67 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/single-attribute-single-line
+description: enforce component opening tags with a single attribute to be on a single line
+since: v7.5.0
+---
+# vue/single-attribute-single-line
+
+> enforce component opening tags with a single attribute to be on a single line
+
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Ensures that component opening tags that only have one attribute are inline to improve readability.
+
+## :book: Rule Details
+
+This rule aims to enforce component opening tags with a single attribute to be place on a single line.
+It checks all the elements in a template and verifies that when the number of attributes for a component is 1, that it is inlined on the opening tag.
+
+<eslint-code-block fix :rules="{'vue/single-attribute-single-line': ['error']}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <MyComponent lorem="1"/>
+  <MyComponent lorem="1" />
+  <MyComponent lorem="1"></MyComponent>
+  <MyComponent lorem="1">
+    <p>content</p>
+  </MyComponent>
+
+  <!-- ✗ BAD -->
+  <MyComponent
+    lorem="1"
+  />
+  <MyComponent
+    lorem="1"
+  ></MyComponent>
+  <MyComponent
+    lorem="1"
+  >
+    <p>content</p>
+  </MyComponent>
+</template>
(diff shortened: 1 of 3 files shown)


ota-meshi:
Thank you for PR.

But I think `single-attribute-single-line` rule will conflict and error resolution will not be possible if you disable the `allowFirstLine` option of the `max-attributes-per-line` rule. So I think it's difficult to incorporate `single-attribute-single-line` rule into this plugin.
Do you have any ideas for resolving rule conflicts?

jacksonGross:
> Thank you for PR.
> 
> But I think `single-attribute-single-line` rule will conflict and error resolution will not be possible if you disable the `allowFirstLine` option of the `max-attributes-per-line` rule. So I think it's difficult to incorporate `single-attribute-single-line` rule into this plugin.
> Do you have any ideas for resolving rule conflicts?

I did consider catering for this behaviour as an option to the `max-attributes-per-line` rule, but opted to separate it as I felt it could be used in combination with that rule. I suppose it could definitely conflict if you disabled the `allowFirstLine` option, but being default enabled I thought this rule being an opt-in that it would be ok. 

Not sure how this has been approached in the past where rules might collide based on options from another rule, so was hoping I could find some insight into that, and possibly some suggestions to improve this, and whether it would be an issue incorporating this rule.

ota-meshi:
Even if it's opt-in, I think rule conflicts can confuse users. I think we most avoid conflicts within a single plugin.

I'm not a fan of your style, so I'm not familiar with the details, I think we can abolish the `max-attributes-per-line` rule and replace it with a new rule related to attribute line breaks.
The new rule gives users the free to use the features of the `max-attributes-per-line` and `single-attribute-single-line` rules in their settings.

ota-meshi:
By the way, I don't think your current rule can resolve the error for the following one attribute:

```vue
<MyComponent :lorem="{
  prop: 42
}"/>
```

jacksonGross:
> Even if it's opt-in, I think rule conflicts can confuse users. I think we most avoid conflicts within a single plugin.
> 
> I'm not a fan of your style, so I'm not familiar with the details, I think we can abolish the `max-attributes-per-line` rule and replace it with a new rule related to attribute line breaks.
> The new rule gives users the free to use the features of the `max-attributes-per-line` and `single-attribute-single-line` rules in their settings.

Yeah that actually sounds like a better idea if we want to avoid conflicts but provide the functionality of both rules. Considering that the `max-attributes-per-line` rule is strongly recommended in the [styleguide](https://v3.vuejs.org/style-guide/#multi-attribute-elements-strongly-recommended) and isn't clear on what to do in the case of a single attribute, what would be the right path forward to achieving this? Personally, I think that part of the style guide is quite lacking, and would appreciate either an official recommendation on this, or having the power to configure it (which is why I made this PR in the first place). Happy to assist in either case 😄

jacksonGross:
> By the way, I don't think your current rule can resolve the error for the following one attribute:
> 
> ```
> <MyComponent :lorem="{
>   prop: 42
> }"/>
> ```

ah yes good catch!

vedmant:
I was searching a rule like this, but forcing some max number of attributes into single line, like 3 attributes is a single line, 4 and more multiline. Looks like this was not merged?

vedmant:
@jacksonGross I'm not able to make it turn multiline tags like:

```vue
<DisclosureButton
  class="flex items-center text-lg font-medium text-left"
/>
```
to single line
```vue
<DisclosureButton class="flex items-center text-lg font-medium text-left" />
```

It works only opposite direction, from single line to multi line.

PR #1437, opened 2021-02-09
Add "ignores" list to prop names casing

yassipad:
Using Vue 3, it is recommended to use the `emits` option to define the events emitted by a component. However, in doing so, the events are not available in the `attrs` object anymore and need to be registered as props. Following the pattern of `update:modelValue`, it is very common to use `:` inside an event's name, which is currently not possible using the `vue/prop-name-casing` rule. This commit allows for adding an `ignores` option (array of regex patterns) just like the `custom-event-name-casing` rule. 

PS: A default value for this "ignores" option should probably be `[".*?:.*?"]`.

### DIFF
--- a/lib/rules/prop-name-casing.js
+++ b/lib/rules/prop-name-casing.js
@@ -16,7 +16,6 @@ function create(context) {
   const options = context.options[0]
   const ignores = (context.options[1] && context.options[1].ignores) || []
 
-
   const caseType =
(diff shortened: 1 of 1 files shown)


ota-meshi:
I don't know what you want to ignore.  First, open an issue and provide some vue files you want to resolve.

yassipad:
Sorry about that, here's the issue: #1438

PR #1449, opened 2021-03-02
docs: remove mention of "prettier/vue"

thorn0:
`prettier/vue`, `prettier/@typescript-eslint`, etc. have been removed in `eslint-config-prettier` v8.0.0.
They are needed no longer.
See https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21

I've also removed the "If [Prettier] conflicts with the rule you have set, turn off that rule" part because that's exactly what `eslint-config-prettier` is for.

### DIFF
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -286,11 +286,7 @@ module.exports = {
 }
 ```
 
-If Prettier conflicts with a rule you have set, [turn off that rule][prettier-linters].
-
-Example **.eslintrc.js**:
-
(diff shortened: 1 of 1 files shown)


ota-meshi:
Thank you for this PR.

> I've also removed the "If [Prettier] conflicts with the rule you have set, turn off that rule" part because that's exactly what eslint-config-prettier is for.

I think this explanation is necessary because users can override the rules by adding them to `rules`.
What do you think?

thorn0:
@ota-meshi Good point. Fixed.

ota-meshi (approved):
Thank you!

PR #1451, opened 2021-03-03
Update: disallow `v-show` on `<template>` tag

### LINKED ISSUE #1410: Template No V-Show

Warn users that putting `v-show` directives on `<template />` tags does nothing.

Unlike `v-if`, `v-show` adds style attributes, which isn't valid for template tags.

[ ] Other (please specify:)

```vue
<template v-show="someCondition" />
```

g-plane:
Fixes #1410


ota-meshi (approved):
Thank you for this PR! LGTM!

PR #1465, opened 2021-03-25
feat(max-attributes-per-line): singleline.allowFirstLine option

privatenumber:
`max-attributes-per-line`

More

Lower the minimum value on `singleline`/`singleline.max` to 0 on the schema.

Update:
Add a `singleline.allowFirstLine` option that defaults to `true`, but prevents single attributes from being on the same line if set to `false`.

When setting `singleline.max = 0`, the following should be marked incorrect:

Update:
When setting `singleline.allowFirstLine = false`, the following should be marked incorrect:

```vue
<template>
	<component name="John Doe" />
</template>
```

Currently, can't set `singleline` to 0 because of the schema validation. 

Correct it to:
```vue
<template>
	<component
		name="John Doe"
	/>
</template>
```

**Additional context**

- This was working in prior versions of ESLint that did not enforce the JSON schema on options. But since upgrading ESLint, this appeared to be a regression. Since it's not a regression in this plugin, I labeled this as a feature request.

- This is style used by a large codebase I'm contributing to.

### DIFF
--- a/docs/rules/max-attributes-per-line.md
+++ b/docs/rules/max-attributes-per-line.md
@@ -57,7 +57,10 @@ There is a configurable number of attributes that are acceptable in one-line cas
 ```json
 {
   "vue/max-attributes-per-line": ["error", {
-    "singleline": 1,
+    "singleline": {
+      "max": 1,
+      "allowFirstLine": true
+    },      
     "multiline": {
       "max": 1,
       "allowFirstLine": false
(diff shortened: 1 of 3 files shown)


ota-meshi:
Thank you for this PR!
I think it easier for users to add a `singleline.allowFirstLine` option similar to `multiline.allowFirstLine` option.
What do you think?

privatenumber:
Yeah that definitely makes more sense when trying to grasp the options.

Will update to `allowFirstLine`. Thanks!

privatenumber:
Now that I'm working on updating the docs, I'm second-guessing whether it makes sense.


It's a little strange that a "singleline" rule enforces the multiline behavior:

```js
['error', {singleline: { allowFirstLine: false }}]
```

```html
<template>
	<component
		name="John Doe"
	/>
</template>
```

I still think it's an improvement over `max: 0` though.

ota-meshi (approved):
Thank you for changing this PR! LGTM

PR #1469, opened 2021-03-28
Fix false negative with require-valid-default-prop on props with type assertion

### LINKED ISSUE #1468: False negative with vue/require-valid-default-prop on props with type assertion
**Checklist**

- **ESLint version:** 7.23.0
- **eslint-plugin-vue version:** 7.8.0 
- **Node version:** 14.16.0
- **Operating System:** Windows 10 Pro

```js
module.exports = {
  root: true,
  env: {
    node: true,
  },
  extends: [
    "plugin:vue/essential",
    "eslint:recommended",
    "@vue/typescript/recommended",
    "@vue/prettier",
    "@vue/prettier/@typescript-eslint",
  ],
  parserOptions: {
    ecmaVersion: 2020,
  },
  rules: {},
};
```

**What did you do?**	

I annotated type to array prop with typescript according to the following documentations.
- https://vuejs.org/v2/guide/typescript.html#Annotating-Props
- https://v3.vuejs.org/guide/typescript-support.html#annotating-return-types

If the prop is an array, the default should be a function, but I set empty array directly by mistake.

```vue
<script lang="ts">
import Vue from "vue";
import { PropType } from "vue";

export default Vue.extend({
  props: {
    foo: {
      type: Array as PropType<string[]>,
      default: [],
    },
  },
});
</script>
```

eslint reports type of the default value for 'foo' prop must be a function.

**What actually happened?**

eslint did not report that. If there is no type assertion, it reports correctly.

frozenbonito:
fix #1468.

This PR ensure that eslint-plugin-vue reports `Type of thye default value for 'foo' prop must be function` in following case.
(It works the same for `Object` or `Function`.)

```vue
<script lang="ts">
import Vue from "vue";
import { PropType } from "vue";

export default Vue.extend({
  props: {
    foo: {
      type: Array as PropType<string[]>,
      default: [],
    },
  },
});
</script>
```

Note: In following case, type cheking based on annotated type of the return value of the default function will not work.

```vue
<script lang="ts">
import Vue from "vue";
import { PropType } from "vue";

export default Vue.extend({
  props: {
    foo: {
      type: Array as PropType<string[]>,
      default: (): number[] => [], // The function should return array of string, but this PR can't check it.
    },
  },
});
</script>
```

### DIFF
--- a/lib/rules/require-valid-default-prop.js
+++ b/lib/rules/require-valid-default-prop.js
@@ -49,10 +49,11 @@ function getPropertyNode(obj, name) {
 }
 
 /**
- * @param {Expression} node
+ * @param {Expression} targetNode
  * @returns {string[]}
  */
-function getTypes(node) {
+function getTypes(targetNode) {
+  const node = utils.skipTSAsExpression(targetNode)
   if (node.type === 'Identifier') {
     return [node.name]
   } else if (node.type === 'ArrayExpression') {
--- a/tests/lib/rules/require-valid-default-prop.js
+++ b/tests/lib/rules/require-valid-default-prop.js
@@ -207,6 +207,51 @@ ruleTester.run('require-valid-default-prop', rule, {
         }
       }`,
       parserOptions
+    },
+    {
+      filename: 'test.vue',
+      code: `export default Vue.extend({
+          props: {
+            foo: {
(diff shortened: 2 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #1487, opened 2021-05-05
Add `vue/require-emit-validator` rule

### LINKED ISSUE #1482: Rule proposal: require-emit-validator

Enforce detailed `emits` option with types.

[ ] Other (please specify:)

```ts
// bad
  emits: ['update'],
})

// good
  emits: {
    update: (payload: Payload) => typeof payload === 'object',
  },
})
```

**Additional context**

- Declaring types of `emits` can provide better TypeScript inference.
- This rule is similar to [require-prop-types](https://eslint.vuejs.org/rules/require-prop-types.html).

g-plane:
Fixes #1482 .

### DIFF
--- a/docs/rules/require-emit-types.md
+++ b/docs/rules/require-emit-types.md
@@ -1,0 +1,61 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/require-emit-types
+description: require type definitions in emits
+---
+# vue/require-emit-types
+
+> require type definitions in emits
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :gear: This rule is included in `"plugin:vue/vue3-strongly-recommended"` and `"plugin:vue/vue3-recommended"`.
+
+## :book: Rule Details
+
+This rule enforces that a `emits` statement contains type definition.
+
+Declaring `emits` with types can bring better maintenance.
+Even if using with TypeScript, this can provide better type inference when annotating parameters with types.
+
+<eslint-code-block :rules="{'vue/require-emit-types': ['error']}">
+
+```vue
+<script>
+/* ✓ GOOD */
+Vue.component('foo', {
+  emits: {
+    // Emit with arguments
+    foo: (payload) => { /* validate payload */ },
+    // Emit without parameters
+    bar: () => true,
+  }
+})
+
+/* ✗ BAD */
+Vue.component('bar', {
+  emits: ['foo']
+})
+
+Vue.component('baz', {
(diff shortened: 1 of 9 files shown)


lib/rules/require-emit-types.js
```diff
@@ -0,0 +1,69 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'require type definitions in emits',
+      categories: ['vue3-strongly-recommended'],
```
ota-meshi, line 23: Could you change the category to undefined as we will not change the configuration except for major version releases?

docs/rules/require-emit-types.md
```diff
@@ -0,0 +1,61 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/require-emit-types
+description: require type definitions in emits
+---
+# vue/require-emit-types
```
ota-meshi, line 7: I noticed that the documentation doesn't describe it as a "type" validation. I think it might be better to change the rule name (e.g. `require-emit-validator`), but what do you think?

https://v3.vuejs.org/guide/component-custom-events.html#validate-emitted-events

tests/lib/rules/require-emit-types.js
```diff
@@ -0,0 +1,266 @@
...
+            foo(payload: string | number) {
+              return true
+            },
+          },
+        })
+      `,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+      parser: require.resolve('@typescript-eslint/parser')
+    }
+  ],
```
ota-meshi, line 143: Could you add the following test cases? I think these should not report errors.

```js
function foo () {}
export default {
  emits: {
    foo
  }
}
```

```js
import {isNumber} from './mod'
export default {
  emits: {
    foo: isNumber
  }
}
```

ota-meshi (changes requested):
Thank you for this PR! I have some change requests.

g-plane:
I have a question about `null` for skipping validation: some new users may just want to define an event without any payload, and they may choose to use `null`, but actual they should use `() => true` which means "no payload". Should we detect for `null` and provide a suggestion which advise them to use `() => true`?

ota-meshi:
Thank you for the question.
I think we need to suggest `() => true` instead of `null`. `() => true` can type check for no payload.
In addition, if the user wants to check for no payload, the user can also set `(...args:[]) => args.length === 0`.

ota-meshi (approved):
LGTM! Thank you!

PR #1498, opened 2021-05-27
fix: detect prop usage in toRef

lukashass:
When using a prop in `toRef`, it is wrongly considered unused. This is due to the call signature of `toRef`:

```js
const propAsRef = toRef(props, 'propName');
```

This is a common use-case suggested in the Vue documentation. See https://v3.vuejs.org/api/refs-api.html#toref

While I'm not sure how viable the detection of this usage is, I would like to try and implement it.
Maybe you can give me some pointers.


ota-meshi:

PR #1501, opened 2021-05-31
update `no-deprecated-destroyed-lifecycle` rule

Alanscut:
in Vue 3.x:
- `destroyed` lifecycle option has been renamed to `unmounted`
- `beforeDestroy` lifecycle option has been renamed to `beforeUnmount`
see in https://v3.vuejs.org/guide/migration/introduction.html#other-minor-changes.

so we should fix them instead of giving suggestion

### DIFF
--- a/tests/lib/rules/no-deprecated-destroyed-lifecycle.js
+++ b/tests/lib/rules/no-deprecated-destroyed-lifecycle.js
@@ -111,42 +111,24 @@ ruleTester.run('no-deprecated-destroyed-lifecycle', rule, {
       }
       </script>
       `,
-      errors: [
-        {
-          message:
-            'The `beforeDestroy` lifecycle hook is deprecated. Use `beforeUnmount` instead.',
-          line: 4,
-          suggestions: [
-            {
-              desc: 'Instead, change to `beforeUnmount`.',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-deprecated-destroyed-lifecycle.js
```diff
@@ -71,8 +71,8 @@ ruleTester.run('no-deprecated-destroyed-lifecycle', rule, {
       code: `
       <script>
       export default {
-        beforeUnmount:beforeDestroy,
-        unmounted:destroyed,
+        beforeUnmount,
+        unmounted,
```
ota-meshi, line 75: Why did you need to change this?

ota-meshi (approved):

ota-meshi (changes requested):
I made a mistake.

tests/lib/rules/no-deprecated-destroyed-lifecycle.js
Alanscut: sorry, that's a misoperation, updated now

ota-meshi (approved):
LGTM! Thank you!

PR #1505, opened 2021-06-01
Fix rules index for `require-emit-validator`

g-plane:
This PR fixes rule `require-emit-validator` doesn't appear at [Available rules](https://eslint.vuejs.org/rules/) page.


ota-meshi (approved):
Thank you!

PR #1512, opened 2021-06-10
Feature/1463

### LINKED ISSUE #1463: vue/component-tags-order - Could it be made fixable?

`vue/component-tags-order`

No

I wish this rule could be auto-changed on `--fix`

None. Rule stays the same.

No changes.

On --fix, it will swap a single-file component's main tags around so they're in the order prescribed by the rule.

**Additional context**

I've been thinking of putting together a regex that would allow me to run a mass replacement on my project to ensure all tags are in the proper order, and it seems like it's probably possible. If a regex can do it, it makes me think that it's probably possible too, and easier, for a tool that "understands" the code like ESLint.

IWANABETHATGUY:
resolved #1463

### DIFF
--- a/lib/rules/component-tags-order.js
+++ b/lib/rules/component-tags-order.js
@@ -24,7 +24,7 @@ module.exports = {
       categories: ['vue3-recommended', 'recommended'],
       url: 'https://eslint.vuejs.org/rules/component-tags-order.html'
     },
-    fixable: null,
+    fixable: 'code',
     schema: [
       {
         type: 'object',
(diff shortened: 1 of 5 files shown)


IWANABETHATGUY:
cc @ota-meshi, ready for review

ota-meshi:
Thank you for this PR.
Could you change it to auto-fix for each reported element, similar to the `order-in-components` rule?
Even if there are multiple out-of-order violations, some of them may be suppressed by the [eslint-disable comment](https://eslint.vuejs.org/rules/comment-directive.html).

Also, could you add test cases with a comment in before of each element?

lib/rules/component-tags-order.js
```diff
@@ -123,9 +162,38 @@ module.exports = {
...
+          const sourceCode = context.getSourceCode().text
+          const reOrderedCode = Array.from(
+            { length: elements.length },
+            (_, i) => {
+              const [start, end] = elements[sortTagList[i].index].range
+              const code = sourceCode.slice(start, end)
+              return code
+            }
+          ).join('\n')
+          reportCandidates.forEach((error) => {
```
IWANABETHATGUY, line 188: @ota-meshi, i report each descriptor here, because i want to share the final souceFile, it could improve a little performance.

ota-meshi: > because i want to share the final souceFile, it could improve a little performance.

I think it's more important that suppressive comments work than performance. We need to change it to move only the reported element.

ota-meshi:

PR #1513, opened 2021-06-10
docs: remove autofix info from vue/eqeqeq

bb-in-hoodie:
`vue/eqeqeq` is marked as autofixable, while it is not.
- [ESLint eqeqeq document](https://eslint.org/docs/2.0.0/rules/eqeqeq)
  - there is no `autofix` phrase
- [a thread related to this issue](https://github.com/eslint/eslint/issues/4578)
  - they have removed autofix from eqeqeq in 2015


ota-meshi:
The latest version v7, seems to support auto-fixing, so I'll close this PR.

https://eslint.org/docs/rules/eqeqeq

image

bb-in-hoodie:
Oh, I'm sorry.
I thought autofix isn't working when I tested it out, but then I just figured out that it autofixes the code only under the very specific circumstance (where two values have the same type 'explicitly'). (and also I was looking at a very very old document).

Thanks for your correction.

Issue #1248, opened 2020-07-17
Supports for <script setup> and <style> variable injection


ota-meshi:
eslint-plugin-vue should be change to support [<script setup>] and [<style> variable injection].
I will list the work needed to support these on this issue.

These features were released as experimental features in Vue.js 3.0. 
So I think that the inclusion or not of the changes listed here will not affect the release of eslint-plugin-vue 7.0.0.

---

### Core 

- [x] Change the parser to parse multiple <script> blocks. (issue https://github.com/mysticatea/vue-eslint-parser/issues/78)
- ~~Change the parser to parse `<style vars="...">`. (issue https://github.com/mysticatea/vue-eslint-parser/issues/78)~~
- ~~Change the parser to parse `<script setup="...">`. (issue https://github.com/mysticatea/vue-eslint-parser/issues/78)~~
- more

### Rules

- [x] #1560 ~~Add `no-refer-vars-script-setup-define-props` rule that disallows referencing variables in `defineProps()` of `<script setup>`~~.
- [x] #1561 ~~Add `no-refer-vars-script-setup-define-emit` rule that disallows referencing variables in `defineEmit()` of `<script setup>`~~.
- [x] #1559 Add `no-export-in-script-setup` rule that disallows ES export in `<script setup>`.
- ~~Change `no-parsing-error` rule to reports parsing errors in `<script setup="...">`~~.
- ~~Change `no-parsing-error` rule to reports parsing errors in `<style vars="...">`~~.
- ~~Add `valid-script-setup` rule that checks whether `<script setup="...">` is valid arguments.~~
- ~~Add `valid-style-vars` rule that checks whether `<style vars="...">` is valid expression.~~
- ~~Add `no-vars-setup-export-default` rule that disallows referencing variables in `export default` of `<script setup>`.~~
- more

---

https://github.com/vuejs/rfcs/pull/227
https://github.com/vuejs/rfcs/blob/script-setup-2/active-rfcs/0000-script-setup.md
https://github.com/vuejs/rfcs/pull/231
https://github.com/vuejs/rfcs/blob/style-vars-2/active-rfcs/0000-sfc-style-variables.md

~~https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-script-setup.md~~
~~https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-style-variables.md~~

[...]

PR #1529, opened 2021-07-01
Add `vue/script-setup-uses-vars` rule

ota-meshi:
This PR adds `vue/script-setup-uses-vars` rule

The `vue/script-setup-uses-vars` rule that will find variables in `<script setup>` used in `<template>` and mark them as used.

refs #1248


ota-meshi:
We need to wait for the release of https://github.com/vuejs/vue-eslint-parser/pull/110.

PR #1540, opened 2021-07-03
Update `vue/no-setup-props-destructure` rule to support `<script setup>`

ota-meshi:
Related to #1248


lvjiaxuan:
Featrue.😀

Would `vue/no-setup-props-destructure` detect `reactivityTransform: true`?

Reference: https://vuejs.org/guide/extras/reactivity-transform.html#explicit-opt-in

PR #1570, opened 2021-07-16
min requirement node version is 10

nazo:
https://github.com/npm/node-semver/blob/v7.3.2/package.json#L36

node-semver v7.3.2 is going to need nodejs 10.


ota-meshi:
Thank you for this PR.
But we need to support node 8 until major version upgrade.
So now we need to downgrade `semver`. I will open a PR for that.

PR #1588, opened 2021-07-30
Drop supports Node.js v8, v10, v13 and v15

ota-meshi:


revolter:
What's wrong with Node 15?

PR #1605, opened 2021-08-12
Add `vue/no-v-text` rule

### LINKED ISSUE #1435: `vue/no-v-text`

This rule would report usage of `v-text`.

[ ] Other (please specify:)

```vue
<template>
  <div v-text="foobar" />
</template>
```

**Additional context**

This issue can be fixable. The above code would be fixed to:
```vue
<template>
  <div>
    {{foobar}}
  </div>
</template>
```

tyankatsu0105:
close #1435

Add rule `vue/no-v-text`.
- Fixable
  - But in the case of selfclose element, it can't be fixed.

### DIFF
--- a/docs/rules/no-v-text.md
+++ b/docs/rules/no-v-text.md
@@ -1,0 +1,47 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-v-text
+description: disallow use of v-text
+since: v7.17.0
+---
+# vue/no-v-text
+
+> disallow use of v-text
+
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+  - However, when using selfClose to element with v-text like `<div v-text="foobar" />`, it can't be fixed.
+
+## :book: Rule Details
+
+This rule reports all uses of `v-text` directive.
+
+
+<eslint-code-block :rules="{'vue/no-v-text': ['error']}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <div>{{ foobar }}</div>
+
+  <!-- ✗ BAD -->
+  <div v-text="foobar"></div>
+  <!-- Reported. However, Not fixable -->
+  <div v-text="foobar" />
+</template>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :rocket: Version
+
+This rule was introduced in eslint-plugin-vue v7.17.0
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-v-text.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-v-text.js)
--- a/docs/rules/no-v-text.md
+++ b/docs/rules/no-v-text.md
@@ -10,6 +10,7 @@ since: v7.17.0
 > disallow use of v-text
 
 - :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
   - However, when using selfClose to element with v-text like `<div v-text="foobar" />`, it can't be fixed.
 
 ## :book: Rule Details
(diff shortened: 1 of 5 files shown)


lib/rules/no-v-text.js
```diff
@@ -0,0 +1,76 @@
...
+    (attribute) =>
+      attribute.key.type === 'VDirectiveKey' &&
+      attribute.key.name.name === 'text'
+  )
+
+  if (!vText) return
+  if (!vText.value) return
+  if (vText.value.type !== 'VExpressionContainer') return
+  if (!vText.value.expression) return
+  if (vText.value.expression.type !== 'Identifier') return
```
ota-meshi, line 28: Why limit it to an Identifier? I think all expressions should be reported.

lib/rules/no-v-text.js
```diff
@@ -0,0 +1,76 @@
...
+      /** @param {VElement} node */
+      "VElement:has(VAttribute[directive=true][key.name.name='text'])"(node) {
+        const vText = getVText(node)
+        if (!vText) return
+
+        context.report({
+          node: vText.node,
+          loc: vText.loc,
+          message: "Don't use 'v-text'.",
+          fix(fixable) {
```
ota-meshi, line 61: I think autofix is very difficult. For example, converting the following expression to `{{}}` will result in a parse error in Vue2.

```html
<div v-text="'</div>'"></div>
```

↓

```html
<div >{{'</div>'}}</div>
```

ota-meshi, line 61: If you want to implement autofix, Could you also add the following test cases.

```html
<!--
  Do not autofix
-->
<div v-text="foo">contents</div>
<div v-text="foo"><div>elements</div></div>
<div v-text="foo"><!--comments--></div>
<div v-text="foo">mixed <br> <!-- comment --> </div>

<!--
  <pre> including whitespaces:
  It is necessary to autofix so that no whitespaces remain.
  Also, since it depends on CSS whether it is pre formatted or not,
  the same can be said for elements other than <pre>.
-->
<pre v-text="foo">
</pre>
<pre v-text="foo">

<!--comments-->

</pre>
```

ota-meshi (changes requested):
Thank you for this PR!
I have some requests.

lib/rules/no-v-text.js
tyankatsu0105: It's my misunderstood.
I don't know how to use `v-text` very much.
I'm going to fix it.

tyankatsu0105: I can use `utils.getDirective`!!!

lib/rules/no-v-text.js
tyankatsu0105: I make sense of what there are some points that I have to think about.
Yeah. I agree with your mention point.
I will remove the fix function from this rule.

docs/rules/no-v-text.md
```diff
@@ -0,0 +1,48 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-v-text
+description: disallow use of v-text
+since: v7.17.0
```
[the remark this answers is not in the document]
tyankatsu0105: done
https://github.com/vuejs/eslint-plugin-vue/pull/1605/commits/623849131c6859ab46a0ccafe945fb630e974945

docs/rules/no-v-text.md
```diff
@@ -0,0 +1,48 @@
...
+---
+# vue/no-v-text
+
+> disallow use of v-text
+
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+  - However, when using selfClose to element with v-text like `<div v-text="foobar" />`, it can't be fixed.
```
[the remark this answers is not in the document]
tyankatsu0105: done
https://github.com/vuejs/eslint-plugin-vue/pull/1605/commits/623849131c6859ab46a0ccafe945fb630e974945

lib/rules/no-v-text.js
tyankatsu0105: I fixed.

lib/rules/no-v-text.js
tyankatsu0105: done
https://github.com/vuejs/eslint-plugin-vue/pull/1605/commits/bc030beb6c98ee463d1981efb0d625b32214abc4

ota-meshi (approved):
LGTM! Thank you!

PR #1607, opened 2021-08-13
Prepare suggestable rules for ESLint v8 (and update `eslint-plugin-eslint-plugin` to v3)

bmish:
* The upcoming [ESLint v8](https://eslint.org/blog/2021/06/whats-coming-in-eslint-8.0.0#rules-with-suggestions-now-require-the-metahassuggestions-property) release will require rules that provide suggestions to enable `meta.hasSuggestions` so I have autofixed that using the [eslint-plugin/require-meta-has-suggestions](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-has-suggestions.md) rule
* A lot of the rules/functionality in `eslint-internal-rules` have been replaced by public rules in [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin)


ota-meshi (approved):
Thank you for your contribution!

PR #1609, opened 2021-08-14
Update README.md

QvanJian:
- Fix the way to **_turn off the Vetur checking_** of Vue files.
 - Provide a way to config ESlint so that VS Code will **_ignore the rules turned off_**.


ota-meshi:
Thank you for this PR.  Hmm.  However, I don't think the explanation you added is a frequently asked question, so I will close this PR.

PR #1610, opened 2021-08-15
Add support for ESLint v8.

ota-meshi:


MichaelDeBoey:
Since we're going to only support ESLint v8-beta (for now), maybe we should release this as a beta too? 🤔

ota-meshi:
I plan to merge this PR and release it in a minor version.

ankurk91:
Stable version has been tagged

https://github.com/eslint/eslint/releases/tag/v8.0.0

PR #1612, opened 2021-08-17
Export rules and configs using `requireindex`

bmish:
* No more need to maintain list of exported configs/rules (either manually or through a script)
  * Reduces barrier to creating/renaming lint rules
* [requireindex](https://www.npmjs.com/package/requireindex) is commonly used by other linters for the same purpose, i.e.:
  * https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/index.js
  * https://github.com/platinumazure/eslint-plugin-qunit/blob/master/index.js
* Renames `lib/rules/syntaxes/` to `lib/rules/_syntaxes/` so requireindex doesn't try to export it


bmish:
I am confused by this build failure...

ota-meshi:
This PR change will stop the DEMO site from working.
e.g. https://eslint.vuejs.org/rules/no-arrow-functions-in-watch.html

So I close this PR.

PR #1639, opened 2021-09-27
Add vue/no-restricted-class rule

taobojlen:
I had to refactor the rule somewhat from my original implementation to make it work with Typescript. This made it a bit harder to read, so any suggestions for improving that are welcome.

### DIFF
--- a/docs/rules/no-restricted-class.md
+++ b/docs/rules/no-restricted-class.md
@@ -2,11 +2,13 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-restricted-class
-description: disallow specific classes
+description: disallow specific classes in Vue components
 ---
-# vue/no-restricted-classes
+# vue/no-restricted-class
 
-> disallow specific classes
+> disallow specific classes in Vue components
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 
 ## :book: Rule Details
 
@@ -70,10 +72,6 @@ export default {
 </script>
 ```
 :::
-
-## :rocket: Version
-
-This rule was introduced in eslint-plugin-vue v7.19.0.
 
 ## :mag: Implementation
 
--- a/docs/rules/no-restricted-class.md
+++ b/docs/rules/no-restricted-class.md
@@ -1,0 +1,104 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-restricted-class
+description: disallow specific classes
+since: v7.19.0
+---
+# vue/no-restricted-classes
+
+> disallow specific classes
+
+## :book: Rule Details
+
+This rule lets you specify a list of classes that you don't want to allow in your templates.
+
+## :wrench: Options
+
+The simplest way to specify a list of forbidden classes is to pass it directly
+in the rule configuration.
+
+```json
+{
+  "vue/no-restricted-props": ["error", { classes: ["forbidden"] }]
+}
+```
+
+<eslint-code-block :rules="{'vue/no-restricted-class': ['error', { classes: ['forbidden'] }]}">
+
+```vue
+<template>
+  <!-- ✗ BAD -->
+  <div class="forbidden" />
+  <div :class="{forbidden: someBoolean}" />
+  <div :class="`forbidden ${someString}`" />
+  <div :class="'forbidden'" />
+  <div :class="'forbidden ' + someString" />
+  <!-- ✗ GOOD -->
+  <div class="allowed-class" />
+</template>
+
+<script>
+export default {
+  props: {
+    someBoolean: Boolean,
+    someString: String,
+  }
+}
+</script>
+```
+
+</eslint-code-block>
+
+
+Alternatively, you can also specify a list of files that contain forbidden classes. Each file
+must contain a JSON-formatted array of strings.
+
+```json
+{
+  "vue/no-restricted-props": ["error",
+    {
+      files: [".eslint/forbidden-classes.json"]
+    },
+  ]
+}
+```
+
+`.eslint/forbidden-classes.json`:
+
+```
+[
+  "forbidden-class",
+  "another-forbidden-class"
+]
+```
+
+::: warning Note
+This rule will only detect classes that are used as strings in your templates. Passing classes via
+variables, like below, will not be detected by this rule.
+
+```vue
+<template>
+  <div :class="classes" />
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      classes: "forbidden"
+    }
+  }
+}
+</script>
+```
+:::
+
+## :rocket: Version
+
+This rule was introduced in eslint-plugin-vue v7.19.0.
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-restricted-class.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-restricted-class.js)
--- a/docs/rules/no-restricted-class.md
+++ b/docs/rules/no-restricted-class.md
@@ -3,7 +3,6 @@ pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-restricted-class
 description: disallow specific classes
-since: v7.19.0
 ---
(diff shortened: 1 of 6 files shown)


docs/rules/no-restricted-class.md
```diff
@@ -0,0 +1,104 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-restricted-class
+description: disallow specific classes
+since: v7.19.0
```
ota-meshi, line 6: Remove it as it will be added automatically by the release script.

```suggestion
```

docs/rules/no-restricted-class.md
```diff
@@ -0,0 +1,104 @@
...
+This rule lets you specify a list of classes that you don't want to allow in your templates.
+
+## :wrench: Options
+
+The simplest way to specify a list of forbidden classes is to pass it directly
+in the rule configuration.
+
+```json
+{
+  "vue/no-restricted-props": ["error", { classes: ["forbidden"] }]
```
ota-meshi, line 23: I think it would be good to be able to accept options in an array like any other `no-restricted-xxx` rule.

```suggestion
  "vue/no-restricted-props": ["error", "forbidden", "forbidden2", "forbidden3"]
```

docs/rules/no-restricted-class.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+Alternatively, you can also specify a list of files that contain forbidden classes. Each file
+must contain a JSON-formatted array of strings.
+
+```json
+{
+  "vue/no-restricted-props": ["error",
+    {
+      files: [".eslint/forbidden-classes.json"]
```
ota-meshi, line 61: I don't think we need the files option. Users can read it in the configuration file.

```js
"vue/no-restricted-props": ["error", {classes: require('./my-forbidden-classes.json')}]
```

lib/rules/no-restricted-class.js
```diff
@@ -0,0 +1,221 @@
...
+      'VAttribute[directive=false][key.name="class"]'(node) {
+        const classes = new Set(node.value.value.split(/\s+/))
+        reportForbiddenClasses(classes, node, context, forbiddenClasses)
+      },
+
+      // <Element :class="{forbidden: someBoolean}" />
+      /**
+       * @param {VAttribute & { value: { expression: ObjectExpression } } } node
+       */
+      'VAttribute[directive=true][key.argument.name="class"][value.expression.type="ObjectExpression"]'(
```
ota-meshi, line 136: `[key.name.name='bind']` is required. If not, `@class="..."` will also be hit.
```suggestion
      'VAttribute[directive=true][key.name.name="bind"][key.argument.name="class"][value.expression.type="ObjectExpression"]'(
```

lib/rules/no-restricted-class.js
```diff
@@ -0,0 +1,221 @@
...
+        if (node.value.expression.left.type === 'PrivateIdentifier') return
+        const classNames = new Set(
+          flattenBinaryExpression(
+            node.value.expression.left,
+            node.value.expression.right
+          ).split(/\s+/)
+        )
+        reportForbiddenClasses(classNames, node, context, forbiddenClasses)
+      }
+    })
```
ota-meshi, line 219: Array-Syntax also needs to be processed.

https://vuejs.org/v2/guide/class-and-style.html#Array-Syntax

lib/rules/no-restricted-class.js
ota-meshi, line 136: What do you think about writing the following so that TemplateLiteral in BinaryExpression can also be processed?

```js

    /**
     * @param {Expression} node
     * @param {boolean} [textOnly]
     * @returns {IterableIterator<{ className:string, reportNode: ESNode }>}
     */
    function* extractClassNames(node, textOnly) {
      if (node.type === 'Literal') {
        yield* `${node.value}`
          .split(/\s+/)
          .map((className) => ({ className, reportNode: node }))
        return
      }
      if (node.type === 'TemplateLiteral') {
        for (const templateElement of node.quasis) {
          yield* templateElement.value.cooked
            .split(/\s+/)
            .map((className) => ({ className, reportNode: templateElement }))
        }
        for (const expr of node.expressions) {
          yield* extractClassNames(expr, true)
        }
        return
      }
      if (node.type === 'BinaryExpression') {
        if (node.operator !== '+') {
          return
        }
        yield* extractClassNames(node.left, true)
        yield* extractClassNames(node.right, true)
        return
      }
      if (!textOnly) {
        return
      }
      if (node.type === 'ObjectExpression') {
        for (const prop of node.properties) {
          if (prop.type !== 'Property') {
            continue
          }
          const classNames = utils.getStaticPropertyName(prop)
          if (!classNames) {
            continue
          }
          yield* classNames
            .split(/\s+/)
            .map((className) => ({ className, reportNode: prop.key }))
        }
        return
      }
      if (node.type === 'ArrayExpression') {
        for (const element of node.elements) {
          if (element == null) {
            continue
          }
          if (element.type === 'SpreadElement') {
            continue
          }
          yield* extractClassNames(element)
        }
        return
      }
    }

// ....


        /** @param {VExpressionContainer} node */
        "VAttribute[directive=true][key.name.name='bind'][key.argument.name='class'] > VExpressionContainer.value"(
          node
        ) {
          if (!node.expression) {
            return
          }

          for (const { className, reportNode } of extractClassNames(
            /** @type {Expression} */ (node.expression)
          )) {
            reportForbiddenClasse(
              className,
              reportNode,
              context,
              forbiddenClasses
            )
          }
        },
```

ota-meshi (changes requested):
Thank you for this PR!
I wrote some comments.

docs/rules/no-restricted-class.md
taobojlen: OK, done. Should I also remove the "This rule was introduced in eslint-plugin-vue v7.19.0." text further down in this file?

docs/rules/no-restricted-class.md
taobojlen: Great suggestion -- fixed!

lib/rules/no-restricted-class.js
```diff
@@ -0,0 +1,154 @@
...
+    return
+  }
+  if (node.type === 'BinaryExpression') {
+    if (node.operator !== '+') {
+      return
+    }
+    yield* extractClassNames(node.left, true)
+    yield* extractClassNames(node.right, true)
+    return
+  }
```
ota-meshi, line 66: I made a mistake in the if statement. We need to add the following:

```suggestion
  }
  if (textOnly) {
    return
  }
```

`<template><div :class="''+{forbidden:true}">Content</div></template>` is actually drawn as `<div class="[object Object]">Content</div>`.

https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPjxkaXYgOmNsYXNzPVwiJycre2ZvcmJpZGRlbjp0cnVlfVwiPkNvbnRlbnQ8L2Rpdj48L3RlbXBsYXRlPiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vc2ZjLnZ1ZWpzLm9yZy92dWUucnVudGltZS5lc20tYnJvd3Nlci5qc1wiXG4gIH1cbn0ifQ==

lib/rules/no-restricted-class.js
```diff
@@ -0,0 +1,154 @@
...
+      if (element.type === 'SpreadElement') {
+        continue
+      }
+      yield* extractClassNames(element)
+    }
+    return
+  }
+  if (!textOnly) {
+    return
+  }
```
ota-meshi, line 96:
```suggestion
```

ota-meshi (changes requested):

docs/rules/no-restricted-class.md
ota-meshi: It will change automatically when you run `npm run update`. However, this is not necessary as it is also done when deploying the document site.

lib/rules/no-restricted-class.js
taobojlen: Thank you; fixed -- and added a test case.

lib/rules/no-restricted-class.js
taobojlen: Amended.

ota-meshi (approved):

PR #1703, opened 2021-11-11
Add `composition-vue2` style in `vue/component-api-style`

### LINKED ISSUE #1698: Add `composition-vue2` style in `vue/component-api-style`
[vue/component-api-style](https://eslint.vuejs.org/rules/component-api-style.html)

Fewer warnings

New option `composition-vue2`

```js
const MyComponent = Vue.extend({
	render(h) {
		return h('div');
	},
});
```

With configuration `"vue/component-api-style": ["error", ["composition"]]`, it warns about using the render function:

> Options API is not allowed in your project. `render` function is the API of Options API. Use Composition API instead.

With configuration `"vue/component-api-style": ["error", ["composition-vue2"]]`, it won't warn about using the render function, but still about using `computed`, `methods`, etc.

**Additional context**
Render functions cannot be replaced by returning `h(…)` from `setup` in Vue 2 with the Composition API plugin: https://github.com/vuejs/composition-api#template-refs

FloEdelmann:
Fixes #1698.

### DIFF
--- a/docs/rules/component-api-style.md
+++ b/docs/rules/component-api-style.md
@@ -73,14 +73,15 @@ export default {
 ```json
 {
   "vue/component-api-style": ["error",
-    ["script-setup", "composition"] // "script-setup", "composition", or "options"
+    ["script-setup", "composition"] // "script-setup", "composition", "composition-vue2", or "options"
   ]
 }
 ```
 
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #1707, opened 2021-11-12
Add `vue/no-child-content` rule

### LINKED ISSUE #1701: Rule proposal: vue/v-text-no-child-content

When using the `v-text` directive, the element's inner content gets overwritten by the result of the `v-text` expression.

[ ] Other (please specify:)

This might just belong to the layout category, not sure.

```vue
<template>
  <!-- ✓ GOOD -->
  <div v-text="foo"></div>

  <!-- ✗ BAD -->
  <div v-text="foo">{{ bar }}</div>
</template>
```

**Additional context**

I'm in the process of localizing an app, and stumbled upon a few elements which already used `v-text`, but *also* had child content:

```vue
<template>
  <label class="control-label col-sm-3" v-text="fields.name.label">
    Name
  </label>
</template>
```

This isn't really a problem, just unnecessary and confusing.

FloEdelmann:
Fixes #1701.

### DIFF
--- a/lib/rules/no-child-content.js
+++ b/lib/rules/no-child-content.js
@@ -21,7 +21,7 @@ module.exports = {
     docs: {
       description:
         "disallow element's child contents which would be overwritten by a directive like `v-html` or `v-text`",
-      categories: [],
+      categories: undefined,
       url: 'https://eslint.vuejs.org/rules/no-child-content.html'
     },
     fixable: null,
--- a/lib/rules/no-child-content.js
+++ b/lib/rules/no-child-content.js
@@ -1,0 +1,93 @@
+/**
+ * @author Flo Edelmann
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+const utils = require('../utils')
+
+/**
+ * @typedef {object} RuleOption
+ * @property {string[]} additionalDirectives
+ */
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    hasSuggestions: true,
+    type: 'problem',
+    docs: {
+      description:
+        "disallow element's child contents which would be overwritten by a directive like `v-html` or `v-text`",
+      categories: [],
+      url: 'https://eslint.vuejs.org/rules/no-child-content.html'
+    },
+    fixable: null,
+    schema: [
+      {
+        type: 'object',
+        additionalProperties: false,
+        properties: {
+          additionalDirectives: {
+            type: 'array',
+            uniqueItems: true,
+            minItems: 1,
+            items: {
+              type: 'string'
+            }
+          }
+        },
+        required: ['additionalDirectives']
+      }
+    ]
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    const directives = new Set(['html', 'text'])
+
+    /** @type {RuleOption | undefined} */
+    const option = context.options[0]
+    if (option !== undefined) {
+      for (const directive of option.additionalDirectives) {
+        directives.add(directive)
+      }
+    }
+
+    return utils.defineTemplateBodyVisitor(context, {
+      /** @param {VDirective} directiveNode */
+      'VAttribute[directive=true]'(directiveNode) {
+        const directiveName = directiveNode.key.name.name
+        const elementNode = directiveNode.parent.parent
+
+        if (directives.has(directiveName) && elementNode.children.length > 0) {
+          const firstChildNode = elementNode.children[0]
+          const lastChildNode =
+            elementNode.children[elementNode.children.length - 1]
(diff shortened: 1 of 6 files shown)


ota-meshi (changes requested):
Thank you for this PR!
Would you please add test cases that includes line breaks, indents, and comments in the content? I think these should be allowed.

tests/lib/rules/no-child-content.js
```diff
@@ -0,0 +1,172 @@
...
+          suggestions: [
+            { output: '<template><div v-html="foo"></div></template>' }
+          ]
+        }
+      ]
+    },
+    {
+      // v-html directive and whitespace-only text content
+      filename: 'test.vue',
+      code: '<template><div v-html="foo"> </div></template>',
```
ota-meshi, line 76: I just noticed this test case 😓.
I think it's better not to report in the case of whitespace or indentation. It's consistent with the vue compiler.


https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcblxuY29uc3QgZm9vID0gcmVmKCdIZWxsbyBXb3JsZCEnKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT48ZGl2IHYtaHRtbD1cImZvb1wiPiA8L2Rpdj48L3RlbXBsYXRlPiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vc2ZjLnZ1ZWpzLm9yZy92dWUucnVudGltZS5lc20tYnJvd3Nlci5qc1wiXG4gIH1cbn0ifQ==

ota-meshi:
I thought we should allow comments, but the vue compiler doesn't seem to allow comments. It may be better for the rule to report comments as well. What do you think?

https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcblxuY29uc3QgZm9vID0gcmVmKCdIZWxsbyBXb3JsZCEnKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT48ZGl2IHYtaHRtbD1cImZvb1wiPjwhLS1jb21tZW50LS0+PC9kaXY+PC90ZW1wbGF0ZT4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3NmYy52dWVqcy5vcmcvdnVlLnJ1bnRpbWUuZXNtLWJyb3dzZXIuanNcIlxuICB9XG59In0=

tests/lib/rules/no-child-content.js
FloEdelmann: That's a helpful website. Let's allow whitespace contents then, and report comments, to stay consistent with the Vue compiler. I'll add the relevant test cases.

lib/rules/no-child-content.js
```diff
@@ -0,0 +1,157 @@
...
+      /** @param {VDirective} directiveNode */
+      'VAttribute[directive=true]'(directiveNode) {
+        const directiveName = directiveNode.key.name.name
+        const elementNode = directiveNode.parent.parent
+
+        const elementComments = templateComments.filter((comment) =>
+          inRange(elementNode.range, comment)
+        )
+
+        const childNodes = [...elementNode.children, ...elementComments]
```
ota-meshi, line 129: I think you can use the TokenStore to extract all of the content tokens as follows:

```js
const tokenStore = context.parserServices.getTemplateBodyTokenStore()
const childContentTokens = tokenStore.getTokensBetween(elementNode.startTag, elementNode.endTag, { includeComments: true, filter:token => token.type !== 'HTMLWhitespace' })
```

ota-meshi (changes requested):

lib/rules/no-child-content.js
FloEdelmann: I simplified the retrieval of comments with this function. However, completely replacing the logic with the tokens returned by this function seemed more effort than it's worth (the suggested fix then doesn't remove whitespace e.g. in `{{ foo }}` anymore). See [`9566ee4`](https://github.com/vuejs/eslint-plugin-vue/pull/1707/commits/9566ee4a20fffdb13feca7cf7c5db58529971504).

ota-meshi: I think your method is better. thank you!

ota-meshi (approved):
LGTM! Thank you!

PR #1708, opened 2021-11-12
Docs: Add hints for rules that provide suggestions

FloEdelmann:
See a similar PR here: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1576


ota-meshi (approved):
Thank you very much!

PR #1716, opened 2021-11-17
Switch CI to GitHub Actions

ota-meshi:


ota-meshi:
Hmm.. I don't have permission to stop building CircleCI.

ota-meshi:
There's no reason to migrate, so let's continue using CircleCI.

PR #1736, opened 2021-12-03
Fix false positives for import binding in `vue/valid-define-emits` and `vue/valid-define-props` rules

### LINKED ISSUE #1734: `defineProps` should be able to work with local imported bindings, but reported as error
**Checklist**

- **ESLint version:** 8.3.0
- **eslint-plugin-vue version:** 8.1.1
- **Node version:** v16.11.0

```js
// http://eslint.org/docs/user-guide/configuring

module.exports = {
  root: true,
  env: {
    browser: true,
  },
  // https://github.com/feross/standard/blob/master/RULES.md#javascript-standard-style
  extends: ['plugin:vue/vue3-recommended', 'plugin:prettier/recommended'],
  plugins: ['html'],
  parserOptions: {
    ecmaVersion: 2020,
    ecmaFeatures: {
      jsx: true,
    },
  },
  // add your custom rules here
  rules: {
    'prettier/prettier': [
      'error',
      {
        endOfLine: 'auto',
        semi: true,
        singleQuote: true,
        trailingComma: 'es5',
      },
    ],
    // allow paren-less arrow functions
    'arrow-parens': 0,
    // allow async-await
    'generator-star-spacing': 0,
    // allow debugger during development
    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0,
    'vue/multi-word-component-names': 0,
    'vue/no-v-html': 0,
  },
};

```

**What did you do?**	

```vue
<script setup>
import { linkProps } from '../../props/link.props';

defineProps({
  ...linkProps,
  active: { type: Boolean, default: false },
});
</script>
```

No error should be reported. as the document https://v3.vuejs.org/api/sfc-script-setup.html#defineprops-and-defineemits said:

> However, it can reference imported bindings since they are in the module scope as well.

It does not require the import statement to inside a normal script instead of setup script.

**What actually happened?**

```
ESLint: `defineProps` are referencing locally declared variables.(vue/valid-define-props)
```

https://github.com/uiv-lib/uiv/blob/dev/src/components/button/Btn.vue

just change the import statement from normal script to setup script to see what happen.

ota-meshi:
close #1734

### DIFF
--- a/lib/rules/valid-define-emits.js
+++ b/lib/rules/valid-define-emits.js
@@ -77,11 +77,18 @@ module.exports = {
               ) {
                 if (
                   variable.defs.length &&
-                  variable.defs.every(
-                    (def) =>
-                      utils.inRange(scriptSetup.range, def.name) &&
-                      !utils.inRange(defineEmits.range, def.name)
-                  )
+                  variable.defs.every((def) => {
+                    if (def.type === 'ImportBinding') {
+                      return false
+                    }
+                    if (
+                      !utils.inRange(scriptSetup.range, def.name) ||
+                      utils.inRange(defineEmits.range, def.name)
+                    ) {
+                      return false
+                    }
+                    return true
+                  })
                 ) {
                   if (utils.withinTypeNode(node)) {
                     continue
--- a/lib/rules/valid-define-emits.js
+++ b/lib/rules/valid-define-emits.js
@@ -77,18 +77,12 @@ module.exports = {
               ) {
                 if (
                   variable.defs.length &&
-                  variable.defs.every((def) => {
-                    if (def.type === 'ImportBinding') {
-                      return false
-                    }
-                    if (
-                      !utils.inRange(scriptSetup.range, def.name) ||
-                      utils.inRange(defineEmits.range, def.name)
-                    ) {
-                      return false
-                    }
-                    return true
-                  })
+                  variable.defs.every(
+                    (def) =>
+                      def.type !== 'ImportBinding' &&
+                      utils.inRange(scriptSetup.range, def.name) &&
+                      !utils.inRange(defineEmits.range, def.name)
+                  )
                 ) {
                   if (utils.withinTypeNode(node)) {
                     continue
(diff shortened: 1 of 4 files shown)


lib/rules/valid-define-emits.js
```diff
@@ -77,11 +77,18 @@ module.exports = {
...
+                      return false
+                    }
+                    if (
+                      !utils.inRange(scriptSetup.range, def.name) ||
+                      utils.inRange(defineEmits.range, def.name)
+                    ) {
+                      return false
+                    }
+                    return true
+                  })
```
FloEdelmann, line 91: Wouldn't this be much more readable and also much more concise?

```suggestion
                  variable.defs.every(
                    (def) =>
                      def.type !== 'ImportBinding' &&
                      utils.inRange(scriptSetup.range, def.name) &&
                      !utils.inRange(defineEmits.range, def.name)
                  )
```

lib/rules/valid-define-props.js
```diff
@@ -78,11 +78,18 @@ module.exports = {
...
+                      return false
+                    }
+                    if (
+                      !utils.inRange(scriptSetup.range, def.name) ||
+                      utils.inRange(defineProps.range, def.name)
+                    ) {
+                      return false
+                    }
+                    return true
+                  })
```
FloEdelmann, line 92: And again:

```suggestion
                  variable.defs.every(
                    (def) =>
                      def.type !== 'ImportBinding' &&
                      utils.inRange(scriptSetup.range, def.name) &&
                      !utils.inRange(defineEmits.range, def.name)
                  )
```

FloEdelmann (changes requested):

ota-meshi:
Thank you for your review! I changed this PR.

FloEdelmann (approved):

baixiaoyu2997:
hi ,does this plugin now support  verify `defineEmits(['onSuccess'])` case?  
like :
```
defineEmits(['onSuccess'])  // bad
defineEmits(['on-success'])  // good
```

PR #1760, opened 2022-01-18
Add `vue/no-v-text-v-html-on-component` rule

### LINKED ISSUE #1724: disallow v-text/v-html on components
Components can have internal html structure that will be destroyed if you use `v-html` or `v-text` on them. Here's codepen example  

<https://codepen.io/jkarczm/pen/MWvRLYa> 

```html
    <v-btn>foo</v-btn>
    <v-btn v-text="'foo'" />
```

and the generated markup:

```html
    <button type="button" class="...">
      <span class="v-btn__content">foo</span>
    </button>
    <button type="button" class="...">
      foo
    </button>
```

As you see the second one misses the `v-btn__content` div, which in this particular case doesn't seem to break anything (EDIT: actually it does <https://github.com/vuetifyjs/vuetify/issues/11792#issuecomment-658767874>) but it could potentially

[ ] Other (please specify:)

```vue
<template>
  <!-- warn -->
  <some-component v-text="'content'" />
  <some-component v-html="'content'" />

  <!-- ok -->
  <div v-text="'content'" />

  <!-- not sure, should probably warn too, could be configurable -->
  <web-component v-text="'content'" />
```

**Additional context**

Initially I thought about a new rule but I think that it could be just a configuration option for `vue/no-v-text` and `vue/no-v-html` rules (`{ allowNativeTags: true }`)

ota-meshi:
This PR adds `vue/no-v-text-v-html-on-component` rule that disallows the use of v-text / v-html on component.

close #1724

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -349,6 +349,7 @@ For example:
 | [vue/no-use-computed-property-like-method](./no-use-computed-property-like-method.md) | disallow use computed property like method |  |
 | [vue/no-useless-mustaches](./no-useless-mustaches.md) | disallow unnecessary mustache interpolations | :wrench: |
 | [vue/no-useless-v-bind](./no-useless-v-bind.md) | disallow unnecessary `v-bind` directives | :wrench: |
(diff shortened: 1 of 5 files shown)


FloEdelmann (approved):
Looks good to me!

PR #1762, opened 2022-01-19
Add `vue/object-shorthand` rule

ota-meshi:
This PR adds `vue/object-shorthand` rule that applies `object-shorthand` rule to expressions in `<template>`.

### DIFF
--- a/docs/rules/object-shorthand.md
+++ b/docs/rules/object-shorthand.md
@@ -1,0 +1,27 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/object-shorthand
+description: require or disallow method and property shorthand syntax for object literals
+---
+# vue/object-shorthand
+
+> require or disallow method and property shorthand syntax for object literals
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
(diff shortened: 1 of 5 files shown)


docs/rules/object-shorthand.md
```diff
@@ -0,0 +1,27 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/object-shorthand
+description: require or disallow method and property shorthand syntax for object literals
```
FloEdelmann, line 5: Maybe to make this clear right from the short description?

```suggestion
description: require or disallow method and property shorthand syntax for object literals in the template
```

FloEdelmann (approved):
Looks good! Only one minor wording suggestion.

docs/rules/object-shorthand.md
ota-meshi: Your point is correct, but it's automatically generated, so it's difficult to fix now :sweat_smile:. We need to copy and replace the description of all extended rules.

FloEdelmann: Ahh I see. I didn't know that the descriptions of extended core rules are also copied. Does it make sense to always append " in `<template>`"? Because that's what the extended core rules always do?

FloEdelmann: Anyway, that should not block this PR. I'll create a new issue for that.

FloEdelmann: Done in #1764.

PR #1763, opened 2022-01-19
Add `vue/no-undef-components` rule and deprecate `vue/no-unregistered-components` rule 

### LINKED ISSUE #1693: Rule Proposal: no undef component in template when using <script setup>
Warn against components that appear in the template but not defined/imported in script-setup

[ ] Other (please specify:)

```vue
// Bad: OtherComponent not defined
<template>
  <div>
    <SomeComponent />
    <OtherComponent />
  </div>
</template>

<script setup>
import SomeComponent from '@/components/SomeComponent.vue'
</script>

//Good:
<template>
  <div>
    <SomeComponent />
    <OtherComponent />
  </div>
</template>

<script setup>
import SomeComponent from '@/components/SomeComponent.vue'
import OtherComponent from '@/components/OtherComponent.vue' // or: const OtherComponent = xxx.
</script>

//Also Good:
<template>
  <div>
    <some-component />
    <other-component />
  </div>
</template>

<script setup>
import SomeComponent from '@/components/SomeComponent.vue'
import OtherComponent from '@/components/OtherComponent.vue' // or: const OtherComponent = xxx.
</script>

```

**Additional context**
This is like the script-setup version of "no-unregistered components" rule

ota-meshi:
This PR adds `vue/no-undef-components` rule and deprecates `vue/no-unregistered-components` rule .
The `vue/no-undef-components` rule warns that the component used in the `<template>` is not defined.  

close #1693


docs/rules/no-undef-components-in-script-setup.md
```diff
@@ -0,0 +1,76 @@
...
+  <!-- ✗ BAD -->
+  <Bar />
+</template>
+```
+
+</eslint-code-block>
+
+## :couple: Related Rules
+
+- [vue/no-unregistered-components](./no-unregistered-components.md)
```
FloEdelmann, line 70: Why are these rules separate? They report the same issues, but only use different places to determine which components are registered/defined. If it's possible to merge them, I'd vote for that.

ota-meshi: I think the word "unregistered" is not correct because the script setup doesn't register the component. So I thought it should be a different rule.
But as you say, I think the rules are reporting the same issue. It may make more sense to mark `vue/no-unregistered-components` as deprecated and merge the new rule as `vue/no-undef-components`.
Do you think it's better to deprecate the old rule and create a new one?

FloEdelmann: Yeah, that sound fine.

ota-meshi: I will change this PR later :+1:

ota-meshi:
I changed this PR. So please check again.

docs/rules/no-undef-components.md
```diff
@@ -0,0 +1,183 @@
...
+
+> disallow use of undefined components in `<template>`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule warns that the component used in the `<template>` is not defined in `<script setup>`.  
+
+This rule reports components that haven't been defined and are being used in the `<template>`.
```
FloEdelmann, line 15:
```suggestion
This rule warns reports component that are used in the `<template>`, but that are not defined in the `<script setup>` or the Options API's `components` section.
```

FloEdelmann (approved):
Looks good to me. Just a docs wording suggestion.

docs/rules/no-undef-components.md
ota-meshi: Thank you!

lib/rules/no-unregistered-components.js
```diff
@@ -44,8 +44,10 @@ module.exports = {
         'disallow using components that are not registered inside templates',
       categories: null,
       recommended: false,
-      url: 'https://eslint.vuejs.org/rules/no-unregistered-components.html'
+      url: 'https://eslint.vuejs.org/rules/no-unregistered-components.html',
+      replacedBy: ['no-undef-components']
```
privatenumber, line 48: Looks like this `replacedBy` needs to be outside of the `doc` object:
https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

FloEdelmann: Thanks, good catch! I fixed the position in https://github.com/vuejs/eslint-plugin-vue/pull/1783.

PR #1769, opened 2022-01-21
Add `vue/quote-props` rule

ota-meshi:
This PR adds `vue/quote-props` rule that applies `quote-props` rule to expressions in `<template>`.


FloEdelmann (approved):
Looks good to me!

PR #1771, opened 2022-01-21
Fix false positives for `v-bind="obj"` with `v-model` in `vue/attributes-order` rule

### LINKED ISSUE #1767: vue/attributes-order doesn't order v-bind correct
**Checklist**

-  I have tried restarting my IDE and the issue persists.
-  I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

- **Vue version":** 3.2.27
- **ESLint version:** 8.7.0
- **eslint-plugin-vue version:**  8.3.0
- **Node version:** 16.13.0
- **Operating System:**  macOS Monterey 12.1

```json5
extends: [
        'eslint:recommended',
        'plugin:vue/vue3-essential',
        'plugin:vue/vue3-strongly-recommended',
        'plugin:vue/vue3-recommended',
],
rules: {
        'vue/attributes-order': 2,
}
```

**What did you do?**	

```vue
<base-menu
        v-model:placement="placement"
        v-bind="menuProps"
        :value="focused"
        :class="className"
        @click:outside="hideOptions"
    />
```

no warning on console

**What actually happened?**

```
Module Warning (from ./node_modules/vue-loader/dist/templateLoader.js):
(Emitted value instead of an instance of Error) (deprecation COMPILER_V_BIND_OBJECT_ORDER) v-bind="obj" usage is now order sensitive and behaves like JavaScript object spread: it will now overwrite an existing non-mergeable attribute that appears before v-bind in the case of conflict. To retain 2.x behavior, move v-bind to make it the first attribute. You can also suppress this warning if the usage is intended.
  Details: https://v3.vuejs.org/guide/migration/v-bind.html
2  |      <base-menu
3  |          v-model:placement="placement"
4  |          v-bind="menuProps"
   |          ^^^^^^^^^^^^^^^^^^
5  |          :value="focused"
6  |          :class="className"

```

**What did you do?**	

```vue
<base-menu
        v-bind="menuProps"
        v-model:placement="placement"
        :value="focused"
        :class="className"
        @click:outside="hideOptions"
    />
```

no warning on console

**What actually happened?**

```
ERROR in component.vue:4:9
vue/attributes-order: Attribute "v-model:placement" should go before "v-bind".
    2 |     <base-menu
    3 |         v-bind="menuProps"
    4 |         v-model:placement="placement"

```

ota-meshi:
fixes #1767

### DIFF
--- a/lib/rules/attributes-order.js
+++ b/lib/rules/attributes-order.js
@@ -340,7 +340,7 @@ function create(context) {
       if (isVBindObject(node)) {
         // node is `v-bind ="object"` syntax
 
-        // In Vue 3, if change the order of `v-bind:foo="..."`, `v-model="..."` and `v-bind ="object"`,
+        // In Vue 3, if change the order of `v-bind:foo="..."`, `v-model="..."` and `v-bind="object"`,
         // the behavior will be different, so adjust so that there is no change in behavior.
 
         const len = attributes.length
--- a/lib/rules/attributes-order.js
+++ b/lib/rules/attributes-order.js
@@ -309,9 +309,9 @@ function create(context) {
         (isVAttributeOrVBindOrVModel(attributes[index - 1]) ||
           isVAttributeOrVBindOrVModel(attributes[index + 1]))
       ) {
-        // In Vue 3, ignore `v-bind ="object"`, which is
-        // a pair of `v-bind:foo="..."` and `v-bind ="object"` and
-        // a pair of `v-model="..."` and `v-bind ="object"`,
+        // In Vue 3, ignore `v-bind="object"`, which is
+        // a pair of `v-bind:foo="..."` and `v-bind="object"` and
+        // a pair of `v-model="..."` and `v-bind="object"`,
         // because changing the order behaves differently.
         return false
       }
--- a/lib/rules/attributes-order.js
+++ b/lib/rules/attributes-order.js
@@ -37,6 +37,14 @@ function isVBind(node) {
   return Boolean(node && node.directive && node.key.name.name === 'bind')
 }
 /**
+ * Check whether the given attribute is `v-model` directive.
+ * @param {VAttribute | VDirective | undefined | null} node
+ * @returns { node is VDirective }
+ */
+function isVModel(node) {
+  return Boolean(node && node.directive && node.key.name.name === 'model')
+}
+/**
  * Check whether the given attribute is plain attribute.
  * @param {VAttribute | VDirective | undefined | null} node
  * @returns { node is VAttribute }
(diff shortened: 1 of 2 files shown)


lib/rules/attributes-order.js
```diff
@@ -298,11 +306,13 @@ function create(context) {
...
-        (isVAttributeOrVBind(attributes[index - 1]) ||
-          isVAttributeOrVBind(attributes[index + 1]))
+        (isVAttributeOrVBindOrVModel(attributes[index - 1]) ||
+          isVAttributeOrVBindOrVModel(attributes[index + 1]))
       ) {
-        // In Vue 3, ignore the `v-bind:foo=" ... "` and `v-bind ="object"` syntax
-        // as they behave differently if you change the order.
+        // In Vue 3, ignore `v-bind ="object"`, which is
+        // a pair of `v-bind:foo="..."` and `v-bind ="object"` and
+        // a pair of `v-model="..."` and `v-bind ="object"`,
```
FloEdelmann, line 314:
```suggestion
        // In Vue 3, ignore `v-bind="object"`, which is
        // a pair of `v-bind:foo="..."` and `v-bind="object"` and
        // a pair of `v-model="..."` and `v-bind="object"`,
```

lib/rules/attributes-order.js
```diff
@@ -330,14 +340,14 @@ function create(context) {
       if (isVBindObject(node)) {
         // node is `v-bind ="object"` syntax
 
-        // In Vue 3, if change the order of `v-bind:foo=" ... "` and `v-bind ="object"`,
+        // In Vue 3, if change the order of `v-bind:foo="..."`, `v-model="..."` and `v-bind ="object"`,
```
FloEdelmann, line 343:
```suggestion
        // In Vue 3, if change the order of `v-bind:foo="..."`, `v-model="..."` and `v-bind="object"`,
```

tests/lib/rules/attributes-order.js
```diff
@@ -1246,6 +1282,50 @@ tester.run('attributes-order', rule, {
...
+      <template>
+        <div
+          v-if="x"
+          v-bind="b"
+          v-bind:id="a"
+          v-model="c">
+        </div>
+      </template>`,
+      errors: ['Attribute "v-bind:id" should go before "v-model".']
+    },
```
FloEdelmann, line 1328: Will this create false positives in existing Vue 2 code? There, the order `v-if, v-model, v-bind, v-bind:id` would be perfectly fine.

FloEdelmann (changes requested):

lib/rules/attributes-order.js
ota-meshi: Thank you!

tests/lib/rules/attributes-order.js
ota-meshi: I don't think there are any false positives because the `id` must be before the `v-model`.

https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended

FloEdelmann: Oh, indeed.

FloEdelmann: Oh, indeed.

FloEdelmann (approved):

PR #1783, opened 2022-01-27
Meta: Fix `replacedBy` position

FloEdelmann:
As noticed by @privatenumber in https://github.com/vuejs/eslint-plugin-vue/pull/1763#discussion_r793596573.


ota-meshi (approved):
@FloEdelmann @privatenumber Thank you!

PR #1792, opened 2022-02-05
Add `no-restricted-imports` rule

### LINKED ISSUE #1780: Rule Proposal: disallow imports from @vue/* dependencies

Disallow or warn about imports from `@vue/*` dependencies. Particularly when using Vue CLI and the build target is a library. 

[ ] Other (please specify:)

```vue
<!-- Fail or Warn -->
import { computed } from "@vue/reactivity";

import {
  something,
  another,
} from "@vue/compiler-dom";
```

```vue
<!-- Pass -->
```

**Additional context**

Full context can be found in [this issue I opened in the core repo](https://github.com/vuejs/core/issues/5327). 

In short, using the `@vue/reactivity` import broke the components I was building for my UI library with Vue CLI and they were failing silently.

jbhammon:
This PR adds a `no-restricted-imports` rule to check for incorrect imports from some `'@vue'` packages. The new rule will tell the user that they should import those APIs directly from `'vue'` instead.

I created some trivial tests to help me check that I extended the existing eslint rule correctly, and they're not all passing so I must have done something wrong. Any advice from maintainers would be appreciated, and I'll happily make updates to get this working. 

Resolves #1780

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -71,6 +71,18 @@ module.exports = {
     'no-proto': 2,
     'no-redeclare': 2,
     'no-regex-spaces': 2,
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          '@vue/reactivity',
+          '@vue/runtime-core',
+          '@vue/runtime-dom',
+          '@vue/shared'
+        ],
+        message: "Please always import these APIs from the main 'vue' package"
+      }
+    ],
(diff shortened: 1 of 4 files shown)


FloEdelmann (changes requested):
See https://github.com/vuejs/eslint-plugin-vue/issues/1780#issuecomment-1030956325.

PR #1796, opened 2022-02-09
Add `prefer-true-attribute-shorthand` rule

### LINKED ISSUE #775: Rule Proposal: Boolean prop shorthand (`prop` instead of `:prop="true"`)

### LINKED ISSUE #1781: New rule: prefer-true-attribute-shorthand

In template, if there's an attribute whose value is `true`, such as `:show="true"`, this rule will warn and provide auto fix (becomes `show`, see the example below). This can reduce verbosity.

Invalid:

```vue
<template>
  <my-dialog :show="true" />
</template>
```

which can be fixed to:

```vue
<template>
  <my-dialog show />
</template>
```

g-plane:
Closes #775 
Closes #1781

### DIFF
--- a/docs/rules/prefer-true-attribute-shorthand.md
+++ b/docs/rules/prefer-true-attribute-shorthand.md
@@ -9,14 +9,13 @@ description: require shorthand form attribute when `v-bind` value is `true`
 > require shorthand form attribute when `v-bind` value is `true`
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
-- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 - :bulb: Some problems reported by this rule are manually fixable by editor [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
 
 ## :book: Rule Details
 
 `v-bind` attribute with `true` value usually can be written in shorthand form. This can reduce verbosity.
 
-<eslint-code-block fix :rules="{'vue/prefer-true-attribute-shorthand': ['error']}">
+<eslint-code-block :rules="{'vue/prefer-true-attribute-shorthand': ['error']}">
 
 ```vue
 <template>
@@ -90,7 +89,7 @@ Default options is `"always"`.
 
 ### `"never"`
 
-<eslint-code-block fix :rules="{'vue/prefer-true-attribute-shorthand': ['error', 'never']}">
+<eslint-code-block :rules="{'vue/prefer-true-attribute-shorthand': ['error', 'never']}">
 
 ```vue
 <template>
--- a/docs/rules/prefer-true-attribute-shorthand.md
+++ b/docs/rules/prefer-true-attribute-shorthand.md
@@ -1,0 +1,68 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/prefer-true-attribute-shorthand
+description: require shorthand form attribute when `v-bind` value is `true`.
+---
+# vue/prefer-true-attribute-shorthand
+
+> require shorthand form attribute when `v-bind` value is `true`.
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :bulb: Some problems reported by this rule are manually fixable by editor [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+## :book: Rule Details
+
+`v-bind` attribute with `true` value usually can be written in shorthand form. This can reduce verbosity.
+
+<eslint-code-block :rules="{'vue/prefer-true-attribute-shorthand': ['error']}">
+
+```vue
+<template>
+  <!-- ✗ BAD -->
+  <MyComponent v-bind:show="true" />
+  <MyComponent :show="true" />
+
+  <!-- ✓ GOOD -->
+  <MyComponent show />
+  <MyComponent another-prop="true" />
+</template>
+```
+
+</eslint-code-block>
+
+However, those two representations are not always equivalent.
+This case will be occurred if the definition of a prop include `String`:
+
+```vue
+<template>
+  <pre>{{ prop }}</pre>
+</template>
+
+<script>
+export default {
+  name: 'MyComponent',
+  props: {
+    prop: [String, Boolean]
+  }
+}
+</script>
+```
+
+```vue
+<template>
+  <MyComponent prop />
+  <MyComponent :prop="true"/>
+</template>
+```
+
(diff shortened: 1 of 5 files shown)


ota-meshi (approved):
LGTM! Thank you!

lib/rules/prefer-true-attribute-shorthand.js
```diff
@@ -0,0 +1,69 @@
...
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'require shorthand form attribute when `v-bind` value is `true`.',
```
FloEdelmann, line 22: To stay consistent with other descriptions:

```suggestion
        'require shorthand form attribute when `v-bind` value is `true`',
```

lib/rules/prefer-true-attribute-shorthand.js
```diff
@@ -0,0 +1,69 @@
...
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
```
FloEdelmann, line 19: It's only a stylistic/conventional choice, not a problem.

```suggestion
    type: 'suggestion',
```

lib/rules/prefer-true-attribute-shorthand.js
```diff
@@ -0,0 +1,69 @@
...
+      report:
+        "Boolean prop with 'true' value can be written in shorthand form.",
+      fix: 'Rewrite this prop into shorthand form.'
+    }
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    // ...
+
+    return utils.defineTemplateBodyVisitor(context, {
```
FloEdelmann, line 39: Please remove the boilerplate comment and empty line:

```suggestion
    return utils.defineTemplateBodyVisitor(context, {
```

docs/rules/prefer-true-attribute-shorthand.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+```vue
+<template>
+  <MyComponent prop />
+  <MyComponent :prop="true"/>
+</template>
+```
+
+Those two calls will introduce different render result. See [this demo](https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBNeUNvbXBvbmVudCBmcm9tICcuL015Q29tcG9uZW50LnZ1ZSdcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxNeUNvbXBvbmVudCBhIGIvPlxuICA8TXlDb21wb25lbnQgOmE9XCJ0cnVlXCIgOmI9XCJ0cnVlXCIvPlxuPC90ZW1wbGF0ZT4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3NmYy52dWVqcy5vcmcvdnVlLnJ1bnRpbWUuZXNtLWJyb3dzZXIuanNcIlxuICB9XG59IiwiTXlDb21wb25lbnQudnVlIjoiPHNjcmlwdD5cbmV4cG9ydCBkZWZhdWx0IHtcbiAgcHJvcHM6IHtcbiAgICBhOiBCb29sZWFuLFxuICAgIGI6IFtTdHJpbmcsIEJvb2xlYW5dXG4gIH1cbn1cbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxwcmU+XG5hOiB7e2F9fVxuYjoge3tifX1cbiAgPC9wcmU+XG48L3RlbXBsYXRlPiJ9).
```
FloEdelmann, line 59: The order matters, see https://github.com/vuejs/eslint-plugin-vue/issues/1781#issuecomment-1022735437.

In this suggestion, I adapted the explanation and example (also the link) and added a warning box:

````suggestion
:::warning Warning
The shorthand form is not always equivalent! If a prop accepts multiple types, but Boolean is not the first one, a shorthand prop won't pass `true`.
:::

```vue
<script>
export default {
  name: 'MyComponent',
  props: {
    bool: Boolean,
    boolOrString: [Boolean, String],
    stringOrBool: [String, Boolean],
  }
}
</script>
```

**Shorthand form:**

```vue
<MyComponent bool bool-or-string string-or-bool />
```

```
bool: true (boolean)
boolOrString: true (boolean)
stringOrBool: "" (string)
```

**Longhand form:**

```vue
<MyComponent :bool="true" :bool-or-string="true" :string-or-bool="true" />
```

```
bool: true (boolean)
boolOrString: true (boolean)
stringOrBool: true (boolean)
```

Those two calls will introduce different render result. See [this demo](https://sfc.vuejs.org/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBNeUNvbXBvbmVudCBmcm9tICcuL015Q29tcG9uZW50LnZ1ZSdcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIFNob3J0aGFuZCBmb3JtOlxuICA8TXlDb21wb25lbnQgYm9vbCBib29sLW9yLXN0cmluZyBzdHJpbmctb3ItYm9vbCAvPlxuICBcbiAgTG9uZ2hhbmQgZm9ybTpcbiAgPE15Q29tcG9uZW50IDpib29sPVwidHJ1ZVwiIDpib29sLW9yLXN0cmluZz1cInRydWVcIiA6c3RyaW5nLW9yLWJvb2w9XCJ0cnVlXCIgLz5cbjwvdGVtcGxhdGU+IiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9zZmMudnVlanMub3JnL3Z1ZS5ydW50aW1lLmVzbS1icm93c2VyLmpzXCJcbiAgfVxufSIsIk15Q29tcG9uZW50LnZ1ZSI6IjxzY3JpcHQ+XG5leHBvcnQgZGVmYXVsdCB7XG4gIHByb3BzOiB7XG4gICAgYm9vbDogQm9vbGVhbixcbiAgICBib29sT3JTdHJpbmc6IFtCb29sZWFuLCBTdHJpbmddLFxuICAgIHN0cmluZ09yQm9vbDogW1N0cmluZywgQm9vbGVhbl0sXG4gIH1cbn1cbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxwcmU+XG5ib29sOiB7e2Jvb2x9fSAoe3sgdHlwZW9mIGJvb2wgfX0pXG5ib29sT3JTdHJpbmc6IHt7Ym9vbE9yU3RyaW5nfX0gKHt7IHR5cGVvZiBib29sT3JTdHJpbmcgfX0pXG5zdHJpbmdPckJvb2w6IHt7c3RyaW5nT3JCb29sfX0gKHt7IHR5cGVvZiBzdHJpbmdPckJvb2wgfX0pXG4gIDwvcHJlPlxuPC90ZW1wbGF0ZT4ifQ==).
````

FloEdelmann (changes requested):
Thanks for implementing this! However, I have a few remarks.

In https://github.com/vuejs/eslint-plugin-vue/issues/775#issuecomment-454081818, it was suggested to make it configurable whether to always or never use shorthand props. I think that's a good idea, given some people may prefer to never use shorthand props because of their quirks or just by taste.

The `never` case could also be auto-fixable, while the suggestion you implemented would be fine for the `always` case.

FloEdelmann (changes requested):
Nice, this looks really good!

Only one problem I just noticed: The `never` option will also report native HTML shorthand attributes: `<input checked>` should become `<input checked="checked">`, not `<input :checked="true">`. Or is this actually allowed?

If it's allowed, please also add this test case to the "allowed" list to document this behavior.

If it isn't, we need a little more logic: Since one can use native HTML shorthand attributes like `hidden` also on components (e.g. `<MyComponent hidden>`), I think we also have to use suggestions instead of an autofix there: One for changing it into a longhand Vue prop, one for changing it into a longhand HTML attribute.

We might be a bit more clever about this and actually do autofix it in one case though: The shorthand attribute is on a known HTML element. Then we know that it should be autofixed to a longhand HTML attribute. But I think this is a nice-to-have improvement; only having suggestions in all cases would be fine for me.

FloEdelmann:
See also https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html

g-plane:
What about skipping native HTML elements, and just check components?

FloEdelmann:
That'd be fine for me, but still not catch the `<MyComponent hidden>` case. `hidden` might be a prop or an HTML attribute.

FloEdelmann (approved):
Nice, thanks for the improvements and the test cases! Please run `npm run update` one more time to update the docs (there is no autofix anymore). But this is not merge-blocking; we can do it afterwards.

FloEdelmann (approved):

PR #1799, opened 2022-02-12
Add Reactivity Transform macros

### LINKED ISSUE #1798: Reactivity Transform macros missing

- **ESLint version:** 8.8.0
- **eslint-plugin-vue version:** 8.4.1
- **Node version:** 16.14.0/17.5.0

The new [Reactivity Transform](https://vuejs.org/guide/extras/reactivity-transform.html) macros are missing from `vue/setup-compiler-macros` environment option. I'd like to add them since they are available globally when using `script setup`.

Explicitly importing the macros as mentioned just above [this section header](https://vuejs.org/guide/extras/reactivity-transform.html#destructuring-with). 👇

image

**Additional context**

Reactivity Transform is still an experimental feature, but I still think it would be nice to include in this plugin for those who want to try it out.

AndreasNasman:
Closes #1798

---

I noticed that the new [Reactivity Transform](https://vuejs.org/guide/extras/reactivity-transform.html) macros were missing in the `vue/setup-compiler-macros` option, resulting in e.g. `'$ref' is not defined  no-undef` errors when not explicitly importing the macros.

I hope this implementation is correct. I looked at #1685 as an example.

### DIFF
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -358,10 +358,10 @@ module.exports = {
 }
 ```
 
-#### Compiler macros such as `defineProps` and `defineEmits` generate `no-undef` warnings
+#### Compiler macros such as `defineProps` and `$ref` generate `no-undef` warnings
 
 You need to enable the compiler macros environment in your ESLint configuration file.
-If you don't want to expose these variables globally, you can use `/* global defineProps, defineEmits */` instead.
+If you don't want to expose these variables globally, you can use `/* global defineProps, $ref */` instead.
 
 Example **.eslintrc.js**:
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -233,9 +233,16 @@ module.exports = {
   environments: {
     'setup-compiler-macros': {
       globals: {
-        defineProps: 'readonly',
+        $$: 'readonly',
+        $: 'readonly',
+        $computed: 'readonly',
+        $customRef: 'readonly',
+        $ref: 'readonly',
+        $shallowRef: 'readonly',
+        $toRef: 'readonly',
         defineEmits: 'readonly',
         defineExpose: 'readonly',
+        defineProps: 'readonly',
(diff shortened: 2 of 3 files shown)


ota-meshi:
Thank you for opening this PR.
However, this package does not support experimental features.
Because if this package tries to support experimental features, it will have to do a major version upgrade every time the RFC changes such as rename `$ref`. Because it's a breaking change for this package.

After the Reactivity Transform RFCs have been merged, we will work to support it.

AndreasNasman:
Ok, thanks for the clarification!

martinszeltins:
> Thank you for opening this PR. However, this package does not support experimental features. Because if this package tries to support experimental features, it will have to do a major version upgrade every time the RFC changes such as rename `$ref`. Because it's a breaking change for this package.
> 
> After the Reactivity Transform RFCs have been merged, we will work to support it.

@ota-meshi But I am getting an error `ESLint: 'id' is never reassigned. Use 'const' instead.(prefer-const)`

I don't want to turn this rule off completely, what can I do?

FloEdelmann:
This is a different ESLint rule: https://eslint.org/docs/rules/prefer-const

Please use `const id = …` instead of `let id = …` or `var id = …`.

FloEdelmann:
At least in your example, both `let` and `const` would work in the compiled output, since only the `.value` part is mutable and the variable itself is constant. So it doesn't really matter what you write in the input, does it? So you could simply use `const` whenever the `prefer-const` rule suggests to do so, and `let` otherwise`.

Modifying the core ESLint `prefer-const` rule to ignore Vue-specific reactivity macros will likely not happen. So the only option is to disable that rule and maybe ([when reactivity macros are not experimental anymore](https://github.com/vuejs/eslint-plugin-vue/pull/1799#issuecomment-1038582067)) add a new rule in eslint-plugin-vue that wraps the core rule but ignores reactivity macros.

PR #1804, opened 2022-02-19
Add `vue/prefer-import-from-vue` rule

### LINKED ISSUE #1780: Rule Proposal: disallow imports from @vue/* dependencies

ota-meshi:
This PR adds `vue/prefer-import-from-vue` rule.
This rule aims to use imports from `'vue'` instead of imports from `'@vue/*'`.

close #1780 
close #1792


ota-meshi:
Thank you for your review!

> Is there any good way to add tests for the update script?

I don't know another way to get Vue3's export name list, so I find it difficult to add a test 😅

FloEdelmann:
I thought of using mock data, i.e. testing that the right kinds of exports are found. But maybe that is too complicated and/or not worth it?

ota-meshi:
Hmm. I don't think we can get the correct names by using mock because the type definition information is not exported in the actual JavaScript.
It may be possible to test for less than the names exported by JS, but it may not be worth it.

FloEdelmann:
Then let's merge this now, and see if this causes any bugs in the future. We can still try to add regression tests then.

PR #1811, opened 2022-03-02
Auto-fix for vue/component-tags-order

### LINKED ISSUE #1463: vue/component-tags-order - Could it be made fixable?

lsdsjy:
Fixes #1463.
This PR reflects the comments from @ota-meshi in #1512 . It yields a fix for each reported error, so eslint-disable comments work normally. Added a test case for eslint-disable comment.

### DIFF
--- a/tests/lib/rules/component-tags-order.js
+++ b/tests/lib/rules/component-tags-order.js
@@ -357,7 +357,9 @@ describe('suppress reporting with eslint-disable-next-line', () => {
     const code = `<style></style><template></template>
     <!-- eslint-disable-next-line vue/component-tags-order -->
     <script></script>`
-    const [{ messages, output }] = await eslint.lintText(code, { filePath: 'test.vue' })
+    const [{ messages, output }] = await eslint.lintText(code, {
+      filePath: 'test.vue'
+    })
     assert.strictEqual(messages.length, 0)
     // should not fix <script>
     assert.strictEqual(
--- a/tests/lib/rules/component-tags-order.js
+++ b/tests/lib/rules/component-tags-order.js
@@ -9,6 +9,25 @@
 
 const rule = require('../../../lib/rules/component-tags-order')
 const RuleTester = require('eslint').RuleTester
+const assert = require('assert')
+const { ESLint } = require('../../eslint-compat')
+
+// Initialize linter.
+const eslint = new ESLint({
+  overrideConfig: {
+    parser: require.resolve('vue-eslint-parser'),
+    parserOptions: {
+      ecmaVersion: 2015
+    },
+    plugins: ['vue'],
+    rules: {
+      'vue/comment-directive': 'error',
+      'vue/component-tags-order': 'error'
+    }
+  },
+  useEslintrc: false,
+  plugins: { vue: require('../../../lib/index') }
+})
 
 // ------------------------------------------------------------------------------
 // Tests
@@ -108,7 +127,8 @@ tester.run('component-tags-order', rule, {
           line: 1,
           column: 37
         }
-      ]
+      ],
+      output: '<template></template><style></style><script></script>'
     },
     {
       code: '<template></template><script></script><style></style>',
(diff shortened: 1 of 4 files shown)


tests/lib/rules/component-tags-order.js
```diff
@@ -268,7 +345,23 @@ tester.run('component-tags-order', rule, {
...
         }
-      ]
+      ],
+      output: '\n        <script></script>\n        <style></style>\n      '
     }
   ]
 })
+
+describe('suppress reporting with eslint-disable-next-line', () => {
+  it('disable all rules if <!-- eslint-disable-line -->', async () => {
```
FloEdelmann, line 355: The test name should probably be adjusted.

FloEdelmann (approved):
Thanks for implementing this!

Looks fine from my side. I have just one minor suggestion.

tests/lib/rules/component-tags-order.js
lsdsjy: Nice catch! It was copied from another test and I forgot to modify the test description.

lsdsjy: Updated.

FloEdelmann (approved):

lsdsjy:
Wait, there's something I can't determine here. If there is a comment right above the element start tag, should we move the comment too? For example,
```vue
<style></style>
<!-- this is a comment -->
<script>
</script>
```

Should we move the comment together with the `<script>` element? What if the comment is a eslint-disable comment, like

```vue
<style></style>
<!-- eslint-disable-next-line vue/multi-word-component-names -->
<script>
export default {
  name: 'Comp'
}
</script>
```

@FloEdelmann @ota-meshi Would you have a look at this plz?

FloEdelmann:
Moving a comment together with the following block would be nice, regardless of whether it is an eslint-disable comment or not. However, I don't think that this would be merge-blocking for me.

tests/lib/rules/component-tags-order.js
```diff
@@ -268,7 +345,23 @@ tester.run('component-tags-order', rule, {
...
+    const code = `<style></style><template></template>
+    <!-- eslint-disable-next-line vue/component-tags-order -->
+    <script></script>`
+    const [{ messages }] = await eslint.lintText(code, { filePath: 'test.vue' })
+    assert.strictEqual(messages.length, 1)
+    // should not fix <script>
+    assert.strictEqual(
+      messages[0].fix.text,
+      '<template></template><style></style>'
+    )
```
ota-meshi, line 365: I think it's better to check the fixed text using the `output` provided by the ESLint class. To use the `output`, add `fix: true` to the ESLint constructor option.

```suggestion
    const [{ messages, output }] = await eslint.lintText(code, { filePath: 'test.vue' })
    assert.strictEqual(messages.length, 0)
    // should not fix <script>
    assert.strictEqual(
      output,
      `<template></template><style></style>
    <!-- eslint-disable-next-line vue/component-tags-order -->
    <script></script>`
    )
```

tests/lib/rules/component-tags-order.js
```diff
@@ -9,6 +9,25 @@
...
+      ecmaVersion: 2015
+    },
+    plugins: ['vue'],
+    rules: {
+      'vue/comment-directive': 'error',
+      'vue/component-tags-order': 'error'
+    }
+  },
+  useEslintrc: false,
+  plugins: { vue: require('../../../lib/index') }
```
ota-meshi, line 29:
```suggestion
  plugins: { vue: require('../../../lib/index') },
  fix: true
```

ota-meshi (changes requested):
Thank you for this PR! I have a change request for testcase.

ota-meshi (approved):
LGTM! Thank you!

PR #1824, opened 2022-03-23
first-attribute-linebreak is not included in plugin:vue/recommended

Marcosaurios:
See [eslint rules](https://eslint.vuejs.org/rules/#priority-c-recommended-minimizing-arbitrary-choices-and-cognitive-overhead-for-vue-js-2-x) and [the repo search](https://github.com/vuejs/eslint-plugin-vue/search?q=first-attribute-linebreak). This attribute is not included in vue/recommended.

Also when [this rule was added](https://github.com/vuejs/eslint-plugin-vue/pull/1587), wasn't suggested to be part of the vue/recommended

### DIFF
--- a/docs/rules/first-attribute-linebreak.md
+++ b/docs/rules/first-attribute-linebreak.md
@@ -9,7 +9,7 @@ since: v8.0.0
 
 > enforce the location of first attribute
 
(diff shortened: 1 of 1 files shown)


FloEdelmann:
It is part of the `strongly-recommended` config though. And the `recommended` config extends the `strongly-recommended` config, so the rule is part of the `recommended` too.

FloEdelmann:
This part of the docs is auto-generated by https://github.com/vuejs/eslint-plugin-vue/blob/master/tools/update-docs.js

So if there would a mistake in the docs, it should be fixed in the update script anyway.

PR #1825, opened 2022-03-24
Fix #1819: Enforce order between script and script setup

### LINKED ISSUE #1819: `vue/component-tags-order` should enforce order between `<script>` and `<script setup>`
vue/component-tags-order

More.

New configuration option. 

Possibly the new default?

**Example**

```vue
<script setup>
  // ...
</script>

<script>
  // ...
</script>
```

Nothing.

Move `<script setup>` below `<script>`.

```vue
<script>
  // ...
</script>

<script setup>
  // ...
</script>
```

doug-wade:
Fixes #1819 . Adds a new tag type to the component-tags-order, script/setup.


ota-meshi (changes requested):
Thank you for this PR!
I wrote comments. I also want @FloEdelmann 's opinion on the option format.

docs/rules/component-tags-order.md
```diff
@@ -14,27 +14,28 @@ since: v6.1.0
...
-This rule warns about the order of the `<script>`, `<template>` & `<style>` tags.
+This rule warns about the order of the `<script>`, `<script setup>`, `<template>` & `<style>` tags.
 
 ## :wrench: Options
 
 ```json
 {
   "vue/component-tags-order": ["error", {
-    "order": [ [ "script", "template" ], "style" ]
+    "order": [ [ "script", "script/setup", "template" ], "style" ]
```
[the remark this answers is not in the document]
ota-meshi: Perhaps if it can support CSS selectors (eg `[arrt]`, `[attr=val]`, `:not()`), I think it's easier for users to learn, but I think implementing is difficult.

```js
"vue/component-tags-order": {
  "order": [
     "script:not([setup])",
     "script[setup]",
     "template",
     "style:not([scoped])",
     "style[scoped]",
     "i18n[locale=en]",
     "i18n[locale=ja]",
  ]
}
```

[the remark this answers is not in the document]
ota-meshi: Hmm. We may not be able to use esquery.
The esquery attribute selector seems to be very different from the CSS selector.

FloEdelmann:
I also think that existing options should continue to work, i.e. `"script"` should report both `<script>` and `<script setup>`.

I really like the CSS selector format suggestion by @ota-meshi. It's really clear and concise, and allows configuring e.g. `<style scoped>` above `<style>`, or any other individual use cases.

The object format is too noisy for me.

ota-meshi:
@FloEdelmann Thank you for your opinion.

> I really like the CSS selector format suggestion by @ota-meshi. It's really clear and concise, and allows configuring e.g. <style scoped> above <style>, or any other individual use cases.
> The object format is too noisy for me.

I agree with you :+1:

@doug-wade Can you change this PR so that user can specify a CSS selector as an option?
I think it's a good to add postcss-selector-parser to the dependency and use it to parse the selector.
https://github.com/postcss/postcss-selector-parser

If it's difficult to parse the CSS selector, could you add `"script:not([setup])"` and `"script[setup]"` as constant options?
I change the rule to be able to use CSS selectors when I have time.
I've used postcss-selector-parser in another project so I can implement it.

lib/rules/component-tags-order.js
```diff
@@ -70,11 +67,74 @@ module.exports = {
...
+     * @return {Boolean} true if the element matches the selector, false otherwise
+     */
+    function matches(ordering, element) {
+      let hasMatch = true
+      let isNegated = false
+
+      parser((selectors) => {
+        selectors.walk((selector) => {
+          switch (selector.type) {
+            case 'tag':
```
doug-wade, line 101: This is in no way a correct implementation of css selectors, but I think it supports all the features we'll need to get this off the ground.

doug-wade: I was wrong, this is busted. I'll get back at it soon; sorry for pushing the re-review button early

lib/rules/component-tags-order.js
```diff
@@ -43,11 +44,7 @@ module.exports = {
         },
         additionalProperties: false
       }
-    ],
-    messages: {
-      unexpected:
-        'The <{{name}}> should be above the <{{firstUnorderedName}}> on line {{line}}.'
-    }
```
ota-meshi, line 50: Could you revert to using messageId?

docs/rules/component-tags-order.md
```diff
@@ -14,29 +14,36 @@ since: v6.1.0
...
 <!-- ✓ GOOD -->
-<script>/* ... */</script>
+<script>
+/* ... */
+</script>
 <template>...</template>
-<style>/* ... */</style>
+<style>
+/* ... */
+</style>
```
FloEdelmann, line 46: Please revert these code style changes in the docs. The more concise form makes it easier to see the effects of this rule.

docs/rules/component-tags-order.md
```diff
@@ -56,59 +67,145 @@ This rule warns about the order of the `<script>`, `<template>` & `<style>` tags
...
+</script>
 <template>...</template>
 ```
 
 </eslint-code-block>
 
 ### `{ "order": ["template", "script", "style"] }`
 
-<eslint-code-block fix :rules="{'vue/component-tags-order': ['error', { 'order': ['template', 'script', 'style'] }]}">
+<eslint-code-block fix :rules="{'vue/component-tags-order': ['error', { 'order': ['template', 'script', 'script[setup]', 'style'] }]}">
```
FloEdelmann, line 83: Please make sure that all code examples match the headings.

lib/rules/component-tags-order.js
```diff
@@ -25,6 +26,10 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/component-tags-order.html'
     },
     fixable: 'code',
+    messages: {
+      unexpected:
+        'The <{{elementName}}{{elementAttributes}}> should be above the <{{firstUnorderedName}}{{firstUnorderedAttributes}}> on line {{line}}.'
```
FloEdelmann, line 31: Better wording:

```suggestion
        '<{{elementName}}{{elementAttributes}}> should be above <{{firstUnorderedName}}{{firstUnorderedAttributes}}> (on line {{line}}).'
```

FloEdelmann (changes requested):

FloEdelmann (approved):
Looks good to me. Thanks for implementing the feedback!

@ota-meshi can you have another look?

lib/rules/component-tags-order.js
```diff
@@ -25,6 +26,10 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/component-tags-order.html'
     },
     fixable: 'code',
+    messages: {
+      unexpected:
```
ota-meshi, line 30: Is there a reason to move the "message"? If not, revart the position to minimize the difference.

docs/rules/component-tags-order.md
```diff
@@ -26,7 +26,7 @@ This rule warns about the order of the `<script>`, `<template>` & `<style>` tags
 }
 ```
 
-- `order` (`(string|string[])[]`) ... The order of top-level element names. default `[ [ "script", "template" ], "style" ]`.
+- `order` (`(string|string[])[]`) ... The order of top-level element names. default `[ [ "script", "template" ], "style" ]`. May also be CSS selectors, such as `script[setup]` and `i18n:not(lang=en)`.
```
ota-meshi, line 29:
```suggestion
- `order` (`(string|string[])[]`) ... The order of top-level element names. default `[ [ "script", "template" ], "style" ]`. May also be CSS selectors, such as `script[setup]` and `i18n:not([lang=en])`.
```

tests/lib/rules/component-tags-order.js
```diff
@@ -102,11 +104,90 @@ tester.run('component-tags-order', rule, {
       output: null,
       options: [{ order: ['docs', 'script', 'template', 'style'] }]
     },
+    {
+      code: '<script setup></script><script></script><template></template><style></style>',
+      output: null,
+      options: [{ order: ['script[setup]', 'script', 'template', 'style'] }]
```
ota-meshi, line 110: I think we should probably test `script:not([setup])`.

```suggestion
      options: [{ order: ['script[setup]', 'script:not([setup])', 'template', 'style'] }]
```

tests/lib/rules/component-tags-order.js
```diff
@@ -102,11 +104,90 @@ tester.run('component-tags-order', rule, {
...
+    },
+    {
+      code: '<template></template><script setup></script><script></script><style></style>',
+      output: null,
+      options: [{ order: [['script[setup]', 'script', 'template'], 'style'] }]
```
ota-meshi, line 115:
```suggestion
      options: [{ order: [['script[setup]', 'script:not([setup])', 'template'], 'style'] }]
```

tests/lib/rules/component-tags-order.js
```diff
@@ -102,11 +104,90 @@ tester.run('component-tags-order', rule, {
...
+    },
+    {
+      code: '<script></script><script setup></script><template></template><style></style>',
+      output: null,
+      options: [{ order: ['script', 'script[setup]', 'template', 'style'] }]
```
ota-meshi, line 120: I think we probably need to test that `'script'` also applies to `<script setup>`.

```suggestion
      options: [{ order: ['script', 'template', 'style'] }]
```

tests/lib/rules/component-tags-order.js
```diff
@@ -102,11 +104,90 @@ tester.run('component-tags-order', rule, {
...
+    },
+    {
+      code: '<template></template><script></script><script setup></script><style></style>',
+      output: null,
+      options: [{ order: [['script', 'script[setup]', 'template'], 'style'] }]
```
ota-meshi, line 125:
```suggestion
      options: [{ order: [['script, 'template'], 'style'] }]
```

ota-meshi (changes requested):
Thank you for changing this PR.
I will follow up about `postcss-selector-parser` later, so I commented on other things.

lib/rules/component-tags-order.js
doug-wade: Sorry about that. I think what happened is that my text editor alphabetized the keys on save. I agree we should try to keep the PR as small as possible, so I've turned off format on save for this workspace because it was causing this kind of trouble in my PRs.

ota-meshi (approved):
LGTM! Thank you.

PR #1826, opened 2022-03-24
Fix #1808: Lint slots in attribute-hyphenation

### LINKED ISSUE #1808: vue/attribute-hyphenation doesn't work on <slot> element
**Checklist**

- **ESLint version:** 8.9.0
- **eslint-plugin-vue version:** 8.5.0
- **Node version:** 16.14.0
- **Operating System:** Windows 11 21H1 22000.493 - (WSL) Ubuntu 20.04.3 LTS(Kernel: 5.10.93.2-microsoft-standard-WSL2)

```js
/* eslint-env node */
module.exports = {
  root: true,
  extends: [
    "plugin:vue/vue3-recommended",
    "eslint:recommended"
  ],
  env: {
    "vue/setup-compiler-macros": true,
  },
};
```

**What did you do?**	

```vue
<template>
  <div style="color: red;">
    <slot camelCase withValue="foo" :withBind="1" fooBar-BAZ_QuX />
  </div>
</template>
```

I expect warnings like this to appear:
> Attribute 'withVariableBind' must be hyphenated. eslint(vue/attribute-hyphenation)

**What actually happened?**

Nothing happens.

https://github.com/yusufkandemir/eslint-plugin-vue-slot-issue

Might be related to https://github.com/vuejs/eslint-plugin-vue/commit/602d80faad9b6004d040209c41fba68e062891d2

`<slot>` is a native HTML element(_related to Web Components_), but the rule should also run on that, just like it used to do before.

doug-wade:
Fixes #1808

### DIFF
--- a/docs/rules/attribute-hyphenation.md
+++ b/docs/rules/attribute-hyphenation.md
@@ -5,7 +5,6 @@ title: vue/attribute-hyphenation
 description: enforce attribute naming style on custom components in template
 since: v3.9.0
 ---
-
 # vue/attribute-hyphenation
 
 > enforce attribute naming style on custom components in template
@@ -36,7 +35,7 @@ This rule enforces using hyphenated attribute names on custom components in Vue 
 ```json
 {
   "vue/attribute-hyphenation": ["error", "always" | "never", {
-    "ignore": [],
+    "ignore": []
   }]
 }
 ```
--- a/docs/rules/attribute-hyphenation.md
+++ b/docs/rules/attribute-hyphenation.md
@@ -5,6 +5,7 @@ title: vue/attribute-hyphenation
 description: enforce attribute naming style on custom components in template
 since: v3.9.0
 ---
+
 # vue/attribute-hyphenation
 
 > enforce attribute naming style on custom components in template
@@ -36,7 +37,6 @@ This rule enforces using hyphenated attribute names on custom components in Vue 
 {
   "vue/attribute-hyphenation": ["error", "always" | "never", {
     "ignore": [],
-    "includeSlots": false,
   }]
 }
 ```
@@ -47,7 +47,6 @@ and all the [SVG attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/At
 - `"always"` (default) ... Use hyphenated name.
 - `"never"` ... Don't use hyphenated name except the ones that are ignored.
 - `"ignore"` ... Array of ignored names
-- `"includeSlots"` ... Will lint attributes on slot elements when set to true
 
 ### `"always"`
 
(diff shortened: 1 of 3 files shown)


docs/rules/attribute-hyphenation.md
```diff
@@ -35,7 +35,8 @@ This rule enforces using hyphenated attribute names on custom components in Vue
 ```json
 {
   "vue/attribute-hyphenation": ["error", "always" | "never", {
-    "ignore": []
+    "ignore": [],
+    "includeSlots": false,
```
doug-wade, line 39: I think the default should be `true`, to be honest, but I don't know the way that we release breaking changes. I thought it would be better to put the behavior behind an option, and then we can change the defaults as part of a major version, should we decide that's the best default.

docs/rules/attribute-hyphenation.md
```diff
@@ -108,6 +110,24 @@ Don't use hyphenated name but allow custom attributes
...
+
+<eslint-code-block fix :rules="{'vue/attribute-hyphenation': ['error', 'always', { includeSlots: true }]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <MyComponent><slot my-prop="prop"></slot></MyComponent>
+
+  <!-- ✗ BAD -->
+  <MyComponent><slot myProp="prop"></slot></MyComponent>
```
FloEdelmann, line 125: Please swap these for consistency with other examples.

```suggestion
  <!-- ✗ BAD -->
  <MyComponent><slot myProp="prop"></slot></MyComponent>

  <!-- ✓ GOOD -->
  <MyComponent><slot my-prop="prop"></slot></MyComponent>
```

FloEdelmann (approved):
Looks good 👍
I have just one suggestion.

docs/rules/attribute-hyphenation.md
ota-meshi: Thank you for this PR!

I think it's a bug that `<slot>` is ignored from the check, so I think it's a good to remove the option and always check `<slot>`.

ota-meshi (changes requested):
https://github.com/vuejs/eslint-plugin-vue/pull/1826#discussion_r834862790

docs/rules/attribute-hyphenation.md
```diff
@@ -35,7 +36,7 @@ This rule enforces using hyphenated attribute names on custom components in Vue
 ```json
 {
   "vue/attribute-hyphenation": ["error", "always" | "never", {
-    "ignore": []
+    "ignore": [],
   }]
```
FloEdelmann, line 40: Something is meesd up here now. Please change it back to how it was before.

FloEdelmann (changes requested):

docs/rules/attribute-hyphenation.md
doug-wade: Sorry about that! Looks like my text editor got a little excited about linting.

FloEdelmann (approved):

ota-meshi (approved):
Thank you!

doug-wade:
@ota-meshi @FloEdelmann Thanks so much for your thorough reviews and quick response time! It's been a great experience to add some features I need!

Issue #1786, opened 2022-02-02
Rule Proposal: Analog of vue/match-component-file-name for imported components


ilyub:
**Please describe what the rule should do:**

This rule should enforce consistent tag names for imported components.

**What category should the rule belong to?**

Enforces code style (layout)

**Provide 2-3 code examples that this rule should warn about:**

Valid example
```vue
import IconPicker from "./IconPicker.vue";

export default defineComponent({
  name: "app",
  components: {
    "icon-picker": IconPicker
```

Invalid example
```vue
import IconPicker from "./IconPicker.vue";

export default defineComponent({
  name: "app",
  components: {
    "some-inconsistent-name": IconPicker
```

FloEdelmann:
Good idea! I'd suggest `vue/match-component-import-name` as a rule name.

ilyub:
Thx for the update

PR #1831, opened 2022-03-27
Fix 1786

doug-wade:
Adds a new rule match-component-import-name as proposed in #1786

### DIFF
--- a/docs/rules/match-component-import-name.md
+++ b/docs/rules/match-component-import-name.md
@@ -17,13 +17,28 @@ This rule reports if the name of a registered component does not match its impor
 
 ## :book: Rule Details
 
-By default, this rule will validate that the imported name is the same casing.
-
-Case can be one of: `"kebab-case"` or `"PascalCase"`
-
-An optional prefix can be provided that must be prepended to all imports.
+By default, this rule will validate that the imported name is the same casing, but it can be configured to enforce kebab-case or PascalCase. An optional prefix can be provided that must be prepended to all imports.
 
 If you are not registering components, this rule will be ignored.
+
+<eslint-code-block :rules="{'vue/match-component-file-name': ['error']}">
+
+```vue
+<script>
+export default {
+  components: {
+    /* ✓ GOOD */
+    AppButton,
+
+    /* ✗ BAD */
+    SomeOtherName: AppButton,
+    'app-button': AppButton
+  }
+}
+</script>
+```
+
+</eslint-code-block>
 
 ## :wrench: Options
 
@@ -39,23 +54,8 @@ If you are not registering components, this rule will be ignored.
 }
 ```
 
-- `"prefix": ""` ... array of file extensions to be verified. Default is set to the empty string.
+- `"prefix": ""` ... required prefix for registered component names. Default is set to an empty string (no prefix).
 - `"case": "PascalCase"` ... one of "kebab-case" or "PascalCase", indicating the casing of the registered name. Default is set to `PascalCase`.
-
-### `{}`
-
-<eslint-code-block :rules="{'vue/match-component-file-name': ['error']}">
-
-```javascript
-/* ✓ GOOD */
-export default { components: { AppButton } }
-
-/* ✗ BAD */
-export default { components: { SomeOtherName: AppButton } }
-export default { components: { 'app-button': AppButton } }
-```
-
-</eslint-code-block>
 
 ### `{ case: 'kebab-case' }`
 
--- a/docs/rules/match-component-import-name.md
+++ b/docs/rules/match-component-import-name.md
@@ -12,7 +12,7 @@ description: require the registered component name to match the imported compone
 
 ## :book: Rule Details
 
-By default, this rule will validate that the imported name matches the name of the components object property identifer. Note that "matches" means that the imported name matches either the PascalCase or kebab-case version of the components object property identifer. If you would like to enforce that it must match only one of PascalCase or kebab-case, use this rule in conjunction with the rule `component-definition-name-casing`.
+By default, this rule will validate that the imported name matches the name of the components object property identifer. Note that "matches" means that the imported name matches either the PascalCase or kebab-case version of the components object property identifer. If you would like to enforce that it must match only one of PascalCase or kebab-case, use this rule in conjunction with the rule [vue/component-definition-name-casing](./component-definition-name-casing.md).
 
 <eslint-code-block :rules="{'vue/match-component-file-name': ['error']}">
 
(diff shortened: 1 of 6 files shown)


FloEdelmann (changes requested):
Preliminary review only.

lib/rules/match-component-import-name.js
```diff
@@ -0,0 +1,131 @@
...
+          casing: {
+            type: 'string'
+          }
+        },
+        additionalProperties: false
+      }
+    ],
+    messages: {
+      // ...
+    }
```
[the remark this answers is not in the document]
doug-wade: From another pr, it seems like maybe we prefer messageIds in general?

docs/rules/match-component-import-name.md
```diff
@@ -0,0 +1,95 @@
...
+      "prefix": "prefix-",
+      "case": "kebab"
+    }
+  ]
+}
+```
+
+By default, this rule will validate that the imported name is the same casing.
+
+Case can be one of: `"kebab"` (`casing-like-this`) or `"pascal"` (`CasingLikeThis`)
```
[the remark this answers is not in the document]
doug-wade: I really like this change a lot.

lib/utils/index.js
```diff
@@ -1015,6 +1015,12 @@ module.exports = {
    * @returns {VElement | null} the element of `<script setup>`
    */
   getScriptSetupElement,
+  /**
+   * Gets the element of `<script>` that is NOT setup
+   * @param {RuleContext} context The ESLint rule context object.
+   * @returns {VElement | null} the element of `<script>`
+   */
+  getScriptElement,
```
FloEdelmann, line 1023: Are the changes in this file really necessary? Other similar rules work with the existing utils.

docs/rules/match-component-import-name.md
```diff
@@ -0,0 +1,102 @@
...
+{
+  "vue/match-component-import-name": [
+    "error",
+    {
+      "prefix": "prefix-",
+      "case": "kebab-case"
+    }
+  ]
+}
+```
```
FloEdelmann, line 32: Please remove these lines (it's already mentioned in the "Options" section) and move the code examples for the options below separate `###` headings below the "Options" section.

See the docs for other rules with options (for consistency).

FloEdelmann (changes requested):

docs/rules/match-component-import-name.md
```diff
@@ -0,0 +1,109 @@
...
+<eslint-code-block :rules="{'vue/match-component-file-name': ['error', { case: 'kebab-case', prefix: 'prefix-' }]}">
+
+```javascript
+/* ✓ GOOD */
+export default { components: { 'prefix-app-button': AppButton } }
+
+/* ✗ BAD */
+export default { components: { SomeOtherName: AppButton } }
+export default { components: { AppButton } }
+```
```
FloEdelmann, line 102:
````suggestion
```vue
<script>
export default {
  components: {
    /* ✓ GOOD */
    'prefix-app-button': AppButton,
    
    /* ✗ BAD */
    SomeOtherName: AppButton,
    AppButton
  }
}
</script>
```
````

**Review by FloEdelmann — CHANGES_REQUESTED** [review]

File: docs/rules/match-component-import-name.md [hunk 6]
```diff
@@ -0,0 +1,109 @@
...
+
+## :wrench: Options
+
+```json
+{
+  "vue/match-component-import-name": [
+    "error",
+    {
+      "prefix": "prefix-",
+      "case": "kebab-case"
```
ota-meshi, line 36: I don't think the rule proposed in #1786 needs to check the casing. I think the casing check should be implemented by another rule. Alternatively, I think you could add an option to the component-definition-name-casing rule.
https://eslint.vuejs.org/rules/component-definition-name-casing.html

I think it's good to make sure that the match-component-import-name rule is defined with the same name as the converted to the kebab-case or PascalCase.



Also, I don't think the `prefix` option is needed. Why did you think we needed `prefix` option?

ota-meshi (changes requested):

docs/rules/match-component-import-name.md
doug-wade: I need the prefix option for the project I'm working on. We prefix every component name with the prefix `sj-` to indicate that its a component that we own, and to avoid name collisions with other libraries. So, for instance, we would have the import:

```
<script>
import Button from '../Button`

export default {
  components: {
    'sj-button': Button,
    'sj-textarea': () => import('../TextArea')
  }
}
</script>
```

Admittedly, I could refactor our product so the imports have the prefix (`import SjButton from '../Button'`), or move all of our files, but this allows me to turn on `match-component-file-name` and `match-component-import-name` without changing any of my templates. If we don't want to maintain this option for `match-component-import-name` going forward, I can remove it.

Thinking about getting options from component-definition-name-casing or match-component-import-name, I'd prefer to make it so this rule can be configured without using/configuring either of them, so I think I'll implement it so that either case is considered matching, and then add a note to the docs that you should use `component-definition-name-casing` to enforce which case.

ota-meshi: > Admittedly, I could refactor our product so the imports have the prefix (import SjButton from '../Button'), or move all of our files, but this allows me to turn on match-component-file-name and match-component-import-name without changing any of my templates. If we don't want to maintain this option for match-component-import-name going forward, I can remove it.

Hmm. If the prefix makes sense, I think it makes more sense to prefix the variable as well. (In addition, it makes sense that the file names also match.) I think it's a consistent rule. So I don't think the prefix option is needed. I think rules should focus on consistency over flexibility.

lib/rules/match-component-import-name.js
```diff
@@ -0,0 +1,136 @@
...
+     */
+    function getExpectedNames(identifier, prefix) {
+      return [
+        `${prefix}${casing.pascalCase(identifier.name)}`,
+        `${prefix}${casing.kebabCase(identifier.name)}`
+      ]
+    }
+
+    return {
+      ExportDefaultDeclaration(node) {
```
ota-meshi, line 93: I think you can use `utils.executeOnVueComponent()`.

ota-meshi (changes requested):

docs/rules/match-component-import-name.md
```diff
@@ -0,0 +1,80 @@
...
+---
+# vue/match-component-import-name
+
+> require the registered component name to match the imported component name
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+By default, this rule will validate that the imported name matches the name of the components object property identifer. Note that "matches" means that the imported name matches either the PascalCase or kebab-case version of the components object property identifer. If you would like to enforce that it must match only one of PascalCase or kebab-case, use this rule in conjunction with the rule `component-definition-name-casing`.
```
FloEdelmann, line 15: Please link the rule docs and also add it to a new "related rules" section (see other rules for an example).

FloEdelmann (changes requested):

FloEdelmann (approved):
Looks good to me :)

PR #1842, opened 2022-04-11
Improve usage of postcss-selector-parser

ota-meshi:
This PR followup for #1825.

~~Also, fixing the linting errors. (Circle CI didn't seem to work well 🤔)~~ Fixed #1843
Also, I made a small change in the documentation.


lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+ */
+function selectorToVElementMatcher(selectorChildren) {
+  const nodes = [...selectorChildren]
+  let node = nodes.shift()
+  /**
+   * @type {VElementMatcher | null}
+   */
+  let result = null
+  while (node) {
+    if (node.type === 'combinator') {
```
ota-meshi, line 134: The component-tags-order rule does not use the combinator, but it does handle the combinator so that it can be used in future changes. It has also been tested.

ota-meshi:
> Circle CI didn't seem to work well 🤔

The CI for pull requests from some forks doesn't seem to work. But I don't know the rules. The settings appear to be set to build pull requests from forks as well.

tests/lib/utils/selector.js
```diff
@@ -0,0 +1,86 @@
...
+    {
+      parser: 'vue-eslint-parser',
+      parserOptions: { ecmaVersion: 2018 },
+      rules: { 'vue/selector-test': 'error' }
+    },
+    undefined,
+    true
+  )
+
+  return result.concat(messages.map((message) => ({ error: message.message })))
```
FloEdelmann, line 61: Why is `selector: inputSelector` part of every result array item?


```suggestion
            elementText: parsed.elementText
          })
        }
      }
    })
  })
  linter.defineParser('vue-eslint-parser', require('vue-eslint-parser'))
  const messages = linter.verify(
    code,
    {
      parser: 'vue-eslint-parser',
      parserOptions: { ecmaVersion: 2018 },
      rules: { 'vue/selector-test': 'error' }
    },
    undefined,
    true
  )

  return {
    selector: inputSelector,
    matches: result,
    errors: messages.map((message) => message.message)
  }
```

ota-meshi: Thank you for checking. The method you suggested is simpler.

lib/rules/component-tags-order.js
```diff
@@ -156,18 +135,34 @@ module.exports = {
...
+              element
+            }
+          })
+          .filter(
+            /**
+             * @param { {order:OrderElement|undefined, element: VElement} } o
+             * @returns {o is {order:OrderElement, element: VElement}}
+             */
+            (o) => Boolean(o.order)
+          )
```
FloEdelmann, line 153: Easier to read (IMHO) and saves one loop:

```suggestion
          .flatMap((element) => {
            const order = getOrderElement(element)
            return order ? [{ order, element }] : []
          })
```

lib/rules/component-tags-order.js
```diff
@@ -156,18 +135,34 @@ module.exports = {
...
-          }
-          const firstUnordered = elements
+        elementWithOrders.forEach(({ order: expected, element }, index) => {
+          const firstUnordered = elementWithOrders
             .slice(0, index)
-            .filter((e) => expectedIndex < getOrderPosition(e))
-            .sort((e1, e2) => getOrderPosition(e1) - getOrderPosition(e2))[0]
+            .filter(({ order }) => {
+              return expected.index < order.index
+            })
```
FloEdelmann, line 160:
```suggestion
            .filter(({ order }) => expected.index < order.index)
```

lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+ * @returns {ChildNode[]}
+ */
+function cleanSelectorChildren(selector) {
+  /** @type {ChildNode[]} */
+  const nodes = []
+  /** @type {ChildNode|null} */
+  let last = null
+  for (const node of selector.nodes) {
+    if (node.type === 'root') {
+      throw new Error('Unexpected state type=root')
```
FloEdelmann, line 89: Shouldn't this be a `SelectorError`?

```suggestion
      throw new SelectorError('Unexpected state type=root')
```

lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+  }
+  if (selector.type === 'tag') {
+    const name = selector.value
+    return (element) => element.rawName === name
+  }
+  if (selector.type === 'universal') {
+    return () => true
+  }
+
+  if (selector.type === 'pseudo') {
```
FloEdelmann, line 265: Maybe extract everything inside this branch into a separate function `pseudoSelectorToVelementMatcher` or similar?

lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+  throw new SelectorError(`Unknown combinator: ${combinator}.`)
+}
+
+/**
+ * Convert node to VElementMatcher
+ * @param {Exclude<parser.Node, {type:'combinator'|'comment'|'root'|'selector'}>} selector
+ * @returns {VElementMatcher}
+ */
+function nodeToVElementMatcher(selector) {
+  if (selector.type === 'attribute') {
```
FloEdelmann, line 234: There are a lot of similar conditions in row here: 

```js
if (selector.type === '…') {
  // …
  return …
}
// similar cases following
```

This might be simplified/clarified to:

```js
switch (selector.type) {
  case '…': {
    // …
    return …
  }
  // similar cases following
}
```

lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+    const name = selector.value
+    return (element) => element.rawName === name
+  }
+  if (selector.type === 'universal') {
+    return () => true
+  }
+
+  if (selector.type === 'pseudo') {
+    const pseudo = selector.value
+    if (pseudo === ':not') {
```
FloEdelmann, line 267: Again, `switch`/`case` might be a little more readable here.

lib/utils/selector.js
```diff
@@ -0,0 +1,582 @@
...
+ * @param {parser.Attribute} selector
+ * @returns {VElementMatcher}
+ */
+function attributeNodeToVElementMatcher(selector) {
+  const key = selector.attribute
+  if (!selector.operator) {
+    return (element) => getAttributeValue(element, key) != null
+  }
+  const value = selector.value || ''
+  if (selector.operator === '=') {
```
FloEdelmann, line 372: Again, `switch`/`case` might be a little more readable here.

tests/lib/utils/selector.js
```diff
@@ -0,0 +1,89 @@
...
+
+  linter.defineRule('vue/selector-test', (context) => {
+    const parsed = selector.parseSelector(inputSelector, context)
+    return utils.defineDocumentVisitor(context, {
+      VElement(node) {
+        if (parsed.test(node)) {
+          result.push({
+            text: context.getSourceCode().text.slice(...node.startTag.range),
+            elementText: parsed.elementText
+          })
```
FloEdelmann, line 43: `elementText` is never used/printed anywhere else. Is that intended? Otherwise, this might be simplified to:

```suggestion
          result.push(context.getSourceCode().text.slice(...node.startTag.range))
```

tests/lib/utils/selector.js
```diff
@@ -0,0 +1,89 @@
...
+    )
+    const code = code0.replace(/^<!--(.+?)-->/, `<!--${name}-->`)
+    const inputSelector = /^<!--(.+?)-->/.exec(code0)[1].trim()
+    return { code, name, inputSelector }
+  })
+}
+
+function extractElements(code, inputSelector) {
+  const linter = new Linter()
+  const result = []
```
FloEdelmann, line 33: Please rename `result` to something more meaningful like `matches`.

FloEdelmann (changes requested):

ota-meshi:
Thank you for checking this PR.
I changed this pull request. So please check it again?

FloEdelmann (approved):
Looks good to me. Thanks!

PR #1843, opened 2022-04-11
Add github workflows

ota-meshi:
This PR adds some simple GitHub actions workflow.

I noticed that CircleCI didn't build some pull requests.

https://github.com/vuejs/eslint-plugin-vue/pull/1842#issuecomment-1094562455

If the problem persists for a long time, we should consider migrating to GitHub actions. However, it may be a temporary problem, so I would like to use it in parallel and see what happens.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,42 @@
+name: CI
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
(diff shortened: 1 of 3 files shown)


ota-meshi:
I don't know the cause, so I contacted CircleCI support.

ota-meshi:
It seems that jobs may be blocked for PR by users who have not recently authenticated to CircleCI.

Issue #1836, opened 2022-04-04
vue/multi-word-component-names doesn't catch SFCs with setup script?


vincerubinetti:
Not sure if this is a bug or a question, but it's certainly strange to me. If it's intentional it should be noted somewhere.

I was linting a project made with `vue-cli`, and noticed that some single-word component files weren't being caught by the linter, and some were. The ones that were being caught were ones with just a `<template>`, and the ones that weren't had a `<script setup>`.

Here's a minimum reproducible example (at least in my project setup); three variations of a file **incorrectly** named `Test.vue`:

Caught ✅  (lint check fails)

```vue
<template>
  <p>test</p>
</template>
```

Caught ✅  (lint check fails)

```vue
<template>
  <AppButton />
</template>

<script lang="ts">
import { defineComponent } from "vue";
import AppButton from "@/components/AppButton.vue";

export default defineComponent({ components: { AppButton } });
</script>
```

Not caught ❌  (lint check passes)

```vue
<template>
  <AppButton />
</template>

<script setup lang="ts">
import AppButton from "@/components/AppButton.vue";
</script>
```

<details>

<summary>Here is my `package.json`:</summary>

```json
{
  "scripts": {
    "serve": "vue-cli-service serve",
    "build": "vue-cli-service build",
    "test:unit": "vue-cli-service test:unit /unit",
    "test:e2e": "vue-cli-service test:e2e --headless",
    "lint": "vue-cli-service lint && prettier --write \"**/*.{scss,js,ts,json,vue,svg,yaml}\" --no-error-on-unmatched-pattern",
    "fresh": "yarn cache clean && rm -rf node_modules yarn.lock && yarn install",
    "test": "yarn test:lint && yarn test:unit && yarn test:e2e && yarn test:axe",
    "test:axe": "vue-cli-service test:unit /axe",
    "test:gui": "vue-cli-service test:e2e",
    "test:lint": "vue-cli-service lint --no-fix && prettier --check \"**/*.{scss,js,ts,json,vue,svg,yaml}\" --no-error-on-unmatched-pattern"
  },
  "dependencies": {
    "@fortawesome/fontawesome-svg-core": "^1.3.0",
    "@fortawesome/free-brands-svg-icons": "^6.0.0",
    "@fortawesome/free-regular-svg-icons": "^6.0.0",
    "@fortawesome/free-solid-svg-icons": "^6.0.0",
    "@fortawesome/vue-fontawesome": "^3.0.0-5",
[...]

PR #1844, opened 2022-04-12
Fix 1836

doug-wade:
Addresses #1836


FloEdelmann (approved):
Looks good to me :)
But maybe @ota-meshi thinks that some `util` function needs to be fixed instead of fixing only this rule?

ota-meshi (approved):
LGTM!

PR #1845, opened 2022-04-13
fix #255: Add new rule require-v-if-else-key

### LINKED ISSUE #255: Rule proposal: `v-if-else-key`
**Style guide:**
https://vuejs.org/v2/style-guide/#v-if-v-if-else-v-else-without-key-use-with-caution

**Description:**
This rule would enforce to add `key` property on elements with `v-if` and `v-else` directives if they're both of the same type.

doug-wade:
This one's probably a smidge later than it could have been -- I can't find it in the style guide anymore -- but better late than never?

### DIFF
--- a/docs/rules/require-v-if-else-key.md
+++ b/docs/rules/require-v-if-else-key.md
@@ -12,7 +12,7 @@ description: require `key` with `v-if/v-else-if/v-else` directives
 
 ## :book: Rule Details
 
-This rule reports the elements which have `v-if`, `v-else-if`, and/or `v-else` and do not have a `v-bind:key` or `key`.
+This rule reports consecutive elements with the same tag name which all have `v-if`, `v-else-if`, or `v-else` and do not have a `v-bind:key` or `key`.
 
 <eslint-code-block :rules="{'vue/require-v-if-else-key': ['error']}">
 
@@ -44,7 +44,7 @@ Nothing.
 
 ## :books: Further Reading
 
-- [Guide (for v2) - v-if without key](https://v2.vuejs.org/v2/style-guide/?redirect=true#v-if-v-else-if-v-else-without-key-use-with-caution)
+- [Guide (for v2) - v-if without key](https://v2.vuejs.org/v2/style-guide/#v-if-v-else-if-v-else-without-key-use-with-caution)
 
 ## :mag: Implementation
 
--- a/docs/rules/require-v-if-else-key.md
+++ b/docs/rules/require-v-if-else-key.md
@@ -14,6 +14,8 @@ description: require `key` with `v-if/v-else-if/v-else` directives
 
 This rule reports consecutive elements with the same tag name which all have `v-if`, `v-else-if`, or `v-else` and do not have a `v-bind:key` or `key`.
 
+This rule is not required in Vue 3, as the key is automatically assigned to the elements.
+
 <eslint-code-block :rules="{'vue/require-v-if-else-key': ['error']}">
 
 ```vue
--- a/docs/rules/require-v-if-else-key.md
+++ b/docs/rules/require-v-if-else-key.md
@@ -1,0 +1,52 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/require-v-if-else-key
+description: require `key` with `v-if/v-else-if/v-else` directives
+---
+# vue/require-v-if-else-key
+
+> require `key` with `v-if/v-else-if/v-else` directives
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule reports the elements which have `v-if`, `v-else-if`, and/or `v-else` and do not have a `v-bind:key` or `key`.
+
+<eslint-code-block :rules="{'vue/require-v-if-else-key': ['error']}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <div v-if="condition"/>
+
+  <div v-if="condition"/>
+  <span v-else>
+
+  <div v-if="condition" key="one"/>
+  <div v-else key="two"/>
+
+  <div v-if="condition" :key="one"/>
+  <div v-else :key="two"/>
+
+  <!-- ✗ BAD -->
+  <div v-if="condition"/>
+  <div v-else/>
+</template>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :books: Further Reading
+
+- [Guide (for v2) - v-if without key](https://v2.vuejs.org/v2/style-guide/?redirect=true#v-if-v-else-if-v-else-without-key-use-with-caution)
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/require-v-if-else-key.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/require-v-if-else-key.js)
--- a/lib/rules/require-v-if-else-key.js
+++ b/lib/rules/require-v-if-else-key.js
@@ -45,7 +45,7 @@ module.exports = {
     function getKey(element) {
       const keyAttribute = utils.getAttribute(element, 'key')
       if (keyAttribute) {
-        return keyAttribute?.value?.value || ''
+        return keyAttribute.value ? keyAttribute.value.value || '' : ''
       }
 
       const keyDirective = utils.getDirective(element, 'bind', 'key')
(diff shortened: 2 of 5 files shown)


lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+      if (!keyDirective) {
+        return ''
+      }
+
+      if (
+        !keyDirective.value ||
+        !keyDirective.value.expression ||
+        keyDirective.value.expression.type !== 'Identifier'
+      ) {
+        return null
```
doug-wade, line 61: I'm not sure about using the empty string and null to represent two different things. I'm definitely open to other suggestions.

docs/rules/require-v-if-else-key.md
```diff
@@ -0,0 +1,52 @@
...
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :books: Further Reading
+
+- [Guide (for v2) - v-if without key](https://v2.vuejs.org/v2/style-guide/?redirect=true#v-if-v-else-if-v-else-without-key-use-with-caution)
```
FloEdelmann, line 47:
```suggestion
- [Guide (for v2) - v-if without key](https://v2.vuejs.org/v2/style-guide/#v-if-v-else-if-v-else-without-key-use-with-caution)
```

lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'require `key` with `v-if/v-else-if/v-else` directives',
```
FloEdelmann, line 27:
```suggestion
      description: 'require `key` with `v-if`/`v-else-if`/`v-else` directives',
```

lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+// ...
+
```
FloEdelmann, line 18: Please remove these empty boilerplate lines.


```suggestion
```

docs/rules/require-v-if-else-key.md
```diff
@@ -0,0 +1,52 @@
...
+---
+# vue/require-v-if-else-key
+
+> require `key` with `v-if/v-else-if/v-else` directives
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule reports the elements which have `v-if`, `v-else-if`, and/or `v-else` and do not have a `v-bind:key` or `key`.
```
FloEdelmann, line 15: An element can't have more than one of these directives. Also, it should be clarified that they have to have the same tag name.

```suggestion
This rule reports consecutive elements with the same tag name which all have `v-if`, `v-else-if`, or `v-else` and do not have a `v-bind:key` or `key`.
```

lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+    docs: {
+      description: 'require `key` with `v-if/v-else-if/v-else` directives',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/require-v-if-else-key.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      unexpected:
+        'Elements in v-if/v-else-if/v-else expect to have distinct keys if they are of the same type.'
```
FloEdelmann, line 35: Maybe _tag name_ is a bit clearer than _element type_?


```suggestion
        'Elements in v-if/v-else-if/v-else should have distinct keys if they have the same tag name.'
```

lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+  create(context) {
+    /**
+     * Gets the key from an element.
+     * @param {VElement} element The element node to get the key from.
+     * @return {String | null}
+     */
+    function getKey(element) {
+      const keyAttribute = utils.getAttribute(element, 'key')
+      if (keyAttribute) {
+        return keyAttribute.value ? keyAttribute.value.value || '' : ''
```
FloEdelmann, line 48:
```suggestion
        return keyAttribute.value?.value || ''
```

tests/lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,124 @@
...
+      filename: 'test.vue',
+      code: `
+      <template>
+        <div>
+          <div v-if="conditional" key="foo">foo</div>
+          <div v-else-if="conditional" key="bar">bar</span>
+        </div>
+      </template>
+      `
+    }
```
FloEdelmann, line 84: Please add one valid test case with two independent `<div v-if="...">`s.

FloEdelmann (changes requested):
Thanks for this rule. I have a few suggestions.

tests/lib/rules/require-v-if-else-key.js
ota-meshi, line 84: We need to add tests using `<template v-else>`. Note that vue2 does not allow you to place a key in `<template>`.

lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,104 @@
...
+        return
+      }
+
+      const key = getKey(element)
+
+      if (key === null) {
+        return
+      }
+
+      if (!key || key === getKey(prevSibling)) {
```
ota-meshi, line 88: I don't think it is necessary to check the contents of the key. If the user uses the same key, I think the user is intentional.
So, I think we can change it to check if the element has a key.

docs/rules/require-v-if-else-key.md
```diff
@@ -0,0 +1,52 @@
...
+
```
ota-meshi, line 16: Can you explain here that Vue3 doesn't require this rule? In Vue3, the key is automatically assigned.

ota-meshi (changes requested):

docs/rules/require-v-if-else-key.md
ota-meshi: Demo: https://vue-next-template-explorer.netlify.app/#eyJzcmMiOiI8ZGl2IHYtaWY9XCJlcnJvclwiPlxuICBFcnJvcjoge3sgZXJyb3IgfX1cbjwvZGl2PlxuPGRpdiB2LWVsc2U+XG4gIHt7IHJlc3VsdHMgfX1cbjwvZGl2PlxuIiwic3NyIjpmYWxzZSwib3B0aW9ucyI6eyJvcHRpbWl6ZUltcG9ydHMiOmZhbHNlLCJvcHRpbWl6ZUJpbmRpbmdzIjpmYWxzZX19

FloEdelmann (approved):
Please run `npm run update` to uodate the docs and fix lint any potential lint issues.

Other than that, this looks good to me. Thanks!

tests/lib/rules/require-v-if-else-key.js
```diff
@@ -0,0 +1,159 @@
...
+            <div>foo</div>
+            <div>bar</span>
+          </template>
+          <template v-else>
+            <div>baz</div>
+            <div>quux</span>
+          </template>
+        </div>
+      </template>
+      `
```
ota-meshi, line 142: Why this is valid? I think this should be reported.

doug-wade: I must have misinterpreted [your comment](https://github.com/vuejs/eslint-plugin-vue/pull/1845#discussion_r850099546):

> Note that vue2 does not allow you to place a key in <template>.

If no keys are allowed, mustn't we allow templates without keys?

ota-meshi: In Vue2 need to put the key on the actual element, not the `<template>`.

e.g.

```vue
<template v-if="foo">
  <div key="a" />
</template>
<template v-else>
  <div key="b" />
</template>
```

```vue
<template v-if="foo">
  <div key="a" />
  <div key="b" />
</template>
<template v-else>
  <div key="c" />
  <div key="d" />
</template>
```

However, with Vue3 you can do the following:



```vue
<template v-if="foo" :key="a">
  <div />
</template>
<template v-else :key="b">
  <div />
</template>
```

```vue
<template v-if="foo" :key="a">
  <div />
  <div />
</template>
<template v-else :key="b">
  <div />
  <div />
</template>
```

doug-wade:
Heya, things have gotten pretty busy for me, and I don't think I'll have time to clean this one up in the near future. I'm going to close this one for now.

PR #1846, opened 2022-04-14
Change default casing of `vue/custom-event-name-casing` rule to camelCase

### LINKED ISSUE #1759: Change default name case for event names to match documentation

https://eslint.vuejs.org/rules/custom-event-name-casing.html#camelcase

Same

Default change

```js
      this.$emit('my-event')
      this.$emit('update:myProp', myProp)

      this.$emit('myEvent')
```

Allows the 2 first lines, fails on the last.

It would do the opposite.

It should inverse the default and enforce camel-cased event names when calling `emit()` like [proposed in docs](https://vuejs.org/guide/components/events.html#emitting-and-listening-to-events)

**Additional context**

ota-meshi:
close #1759

### DIFF
--- a/docs/rules/custom-event-name-casing.md
+++ b/docs/rules/custom-event-name-casing.md
@@ -13,7 +13,7 @@ Define a style for custom event name casing for consistency purposes.
 
 ## :book: Rule Details
 
-This rule aims to warn the custom event names other than the configured casing.
+This rule aims to warn the custom event names other than the configured casing. (Default is **camelCase**.)
 
 Vue 2 recommends using kebab-case for custom event names.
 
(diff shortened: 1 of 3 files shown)


FloEdelmann:
Should the default of `vue/v-on-event-hyphenation` also change accordingly?

ota-meshi:
I don't think we need to change `vue/v-on-event-hyphenation`. Vue3 allows users to listen to camelCase events using kebab-case.

The documentation recommends a kebab-case.

> we recommend using kebab-cased event listeners in templates.

https://vuejs.org/guide/components/events.html#emitting-and-listening-to-events

FloEdelmann (approved):

PR #1847, opened 2022-04-14
Drop support for Node.js v12

ota-meshi:


FloEdelmann (changes requested):
This should also increase the ECMA version in our ESLint config to ES2020. See

* https://github.com/vuejs/eslint-plugin-vue/pull/1856#discussion_r853112971
* https://stackoverflow.com/questions/64545963/mapping-of-node-js-version-to-ecmascript-version
* https://node.green/#ES2020

FloEdelmann (approved):

PR #1848, opened 2022-04-14
Change presets configs and remove unused internal methods

ota-meshi:
### Change Vue 3 presets configs

* Change `plugin:vue/vue3-essential` config:
  - Add [`vue/no-child-content`](https://eslint.vuejs.org/rules/no-child-content.html) rule
  - Add [`vue/no-expose-after-await`](https://eslint.vuejs.org/rules/no-expose-after-await.html) rule
  - Add [`vue/no-reserved-component-names`](https://eslint.vuejs.org/rules/no-reserved-component-names.html) rule
  - Add [`vue/no-use-computed-property-like-method`](https://eslint.vuejs.org/rules/no-use-computed-property-like-method.html) rule
  - Add [`vue/no-v-text-v-html-on-component`](https://eslint.vuejs.org/rules/no-v-text-v-html-on-component.html) rule
  - Add [`vue/prefer-import-from-vue`](https://eslint.vuejs.org/rules/prefer-import-from-vue.html) rule
* Change `plugin:vue/vue3-strongly-recommended` config:
  - Same changes as above
* Change `plugin:vue/vue3-recommended` config:
  - Same changes as above

### Change Vue 2 preset configs

* Change `plugin:vue/essential` config:
  - Add [`vue/no-child-content`](https://eslint.vuejs.org/rules/no-child-content.html) rule
  - Add [`vue/no-reserved-component-names`](https://eslint.vuejs.org/rules/no-reserved-component-names.html) rule
  - Add [`vue/no-use-computed-property-like-method`](https://eslint.vuejs.org/rules/no-use-computed-property-like-method.html) rule
  - Add [`vue/no-v-text-v-html-on-component`](https://eslint.vuejs.org/rules/no-v-text-v-html-on-component.html) rule
  - Add [`vue/valid-model-definition`](https://eslint.vuejs.org/rules/valid-model-definition.html) rule
  - Add [`vue/no-export-in-script-setup`](https://eslint.vuejs.org/rules/no-export-in-script-setup.html) rule \*
  - Add [`vue/no-ref-as-operand`](https://eslint.vuejs.org/rules/no-ref-as-operand.html) rule \*
  - Add [`vue/no-setup-props-destructure`](https://eslint.vuejs.org/rules/no-setup-props-destructure.html) rule \*
  - Add [`vue/return-in-emits-validator`](https://eslint.vuejs.org/rules/return-in-emits-validator.html) rule \*
  - Add [`vue/valid-define-emits`](https://eslint.vuejs.org/rules/valid-define-emits.html) rule \*
  - Add [`vue/valid-define-props`](https://eslint.vuejs.org/rules/valid-define-props.html) rule \*
* Change `plugin:vue/strongly-recommended` config:
  - Same changes as above
* Change `plugin:vue/recommended` config:
  - Same changes as above

*Note:* The rules marked with \* are now included in the `plugin:vue/essential` config because [`@vue/composition-api`](https://github.com/vuejs/composition-api) and [`unplugin-vue2-script-setup`](https://github.com/antfu/unplugin-vue2-script-setup/) add (limited) support for these features in Vue 2. If you don't use those libraries, enabling these rules shouldn't affect you.

### Rename rules

* Rename `vue/no-invalid-model-keys` rule to [`vue/valid-model-definition`](https://eslint.vuejs.org/rules/valid-model-definition.html)

### Internal changes

Also, remove unused internal methods:

* `getComponentProps` util method: use `getComponentPropsFromOptions` instead.
* `getComponentEmits` util method: use `getComponentEmitsFromOptions` instead.


FloEdelmann:
Maybe also include the following uncategorized rules in presets?

* `plugin:vue/vue3-essential`:
  - `vue/no-invalid-model-keys` (could be renamed to `vue/valid-model-definition` for consistency)
* `plugin:vue/vue3-essential` and `plugin:vue/essential`:
  - `vue/no-use-computed-property-like-method`
  - `vue/no-reserved-component-names`
  - `vue/no-child-content`
* `plugin:vue/vue3-strongly-recommended`:
  - `vue/require-expose`
  - `vue/require-emit-validator`
* `plugin:vue/vue3-recommended` and `plugin:vue/recommended`:
  - `vue/component-options-name-casing`
  - `vue/custom-event-name-casing`
  - `vue/match-component-import-name`
  - `vue/no-multiple-objects-in-class`
  - `vue/no-useless-v-bind`
  - `vue/no-useless-mustaches`
  - `vue/no-v-text`
  - `vue/prefer-separate-static-class`

ota-meshi:
Thank you for checking and listing the rules.

> * `plugin:vue/vue3-essential`:
>   - `vue/no-invalid-model-keys` (could be renamed to `vue/valid-model-definition` for consistency)

I think it's a good idea to add the `no-invalid-model-keys` rule to `plugin:vue/essential`.

> * `plugin:vue/vue3-essential` and `plugin:vue/essential`:
>   - `vue/no-use-computed-property-like-method`
>   - `vue/no-reserved-component-names`
>   - `vue/no-child-content`

I think it is a good idea to include these rules in the preset since they are intended to prevent errors.

> * `plugin:vue/vue3-strongly-recommended`:
>   - `vue/require-expose`
>   - `vue/require-emit-validator`
> * `plugin:vue/vue3-recommended` and `plugin:vue/recommended`:
>   - `vue/component-options-name-casing`
>   - `vue/custom-event-name-casing`
>   - `vue/match-component-import-name`
>   - `vue/no-multiple-objects-in-class`
>   - `vue/no-useless-v-bind`
>   - `vue/no-useless-mustaches`
>   - `vue/no-v-text`
>   - `vue/prefer-separate-static-class`

I'm wondering if these should be included in the preset. Since presets are based on style guide, we may not want to include rules that aren't written in the style guide (and Vue official documentation). 
(However, the style guide has not been maintained recently 😓)

It already contains rules that are not written in the style guide. However, many users have complained when new rules (even if the rules are written in the style guide) have been added until now.

ota-meshi:
I forgot to write a comment about renaming.
I think it's a good idea to rename the rule as you say :+1:

lib/rules/valid-model-definition.js
```diff
@@ -0,0 +1,57 @@
...
+// Rule Definition
+// ------------------------------------------------------------------------------
+const VALID_MODEL_KEYS = ['prop', 'event']
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'require valid keys in model option',
+      categories: ['essential'],
```
FloEdelmann, line 19: `vue/valid-model-definition` does definitely make more sense for Vue 3. So either it should be in `vue3-essential` only, or in `vue3-essential` and `essential`.

lib/rules/no-expose-after-await.js
```diff
@@ -40,8 +40,7 @@ module.exports = {
     type: 'problem',
     docs: {
       description: 'disallow asynchronously registered `expose`',
-      categories: undefined,
-      // categories: ['vue3-essential'], TODO Change with the major version
+      categories: ['vue3-essential'],
```
FloEdelmann, line 43: This rule (and some others) work fine on Vue 2 with [`@vue/composition-api`](https://github.com/vuejs/composition-api). Should we also add them to the respective Vue 2 presets, to help smooth the transition to Vue 3?

Or maybe create a new preset for those rules only that can be applied on top of any Vue 2 preset?

lib/rules/valid-model-definition.js
ota-meshi: As far as I know, the model option has been removed in vue3. It is fixed with `modelValue` and `update:modelValue`. Am I misunderstanding something?

FloEdelmann: Ah, again you're right.

lib/rules/no-expose-after-await.js
ota-meshi: I'm not sure about `@vue/composition-api` because I'm not using `@vue/composition-api`. 
I think it's okay if we add a rule to the preset if you think users need it.
However, I've heard before that `@vue/composition-api` doesn't support async, so if that's correct, I think the `no-expose-after-await` rule should only be included in the preset for Vue3.

FloEdelmann: You're right, `@vue/composition-api` doesn't support async `setup`, so this rule (and some others) are not required on Vue 2. However, there are some rules from which Vue 2 users would profit:

#### For [`@vue/composition-api`](https://github.com/vuejs/composition-api) users

* `vue/no-setup-props-destructure`
* `vue/no-ref-as-operand`

#### For https://github.com/antfu/unplugin-vue2-script-setup/ users

* `vue/no-export-in-script-setup`
* `vue/return-in-emits-validator`
* `vue/valid-define-emits`
* `vue/valid-define-props`

ota-meshi: Thank you for listing the rules!
I will include them in the preset for vue2.

FloEdelmann (approved):

PR #1849, opened 2022-04-14
Upgrade vue-eslint-parser

ota-meshi:
This PR upgrades the parser to v9.

As a result, the following features have been deprecated.

- `vue/script-setup-uses-vars` rule
- `vue/setup-compiler-macros` environment

### DIFF
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -344,7 +344,7 @@ See also: "[Visual Studio Code](#editor-integrations)" section and [Vetur - Lint
 
 You need to use [vue-eslint-parser] v9.0.0 or later.
 
-Previously you had to use the [vue/script-setup-uses-vars](../rules/script-setup-uses-vars.md) rule, but now you don't.
+Previously you had to use the [vue/script-setup-uses-vars](../rules/script-setup-uses-vars.md) rule, this is no longer needed.
 
 #### Compiler macros such as `defineProps` and `defineEmits` generate `no-undef` warnings
 
--- a/docs/user-guide/README.md
+++ b/docs/user-guide/README.md
@@ -342,36 +342,15 @@ See also: "[Visual Studio Code](#editor-integrations)" section and [Vetur - Lint
 
 #### The variables used in the `<template>` are warned by `no-unused-vars` rule
 
-You must use [vue/script-setup-uses-vars](../rules/script-setup-uses-vars.md) rule.  
-In your configuration, use the rule set provided by `eslint-plugin-vue` or enable it rule.
+You need to use [vue-eslint-parser] v9.0.0 or later.
 
-Example **.eslintrc.js**:
-
-```js
-module.exports = {
-  // Use the rule set.
-  extends: ['plugin:vue/base'],
-  rules: {
-    // Enable vue/script-setup-uses-vars rule
-    'vue/script-setup-uses-vars': 'error',
-  }
-}
-```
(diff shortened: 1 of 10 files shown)


docs/user-guide/README.md
```diff
@@ -342,36 +342,15 @@ See also: "[Visual Studio Code](#editor-integrations)" section and [Vetur - Lint
...
-module.exports = {
-  // Use the rule set.
-  extends: ['plugin:vue/base'],
-  rules: {
-    // Enable vue/script-setup-uses-vars rule
-    'vue/script-setup-uses-vars': 'error',
-  }
-}
-```
+Previously you had to use the [vue/script-setup-uses-vars](../rules/script-setup-uses-vars.md) rule, but now you don't.
```
FloEdelmann, line 347: Nicer wording:

```suggestion
Previously you had to use the [vue/script-setup-uses-vars](../rules/script-setup-uses-vars.md) rule, this is no longer needed.
```

docs/user-guide/README.md
```diff
@@ -342,36 +342,15 @@ See also: "[Visual Studio Code](#editor-integrations)" section and [Vetur - Lint
...
-Example **.eslintrc.js**:
-
-```js
-module.exports = {
-  env: {
-    'vue/setup-compiler-macros': true
-  }
-}
-```
+Previously you had to use the `vue/setup-compiler-macros` environment, but now you don't.
```
FloEdelmann, line 353:
```suggestion
Previously you had to use the `vue/setup-compiler-macros` environment, this is no longer needed.
```

tests/lib/script-setup-vars.js
```diff
@@ -0,0 +1,384 @@
...
+    ecmaVersion: 6,
+    sourceType: 'module'
+  }
+})
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+describe('vue-eslint-parser should to properly mark the variables used in the template', () => {
```
FloEdelmann, line 25:
```suggestion
describe('vue-eslint-parser should properly mark the variables used in the template', () => {
```

tests/lib/script-setup-vars.js
```diff
@@ -0,0 +1,384 @@
...
+  ruleTester.run('no-unused-vars', ruleNoUnusedVars, {
```
FloEdelmann, line 26: Doesn't it make more sense to put the tests into the respective rule tests?

docs/user-guide/README.md
ota-meshi: Thank you so much!

tests/lib/script-setup-vars.js
ota-meshi: These test that the ESLint core rules work as expected, so I don't think it makes sense to put them in the same place as the plugin rules.

FloEdelmann: Ahh, yeah you're right.

tools/update-lib-index.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     '.vue': require('./processor')
   },
   environments: {
+    // Deprecated
```
FloEdelmann, line 42:
```suggestion
    // TODO Remove in the next major version
    /** @deprecated */
```

(and then run `npm run update`)

FloEdelmann (approved):
Looks good to me. I have just one suggestion, so we don't forget removing the environment in v10.

Issue #1373, opened 2020-12-10
Rule Proposal: `no-invalid-attribute-name`


frederikadamse22:
**Please describe what the rule should do:**
`no-invalid-attribute-name` should disallow invalid attribute names for elements in the `<template>` tag.

**What category should the rule belong to?**

[ ] Enforces code style (layout)
[X] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<template>
  <div>
    <p v-if="condition" 0abc>
      {{ content }}
    </p>
  </div>
</template>
```
Here the linter should warn about the attribute `0abc`, which is invalid and will cause a runtime error when vue attempts to insert it onto the dom element

```vue
<template>
  <div>
    <p v-if="condition" -->
      {{ content }}
    </p>
  </div>
</template>
```
Here the linter should warn about the attribute `--`, which is invalid and will cause a runtime error when vue attempts to insert it onto the dom element

**Additional context**
My team discovered this issue when a rogue `0` snuck its way through PR-review and onto an element behind a somewhat rare `v-if` condition. A linting rule like this would have immediately caught the problem.

ota-meshi:
Thank you for proposing this rule.

Do you know the specification of valid attribute names? If you know, please let me know. As far as I can see, `0abc` also looks like a valid attribute name. (However, it does not work in Chrome.)
https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

You can also report any attribute.
https://eslint.vuejs.org/rules/no-restricted-static-attribute.html

frederikadamse22:
Yes, during my research before suggesting this rule I did read the html spec. It's also interesting to note that the spec specifically mentions that ambigious ampersands eg. `&#169;` are not allowed as an attribute name, but this is not caught either.

frederikadamse22:
The WHATWG

frederikadamse22:
According to the specification, the HTML name **may** be written using anything from in the ascii range a-z, and A-Z, but nothing is more specific. `0abc `  throws an error when used in `setAttribute` but `ab0c` does not.

ota-meshi:
Could you provide a link to the specs?

frederikadamse22:
It's right there in the specs you linked, https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

[...]

PR #1851, opened 2022-04-16
Add `vue/no-invalid-attribute-name` rule

doug-wade:
Adds the rule proposed in #1373


lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,54 @@
...
+'use strict'
+
+const utils = require('../utils')
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'require the attributes to match the imported component name',
```
FloEdelmann, line 14:
```suggestion
        'report invalid attribute names',
```

Please run `npm run update` to update the docs afterwards.

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,54 @@
...
+     * @param {string | VIdentifier} key
+     * @return {string}
+     */
+    const getName = (key) => {
+      if (typeof key === 'string') {
+        return key
+      }
+
+      return key.name
+    }
```
FloEdelmann, line 36:
```suggestion
    const getName = (key) => (typeof key === 'string' ? key : key.name)
```

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,54 @@
...
+
+    return utils.defineTemplateBodyVisitor(context, {
+      VAttribute(node) {
```
FloEdelmann, line 39:
```suggestion
      /** @param {VNode} node */
      VAttribute(node) {
```

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,158 @@
...
+      filename: 'test.vue',
+      code: `
+      <template>
+        <div>
+          <p foo>
+            {{ content }}
+          </p>
+        </div>
+      </template>
+      `
```
FloEdelmann, line 30: Please simplify all test cases to something like this, since the content is not important to this rule:

```suggestion
      code: '<template><p foo /></template>'
```

docs/rules/no-invalid-attribute-name.md
```diff
@@ -0,0 +1,43 @@
...
+---
+# vue/no-invalid-attribute-name
+
+> require the attributes to match the imported component name
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule detects invalid html attributes.
```
FloEdelmann, line 15:
```suggestion
This rule detects invalid HTML attributes.
```

FloEdelmann (changes requested):

ota-meshi:
I don't think the issue has reached consensus. Would you please write your thoughts in the issue?

doug-wade:
Sorry @ota-meshi ; I thought the help wanted tag meant this was shovel-ready. I've updated the ticket with what I found in my investigation, and I'll reopen if we decide if the rule is something we want.

lib/rules/no-invalid-attribute-name.js
doug-wade: I tried to use this error message, but it doesn't seem to lint:

```
/path/to/eslint-plugin-vue/lib/rules/no-invalid-attribute-name.js
  14:20  error  `meta.docs.description` must match the regexp /^(enforce|require|disallow)/  eslint-plugin/require-meta-docs-description
```

I substituted a different one that does, but I am certainly open to changing it.

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @author Doug Wade <douglas.b.wade@gmail.com>
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+const utils = require('../utils')
+const xnv = require('xml-name-validator')
```
doug-wade, line 8: This line has a red squiggly in my text editor because there aren't types for it so I [opened a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59989) to add some

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,49 @@
...
+     */
+    const getName = (key) => (typeof key === 'string' ? key : key.name)
+
+    return utils.defineTemplateBodyVisitor(context, {
+      /** @param {VDirective | VAttribute} node */
+      VAttribute(node) {
+        const name = getName(node.key.name)
+
+        if (!xnv.name(name)) {
+          context.report({
```
ota-meshi, line 38: I think it is necessary to exclude the component from the elements to be checked. Attributes may work on components.

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * @param {string | VIdentifier} key
+     * @return {string}
+     */
+    const getName = (key) => (typeof key === 'string' ? key : key.name)
+
+    return utils.defineTemplateBodyVisitor(context, {
+      /** @param {VDirective | VAttribute} node */
+      VAttribute(node) {
+        const name = getName(node.key.name)
```
ota-meshi, line 35: We need to exclude the directive from the check or check if it is v-bind.

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+    {
+      filename: 'test.vue',
+      code: `<template><p !ghi /></template>`,
+      errors: [
+        {
+          message: 'Attribute name !ghi is not valid.'
+        }
+      ]
+    }
```
ota-meshi, line 76: If the rule also checks for directives, add tests for v-bind and tests for directives other than v-bind.

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,78 @@
...
+      code: `<template><p quux-.9 /></template>`
+    }
+  ],
+  invalid: [
+    {
+      filename: 'test.vue',
+      code: `<template><p 0abc /></template>`,
+      errors: [
+        {
+          message: 'Attribute name 0abc is not valid.'
```
ota-meshi, line 55: Add column and test the reported location.

ota-meshi (changes requested):
Thank you for changing this PR! I have some requests.

package.json
```diff
@@ -54,12 +54,14 @@
     "eslint": "^6.2.0 || ^7.0.0 || ^8.0.0"
   },
   "dependencies": {
+    "@types/xml-name-validator": "^4.0.0",
```
FloEdelmann, line 57: This should be a `devDependency`.

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,53 @@
...
+  create(context) {
+    /**
+     * @param {string | VIdentifier} key
+     * @return {string}
+     */
+    const getName = (key) => (typeof key === 'string' ? key : key.name)
+
+    return utils.defineTemplateBodyVisitor(context, {
+      /** @param {VDirective | VAttribute} node */
+      'VAttribute[directive=false]'(node) {
```
FloEdelmann, line 34: I think we should check the `v-bind` directive, i.e.

* `v-bind:href="…"` (valid)
* `:href="…"` (valid)
* `v-0abc:href="…"` (invalid)
* `:0abc="…"` (invalid)
* don't check `v-bind="…"`

Please also add tests for all these cases.

FloEdelmann (changes requested):

lib/rules/no-invalid-attribute-name.js
doug-wade: Thanks so much for including the test cases!

FloEdelmann: Ah, you're right, it seems that I double-clicked the wrong word :sweat_smile:

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,146 @@
...
+      code: `<a @[event]="doSomething"> ... </a>`
+    },
+    {
+      filename: 'test.vue',
+      code: `<template><div v-bind:href="..."></div></template>`
+    },
+    {
+      filename: 'test.vue',
+      code: `<template><div :href="..."></div></template>`
+    }
```
FloEdelmann, line 87: These cases are already by the test cases above (`<a :href="url">` and `<div v-bind:class="{ active: isActive }">`). But I'd add these test cases instead:

```suggestion
    {
      filename: 'test.vue',
      code: `<template><div v-bind="..."></div></template>`
    },
    {
      filename: 'test.vue',
      code: `<template><div v-0abc="..."></div></template>`
    }
```

FloEdelmann (approved):

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,70 @@
...
+    docs: {
+      description: 'require valid attribute names',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-invalid-attribute-name.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      attribute: 'Attribute name {{name}} is not valid.',
+      directive: 'Directive name {{name}} is not valid.'
```
ota-meshi, line 22: I think it's `'v-bind' argument`, not `Directive name`.

ota-meshi (changes requested):

lib/rules/no-invalid-attribute-name.js
ota-meshi: When use v-bind on native HTML element, it translates to an attribute, so I reconsidered that it would be better to use the same message as `attribute`.
What do you think?

ota-meshi, line 22:
```suggestion
      attribute: 'Attribute name {{name}} is not valid.'
```

lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,70 @@
...
+        if (
+          node.directive &&
+          name === 'bind' &&
+          node.key.argument &&
+          node.key.argument.type === 'VIdentifier' &&
+          !xnv.name(node.key.argument.name)
+        ) {
+          context.report({
+            node,
+            messageId: 'directive',
```
ota-meshi, line 51:
```suggestion
            messageId: 'attribute',
```

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,146 @@
...
+          column: 14
+        }
+      ]
+    },
+    {
+      filename: 'test.vue',
+      code: `<template><p v-bind:0abc=""></template>`,
+      errors: [
+        {
+          message: 'Directive name 0abc is not valid.',
```
ota-meshi, line 128:
```suggestion
          message: 'Attribute name 0abc is not valid.',
```

tests/lib/rules/no-invalid-attribute-name.js
```diff
@@ -0,0 +1,146 @@
...
+          column: 14
+        }
+      ]
+    },
+    {
+      filename: 'test.vue',
+      code: `<template><p :0abc="..." /></template>`,
+      errors: [
+        {
+          message: 'Directive name 0abc is not valid.',
```
ota-meshi, line 139:
```suggestion
          message: 'Attribute name 0abc is not valid.',
```

ota-meshi (changes requested):

ota-meshi (approved):
I changed this PR because I want to include this rule in the v9 release. Thank you for much work on this PR.

doug-wade:
@ota-meshi thank you so much for taking this one over the line for me. Life just got out of hand, and I really appreciate you getting this into v9.

Issue #1855, opened 2022-04-19
vue/define-macros-order


edikdeisling:
**Please describe what the rule should do:**

There are `defineEmits` and `defineProps` macros in script setup.
It would be nice to have a rule that moves these macros on top of the file.

**What category should the rule belong to?**

[x] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<!-- ✓ GOOD -->
<script setup>
defineEmits(/* ... */)
defineProps(/* ... */)
/* some code here */
</script>
```

```vue
<!-- ✗ BAD -->
<script setup>
/* some code here */
const bar = ref()
/* some code here */
defineEmits(/* ... */)
/* some code here */
defineProps(/* ... */)
</script>
```

**Additional context**

We write code this way in my company. Maybe it's a convenient way to define props/emits on the top.
If it would be helpful I'd like to create this eslint rule.

PR #1856, opened 2022-04-19
Add new rule: vue/define-macros-order (#1855)

edikdeisling:


tests/lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,275 @@
...
+        </script>
+      `,
+      output: `
+        <script lang="ts" setup>
+          const emit = defineEmits<{(e: 'update:test'): void}>()
+          const props = withDefaults(defineProps<Props>(), {
+            msg: 'hello',
+            labels: () => ['one', 'two']
+          })
+          interface Props {
```
edikdeisling, line 257: For now `defineProps` and `defineEmits` moves on the top.
Maybe it would be convenient to move it after top import statements?

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,248 @@
...
+  )
+
+  /**
+   * @param {ASTNode} shouldFirstNode
+   * @param {ASTNode} shouldSecondNode
+   * @param {ASTNode} before
+   */
+  function reportBothNotOnTop(shouldFirstNode, shouldSecondNode, before) {
+    context.report({
+      node: shouldFirstNode,
```
edikdeisling, line 141: Here we show error only at `shouldFirstNode`, but move both `shouldFirstNode` and `shouldSecondNode`.
Because of this RuleTester behavior described here(https://github.com/eslint/eslint/issues/11187) I can't make two errors with two fixes(I tried and had the same problem)

FloEdelmann: I think that's fine.

tests/lib/rules/define-macros-order.js
FloEdelmann: Yes, I think `interface` and `type` statements should be allowed before `defineProps` and `defineEmits`, just like `import` statements are. The auto-fix should also move `defineProps` and `defineEmits` below all (consecutive) `import`, `interface` and `type` statements.

Consecutive meaning that it should be moved here:
```ts
import bla from 'bla';
interface Foo {};
// <--- auto-fix should move `defineProps` here
const someOtherCode = '';
import 'foo' from 'bar'; // not idiomatic, but allowed
interface SomeOtherInterface {};
defineProps();
```

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,248 @@
...
+            },
+            uniqueItems: true,
+            additionalItems: false
+          }
+        },
+        additionalProperties: false
+      }
+    ],
+    messages: {
+      macrosNotOnTop: '{{macro}} must be on top.'
```
FloEdelmann, line 244:
```suggestion
      macrosNotOnTop: '{{macro}} should be the first statement in `<script setup>` (after any potential import statements or type definitions).'
```

FloEdelmann (changes requested):
Thank you for this suggestion and implementation. I've only done a preliminary review for now.

tests/lib/rules/define-macros-order.js
edikdeisling: Pushed this case(https://github.com/vuejs/eslint-plugin-vue/pull/1856/files#diff-6ebdd3c21be70e29ce94a2f399884fd220d4fe9bba358ade66b0bceb325d3102R280)

edikdeisling:
There are a few thoughts:
1. Maybe it would be better to have `defineEmitsFirst: true` rather than `order: ['defineEmits', 'defineProps']`? Or there are maybe some other `define*` that could come in?
2. Do we need to give an ability to specify a new line between these statements? `newLineBetween: true`

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,255 @@
...
+  const scriptSetup = utils.getScriptSetupElement(context)
+
+  if (!scriptSetup) {
+    return {}
+  }
+
+  const sourceCode = context.getSourceCode()
+  const options = context.options
+  /** @type {[string, string]} */
+  const order = (options[0] && options[0].order) || DEFAULT_ORDER
```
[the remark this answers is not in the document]
edikdeisling: There is some problem
image

We need to set `ecmaVersion: 2020` in `.eslintrc.js` to make this works.

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,255 @@
...
+  const macrosNodes = new Map()
+
+  return utils.compositingVisitors(
+    utils.defineScriptSetupVisitor(context, {
+      onDefinePropsExit(node) {
+        macrosNodes.set(MACROS_PROPS, getDefineMacrosStatement(node))
+      },
+      onDefineEmitsExit(node) {
+        macrosNodes.set(MACROS_EMITS, getDefineMacrosStatement(node))
+      }
```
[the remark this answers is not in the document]
edikdeisling: I've tested it. Seems that it's impossible
image

lib/rules/define-macros-order.js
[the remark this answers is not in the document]
FloEdelmann: Indeed. I've added this to https://github.com/vuejs/eslint-plugin-vue/pull/1847#issuecomment-1102700623. Then just ignore this comment :)

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,255 @@
...
+  const blockStatements = new Set([
+    'ImportDeclaration',
+    'TSInterfaceDeclaration',
+    'TSTypeAliasDeclaration',
+    'EmptyStatement'
+  ])
+
+  for (const [index, item] of program.body.entries()) {
+    if (!blockStatements.has(item.type)) {
+      return index
```
ota-meshi, line 37: I don't think most people will do that, but could you check directives like `"use strict"` just in case?

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * Get index of first new line before the "node"
+   * @param {ASTNode} node
+   * @return {number}
+   */
+  function getNewLineIndex(node) {
+    const after = sourceCode.getTokenBefore(node, { includeComments: true })
+    const hasWhitespace = node.loc.start.line - after.loc.end.line > 1
```
ota-meshi, line 210:
```suggestion
    const hasWhitespace = after.loc.end.line < node.loc.start.line
```

tests/lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,316 @@
...
+      parserOptions: {
+        parser: require.resolve('@typescript-eslint/parser')
+      },
+      errors: [
+        {
+          message: message('defineProps'),
+          line: 10
+        }
+      ]
+    }
```
ota-meshi, line 314: Add test with semicolon. I want to check if the semicolon moves correctly.

ota-meshi, line 314: Add test with comments. I think that the comment just before the define macro needs to be moved together with the define macro.

```ts
          console.log('test')
          /** Description for props */
          const props = withDefaults(defineProps<Props>(), {
            msg: 'hello'
          })
          console.log('test')
          /** Description for emit */
          const emit = defineEmits<{(e: 'test'): void}>()
```

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,255 @@
...
+ * Get an index of the first statement after imports and interfaces in order
+ * to place defineEmits and defineProps before this statement
+ * @param {Program} program
+ */
+function getTargetStatementPosition(program) {
+  const blockStatements = new Set([
+    'ImportDeclaration',
+    'TSInterfaceDeclaration',
+    'TSTypeAliasDeclaration',
+    'EmptyStatement'
```
ota-meshi, line 32: I think it's probably better to add `'DebuggerStatement'` as well.

ota-meshi (changes requested):
Thank you for this PR! I wrote some comments.

tests/lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,389 @@
...
+          const props = defineProps({ test: Boolean });        </script>
+      `,
+      options: optionsEmitsFirst,
+      errors: [
+        {
+          message: message('defineEmits'),
+          line: 3
+        }
+      ]
+    }
```
edikdeisling, line 387: @ota-meshi @FloEdelmann 
There is the moment I need help.
Need to decide what we "move".
image

1. We can move only the node itself(without newlines).
This breaks the code formatting almost at every case
```vue
<-- BEFORE -->
<script setup>
    defineEmits(['update:test']);
    const props = defineProps({ test: Boolean });
    console.log('test')
</script>

<-- AFTER -->
<script setup>
    const props = defineProps({ test: Boolean });defineEmits(['update:test']);

    console.log('test')
</script>
```

2. We can move node with space before and space with a new line after
In most cases, this behavior doesn't break formatting drastically but does make some double new lines
```vue
<-- BEFORE -->
<script setup>
    defineEmits(['update:test'])

    console.log('test')

    const props = defineProps({ test: Boolean })

    console.log('test2')
</script>

<-- AFTER -->
<script setup>
    const props = defineProps({ test: Boolean })
    defineEmits(['update:test'])
    
    console.log('test')


    console.log('test2')
</script>
```

3. Cut node with newlines before\after
This thing has the same downsides as 2nd

I took the 2nd variant. I.e. there would be cases where the result code contains multiple new lines

There is an option to make some heuristics about the newlines, but it feels that this is not the right direction

lib/rules/define-macros-order.js
edikdeisling: I removed this heuristic about new lines. Here are the details: https://github.com/vuejs/eslint-plugin-vue/pull/1856#discussion_r854017385

tests/lib/rules/define-macros-order.js
ota-meshi: How about moving from the start position of the node to the position immediately before the next token?

```vue
<script setup>
    defineEmits(['update:test'])

    console.log('test')

    |<const props = defineProps({ test: Boolean })

    >|console.log('test2')
</script>
```


```vue
<script setup>
    |<const props = defineProps({ test: Boolean })

    >|defineEmits(['update:test'])

    console.log('test')

    console.log('test2')
</script>
```

docs/rules/define-macros-order.md
```diff
@@ -0,0 +1,72 @@
...
+{
+  "vue/define-macros-order": ["error", {
+    "order": ["defineEmits", "defineProps"]
+  }]
+}
+```
+
+- `order` (`string[]`) ... The order of defineEmits and defineProps macros
+
+### `{ "order": ["defineEmits", "defineProps"] }` (default)
```
ota-meshi, line 30: The default order has changed.

tests/lib/rules/define-macros-order.js
edikdeisling: There would be problem with this case
```vue
<script setup>
    defineEmits(['update:test'])

    console.log('test')

    |<const props = defineProps({ test: Boolean })
>|</script>
```

Here is the expected behavior:
https://user-images.githubusercontent.com/41228762/164386145-98ae4695-29a8-4023-9f63-e792141165c5.mp4
It seems that it's too hard to move every node one by one...

It feels that we need smth like this:
1. Cut node without newlines
2. Remove all empty lines after the node
3. Paste the node on the top with the same new lines that the top node have

I'll try to write it. Sounds not so fearful...

ota-meshi: Oh. I didn't consider that case.

I think it might work if you cut from immediately after the previous token to the end of the node, and move it immediately after the token before the insert node. What do you think?
It may look a bit out of format because it looks like more line breaks at the beginning, but I don't think it's a problem for the scope of auto-fix.
I want to avoid leaving new multiple blank lines as much as possible.

```vue
<script setup>|target
    defineEmits(['update:test'])

    console.log('test')|<

    const props = defineProps({ test: Boolean })>|
</script>

<script setup>|<

    const props = defineProps({ test: Boolean })>|
    defineEmits(['update:test'])

    console.log('test')
</script>
```

```vue
<script setup>|target
    defineEmits(['update:test'])

    console.log('test')|<

    const props = defineProps({ test: Boolean })>|

    console.log('test2')
</script>

<script setup>|<

    const props = defineProps({ test: Boolean })>|
    defineEmits(['update:test'])

    console.log('test')

    console.log('test2')
</script>
```

ota-meshi: > Cut node without newlines
Remove all empty lines after the node
Paste the node on the top with the same new lines that the top node have

Oh. I now understand this. It may work.

tests/lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,392 @@
...
+            test: Boolean
+          })
+
+          defineEmits(['update:test'])
+
+          console.log('test1')
+          console.log('test2')
+          console.log('test3')
+        </script>
+      `,
```
edikdeisling, line 165: @ota-meshi here is the next iteration :)
This logic introduces new lines(but only if these new lines were before)

edikdeisling: For me, it feels more natural variant. Because it keeps a new line between import and next token as it was before the rearrangement. 
But it may introduce newline between defineProps and defineEmits if they moved in one time

lib/rules/define-macros-order.js
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * @param {ASTNode} tokenBefore
+   * @param {ASTNode} tokenAfter
+   */
+  function getTextBetweenTokens(tokenBefore, tokenAfter) {
+    const delta = tokenBefore.range[1] - tokenBefore.range[0]
+    const count = tokenAfter.range[0] - tokenBefore.range[1]
+
+    return sourceCode.getText(tokenBefore, -delta, count)
```
ota-meshi, line 238: It may be easier to get it directly from `text`.

```suggestion
    return sourceCode.text.slice(tokenBefore.range[1], tokenAfter.range[0])
```

tests/lib/rules/define-macros-order.js
edikdeisling: Maybe we should try your variant... Don't know) Both have tradeoffs

ota-meshi (approved):
LGTM! Thank you!

FloEdelmann (approved):

edikdeisling:
@ota-meshi I found a bug :(
This plugin breaks code with `<script>` and `<script setup>` tags in one file.

I'll create PR soon

PR #1858, opened 2022-04-21
Improve warning wording in `vue/no-v-for-template-key` docs

yammyxing:
fix the warning note on rule conflicts between  vue/no-v-for-template-key and vue/no-v-for-template-key-on-child

### DIFF
--- a/docs/rules/no-v-for-template-key.md
+++ b/docs/rules/no-v-for-template-key.md
@@ -18,8 +18,8 @@ This rule reports the `<template v-for>` elements which have `key` attribute.
 In Vue.js 2.x, disallows `key` attribute on `<template>` elements.
 
 ::: warning Note
-Do not use with the [vue/no-v-for-template-key] rule for Vue.js 3.x.  
-This rule conflicts with the [vue/no-v-for-template-key-on-child] rule.
+This rule is targeted at Vue.js 2.x.
+If you are using Vue.js 3.x, enable the [vue/no-v-for-template-key-on-child] rule instead. Don't enable both rules together; they are conflicting.
 :::
 
 <eslint-code-block :rules="{'vue/no-v-for-template-key': ['error']}">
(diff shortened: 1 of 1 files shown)


ota-meshi (changes requested):
I think the original note is correct.
However, I am not familiar with English, so I may have misunderstood.
Could you give me the details on this fix?

docs/rules/no-v-for-template-key.md
```diff
@@ -18,7 +18,7 @@ This rule reports the `<template v-for>` elements which have `key` attribute.
 In Vue.js 2.x, disallows `key` attribute on `<template>` elements.
 
 ::: warning Note
-Do not use with the [vue/no-v-for-template-key-on-child] rule for Vue.js 3.x.  
+Do not use with the [vue/no-v-for-template-key] rule for Vue.js 3.x.  
 This rule conflicts with the [vue/no-v-for-template-key-on-child] rule.
```
FloEdelmann, line 22:
```suggestion
This rule is targeted at Vue.js 2.x.
If you are using Vue.js 3.x, enable the [vue/no-v-for-template-key-on-child] rule instead. Don't enable both rules together; they are conflicting.
```

FloEdelmann (changes requested):
I agree with @ota-meshi, the previous note was correct. But it was difficult to understand, so here is my approach to clear that up. See also the similar https://github.com/vuejs/eslint-plugin-vue/pull/1859#discussion_r855007958

yammyxing:
ok, the approach is clear enough.

yammyxing:
> I think the original note is correct. However, I am not familiar with English, so I may have misunderstood. Could you give me the details on this fix?

seems like i get the meaning of 'use with' wrong, my bad [Lol]

docs/rules/no-v-for-template-key.md
yammyxing: ok, the approach is clear enough.

FloEdelmann (approved):

PR #1860, opened 2022-04-22
Improved rule list doc

ota-meshi:
This PR improves the rule list documentation.

There are quite a lot of rules now. I have made the following changes to make the rules easier for users to access.

- Changed layout so that rules do not duplicate in multiple tables.
- Placed icon on the rule to indicate the type of the rule.
- Added checkboxes so that user can highlight by the type of rule.

https://deploy-preview-1860--eslint-plugin-vue.netlify.app/rules/


tools/lib/utils.js
```diff
@@ -0,0 +1,36 @@
...
+  'vue3-recommended': 'vue3-strongly-recommended'
+  // 'use-with-caution': 'recommended',
+  // 'vue3-use-with-caution': 'vue3-recommended'
+}
+
+function formatItems(items) {
+  if (items.length <= 2) {
+    return items.join(' and ')
+  }
+  return `all of ${items.slice(0, -1).join(', ')} and ${last(items)}`
```
FloEdelmann, line 21: With `[...items].pop()` or `items.at(-1)`, we could avoid the lodash dependency.

tools/lib/utils.js
```diff
@@ -0,0 +1,36 @@
...
+}
+function getPresetIds(categoryIds) {
+  const subsetCategoryIds = []
+  for (const categoryId of categoryIds) {
+    for (const subsetCategoryId in presetCategories) {
+      if (presetCategories[subsetCategoryId] === categoryId) {
+        subsetCategoryIds.push(subsetCategoryId)
+      }
+    }
```
FloEdelmann, line 30:
```suggestion
    for (const [subsetCategoryId, supersetCategoryId] of Object.entries(presetCategories)) {
      if (supersetCategoryId === categoryId) {
        subsetCategoryIds.push(subsetCategoryId)
      }
    }
```

tools/update-docs-rules-index.js
```diff
@@ -46,25 +62,95 @@ function toDeprecatedRuleRow(rule) {
...
+  .map((rule) => {
+    const mark = group.useMark
+      ? [
+          rule.meta.docs.categories.includes(group.categoryIdForVue3)
+            ? [':green_heart:']
+            : [],
+          rule.meta.docs.categories.includes(group.categoryIdForVue2)
+            ? [':yellow_heart:']
+            : []
+        ].flatMap((e) => e)
```
FloEdelmann, line 143:
```suggestion
        ].flat()
```

FloEdelmann (changes requested):
Really cool changes!

Maybe the layout of only this docs page can be adjusted, so it is wider (on screens with more horizontal space) and thus better uses the available space?

tools/update-docs-rules-index.js
```diff
@@ -46,25 +62,95 @@ function toDeprecatedRuleRow(rule) {
...
+${rules
+  .map((rule) => {
+    const mark = group.useMark
+      ? [
+          rule.meta.docs.categories.includes(group.categoryIdForVue3)
+            ? [':green_heart:']
+            : [],
+          rule.meta.docs.categories.includes(group.categoryIdForVue2)
+            ? [':yellow_heart:']
+            : []
```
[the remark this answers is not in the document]
ota-meshi: I tried it once and found it easier to see in different colors. However, since I added the highlight tool after that, it may not be a problem anymore. I will try it.

tools/lib/utils.js
```diff
@@ -0,0 +1,36 @@
...
+  // 'use-with-caution': 'recommended',
+  // 'vue3-use-with-caution': 'vue3-recommended'
+}
+
+function formatItems(items) {
+  if (items.length <= 2) {
+    return items.join(' and ')
+  }
+  return `all of ${items.slice(0, -1).join(', ')} and ${last(items)}`
+}
```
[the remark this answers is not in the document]
ota-meshi: I noticed that this function is used elsewhere. I will do it in another way.

ota-meshi:
I think I made the changes you requested. Could you check it again?

docs/.vuepress/components/rules-table.vue
```diff
@@ -0,0 +1,90 @@
...
+      </label>
+    </div>
+    <div class="table-root" ref="tableRoot">
+      <slot />
+    </div>
+  </div>
+</template>
+
+<script>
+const emojis = ['3️⃣', '2️⃣', '⚠️']
```
ota-meshi, line 17: `Intl.Segmenter` is required to properly split these emojis, but FireFox doesn't support it, so we'll handle it ourselves.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter

FloEdelmann (approved):

PR #1861, opened 2022-04-22
vue/define-macros-order bug

edikdeisling:
Fix bugs from https://github.com/vuejs/eslint-plugin-vue/pull/1856

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -20,6 +20,17 @@ const ORDER = [MACROS_EMITS, MACROS_PROPS]
 const DEFAULT_ORDER = [MACROS_PROPS, MACROS_EMITS]
 
 /**
+ * @param {VElement} scriptSetup
+ * @param {ASTNode} node
+ */
+function inScriptSetup(scriptSetup, node) {
+  return (
+    scriptSetup.range[0] <= node.range[0] &&
+    node.range[1] <= scriptSetup.range[1]
+  )
+}
+
+/**
  * @param {ASTNode} node
  */
 function isUseStrictStatement(node) {
@@ -33,9 +44,10 @@ function isUseStrictStatement(node) {
 /**
  * Get an index of the first statement after imports and interfaces in order
  * to place defineEmits and defineProps before this statement
+ * @param {VElement} scriptSetup
  * @param {Program} program
  */
-function getTargetStatementPosition(program) {
+function getTargetStatementPosition(scriptSetup, program) {
   const skipStatements = new Set([
     'ImportDeclaration',
     'TSInterfaceDeclaration',
@@ -45,7 +57,11 @@ function getTargetStatementPosition(program) {
   ])
 
   for (const [index, item] of program.body.entries()) {
-    if (!skipStatements.has(item.type) && !isUseStrictStatement(item)) {
+    if (
+      inScriptSetup(scriptSetup, item) &&
+      !skipStatements.has(item.type) &&
+      !isUseStrictStatement(item)
+    ) {
       return index
     }
   }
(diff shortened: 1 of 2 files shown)


lib/rules/define-macros-order.js
```diff
@@ -45,7 +57,11 @@ function getTargetStatementPosition(program) {
   ])
 
   for (const [index, item] of program.body.entries()) {
-    if (!skipStatements.has(item.type) && !isUseStrictStatement(item)) {
+    if (
+      inScriptSetup(scriptSetup, item) &&
```
edikdeisling, line 61: Script setup stuff fixes multiple scripts bug(test is here: https://github.com/vuejs/eslint-plugin-vue/pull/1861/files#diff-6ebdd3c21be70e29ce94a2f399884fd220d4fe9bba358ade66b0bceb325d3102R387)

lib/rules/define-macros-order.js
```diff
@@ -213,13 +232,12 @@ function create(context) {
     const targetComment = sourceCode.getTokenAfter(beforeTargetToken, {
       includeComments: true
     })
-    const textSpace = getTextBetweenTokens(beforeTargetToken, targetComment)
```
edikdeisling, line 216: This fix fixes these newlines https://github.com/vuejs/eslint-plugin-vue/pull/1861/files#diff-6ebdd3c21be70e29ce94a2f399884fd220d4fe9bba358ade66b0bceb325d3102L158

lib/rules/define-macros-order.js
```diff
@@ -228,17 +246,28 @@ function create(context) {
...
+  function getInsertText(textNode, target) {
+    const afterTargetComment = sourceCode.getTokenAfter(target, {
+      includeComments: true
+    })
+    const afterText = sourceCode.text.slice(
+      target.range[1],
+      afterTargetComment.range[0]
+    )
+    // handle case when a();b() -> b()a();
+    const invalidResult = !textNode.endsWith(';') && !/[\n;]/.test(afterText)
```
edikdeisling, line 262: This is the third corner case I found

edikdeisling:
I ran it in my project and the result is great. The only annoying thing left is this:
image

There is an option to introduce "newLineAfterDefineMacros: true" option maybe... For now, I've done nothing

lib/rules/define-macros-order.js
FloEdelmann, line 262: I guess it should check the start only, right?

```suggestion
    const invalidResult = !textNode.endsWith(';') && !/^[\n;]/.test(afterText)
```

lib/rules/define-macros-order.js
ota-meshi: Hmmm. I think this change is an improvement. I don't think this should be included in the patch, so could you revert this change in this PR and open another PR?

ota-meshi (changes requested):

lib/rules/define-macros-order.js
edikdeisling: No, this value also get valid result
```ts
const afterText = '   ;   '
```

edikdeisling:
@ota-meshi sure

lib/rules/define-macros-order.js
FloEdelmann: But wouldn't it also return true for `b();c()`? Or `(b();)`?

ota-meshi:
If you write an explanation in the PR, you do not need to open an issue.

ota-meshi (approved):
Thank you!

lib/rules/define-macros-order.js
edikdeisling: The thing here is what is inside afterText. afterText contains text between the node and the next comment. So it can contain only whitespaces and new lines.
While I was writing this I understood that there is no need to look for `;`. There is no change afterText can contain it :)

edikdeisling: hm

edikdeisling: maybe i need to remove `;` from regexp

edikdeisling: Changed it here https://github.com/vuejs/eslint-plugin-vue/pull/1863/files#diff-8293acf6fd19c2affb306c4d724be1279c64123427e207a444b5ecdab4789940R262

PR #1863, opened 2022-04-22
Improve auto-fix for `vue/define-macros-order` rule

edikdeisling:
In PR https://github.com/vuejs/eslint-plugin-vue/pull/1861 was decided to create separate PR for these improvements

There are two fixes here:
1. Fix extra newlines appear
2. Fix corner case `a();b() -> b()a();`

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -259,7 +259,7 @@ function create(context) {
       afterTargetComment.range[0]
     )
     // handle case when a();b() -> b()a();
-    const invalidResult = !textNode.endsWith(';') && !/[\n;]/.test(afterText)
+    const invalidResult = !textNode.endsWith(';') && !afterText.includes('\n')
 
     return textNode + afterText + (invalidResult ? ';' : '')
   }
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -232,13 +232,12 @@ function create(context) {
     const targetComment = sourceCode.getTokenAfter(beforeTargetToken, {
       includeComments: true
     })
-    const textSpace = getTextBetweenTokens(beforeTargetToken, targetComment)
     // make insert text: comments + node + space before target
     const textNode = sourceCode.getText(
       node,
       node.range[0] - nodeComment.range[0]
     )
-    const insertText = textNode + textSpace
+    const insertText = getInsertText(textNode, target)
 
     return [
       fixer.insertTextBefore(targetComment, insertText),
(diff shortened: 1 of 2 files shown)


lib/rules/define-macros-order.js
```diff
@@ -232,13 +232,12 @@ function create(context) {
     const targetComment = sourceCode.getTokenAfter(beforeTargetToken, {
       includeComments: true
     })
-    const textSpace = getTextBetweenTokens(beforeTargetToken, targetComment)
```
edikdeisling, line 235: This thing fixes these newlines https://github.com/vuejs/eslint-plugin-vue/pull/1861/files#diff-6ebdd3c21be70e29ce94a2f399884fd220d4fe9bba358ade66b0bceb325d3102L158

Before we took space **before** the target token and added it after the moving token.
Now we take space **after** the target token and added it after the moving token.

lib/rules/define-macros-order.js
```diff
@@ -247,17 +246,28 @@ function create(context) {
...
+  function getInsertText(textNode, target) {
+    const afterTargetComment = sourceCode.getTokenAfter(target, {
+      includeComments: true
+    })
+    const afterText = sourceCode.text.slice(
+      target.range[1],
+      afterTargetComment.range[0]
+    )
+    // handle case when a();b() -> b()a();
+    const invalidResult = !textNode.endsWith(';') && !afterText.includes('\n')
```
edikdeisling, line 262: This is corner case `a();b() -> b()a();`.
`afterText` contains symbols between target token(`a();`) and next token(`b()`). If there are no new lines and `b()` doesn't ends with `;` - we add `;` manually

edikdeisling:
I ran it in my project and the result was great. The only annoying thing left is this:
image
There is an option to introduce "newLineAfterDefineMacros: true" option maybe... For now, I've done nothing

lib/rules/define-macros-order.js
FloEdelmann: Can you please add a test case for this edge case?

FloEdelmann (changes requested):

lib/rules/define-macros-order.js
edikdeisling: Sure. Done

FloEdelmann (approved):
Thanks!

ota-meshi (approved):
LGTM!

ota-meshi:
> There is an option to introduce "newLineAfterDefineMacros: true" option maybe... For now, I've done nothing

I think we shouldn't add options for auto-fix. Because the aims of the rule is to check.

Issue #1865, opened 2022-04-23
`vue/sort-keys runOutsideVue` option


lukethompson:
**What rule do you want to change?**
`vue/sort-keys` add option to give the ability to only run inside of Vue. My team likes to alphabetize our `props`, `computed`, `methods` etc... but we do not want to enforce all object keys being sorted.

`vue/sort-keys` does already have a `runOutsideVue` option but it is not used in the rule.

**Does this change cause the rule to produce more or fewer warnings?**
Fewer if the option is set.

**How will the change be implemented? (New option, new default behavior, etc.)?**
`vue/sort-keys` already has `options.runOutsideVue`, the default value is `true` passing `false` doesn't do anything. The change will use the value passed by this option to allow users to ignore sorting outside of vue.

**Please provide some example code that this change will affect:**

```vue
// runOutsideVue: false

export default {
    methods: {
       // Will sort keys inside of `methods`, `props`, `computed` etc...
        createArticle () {
            return this.createEntity({
                // Will ignore keys not on a Vue instance
                id: 1,
                type: 'article',
                attributes: { createdAt: new Date() }
            })
        },
        deleteArticle (id) {}
    },
}
```

**What does the rule currently do for this code?**
`options.runOutsideVue` exists but is not used. Currently all object keys are sorted by this rule. 

**What will the rule do after it's changed?**
Use the `options.runOutsideVue` option, currently is defaulted to `true` so this change will have no impact on existing uses.

**Additional context**
This change will somewhat address the issue here https://github.com/vuejs/eslint-plugin-vue/issues/630

I have already implemented this change and will link the PR below.

PR #1866, opened 2022-04-23
Add runOutsideVue option to vue/sort-keys

lukethompson:
Address issue raised here https://github.com/vuejs/eslint-plugin-vue/issues/1865

Update `vue/sort-keys` to use existing `options.runOutsideVue` option, it seems like previously this option did not do anything.

If this change gets accepted we should update the docs to state which version this option will be available from 😄

### DIFF
--- a/tests/lib/rules/sort-keys.js
+++ b/tests/lib/rules/sort-keys.js
@@ -510,7 +510,7 @@ ruleTester.run('sort-keys', rule, {
 
     // runOutsideVue (false) should ignore unsorted keys outside of vue
     {
-      code: 'var obj = { c:3, a:1, b:2}',
+      code: 'var obj = {c:3, a:1, b:2}',
       options: ['asc', { runOutsideVue: false }]
     },
     {
--- a/tests/lib/rules/sort-keys.js
+++ b/tests/lib/rules/sort-keys.js
@@ -506,6 +506,50 @@ ruleTester.run('sort-keys', rule, {
     {
       code: 'var obj = {a:1, _:2, b:3}',
       options: ['desc', { natural: true, caseSensitive: false, minKeys: 4 }]
+    },
+
+    // runOutsideVue (false) should ignore unsorted keys outside of vue
+    {
+      code: 'var obj = { c:3, a:1, b:2}',
+      options: ['asc', { runOutsideVue: false }]
+    },
+    {
+      filename: 'test.js',
+      code: `
+        const { component } = Vue;
+        component('test', {
+          name: 'app',
+          data () {
+            return {
+              c: 3,
+              a: 1,
+              b: 2
+            }
+          }
+        })
+      `,
+      parserOptions: { ecmaVersion: 6 },
+      options: ['asc', { runOutsideVue: false }]
+    },
+    {
+      filename: 'test.js',
+      code: `
+        const { component } = Vue;
+        component('test', {
+          name: 'app',
+          computed: {
+            test () {
+              return {
+                c: 3,
+                a: 1,
+                b: 2
+              }
+            }
+          }
+        })
+      `,
+      parserOptions: { ecmaVersion: 6 },
+      options: ['asc', { runOutsideVue: false }]
     }
   ],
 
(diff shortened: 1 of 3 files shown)


Mighty-Tulip (approved):

FloEdelmann (approved):
Looks good! Interesting that the option existed before but was unused and undocumented.

I don't think we need to specify in the docs the version in which this "new" option is released, since it wasn't documented before.

tests/lib/rules/sort-keys.js
```diff
@@ -506,6 +506,50 @@ ruleTester.run('sort-keys', rule, {
...
+      code: `
+        const { component } = Vue;
+        component('test', {
+          name: 'app',
+          data () {
+            return {
+              c: 3,
+              a: 1,
+              b: 2
+            }
```
ota-meshi, line 527: Thank you for this PR!
Why is this not reported? I think `a`, `b`, and `c` are Vue component properties and should be sorted.

ota-meshi (changes requested):

tests/lib/rules/sort-keys.js
lukethompson: Thank you for the review.

I have pushed a change that will ensure objects returned by the `data()` method are sorted.

Just clarifying here, we only want to sort the keys of `data()` objects and we can leave objects defined within `computed`, `watchers` and `methods` unsorted?

If that is the case then this is all good.

If we want to also sort all new objects defined within a vue instance but not directly attached to the component maybe we should look towards implementing a new option that only sorts keys that are directly attached to the vue instance.

For example
```
...
computed: {
  // I would like the computed keys to be sorted.
  a () {
    // But I do not want to sort new objects defined within my computed.
     return {
       b: null,
       a: null,
     }
  },
}
```

ota-meshi: Sorry for late reply.

I think this rule should sort the keys of all objects that aren't reserved as Vue component definitions.
Therefore, I think the following `b` and `a` should also be sorted. Because they are *inside* the Vue component.

```vue
computed: {
  a () {
     return {
       b: null,
       a: null,
     }
  },
}
```

If you want to sort only the properties of Vue components (e.g. props, data, computed, watch, and methods),  I think we can add another rule to exclude it from being checked by the `vue/sort-keys` rule.
I think it would be better to have a separate rule for sorting related to Vue components than `vue/sort-keys` rule,  like `vue/order-in-components` rule.
https://eslint.vuejs.org/rules/order-in-components.html
What do you think?

lukethompson: Hello, sorry for the slow reply, been very busy with work lately.

This rule already accepts an `ignoreGrandchildrenOf` option that might be applicable here, iirc it only considers grandchildren keys of the same object, not newly created objects. Let me confirm this behaviour when I get a chance and maybe we could look towards making that ignore all grandchildren, not just grandchildren that are directly attached to the grandparent key.

ota-meshi: The ignoreGrandchildrenOf option is an option to avoid conflicting with ordering conventions with Vue components.    
If we need an option to ignore all grandchildren, I don't think that option should ignore the outside of the Vue component.  

I think `vue/sort-keys` rule should have a role to enforce an order other than the Vue component ordering convention.  
I think the order of Vue properties is the ordering convention of Vue components, so I think we need another rule.

I think `runOutsideVue=false` is just an option to ignore the outside of Vue. (However, I'm not sure if it's necessary for the user.)

I think it should be another rule if you want to enforce the order of Vue properties.
Where do you want to enforce the order?

FloEdelmann:
Ping @lukethompson

lukethompson:
> Ping @lukethompson

Sorry @FloEdelmann – I have been too busy to pick this one back up. Will close the PR for now and re-open when I get a chance to work on this again 😃

tests/lib/rules/sort-keys.js
lukethompson: @ota-meshi I was hoping to enforce order on all `props`, `computed` props, `methods` and `watch` child keys, eg.

```
export default {
  // These keys would not be enforced by this rule/option combination.
  name: 'ComponentName',

  props: {
    // These keys order would be enforced by this rule/option combination
    a: { type: String, default: null },
    b: { type: String, default: null },
  },
  
  computed: {
    c () {
      return {
        // These keys would not be enforced by this rule/option combination.
        e: null,
        d: null,
      }
    }
  },
}
```
I agree, I believe a new rule would be more appropriate for this. I have closed this PR for now and if I get a chance in the future I will submit a new PR, thanks.

PR #1871, opened 2022-04-26
Improve `vue/multi-word-component-names` rule docs

zjfresh:
Rule `vue/multi-word-component-names`

### DIFF
--- a/docs/rules/multi-word-component-names.md
+++ b/docs/rules/multi-word-component-names.md
@@ -76,6 +76,47 @@ export default {
 
 </eslint-code-block>
 
+<eslint-code-block filename="src/Todo.vue" :rules="{'vue/multi-word-component-names': ['error']}">
+
+```vue
+<!-- filename: Todo.vue -->
+<!-- ✗ BAD -->
+<script setup>
+  // ...
+</script>
+```
+
+</eslint-code-block>
+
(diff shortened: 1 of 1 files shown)


4tj:
I think this rule is not friendly, component naming good or bad should be **directly** decided by Vue? Enable by default and throw an Error. 

Please google how many developers it causes trouble

FloEdelmann:
You are free to disable this rule. But it is the first recommendation in the Style Guide, so the rule makes sense: https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names

4tj:
> You are free to disable this rule. But it is the first recommendation in the Style Guide, so the rule makes sense: https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names

So why create a rule with the highest **disable rate** and enable it by default?

FloEdelmann:
Because it is in the style guide and prevents errors. Note that you can also ignore single component names:

https://eslint.vuejs.org/rules/multi-word-component-names.html#options

4tj:
I know it can be disabled and just saying it's BAD design. 

At least it should just be a warning

FloEdelmann:
That is something that can be discussed. Please open a new issue for that, with some evidence that this rule is really often disabled.

PR #1874, opened 2022-04-29
feat: fixable match-component-file-name rule

### LINKED ISSUE #1816: Add suggestion(s) to `vue/match-component-file-name`
[vue/match-component-file-name](https://eslint.vuejs.org/rules/match-component-file-name.html)

Same.

New [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) that can be applied by the user via the editor:

* "Rename component to match file name." (without file extension)
* "Rename file to match component name." (not sure if ESLint suggestions can rename files though)

Note that we can't add an auto-fix here because the user intent is not clear.

```js
// file name: "MySuperCoolComponent.vue"
export default {
  name: 'MyAwesomeComponent',
}
```

Report, but no auto-fix and no suggestions.

Report, still no auto-fix, but suggestions.

* Rename component to match file name.
  ```js
  // file name: "MySuperCoolComponent.vue"
  export default {
    name: 'MySuperCoolComponent',
  }
  ```
* Rename file to match component name.
  ```js
  // file name: "MyAwesomeComponent.vue"
  export default {
    name: 'MyAwesomeComponent',
  }
  ```

mahaker:
fix #1816 

This plugin has always helped me.
Thank you very much.

There are two fixes.

1. Add 'Literal#raw' property
2. fixable match-component-file-name rule

### DIFF
--- a/lib/rules/match-component-file-name.js
+++ b/lib/rules/match-component-file-name.js
@@ -31,6 +31,8 @@ function canVerify(node) {
 
 module.exports = {
   meta: {
+    // eslint-disable-next-line eslint-plugin/require-meta-has-suggestions
+    hasSuggestions: true,
     type: 'suggestion',
     docs: {
       description: 'require component name property to match its file name',
@@ -114,7 +116,17 @@ module.exports = {
           node,
           message:
             'Component name `{{name}}` should match file name `{{filename}}`.',
-          data: { filename, name }
+          data: { filename, name },
+          suggest: [
+            {
+              desc: 'Rename component to match file name.',
+              fix(fixer) {
+                const quote =
+                  node.type === 'TemplateLiteral' ? '`' : node.raw[0]
+                return fixer.replaceText(node, `${quote}${filename}${quote}`)
+              }
+            }
+          ]
         })
       }
     }
--- a/tests/lib/rules/match-component-file-name.js
+++ b/tests/lib/rules/match-component-file-name.js
@@ -564,7 +564,18 @@ ruleTester.run('match-component-file-name', rule, {
       errors: [
         {
           message:
-            'Component name `MComponent` should match file name `MyComponent`.'
+            'Component name `MComponent` should match file name `MyComponent`.',
+          suggestions: [
+            {
+              desc: 'Rename component to match file name.',
+              output: `
+        export default {
+          name: 'MyComponent',
+          render() { return <div /> }
+        }
+      `
+            }
+          ]
         }
       ]
     },
@@ -581,7 +592,46 @@ ruleTester.run('match-component-file-name', rule, {
       errors: [
         {
           message:
-            'Component name `MComponent` should match file name `MyComponent`.'
+            'Component name `MComponent` should match file name `MyComponent`.',
+          suggestions: [
+            {
+              desc: 'Rename component to match file name.',
+              output: `
+        export default {
+          name: 'MyComponent',
+          render() { return <div /> }
+        }
+      `
+            }
+          ]
+        }
+      ]
+    },
+    {
+      filename: 'MyComponent.jsx',
+      code: `
+        export default {
+          name: "MComponent",
+          render() { return <div /> }
+        }
+      `,
+      options: [{ extensions: ['jsx'] }],
+      parserOptions: jsxParserOptions,
+      errors: [
+        {
+          message:
+            'Component name `MComponent` should match file name `MyComponent`.',
+          suggestions: [
+            {
+              desc: 'Rename component to match file name.',
+              output: `
+        export default {
+          name: "MyComponent",
+          render() { return <div /> }
+        }
+      `
+            }
+          ]
         }
       ]
     },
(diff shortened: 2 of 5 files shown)


lib/rules/match-component-file-name.js
```diff
@@ -102,19 +102,25 @@ module.exports = {
      */
     function verifyName(node) {
       let name
+      let quote = ''
```
mahaker, line 105: Defining  `let quote`, resulted in a typescript error. 

```
Variable 'quote' implicitly has type 'any' in some locations where its type cannot be determined.
```

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
```diff
@@ -347,6 +347,7 @@ export interface PrivateIdentifier extends HasParentNode {
 export interface Literal extends HasParentNode {
   type: 'Literal'
   value: string | boolean | null | number | RegExp | BigInt
+  raw: string
```
mahaker, line 350: `console.log(node)`, the following object can be displayed.
The fixer use the `raw` property to determine the single/doublequotes.

```sh
<ref *2> Node {
  type: 'Literal',
  start: 38,
  end: 51,
  loc: SourceLocation {
    start: { line: 3, column: 10 },
    end: { line: 3, column: 23 }
  },
  range: [ 38, 51 ],
  value: 'MyComponent',
  raw: "'MyComponent'",
  parent: <ref *1> Node {
    type: 'Property',
    start: 32,
    end: 51,
    loc: SourceLocation { start: [Object], end: [Object] },
    range: [ 32, 51 ],
    method: false,
    shorthand: false,
    computed: false,
    key: Node {
      type: 'Identifier',
      start: 32,
      end: 36,
      loc: [SourceLocation],
      range: [Array],
      name: 'name',
      parent: [Circular *1]
    },
    value: [Circular *2],
    kind: 'init',
    parent: Node {
      type: 'ObjectExpression',
      start: 26,
      end: 110,
      loc: [SourceLocation],
      range: [Array],
      properties: [Array],
      parent: [Node]
    }
  }
}
```

lib/rules/match-component-file-name.js
```diff
@@ -102,19 +102,25 @@ module.exports = {
...
       if (!compareNames(name, filename)) {
         errors.push({
           node,
           message:
             'Component name `{{name}}` should match file name `{{filename}}`.',
-          data: { filename, name }
+          data: { filename, name },
+          fix(fixer) {
+            return fixer.replaceText(node, `${quote}${filename}${quote}`)
+          }
```
FloEdelmann, line 123: I think this should not be an auto-fix, since it's sometimes the case that the filename is wrong and the name property is correct.

Instead, please change this to be a [suggestion](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions) instead. Then cases where the filename is correct and the name property is wrong can be easily corrected in the editor, without causing wrong fixes when running from the commandline.

lib/rules/match-component-file-name.js
FloEdelmann: You could instead set a type without initializing the variable:

```ts
/** @type {string} */
let quote

// equivalent to the following Typescript syntax:
let quote: string
```

FloEdelmann: But maybe it would make sense to only compute the quote in the fixer function:

```js
fix(fixer) {
  const quote = node.type === 'TemplateLiteral' ? '`' : node.raw[0]
  return fixer.replaceText(node, `${quote}${filename}${quote}`)
}
```

Then you don't have to initialize it and the meaning of the variable is immediately clear from the context.

FloEdelmann (changes requested):

FloEdelmann:
Thanks for your first contribution! I have a few comments, see above. After you're done, please run `npm run update` to update the docs (which adds a hint that the rule has suggestions).

mahaker:
Very thanks for the review!!

I will confirm comments, and fix them.
Also run `npm run update` to update the docs.

mahaker:
I force pushed commit. Could you please review it?
If the review is ok, i will run `npm run update` to update the docs.

I have checked the operation of `suggest` with my VSCode and [ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).
It seems to be working correctly.

#### 1
image

#### 2: `Quick Fix` clicked
image

#### 3: `Rename component to...` clicked
image

lib/rules/match-component-file-name.js
mahaker: Thank you very much. I understand the reason to using suggest instead of fix.

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
mahaker: I found `Literal`'s `raw` property is defined [here](https://github.com/vuejs/vue-eslint-parser/blob/c476e7c25ab161b336339448765085e6219d0e04/src/ast/nodes.ts#L392).

tests/lib/rules/match-component-file-name.js
```diff
@@ -792,9 +1093,21 @@ ruleTester.run('match-component-file-name', rule, {
...
+          name: 'my-component',
+          render() { return <div /> }
+        }
+      `
+            }
+          ]
         }
-      ]
+      ],
+      output: null
```
ota-meshi, line 1110: Remove `output` as it is no longer needed.

ota-meshi (changes requested):
Thank you for this PR!
Almost LGTM! I have only one request to change the test script.

tests/lib/rules/match-component-file-name.js
mahaker: Very thanks for the review!!

I removed `output`, and run update the document.

FloEdelmann (approved):
Looks good to me, thanks!

ota-meshi (approved):
LGTM! thank you!

PR #1881, opened 2022-05-07
Remove deprecated rules

FloEdelmann:
Previously deprecated rules were removed completely:

* `vue/experimental-script-setup-vars` (deprecated in v7.13.0): not needed anymore
* `vue/name-property-casing` (deprecated in v7.0.0): use [`vue/component-definition-name-casing`](https://eslint.vuejs.org/rules/component-definition-name-casing.html) instead
* `vue/no-confusing-v-for-v-if` (deprecated in v5.0.0): use [`vue/no-use-v-if-with-v-for`](https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html) instead
* `vue/no-unregistered-components` (deprecated in v8.4.0): use [`vue/no-undef-components`](https://eslint.vuejs.org/rules/no-undef-components.html) instead


FloEdelmann:
The test fails because `vue/experimental-script-setup-vars` is still used in [tests/lib/script-setup.js](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/script-setup.js). See #1882.

ota-meshi:
I wanted to discuss with you about removing the rule.
ESLint core doesn't seem to remove deprecation rules.
https://eslint.org/docs/user-guide/rule-deprecation
Do you think we should adopt this policy too?

FloEdelmann:
Hmm, I tend to disagree here. Not removing deprecated rules might sound like it's making user's lives easier (since they don't have to make any changes in the next major release), but I'd argue it's making it harder for users and maintainers:

* First, users have to realize that a rule is actually unmaintained. If it's not there anymore, users will have to look at the release notes and docs to find out what to do now.
* If they encounter a bug with a deprecated rule, it's likely already fixed in the replacement rule. If users don't notice that themselves, they'll open an issue, which causes more work on both sides than just switching to the new rule.
* What does "no maintenance" mean exactly? When we introduce a new ESLint rule in our codebase (e.g. some time ago I introduced eslint-plugin-unicorn), shouldn't we fix these new lint errors in the whole codebase? I.e. also in deprecated/unmaintained rules?

That's why I think that deprecating a rule in a minor release and removing it in the next major release is a viable and fair strategy that saves time and work.

ota-meshi:
Thank you for your opinion! I agree with you. Let us remove the deprecated rules.

I think it's better to leave a document about the remove rules. I think maybe we need to move the list in the `## Deprecated` section somewhere.

FloEdelmann:
@ota-meshi I have added the rules docs back, and made clear in the docs that the rules are removed.

Please see https://deploy-preview-1881--eslint-plugin-vue.netlify.app/rules/#removed

ota-meshi (approved):
LGTM! Thank you very much

PR #1885, opened 2022-05-11
Update `vue/no-expose-after-await` rule to support `<script setup>`

### LINKED ISSUE #1872: Disallow defineExpose before any top-level await
The rule should notify when `defineExpose` is used before any top-level `await`. Doing so causes problems as described in this core Vue.js issue:
https://github.com/vuejs/core/issues/4930#issuecomment-966544809

```vue
// Example 1
...
<script setup>
  ...
  const foo = { bar: 'baz' };
  await asyncOperation();
  defineExpose({ foo }); // `defineExpose` should be called before any top-level `await`.
  ...
</script>
...

// Example 2
...
<script setup>
  ...
  await asyncOperation();
  const foo = { bar: 'baz' };
  defineExpose({ foo }); // `defineExpose` should be called before any top-level `await`.
  ...
</script>
...
```

**Additional context**
The current workaround is to manually remember to place all `await`s after `defineExpose`. Forgetting to do so leads to problems that are hard to debug. In the context of my examples, a JavaScript error would be thrown when `foo` is used (`foo is not defined` or similar) in some other component and it's unclear the source of the problem is actually in the component defining `foo`.

The problem seems challenging to fix in Vue.js itself:
https://github.com/vuejs/core/issues/4930#issuecomment-966544809

ota-meshi:
close #1872

### DIFF
--- a/lib/rules/no-expose-after-await.js
+++ b/lib/rules/no-expose-after-await.js
@@ -47,7 +47,7 @@ module.exports = {
     fixable: null,
     schema: [],
     messages: {
-      forbidden: 'The `expose` after `await` expression are forbidden.'
+      forbidden: 'The `{{name}}` after `await` expression are forbidden.'
     }
   },
   /** @param {RuleContext} context */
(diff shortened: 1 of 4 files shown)


lib/rules/no-expose-after-await.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
     fixable: null,
     schema: [],
     messages: {
-      forbidden: 'The `expose` after `await` expression are forbidden.'
+      forbidden: 'The `{{name}}` after `await` expression are forbidden.'
```
FloEdelmann, line 50: Better wording (might also apply to similar `no-*-after-await` rules):

```suggestion
      forbidden: '`{{name}}` is forbidden after an `await` expression.'
```

lib/rules/no-expose-after-await.js
```diff
@@ -56,147 +56,195 @@ module.exports = {
...
+              .references.find((ref) => ref.identifier.name === 'defineExpose')
+            setupScopes.set(node, {
+              afterAwait: false,
+              range: scriptSetup.range,
+              isExposeReferenceId: (_id, callNode) => {
+                return (
+                  callNode.parent.type === 'ExpressionStatement' &&
+                  callNode.parent.parent === node
+                )
+              },
```
FloEdelmann, line 176: A bit shorter (though it may be worth to check how Prettier would format it):

```suggestion
              isExposeReferenceId: (_id, callNode) => (
                callNode.parent.type === 'ExpressionStatement' &&
                callNode.parent.parent === node
              ),
```

FloEdelmann (approved):
Looks good, but very complicated :astonished: 

I have two very minor suggestions.

ota-meshi:
Thank you for your review! I have merged your suggestions.
I also updated eslintrc to use arrow-body-style rule.
https://eslint.org/docs/rules/arrow-body-style

FloEdelmann (approved):
Looks good to me! Thanks for fixing the other rules, too :)

cjpearson:
Could a similar change be applied to `no-lifecycle-after-await` so it also supports script setup?

FloEdelmann:
Sounds like a valid improvement. Could you please open an issue for that? And maybe you'd like to implement that change? It should be quite similar to this PR.

cjpearson:
@FloEdelmann Added https://github.com/vuejs/eslint-plugin-vue/issues/2291

@ota-meshi I'm using vue@3.3.4, but I still get the console warning `Lifecycle injection APIs can only be used during execution of setup(). If you are using async setup(), make sure to register lifecycle hooks before the first await statement. ` at runtime when using `onMounted` after `await`. Is that something different?

ota-meshi:
You can check it on SFC Playground. You can see that the compilation result of await `doSomething()` looks like this:

```js
;(
  ([__temp,__restore] = _withAsyncContext(() => doSomething())),
  await __temp,
  __restore()
)
```

`__restore()` should now allow you to use lifecycle methods and others then `defineExpose`. If it doesn't work, I think it's a bug in Vue.


SFC Playground:

https://play.vuejs.org/#eNqNUk1rwkAQ/SvLXkxAotabpEJbPLTQVqrHvYRkjGuT3WU/VJD8984kmlaopadk5r03vHmzJ/5gTLIPwGc8dbmVxjMHPpi5ULI22nr2pGvDNlbXbJCMqCD6QKh01PGRiYWH2lSZB6wYS1fBGVAO5ikJ5mmrw0/fJ3kv4UPuXa7VRpbJzmmFXk40RvAcZbIC+2681MoJPmMtQlhWVfrw0va8DTC89PMt5J+/9HfuSD3BlxYc2D0I3mM+syX4Dl6s3uCI/z1Y6yJUyP4D/ACnq0AeO9pjUAXa/sFr3T63iUpVrt3i6DGIy1JklJhNyxccE6bEbq3+bXeaTFudUA2meLnO7WOemFavOigPBWvOV+3OKVSPRFHM7ufIxZvgXpBUuowG9Vk2GcSsiUmQHTLpWaFXuga/xbWitv3POXf9nE1QOS15PQp1Fs1bxRQc2BKtSgcR3o5m4lprWYMOnjpDNhmPx3GMMdx8lZTN1asjAm++AK0DBNA=

ota-meshi:
Note that your own implementation of `async setup(props)` will not do this. This only applies to `<script setup>`.

cjpearson:
Ok, that's interesting. The warning I saw must be caused by something else then. Thank you for the explanation!

Issue #1884, opened 2022-05-09
Rule proposal: Allow only one return inside of the `setup`


pikax:
**Please describe what the rule should do:**

**What category should the rule belong to?**

[ ] Enforces code style (layout)
[x] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```ts
defineComponent({
 setup(){
   // code
   If (something) {
     // do something
     return; // <- Should warn about this  
   }
    return { /* ... */ }
  }
});

defineComponent({
 setup(){
   // code
 
   If (something) {
     // do something
     return {
       other: 1
     }; // <-- Should warn about this 
   }
   
    return {
      // ...
    }
  }
})

```

**Additional context**

If we return multiple times on the `setup` typescript will infer the `setup` type to be possible all the returns, the only case where multiple returns might make sense if on SSR (to prevent reactive objects), but even in SSR I would not multiple return.

FloEdelmann:
Yes, you are right that an early `return` should be avoided in the `setup` function. It's a potential problem and (even if used correctly) can make the code harder to read.

Actually, it should also be avoided in the `data` function and Nuxt's `asyncData` function.

Thus, I suggest the name `vue/no-early-return`.

PR #1889, opened 2022-05-13
Add rule vue/no-early-return

jesusgn90:
fixes https://github.com/vuejs/eslint-plugin-vue/issues/1884

### DIFF
--- a/docs/rules/no-early-return.md
+++ b/docs/rules/no-early-return.md
@@ -1,0 +1,103 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-early-return
+description: disallow early returns in setup and data functions
+---
+# vue/no-early-return
+
+> disallow early returns in setup and data functions
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule helps to identify accidental `return` statements in the `setup` block of a Vue component, these are `return` statements that would exit the `setup` block skipping part of the block itself (not allowing to each the end of the block).
+
+On the other hand, the rule checks the `data` block too, applying the same logic.
+
+<eslint-code-block :rules="{'vue/no-early-return': ['error']}">
+
+```vue
+<!-- ✓ GOOD -->
+<script>
+  export default {
+    setup () {
+      const foo = ref()
+      return { foo }
+    }
+  }
+</script>
+```
+
+</eslint-code-block>
+
+<eslint-code-block :rules="{'vue/no-early-return': ['error']}">
+
+```vue
+<!-- ✗ BAD -->
+<script>
+  export default {
+    setup () {
+      const foo = ref()
+
+      if (maybe) {
+        return
+      }
+
+      for (const t of foo.value) {
+        if (t) {
+          return
+        }
+      }
+
+      return { foo }
+    }
+  }
+</script>
+```
+
+</eslint-code-block>
+
+<eslint-code-block :rules="{'vue/no-early-return': ['error']}">
+
+```vue
+<!-- ✓ GOOD -->
+<script>
+  export default {
+    data () {
+      return { foo: true }
+    }
+  }
+</script>
+```
+
+</eslint-code-block>
+
+<eslint-code-block :rules="{'vue/no-early-return': ['error']}">
+
+```vue
+<!-- ✗ BAD -->
+<script>
+  export default {
+    data () {
+      if (maybe) {
+        return { foo: false }
+      }
+
+      return { foo: true }
+    }
+  }
+</script>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-early-return.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-early-return.js)
--- a/docs/rules/no-early-return.md
+++ b/docs/rules/no-early-return.md
@@ -2,19 +2,17 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-early-return
-description: disallow early returns in setup and data functions
+description: disallow early `returns` in `setup` and `data` functions
 ---
 # vue/no-early-return
 
-> disallow early returns in setup and data functions
+> disallow early `returns` in `setup` and `data` functions
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 
 ## :book: Rule Details
 
-This rule helps to identify accidental `return` statements in the `setup` block of a Vue component, these are `return` statements that would exit the `setup` block skipping part of the block itself (not allowing to each the end of the block).
-
-On the other hand, the rule checks the `data` block too, applying the same logic.
+The `setup` and `data` functions of Vue components should only have a single (unconditional) return statement at the end. This rule reports early `return` statements, i.e. those that would exit the function early without allowing to reach the end of the function body.
 
 <eslint-code-block :rules="{'vue/no-early-return': ['error']}">
 
--- a/docs/rules/no-early-return.md
+++ b/docs/rules/no-early-return.md
@@ -12,7 +12,7 @@ description: disallow early `returns` in `setup` and `data` functions
 
 ## :book: Rule Details
 
-The `setup` and `data` functions of Vue components should only have a single (unconditional) return statement at the end. This rule reports early `return` statements, i.e. those that would exit the function early without allowing to reach the end of the function body.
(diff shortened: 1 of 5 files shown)


lib/rules/no-early-return.js
```diff
@@ -0,0 +1,76 @@
...
+
+        onSetupFunctionExit() {
+          if (rStatementLocs.length === 0) {
+            return
+          }
+
+          for (const loc of rStatementLocs) {
+            context.report({
+              loc,
+              message: 'Extra return statement in setup function.'
```
FloEdelmann, line 49: Please change this message to:

```suggestion
              message: 'Only one (unconditional) return statement is allowed in setup function.'
```

Even better: Create a shared message with a `messageId`: https://eslint.org/docs/developer-guide/working-with-rules#messageids

```js
messages: {
  extraReturnStatement: 'Only one (unconditional) return statement is allowed in {{ functionType }} function.'
}
```

Then reuse this message here:

```js
messageId: 'extraReturnStatement',
data: {
  functionType: 'setup'
},
```

lib/rules/no-early-return.js
```diff
@@ -0,0 +1,76 @@
...
+            return
+          }
+          if (
+            context
+              .getSourceCode()
+              .ast.tokens.filter((t) => t.value === 'return').length > 1
+          ) {
+            context.report({
+              node,
+              message: 'Extra return statement in data function.'
```
FloEdelmann, line 69:
```suggestion
              message: 'Only one (unconditional) return statement is allowed in data function.'
```

lib/rules/no-early-return.js
```diff
@@ -0,0 +1,76 @@
...
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-early-return.html'
+    },
+    fixable: null,
+    schema: []
+  },
+
+  /** @param {RuleContext} context */
+  create(context) {
+    const rStatementLocs = []
```
FloEdelmann, line 31: Please rename this function to `returnStatementLocations`.

lib/rules/no-early-return.js
```diff
@@ -0,0 +1,76 @@
...
+
+          for (const loc of rStatementLocs) {
+            context.report({
+              loc,
+              message: 'Extra return statement in setup function.'
+            })
+          }
+        },
+        onVueObjectEnter(node) {
+          const dataProperty = utils.findProperty(node, 'data')
```
FloEdelmann, line 54: The same logic should also apply to Nuxt.js's `asyncData` property: https://nuxtjs.org/docs/directory-structure/pages#asyncdata

lib/rules/no-early-return.js
```diff
@@ -0,0 +1,76 @@
...
+      url: 'https://eslint.vuejs.org/rules/no-early-return.html'
+    },
+    fixable: null,
+    schema: []
+  },
+
+  /** @param {RuleContext} context */
+  create(context) {
+    const rStatementLocs = []
+    return utils.compositingVisitors(
```
FloEdelmann, line 32: Is this needed? Can't you return `utils.defineVueVisitor(…)` directly?

FloEdelmann (changes requested):
Wow, that was fast, thanks! I have a few remarks.

docs/rules/no-early-return.md
```diff
@@ -0,0 +1,103 @@
...
+
+> disallow early returns in setup and data functions
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule helps to identify accidental `return` statements in the `setup` block of a Vue component, these are `return` statements that would exit the `setup` block skipping part of the block itself (not allowing to each the end of the block).
+
+On the other hand, the rule checks the `data` block too, applying the same logic.
```
FloEdelmann, line 17:
```suggestion
The `setup` and `data` functions of Vue components (and [`asyncData`](https://nuxtjs.org/docs/directory-structure/pages/#asyncdata) of Nuxt.js page components) should only have a single (unconditional) return statement at the end. This rule reports early `return` statements, i.e. those that would exit the function early without allowing to reach the end of the function body.
```

jesusgn90:
@FloEdelmann thanks for your review, I think I've addressed your comments. There is one point however, the one related to Nuxt that I'm not super sure so I'd prefer to leave the rule as it is for now. Perhaps, it's worth opening a new issue specific for that so others can comment on making this rule cover that for Nuxt or not. For me, even covering the `data` part was an extra, the original purpose was to cover just extra returns on `setup` blocks :smile:

docs/rules/no-early-return.md
jesusgn90: applied this without the part that mentions Nuxt

FloEdelmann:
The logic for `asyncData` would be exactly the same as for `data`.

jesusgn90:
> The logic for asyncData would be exactly the same as for data.

@FloEdelmann ok, if they work the same, I just pushed a small change to cover it (and in a way that if we need to add more properties in the future, it's easily extendable)

tests/lib/rules/no-early-return.js
```diff
@@ -0,0 +1,302 @@
...
+      `,
+      errors: [
+        {
+          messageId: 'extraReturnStatement',
+          data: { functionType: 'data' },
+          line: 3
+        }
+      ]
+    },
+    {
```
FloEdelmann, line 235: Please add one more test case. I'm afraid this will produce too many errors :slightly_frowning_face: 

```suggestion
    {
      filename: 'test.vue',
      code: `
      <script>
      export default {
        setup: () => {
          if (maybe) {
            return { foo: false }
          }
          return { foo: true }
        },
        data() {
          return {
            foo: true
          }
        },
        computed: {
          foo() {
            return true
          }
        }
      }
      </script>
      `,
      errors: [
        {
          messageId: 'extraReturnStatement',
          data: { functionType: 'setup' },
          line: 3
        }
      ]
    },
    {
```

tests/lib/rules/no-early-return.js
```diff
@@ -0,0 +1,302 @@
...
+          const foo = computed(() => {
+            if (s) {
+              return
+            }
+          })
+        }
+      }
+      </script>
+      `
+    }
```
FloEdelmann, line 147: Unfortunately, this test case fails although it should pass:

https://github.com/vuejs/eslint-plugin-vue/runs/6426396807?check_suite_focus=true#step:5:187

lib/rules/no-early-return.js
```diff
@@ -0,0 +1,85 @@
...
+
+    return utils.defineVueVisitor(context, {
+      ReturnStatement(node) {
+        if (context.getScope().type === 'function') {
+          return
+        }
+        returnStatementLocations.push(node.loc)
+      },
+
+      onSetupFunctionExit() {
```
FloEdelmann, line 47: Can't the logic for `data`/`asyncData` and for `setup` be shared? I think they should work/behave exactly the same.

FloEdelmann (changes requested):

tests/lib/rules/no-early-return.js
```diff
@@ -0,0 +1,302 @@
...
+          return { foo: true }
+        }
+      }
+      </script>
+      `,
+      errors: [
+        {
+          messageId: 'extraReturnStatement',
+          data: { functionType: 'data' },
+          line: 3
```
ota-meshi, line 231: I think the report should point to line 6.

tests/lib/rules/no-early-return.js
ota-meshi: I found a false positive in the next test source.

```vue
<script>
export default {
  methods: {
    fn () {
      if (a) {
        return {}; // <- false positive
      }
    }
  },
  setup () {
    const foo = ref()
    return { foo }
  }
}
</script>
```

I think we need to check the scope of ReturnStatement correctly.
The implementation of the return-in-emits-validator rule may be helpful.
https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/return-in-emits-validator.js

tests/lib/rules/no-early-return.js
```diff
@@ -0,0 +1,302 @@
...
+            if (s) {
+              return
+            }
+          })
+        }
+      }
+      </script>
+      `
+    }
+  ],
```
ota-meshi, line 148: I found a false positive in the next test source.

```vue
<script>
export default {
  methods: {
    fn () {
      if (a) {
        return {}; // <- false positive
      }
    }
  },
  setup () {
    const foo = ref()
    return { foo }
  }
}
</script>
```

I think we need to check the scope of ReturnStatement correctly.
The implementation of the return-in-emits-validator rule may be helpful.
https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/return-in-emits-validator.js

ota-meshi (changes requested):

tests/lib/rules/no-early-return.js
ota-meshi: Or I think you can use code path analysis like `consistent-return` rule.
https://eslint.org/docs/developer-guide/code-path-analysis
https://github.com/eslint/eslint/blob/main/lib/rules/consistent-return.js

jesusgn90:
I'll resume this with more time

PR #1890, opened 2022-05-13
Bump devDependencies to their newest versions

FloEdelmann:
Note that [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/) was not updated because v4.x doesn't seem to recognize ESLint rules in our codebase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "markdownlint-cli": "^0.31.1",
     "mocha": "^10.0.0",
     "nyc": "^15.1.0",
-    "prettier": "^2.4.1",
+    "prettier": "^2.6.2",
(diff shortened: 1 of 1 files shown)


FloEdelmann:
@ota-meshi The deploy on Netlify failed, but I don't know why. Can you try clearing the npm cache there and retry the build?

I noticed that I sometimes had to run `npm install` twice on my local machine as well, otherwise I got "MODULE_NOT_FOUND" issues.

ota-meshi:
I deleted the cache and re-run it and it was successful😀

ota-meshi (approved):
Thank you

Issue #1850, opened 2022-04-14
Summary of eslint-plugin-vue v9


ota-meshi:
I am planning a major version upgrade of eslint-plugin-vue.
Mainly due to the breaking changes caused by the parser upgrade, a major version upgrade is required.

I started some necessary work with the following PR.

- #1846
- #1847
- #1848
- #1849
- #1881
- #1883
- #1851
- #1892

Please leave a comment on this issue if you think we have other work to do.

@FloEdelmann

ota-meshi:
It will be released after May as it must be after Node.js v12 is no longer supported.
However, we may release an alpha version before that.

FloEdelmann:
This issue should also be fixed in the next major release (since the bug fix could also be considered a breaking change):

* #1879

FloEdelmann:
Let me try to draft the release notes here (feel free to edit this comment @ota-meshi!):

---

# v9.0.0

## 💥 Breaking Changes

* [#1847](https://github.com/vuejs/eslint-plugin-vue/pull/1847) Drop support for Node.js v12.
* [#1846](https://github.com/vuejs/eslint-plugin-vue/pull/1846) Change default casing of [`vue/custom-event-name-casing`](https://eslint.vuejs.org/rules/custom-event-name-casing.html) rule to `camelCase`.
* [#1883](https://github.com/vuejs/eslint-plugin-vue/pull/1883) Potentially breaking bug fix: Enable [`vue/html-closing-bracket-spacing`](https://eslint.vuejs.org/rules/html-closing-bracket-spacing.html) and [`vue/html-closing-bracket-newline`](https://eslint.vuejs.org/rules/html-closing-bracket-newline.html) rules for top-level tags.
* [#1848](https://github.com/vuejs/eslint-plugin-vue/pull/1848) Change presets configs and remove unused internal methods. *(see below)*
* [#1849](https://github.com/vuejs/eslint-plugin-vue/pull/1849) Upgrade [vue-eslint-parser](https://github.com/vuejs/vue-eslint-parser) to v9. *(see below)*
* [#1881](https://github.com/vuejs/eslint-plugin-vue/pull/1881) Remove deprecated rules. *(see below)*

### Change Vue 3 presets configs

* Change `plugin:vue/vue3-essential` config:
  - Add [`vue/no-child-content`](https://eslint.vuejs.org/rules/no-child-content.html) rule
[...]

PR #1892, opened 2022-05-17
Rename `vue/no-invalid-attribute-name` rule to `vue/invalid-attribute-name` and add it to presets

ota-meshi:


FloEdelmann (approved):

FloEdelmann:
Maybe before we release v9, should we rename `no-invalid-attribute-name` to `valid-attribute-name` for consistency?

ota-meshi:
I think it's a good idea to rename the rule. I will rename in this PR.

FloEdelmann (approved):

ota-meshi:
Thank you for checking! I will release v9 tomorrow if I have time.

FloEdelmann:
I already updated the changelog draft in [#1850](https://togithub.com/vuejs/eslint-plugin-vue/issues/1850), so releasing should be as straightforward as possible.

FloEdelmann:
@ota-meshi The documentation for this rule is not yet published. Maybe the npm cache needs to be cleared for the production build, too?

https://eslint.vuejs.org/rules/valid-attribute-name.html

ota-meshi:
Thank you for letting me know. I cleared the cache and redeployed.

PR #1907, opened 2022-05-29
fix(prefer-import-from-vue): skip side-effect import in `.d.ts` files

antfu:
For example, in the dts generated by unplugin-vue-components, we use `import '@vue/runtime-core'` to let TypeScript aware of the context. I guess similar cases where ppl import without names, they are probably intentional to rely on its side-effects.

https://github.com/antfu/vitesse/blob/168fe3560999b1132f4b7a4ffa0a3a67859b0ca7/src/components.d.ts

### DIFF
--- a/tests/lib/rules/prefer-import-from-vue.js
+++ b/tests/lib/rules/prefer-import-from-vue.js
@@ -22,7 +22,8 @@ tester.run('prefer-import-from-vue', rule, {
     `export * from 'vue'`,
     `import Foo from 'foo'`,
     `import { createApp } from 'vue'
-    export { createApp }`
+    export { createApp }`,
+    `import '@vue/runtime-dom'`
   ],
   invalid: [
     {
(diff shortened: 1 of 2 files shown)


ota-meshi:
Thank you for this PR!
However, prefer-import-from-vue rule is intended to prevent imports from failing in a stricter package manager.

https://github.com/vuejs/eslint-plugin-vue/issues/1780#issuecomment-1028924275

So I think we need to limit the behavior changes to `d.ts`. What do you think?

antfu:
Yeah, that makes sense to me. I have updated the implementation, however I might need some help on testing against it.

tests/lib/rules/prefer-import-from-vue.js
```diff
@@ -22,7 +22,8 @@ tester.run('prefer-import-from-vue', rule, {
     `export * from 'vue'`,
     `import Foo from 'foo'`,
     `import { createApp } from 'vue'
-    export { createApp }`
+    export { createApp }`,
+    `import '@vue/runtime-dom'`
```
ota-meshi, line 26: To test the filename, define a test case as follows:

```suggestion
    {
      filename: 'test.d.ts',
      code: `import '@vue/runtime-dom'`
    }
```

FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thank you!

PR #1909, opened 2022-06-09
Fix regression on no-unused-components rule

matthieumota:
Hi,

I just saw that ES Lint no longer detects unused components in Vue. So I checked in the no-unused-components rule and noticed a change [here](https://github.com/vuejs/eslint-plugin-vue/commit/ba2911413acd6c61cd966ab8471037ea4563451e#diff-431a19a18ef6f160af08eeb4967020d8d755928dd7091471ea32754aa419d684).

Before, a filter was applied to get all unused components and throw an error in context for everyone.

But after PR, return stop for and check only first component if it is used. Rule stop and is useless at the moment. I think we want to continue in for for all registered components.

### DIFF
--- a/lib/rules/no-unused-components.js
+++ b/lib/rules/no-unused-components.js
@@ -120,13 +120,13 @@ module.exports = {
                       name === casing.camelCase(n))
                 )
               ) {
-                return
+                continue
               }
             } else {
               // In any other case the used component name must exactly match
               // the registered name
               if (usedComponents.has(name)) {
-                return
+                continue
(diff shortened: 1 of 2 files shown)


ota-meshi (changes requested):
Thank you for this PR!!
Could you add test cases to make sure this works?

matthieumota:
I've added a test when a component in middle is no present in template.

ota-meshi (approved):
Thank you very much!

PR #1917, opened 2022-06-20
feat: support script setup and css var inj for vue 2.7

### LINKED ISSUE #1916: Support `<script setup>` and SFC css var injection for Vue 2.7.0+ in `no-unsupported-features` rule
`no-unsupported-features`

Fewer warnings, as `<script setup>` and CSS var injection are both supported features in Vue `>=2.7.0`

I plan to update the rule to include `<script setup>` and CSS var injection as part of the `>=2.7.0` version.

```vue
<script setup>
</script>
```

Displays an error/warning with a `{ version: '2.7.0' }` rule option set

Will pass with a `{ version: '2.7.0' }` rule option set

**Additional context**

danielwaltz:
Closes #1916


ota-meshi:
Thank you for this PR! This PR looks good now.
We will check this PR again once the stable version of Vue 2.7 is released.

ota-meshi (approved):

FloEdelmann (approved):

PR #1943, opened 2022-08-08
feat: add `no-required-prop-with-default` rule

neferqiqi:
This rule enforce props with default values ​​to be optional.
Because when a required prop declared with a default value, but it doesn't be passed value when using it, it will be assigned the default value. So a required prop with default value is same as a optional prop.


xiaoxiangmoe:
Maybe we should add  this rule in vue2 and vue3 presets.

docs/rules/prefer-optional-props-using-with-defaults.md
```diff
@@ -0,0 +1,56 @@
...
+
+> enforce props with default values ​​to be optional
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule enforce props with default values ​​to be optional.
+Because when a required prop declared with a default value, but it doesn't be passed value when using it, it will be assigned the default value. So a required prop with default value is same as a optional prop.
```
sodatea, line 17: I rephrased it to make it a little bit clearer:
```suggestion
If a prop is declared with a default value, whether it is required or not, we can always skip it in actual use. In that situation, the default value would be applied.
So, a required prop with a default value is essentially the same as an optional prop.
This rule enforces all props with default values ​​to be optional.
```

sodatea (dismissed):

docs/rules/prefer-optional-props-using-with-defaults.md
```diff
@@ -0,0 +1,56 @@
...
+  );
+</script>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
```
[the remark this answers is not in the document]
neferqiqi: done:)

docs/rules/prefer-optional-props-using-with-defaults.md
neferqiqi: edited

ota-meshi:
Thank you for this PR.

I think a rule that checks for unnecessary defaults would be useful. However, I think it would be better to be able to check other than `<script setup lang="ts">`.

```vue
<script>
export default {
  props: {
    foo: {
      required: true,
      default: 42
    }
  }
}
</script>
```

```vue
<script setup>
defineProps({
  foo: {
    required: true,
    default: 42
  }
})
</script>
```

In this case, don't use `withDefaults`, so maybe we should rename the rule.

lib/rules/prefer-optional-props-using-with-defaults.js
```diff
@@ -0,0 +1,96 @@
...
+            )
+            if (!keyToken) return
+            context.report({
+              node: prop.node,
+              loc: prop.node.loc,
+              data: {
+                key: prop.propName
+              },
+              message: `Prop "{{ key }}" should be optional.`,
+              fix: (fixer) => fixer.insertTextAfter(keyToken, '?')
```
ota-meshi, line 89: We don't know if it's right to make it optional or to remove the default, so I think it makes more sense to use the suggestion API to provide suggestions instead of autofix.

sodatea:
Regarding the design:
`eslint-plugin-react` uses an option in `require-default-props` for a similar purpose: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/require-default-props.md#forbiddefaultforrequired

I think that design might be more performant as all the checks are done in one pass. What do you think?

ota-meshi:
Hmm... I think it's a good idea to follow the precedent,  but I think the name `require-default-props` can be confusing to report unnecessary defaults. If I had to choose one, I think the new rule is better.

ota-meshi:
@neferqiqi can you change this PR?

neferqiqi:
> 

ok, i will change the rule to supoort the case
```
<script>
export default {
  props: {
    foo: {
      required: true,
      default: 42
    }
  }
}
</script>
```
but i want to reserve the auto-fix. Because firstly, it only fix the props in current file, so it's safe to auto fix; secondly, for example in my project, there are many cases that assign default values ​​to required props, i think need a auto fix way to fix these cases. what do you think? :)

ota-meshi:
> but i want to reserve the auto-fix. Because firstly, it only fix the props in current file, so it's safe to auto fix; secondly, for example in my project, there are many cases that assign default values ​​to required props, i think need a auto fix way to fix these cases. what do you think? :)

I don't think it should auto-fix. As I wrote earlier, it's because we don't know which of the two methods the user intended.
I think it's the same reason no-unused-vars doesn't automatically remove variables. It's safe to remove the variable, but we shouldn't, as the user may intend to use it.

However, I do understand that your team need auto-fix. What do you think of adding an option to enable auto-fix?
(However, that in most cases I don't like that option as it just complicates the implementation of the rules.)

neferqiqi:
> > but i want to reserve the auto-fix. Because firstly, it only fix the props in current file, so it's safe to auto fix; secondly, for example in my project, there are many cases that assign default values ​​to required props, i think need a auto fix way to fix these cases. what do you think? :)
> 
> I don't think it should auto-fix. As I wrote earlier, it's because we don't know which of the two methods the user intended. I think it's the same reason no-unused-vars doesn't automatically remove variables. It's safe to remove the variable, but we shouldn't, as the user may intend to use it.
> 
> However, I do understand that your team need auto-fix. What do you think of adding an option to enable auto-fix? (However, that in most cases I don't like that option as it just complicates the implementation of the rules.)

ok, i will add an option to enable auto-fix.

FloEdelmann:
Maybe better rename the rule to `no-required-prop-with-default`?

docs/rules/prefer-optional-props-using-with-defaults.md
```diff
@@ -0,0 +1,61 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/prefer-optional-props-using-with-defaults
```
FloEdelmann, line 4: Maybe better rename the rule to `vue/no-required-prop-with-default`?

neferqiqi: thank you! I have improved name to that

lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,154 @@
...
+
+// ...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
```
FloEdelmann, line 27: I think this should be an error. @ota-meshi what do you think?

```suggestion
    type: 'error',
```

ota-meshi: I agree with you 👍

neferqiqi: Eslint prompts "`meta.type` must be either `problem`, `suggestion`, or `layout`"; do you mean this should be 'problem'?

FloEdelmann: Ah, yeah, I meant `problem`.

neferqiqi: done

FloEdelmann (approved):
Looks good to me, apart from some smaller remarks. Is this now ready to be merged?

lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,154 @@
...
+ * @typedef {import('../utils').ComponentTypeProp} ComponentTypeProp
+ */
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+// ...
+
+// ------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
neferqiqi: Thank you, fixed

lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,148 @@
...
+                !!requiredNode.value
+              ) {
+                context.report({
+                  node: prop.node,
+                  loc: prop.node.loc,
+                  data: {
+                    key: prop.propName
+                  },
+                  message: `Prop "{{ key }}" should be optional.`,
+                  fix: canAutoFix
```
ota-meshi, line 92: Could you change to provide suggestions when auto-fix is not provided?

tests/lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,879 @@
...
+        })
+        </script>
+      `,
+      errors: [
+        {
+          message: 'Prop "name" should be optional.',
+          line: 6
+        }
+      ]
+    }
```
ota-meshi, line 877: Could you add test cases that uses the object with defineProps?

```js
defineProps({
  name: {
    required: true,
    default: 'Hello'
  }
})
```

lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,148 @@
...
+            item.type === 'type' && item.required ? [item] : []
+          )
+
+          for (const prop of requiredProps) {
+            if (withDefaultsProps.includes(prop.propName)) {
+              const firstToken = context
+                .getSourceCode()
+                .getFirstToken(prop.node)
+              // skip setter & getter case
+              if (firstToken.value === 'get' || firstToken.value === 'set') {
```
ota-meshi, line 119: I think you could have written it like this:

```suggestion
              if (prop.node.type === "TSMethodSignature") {
```

lib/rules/no-required-prop-with-default.js
```diff
@@ -0,0 +1,148 @@
...
+              if (!keyToken) return
+              context.report({
+                node: prop.node,
+                loc: prop.node.loc,
+                data: {
+                  key: prop.propName
+                },
+                message: `Prop "{{ key }}" should be optional.`,
+                fix: canAutoFix
+                  ? (fixer) => fixer.insertTextAfter(keyToken, '?')
```
ota-meshi, line 139: I don't think we need to find the keyToken.

```suggestion
                  ? (fixer) => fixer.insertTextAfter(prop.key, '?')
```

ota-meshi (changes requested):

lib/rules/no-required-prop-with-default.js
neferqiqi: If change to this, the following two cases will be ignored:
```ts
interface TestPropType {
    foo(): void
    age?: number
}
const props = withDefaults(
    defineProps<TestPropType>(),
    {
        foo() {console.log('foo')},
    }
);
```

```ts
interface TestPropType {
    readonly foo(): void
    age?: number
}
const props = withDefaults(
   defineProps<TestPropType>(),
   {
       foo() {console.log('foo')},
   }
);
```

ota-meshi: Thank you for letting me know. We should have checked `get` and `set` as you said.

```suggestion
              if (prop.node.type === "TSMethodSignature" && (prop.node.kind === 'get' || prop.node.kind === 'set')) {
```

lib/rules/no-required-prop-with-default.js
neferqiqi: done

PR #1955, opened 2022-08-31
fix(order-in-components): allow autofix with TypeScript PropType

acupofspirt:
In the following code `vue/order-in-components` rule can't be autofixed because of `as PropType`:
```ts
import { defineComponent, PropType } from 'vue'

export defineComponent({
  setup () {},

  props: {
    someProp: { 
      type: Array as PropType<string[]>
    }
  }
})
```
TSAsExpression node is considered as not side-effects free. This PR fixes it.


ota-meshi:
Thank you for this PR. Could you please add test cases?

acupofspirt:
[ota-meshi](https://github.com/ota-meshi) tests done

tests/lib/autofix.js
```diff
@@ -172,3 +172,46 @@ describe('Complex autofix test cases', () => {
     })
   })
 })
+
+describe('Autofix test cases with Typescript', () => {
```
ota-meshi, line 176: Could you change it to test case using rule tester?
https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/order-in-components.js

acupofspirt: done!

ota-meshi (approved):
LGTM! Thank you!

ota-meshi:
ts-eslint-v4 error is fixed in #1956.

FloEdelmann (approved):

PR #1956, opened 2022-09-01
Chore: fix typescript-eslint v4 test

ota-meshi:
It seems that using typescript-eslint v4 and typescript v4.8 together breaks the AST due to incompatibility.

This PR changes typescript-eslint v4 tests to use typescript v4.7.


FloEdelmann (approved):
LGTM! :)

PR #1965, opened 2022-09-10
Add `vue/no-ref-object-destructure` rule

### LINKED ISSUE #1949: Rule Proposal: `vue/no-ref-object-destructure` rule

A new rule reports usage of ref objects that can lead to loss of reactivity.

[ ] Other (please specify:)

```js
// Ref Object
const count = ref(0)
const cnt1 = count.value /* ✗ BAD */
const { value: cnt2 } = count /* ✗ BAD */
const cnt3 = computed(() => count.value /* ✓ GOOD */)
const value1 = fn(count.value) /* ✗ BAD */
const value2 = fn(count) /* ✓ GOOD */
const value3 = computed(() => fn(count.value) /* ✓ GOOD */)
```

```js
// Reactivity Transform
const count = $ref(0)
const cnt1 = count /* ✗ BAD */
const cnt2 = computed(() => count /* ✓ GOOD */)
const value1 = fn(count) /* ✗ BAD */
const value2 = fn($$(count)) /* ✓ GOOD */
const value3 = computed(() => fn(count) /* ✓ GOOD */)
```

**Additional context**

I think it's useful for users who use Reactivity Transform because it prevents them from forgetting to add `$$()`.
So it is related to Reactivity Transform #1948.

Regarding Reactive Objects, the following code loses its reactivity:

```js
const state = reactive({ count: 0 })
const cnt1 = state.count
const { count: cnt2 } = state
const value1 = fn(state.count)
```

But when count is an object, it may not lose reactivity, so it's hard to report:

```js
const state = reactive({ count: { data: 0 } })
const cnt1 = state.count
const { count: cnt2 } = state
const value1 = fn(state.count)
```

ota-meshi:
This PR adds `vue/no-ref-object-destructure` rule that reports the destructuring of ref objects causing the value to lose reactivity.

close #1949

### DIFF
--- a/docs/rules/no-ref-object-destructure.md
+++ b/docs/rules/no-ref-object-destructure.md
@@ -29,7 +29,8 @@ const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
 
 </eslint-code-block>
 
-This rule also supports Reactivity Transform, but Reactivity Transform is an experimental feature and may have false positives due to future Vue changes.
+This rule also supports Reactivity Transform, but Reactivity Transform is an experimental feature and may have false positives due to future Vue changes.  
+See the [RFC](https://github.com/vuejs/rfcs/pull/420) for more information on Reactivity Transform.
 
 <eslint-code-block :rules="{'vue/no-ref-object-destructure': ['error']}" language="javascript" filename="example.js" >
 
--- a/docs/rules/no-ref-object-destructure.md
+++ b/docs/rules/no-ref-object-destructure.md
@@ -1,0 +1,54 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-ref-object-destructure
+description: disallow destructuring of ref objects that can lead to loss of reactivity
+---
+# vue/no-ref-object-destructure
+
+> disallow destructuring of ref objects that can lead to loss of reactivity
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule reports the destructuring of ref objects causing the value to lose reactivity.
+
+<eslint-code-block :rules="{'vue/no-ref-object-destructure': ['error']}" language="javascript" filename="example.js" >
+
+```js
+import { ref } from 'vue'
+const count = ref(0)
+const v1 = count.value /* ✗ BAD */
+const { value: v2 } = count /* ✗ BAD */
+const v3 = computed(() => count.value /* ✓ GOOD */)
+const v4 = fn(count.value) /* ✗ BAD */
+const v5 = fn(count) /* ✓ GOOD */
+const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
+```
+
+</eslint-code-block>
+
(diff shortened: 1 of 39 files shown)


docs/rules/no-ref-object-destructure.md
```diff
@@ -0,0 +1,54 @@
...
+const { value: v2 } = count /* ✗ BAD */
+const v3 = computed(() => count.value /* ✓ GOOD */)
+const v4 = fn(count.value) /* ✗ BAD */
+const v5 = fn(count) /* ✓ GOOD */
+const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
+```
+
+</eslint-code-block>
+
+This rule also supports Reactivity Transform, but Reactivity Transform is an experimental feature and may have false positives due to future Vue changes.
```
FloEdelmann, line 32: Please add a link to the RFC or other official docs here.

tests/lib/rules/no-ref-object-destructure.js
```diff
@@ -0,0 +1,449 @@
...
+})
+
+tester.run('no-ref-object-destructure', rule, {
+  valid: [
+    `
+    import { ref } from 'vue'
+    const count = ref(0)
+    const value1 = computed(() => count.value /* ✓ GOOD */)
+    const value2 = fn(count) /* ✓ GOOD */
+    const value3 = computed(() => fn(count.value) /* ✓ GOOD */)
```
FloEdelmann, line 24:
```suggestion
    const value1 = computed(() => count.value)
    const value2 = fn(count)
    const value3 = computed(() => fn(count.value))
```

tests/lib/rules/no-ref-object-destructure.js
```diff
@@ -0,0 +1,449 @@
...
+    {
+      code: `
+      import { ref } from 'vue'
+      const count = ref(0)
+      const value1 = count.value /* ✗ BAD */
+      const { value: value2 } = count /* ✗ BAD */
+      const value3 = fn(count.value) /* ✗ BAD */
+      const { value: value4 = 42 } = count /* ✗ BAD */
+      if (foo) {
+        const value1 = count.value /* ✗ BAD */
```
FloEdelmann, line 150:
```suggestion
      const value1 = count.value
      const { value: value2 } = count
      const value3 = fn(count.value)
      const { value: value4 = 42 } = count
      if (foo) {
        const value1 = count.value
```

FloEdelmann (changes requested):
Just a quick look because I don't have time for a full review of this big PR.

ota-meshi:
Thank you for the quick review! I have changed the part you commented on.

FloEdelmann (approved):
Thanks, that looks fine now. As noted earlier, I have not looked deeper into the other parts though.

Tofandel:
This seems to have caused a regression on the `vue/no-ref-as-operand` rule

```js
import {ref} from 'vue';

const someting = false;
let columns;

if (something) {
  const cols = ref([1]);
  columns = cols;
} else {
  columns = ref([2]);
}
```

This used to be fine, with the new version this causes 
```
 error  Must use `.value` to read or write the value wrapped by `ref()`  vue/no-ref-as-operand
 error  Must use `.value` to read or write the value wrapped by `ref()`  vue/no-ref-as-operand
```

Issue #1832, opened 2022-03-28
Rule proposal: `padding-lines-between-tags`


amiranagram:
**Please describe what the rule should do:**

The rule, as demonstrated in https://github.com/vuejs/eslint-plugin-vue/issues/855#issuecomment-538750738 and #1829 should enforce line breaks between HTML tags in template. Not all tags, just siblings.

**What category should the rule belong to?**

- [x] Enforces code style (layout)
- [ ] Warns about a potential error (problem)
- [ ] Suggests an alternate way of doing something (suggestion)
- [ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<!-- GOOD -->
<template>
    <div>
        <p>Item 1</p>

        <p>Item 2</p>
    </div>
</template>

<!-- BAD -->
<template>
    <div>
        <p>Item 1</p>
        <p>Item 2</p>
    </div>
</template>
```

```vue
<!-- GOOD -->
<template>
    <ul>
        <li>1</li>
        
        <li>2</li>

        <li>
            <ul>
                <li>3.1</li>
                
                <li>3.2</li>
            </ul>
        </li>

        <li>
            <ul>
                <li>4.1</li>
            </ul>
        </li>
    </ul>
</template>

<!-- BAD-->
<template>
    <ul>
        <li>1</li>
        <li>2</li>
        <li>
            <ul>
                <li>3.1</li>
                <li>3.2</li>
            </ul>
        </li>
        <li>
            <ul>
                <li>4.1</li>
            </ul>
        </li>
    </ul>
</template>
```

**Additional context**

Similar to `vue/padding-line-between-blocks` https://eslint.vuejs.org/rules/padding-line-between-blocks.html

ota-meshi:
Thank you for the rule proposal.
I think it's probably better to be able to set options for each element before and after, like `padding-line-between-statements`. For example, I don't want to have padding before and after `<br>`.

https://eslint.org/docs/rules/padding-line-between-statements

amiranagram:
Agree @ota-meshi. What do you think about "ignores" option where you can specify tags?

Example: https://eslint.vuejs.org/rules/multiline-html-element-content-newline.html#vue-multiline-html-element-content-newline

Also, here (https://github.com/vuejs/eslint-plugin-vue/issues/1834#issuecomment-1084421621) it's suggested for components, but I'd like for both with the same rule.

Evgenios95:
[...]

PR #1966, opened 2022-09-10
Create `padding-line-between-tags` rule

dev1437:
Add rule for #1832


FloEdelmann:
Could you please have a look at the CI failures?

* https://app.circleci.com/pipelines/github/vuejs/eslint-plugin-vue/1666/workflows/9f5355a9-5677-4568-9930-23bd379ec8e3/jobs/12552
* https://app.netlify.com/sites/eslint-plugin-vue/deploys/631c3a7fd3c155000899f827#L91

docs/rules/space-between-siblings.md
```diff
@@ -0,0 +1,123 @@
...
+
+## :wrench: Options
+
+```json
+{
+  "vue/space-between-siblings": ["error", {
+      "ignoreNewlinesBefore": [],
+      "ignoreNewlinesAfter": []
+  }]
+}
```
FloEdelmann, line 64: I think the rule should have a primary option `"always" | "never"`, just like https://eslint.vuejs.org/rules/padding-line-between-blocks.html.

lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,117 @@
...
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Helpers
+// ------------------------------------------------------------------------------
+
+// ...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
```
FloEdelmann, line 20:
```suggestion
const utils = require('../utils')

// ------------------------------------------------------------------------------
// Rule Definition
```

tests/lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,384 @@
+/**
+ * @author *****your name*****
```
FloEdelmann, line 2: Please fill in your name or username.

lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,117 @@
+/**
+ * @author *****your name*****
```
FloEdelmann, line 2: Again.

tests/lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,384 @@
...
+            </li>
+
+            <li>
+            </li>
+
+            <li>
+            </li>
+          </ul>
+        </div>
+      </template>
```
FloEdelmann, line 36: Maybe the code snippets in the examples can be a bit shorter? It's quite noisy to read through all of them.

I think this should also suffice (for this test case):

```suggestion
      <template>
        <div>
          <div></div>

          <div>
          </div>

          <div />
        </div>
      </template>
```

lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,117 @@
...
+            uniqueItems: true,
+            additionalItems: false
+          }
+        },
+        additionalProperties: false
+      }
+    ],
+    messages: {
+      never: 'Unexpected blank line before this block.',
+      always: 'Expected blank line after this block.'
```
FloEdelmann, line 108:
```suggestion
      never: 'Unexpected blank line before this tag.',
      always: 'Expected blank line after this tag.'
```

tests/lib/rules/space-between-siblings.js
```diff
@@ -0,0 +1,384 @@
...
+        },
+        {
+          message: 'Expected blank line after this block.',
+          line: 9,
+          column: 13
+        }
+      ],
+      options: [{ ignoreNewlinesBefore: ['br'], ignoreNewlinesAfter: ['br'] }]
+    }
+  ]
```
FloEdelmann, line 383: What happens for this code?

```html
<template>
  <p>Foo <b>bar</b> baz.</p>
</template>
```

I think it should be valid. Please add a test case for that.

FloEdelmann (changes requested):

docs/rules/README.md
```diff
@@ -257,6 +257,7 @@ For example:
 | [vue/require-name-property](./require-name-property.md) | require a name property in Vue components |  | :hammer: |
 | [vue/script-indent](./script-indent.md) | enforce consistent indentation in `<script>` | :wrench: | :lipstick: |
 | [vue/sort-keys](./sort-keys.md) | enforce sort-keys in a manner that is compatible with order-in-components |  | :hammer: |
+| [vue/space-between-siblings](./space-between-siblings.md) | Insert newlines between sibling tags in template  | :wrench: | :warning: |
```
ota-meshi, line 260: I think the rule name should start with `padding-line-between-`. I don't think `space` mean just blank lines.
I also think `siblings` is ambiguous. I think `padding-line-between-tags` is probably better.

dev1437:
Anything else? Let me know @ota-meshi @FloEdelmann thank you

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,128 @@
...
+  }]
+}
+```
+
+- `always` (default) Requires one or more blank lines.
+- `never` Disallows blank lines between tags
+- `ignoreNewlinesBefore` ignores newlines before specified elements. Only applies when using `always`.
+    default `[]`
+- `ignoreNewlinesAfter` ignores newlines after specified elements. Only applies when using `always`.
+    default `[]`
```
ota-meshi, line 64: Hmm. With the current options, I don't think a rule can both always force blank lines and always disallow blank lines.
I think it would be better to refer to the padding-line-between-statements rule option and make it possible to set "never" and "always" in more detail.

https://eslint.org/docs/latest/rules/padding-line-between-statements

For example:

```js
: [
        "error",
        { "blankLine": "always", "prev": "*", "next": "*" },
        { "blankLine": "never", "prev": "<br>", "next": "*" },
        { "blankLine": "never", "prev": "*", "next": "<br>" },
        // ...
    ]
```

ota-meshi (changes requested):

dev1437:
Hi @ota-meshi let me know if there's anything else. Maybe more test cases that you can think of or improvement for docs? Thanks

FloEdelmann (changes requested):
Looks good overall. Just some minor docs and test changes. Thanks for your effort!

dev1437:
Changes are done @FloEdelmann thank you

lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,217 @@
...
+      ) {
+        insertNewLine(context, block, element)
+      }
+    }
+  }
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
```
FloEdelmann, line 183: This should be also a layout rule, just like `padding-line-between-blocks`:

grafik

(see https://deploy-preview-1966--eslint-plugin-vue.netlify.app/rules/)

```suggestion
    type: 'layout',
```

tests/lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,812 @@
...
+      `,
+      errors: [
+        {
+          message: 'Expected blank line before this tag.',
+          line: 7,
+          column: 11
+        }
+      ],
+      options: [[{ blankLine: 'always', prev: '*', next: 'br' }]]
+    }
```
FloEdelmann, line 810: Should `*` maybe also catch all closing tags? Otherwise, this situation can't be prevented by this rule:

```html
<template>
  <div>
    <div />
    <div />
    <div />
    
  </div>
</template>
```

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,168 @@
...
+
+```json
+{
+  "vue/padding-line-between-tags": ["error", [
+    { "blankLine": "always", "prev": "*", "next": "*" }
+  ]]
+}
+```
+
+This rule requires blank lines between each sibling HTML tag by default.
```
FloEdelmann, line 58: The default is now to check nothing at all: https://deploy-preview-1966--eslint-plugin-vue.netlify.app/rules/padding-line-between-tags.html#rule-details

I think it should be changed to require a blank space between all tags like noted here. Please also add a test case for that.

FloEdelmann (changes requested):
I now looked at the preview deployment and noticed some more issues. Sorry that these were not surfaced earlier.

docs/rules/padding-line-between-tags.md
dev1437: I'm setting the default here [https://github.com/vuejs/eslint-plugin-vue/pull/1966/files/e210bc9f71b9dc7266a3b081d49430e3b638e5e3#diff-b8a55875caa63e9a2b2ac62113a599199f76d4f4f54a8f8c554a4582e2d412b4R68](url)

Should it be done differently?

There is also the first invalid test which tests the default behaviour [https://github.com/vuejs/eslint-plugin-vue/pull/1966/files/e210bc9f71b9dc7266a3b081d49430e3b638e5e3#diff-011f1880886531880d77f18276fbc8cbda2dd6954a775f62b318d6445c0be50aR150](url)

FloEdelmann: Ah, you're right. Just the example is wrong, since it needs to be split into two separate code blocks containing only one `<template>` each.

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,168 @@
...
+# vue/padding-line-between-tags
+
+> Require or disallow newlines between sibling tags in template
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule requires or disallows newlines between sibling HTML tags
```
FloEdelmann, line 16:
```suggestion
This rule requires or disallows newlines between sibling HTML tags.
```

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,168 @@
...
+  <div>
+    <div></div>
+    <div>
+    </div>
+    <div />
+  </div>
+</template>
+```
+
+</eslint-code-block>
```
FloEdelmann, line 46: To make the example work:

````suggestion
<eslint-code-block fix :rules="{'vue/padding-line-between-tags': ['error']}">

```vue
<template>
  <div>
    <!-- ✓ GOOD: -->
    <div></div>

    <div>
    </div>

    <div />
    <!-- ✗ BAD: -->
    <div></div>
    <div>
    </div>
    <div />
  </div>
</template>
```

</eslint-code-block>
````

**Review by FloEdelmann — CHANGES_REQUESTED** [review]

File: tests/lib/rules/padding-line-between-tags.js (hunk 11 above)
Comment (dev1437) [reply tests/lib/rules/padding-line-between-tags.js:810]: I don't understand, would you like 
```html
<template>
  <div>
    <div />
    <div />
    <div />
    
  </div>
</template>
```
to become 
```html
<template>
  <div>
    <div />
    <div />
    <div />  
  </div>
</template>
```

FloEdelmann: I'm not sure. I just saw that this is also not reported:

```
<template>
  <div>

    <div />
    <div />
    <div />
  </div>
</template>
```

So maybe we should indeed leave it like it is. `between-tags` in the rule name and "siblings" in the docs also make it clear.

dev1437: I'd agree, a separate rule should be done for parent tags and their children (although I don't know how in-demand that would be)

FloEdelmann (approved):
Looks good to me now. Thanks again!

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,161 @@
...
+{
+  "vue/padding-line-between-tags": ["error", [
+    { "blankLine": "always", "prev": "*", "next": "*" }
+  ]]
+}
+```
+
+This rule requires blank lines between each sibling HTML tag by default.
+
+A configuration is an object which has 3 properties; blankLine, prev and next. For example, { blankLine: "always", prev: "br", next: "div" } means “one or more blank lines are required between a br tag and a div tag.” You can supply any number of configurations.
```
ota-meshi, line 53: As with padding-line-between-statements, could you change the priority so that the last matched configuration will be used?

https://eslint.org/docs/latest/rules/padding-line-between-statements

It looks to me that the current rules are implementation dependent.

tests/lib/rules/padding-line-between-tags.js
ota-meshi, line 810: Is it intentional not to insert an empty line if there is no newline between elements? I think it should also insert a blank line.
Could you change it so that it is checked? We may need a new option if you don't want it included in the check.

```vue
<template>
  <div>
    <div /><div />
  </div>
</template>

<script>
  /* eslint "vue/padding-line-between-tags": ["error", [
    { "blankLine": "always", "prev": "*", "next": "*" },
  ]] */
</script>
```

ota-meshi (changes requested):
Thank you for the many changes. I still have change requests.

dev1437:
Hi @ota-meshi I've updated the rule, let me know if anything else thank you

docs/rules/padding-line-between-tags.md
```diff
@@ -0,0 +1,163 @@
...
+{
+  "vue/padding-line-between-tags": ["error", [
+    { "blankLine": "always", "prev": "*", "next": "*" }
+  ]]
+}
+```
+
+This rule requires blank lines between each sibling HTML tag by default.
+
+A configuration is an object which has 3 properties; blankLine, prev and next. For example, { blankLine: "always", prev: "br", next: "div" } means “one or more blank lines are required between a br tag and a div tag.” You can supply any number of configurations. If a statement pair matches multiple configurations, the last matched configuration will be used.
```
ota-meshi, line 55: I don't think it's a `statement pair`.

```suggestion
A configuration is an object which has 3 properties; blankLine, prev and next. For example, { blankLine: "always", prev: "br", next: "div" } means “one or more blank lines are required between a br tag and a div tag.” You can supply any number of configurations. If a tag pair matches multiple configurations, the last matched configuration will be used.
```

lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,193 @@
...
+              fix(fixer) {
+                if (!templateTag) {
+                  return
+                }
+                const start = templateTag.range[0]
+                const end = endTag.range[0]
+                const paddingText = context
+                  .getSourceCode()
+                  .text.slice(start, end)
+                const lastSpaces = splitLines(paddingText).pop()
```
ota-meshi, line 140: I did some tests to understand this code, but the value it processed is not necessarily spaces.
For example, if the following code is auto-fixed, the string will be doubled.

```vue
<template>
  <div>
  Foo  <div /><div />
  </div>
</template>
```

↓

```vue
<template>
  <div>
  Foo  <div />

  Foo  <div />
  </div>
</template>
```

I think we can probably fix it by doing:

```js
const lastSpaces = /** @type {RegExpExecArray} */(/^\s*/.exec(context.getSourceCode().lines[endTag.loc.start.line-1]))[0]
```

lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,193 @@
...
+  context.report({
+    messageId: 'never',
+    loc: sibling.loc,
+    // @ts-ignore
+    fix(fixer) {
+      const start = endTag.range[1]
+      const end = sibling.range[0]
+      const paddingText = context.getSourceCode().text.slice(start, end)
+      const lastSpaces = splitLines(paddingText).pop()
+      return fixer.replaceTextRange([start, end], `\n${lastSpaces}`)
```
ota-meshi, line 54: I've been testing this and noticed a loss of strings.

```vue
<template>
  <div>
    </div>A
    B
    C<div />
  </div>
</template>

<script>
/* eslint "vue/padding-line-between-tags": ["error", [
    { "blankLine": "never", "prev": "*", "next": "*" }
  ]]
*/
</script>
```

↓

```vue
<template>
  <div>
    </div>
    C<div />
  </div>
</template>

<script>
/* eslint "vue/padding-line-between-tags": ["error", [
    { "blankLine": "never", "prev": "*", "next": "*" }
  ]]
*/
</script>
```

Perhaps it makes more sense to exclude from the check if there is anything other than element between tags.
What do you think?

ota-meshi (changes requested):
Thank you for the changes! I tested the autofix and found some bugs.

tests/lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,968 @@
...
+      `,
+      errors: [
+        {
+          message: 'Unexpected blank line before this tag.',
+          line: 7,
+          column: 12
+        }
+      ],
+      options: [[{ blankLine: 'never', prev: '*', next: '*' }]]
+    }
```
ota-meshi, line 966: Autofix still seems to have bugs.

```vue
<template>
  <div>
    <div />
    Foo
    <div />
    <div />Foo

    <div />
  </div>
</template>

<script>
/* eslint "vue/padding-line-between-tags": ["error", [
    { "blankLine": "never", "prev": "*", "next": "*" }
  ]] */
</script>
```

↓

```vue
<template>
  <div>
    <div />
    <div />
    <div />
    <div />
  </div>
</template>

<script>
/* eslint "vue/padding-line-between-tags": ["error", [
    { "blankLine": "never", "prev": "*", "next": "*" }
  ]] */
</script>
```

ota-meshi (changes requested):

tests/lib/rules/padding-line-between-tags.js
ota-meshi: Sorry, I was checking it with old source code.

lib/rules/padding-line-between-tags.js
```diff
@@ -0,0 +1,189 @@
...
+            let hasOnlyTextBetween = false
+            for (
+              let i = endTag.loc.start.line;
+              i < closestSibling.loc.start.line - 1;
+              i++
+            ) {
+              hasOnlyTextBetween = !/^\s*$/.test(
+                context.getSourceCode().lines[i]
+              )
+            }
```
ota-meshi, line 141: I think it's the same as checking only the last line.
Or does it require additional processing?

ota-meshi (changes requested):

lib/rules/padding-line-between-tags.js
dev1437: Hi, sorry, this was a mistake from programming while tired :)
Thank you for finding this

dev1437:
Fixed, let me know if anything else thank you

ota-meshi (approved):
LGMT! Thank you!

Issue #1963, opened 2022-09-08
vue/no-undef-components is expected to report error if component is type only


Amorites:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:**  8.23.0
- **eslint-plugin-vue version:** 9.4.0
- **Node version:** 16.17.0
- **Operating System:** MacOS 12.5

**Please show your full configuration:**

```json5
module.exports = {
  extends: [
    // add more generic rulesets here, such as:
    // 'eslint:recommended',
    'plugin:vue/vue3-recommended',
    // 'plugin:vue/recommended' // Use this if you are using Vue.js 2.x.
  ],
  rules: {
    // override/add rules settings here, such as:
    // 'vue/no-unused-vars': 'error'
    'vue/no-undef-components': 'error',
  },
  parser: 'vue-eslint-parser',
  parserOptions: {
    extraFileExtensions: ['.vue'],
    parser: '@typescript-eslint/parser',
    ecmaVersion: 2022, // Allows for the parsing of modern ECMAScript features
    sourceType: 'module', // Allows for the use of imports
  },
};

```

**What did you do?**	

```vue
<script setup lang="ts">
import type HelloWorld from './components/HelloWorld.vue';
</script>

<template>
  <div>
    <a href="https://vitejs.dev" target="_blank">
      <img src="/vite.svg" class="logo" alt="Vite logo" />
    </a>
    <a href="https://vuejs.org/" target="_blank">
      <img src="./assets/vue.svg" class="logo vue" alt="Vue logo" />
    </a>
  </div>
  <HelloWorld msg="Vite + Vue" />
  <Foo></Foo>
</template>
```

**What did you expect to happen?**

` <HelloWorld msg="Vite + Vue" />` is expected to report no-undef-component error like `<Foo></Foo>`

**What actually happened?**

`<HelloWorld msg="Vite + Vue" />` doesn't trigger a eslint error

**Repository to reproduce this issue**

https://github.com/Amorites/eslint-plugin-vue-repro/blob/main/src/App.vue

Amorites:
After doing some inspect, I found that it might be possible to use `importKind` to distinguish whether it's a type-only import or not.

With any luck, I can make a PR today.

PR #1967, opened 2022-09-10
Report type-only imports in `no-undef-components`

Amorites:
fix https://github.com/vuejs/eslint-plugin-vue/issues/1963

### DIFF
--- a/lib/rules/no-undef-components.js
+++ b/lib/rules/no-undef-components.js
@@ -50,7 +50,9 @@ module.exports = {
       }
     ],
     messages: {
-      undef: "The '<{{name}}>' component has been used, but not defined."
+      undef: "The '<{{name}}>' component has been used, but not defined.",
+      typeOnly:
+        "The '<{{name}}>' component has been used, but '{{name}}' only refers to a type."
     }
   },
   /** @param {RuleContext} context */
@@ -121,6 +123,7 @@ module.exports = {
       // For <script setup>
       /** @type {Set<string>} */
       const scriptVariableNames = new Set()
+      const scriptTypeOnlyNames = new Set()
       const globalScope = context.getSourceCode().scopeManager.globalScope
       if (globalScope) {
         for (const variable of globalScope.variables) {
@@ -130,7 +133,11 @@ module.exports = {
           (scope) => scope.type === 'module'
         )
         for (const variable of (moduleScope && moduleScope.variables) || []) {
-          scriptVariableNames.add(variable.name)
+          if (variable.isTypeVariable) {
+            scriptTypeOnlyNames.add(variable.name)
+          } else {
+            scriptVariableNames.add(variable.name)
+          }
         }
       }
       /**
@@ -167,7 +174,7 @@ module.exports = {
 
         context.report({
           node: reportNode,
-          messageId: 'undef',
+          messageId: scriptTypeOnlyNames.has(rawName) ? 'typeOnly' : 'undef',
           data: {
             name: rawName
           }
--- a/tests/lib/rules/no-undef-components.js
+++ b/tests/lib/rules/no-undef-components.js
@@ -661,6 +661,66 @@ tester.run('no-undef-components', rule, {
         }
       ]
     },
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup lang="ts">
+        import type Foo from './Foo.vue'
+        import type {HelloWorld1} from './components/HelloWorld'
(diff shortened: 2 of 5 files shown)


Amorites:
Two issues in CI:

1. We can't report an error for `import  { type HelloWorld2 } from './components/HelloWorld2'` with `@typescript-eslint/parser@4`. This is not sound but is acceptable for legacy users. But how to write test here.
2. Current vuepress config doesn't support `optional chaining operator`. https://github.com/vuejs/vue-loader/issues/1697. I will look into the config with least influence to other codes.(Or when we can migrate directly to vitepress

tests/lib/rules/no-undef-components.js
```diff
@@ -661,6 +661,65 @@ tester.run('no-undef-components', rule, {
...
+      `,
+      parserOptions: {
+        ecmaVersion: 6,
+        sourceType: 'module',
+        parser: require.resolve('@typescript-eslint/parser')
+      },
+      parser: require.resolve('vue-eslint-parser'),
+      errors: [
+        {
+          message: "The '<Foo>' component has been used, but not defined.",
```
ota-meshi, line 693: I think it would be better to make the message clearer. For example:

>  '\<Foo\>' component has been used, but 'Foo' only refers to a type.

What do you think?

lib/rules/no-undef-components.js
```diff
@@ -130,7 +130,16 @@ module.exports = {
           (scope) => scope.type === 'module'
         )
         for (const variable of (moduleScope && moduleScope.variables) || []) {
-          scriptVariableNames.add(variable.name)
+          const definition = variable.defs[0]
+          if (
```
ota-meshi, line 134: I think we can more easily check if it's a value by doing:

```js
if (!variable.isValueVariable || variable.isValueVariable())
```

https://github.com/typescript-eslint/typescript-eslint/blob/77d76e21cdc2e100c729c839c292e82ab7c554c5/packages/scope-manager/src/variable/Variable.ts#L24

ota-meshi (changes requested):
Thank you for this PR!

ota-meshi:
We need to branch using semver regarding test cases failing in tseslint v4. We have already done that in some test cases.

https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-export-in-script-setup.js#L54

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
```
Amorites, line 3: This time I made vuepress support `optional chaining operator` by lowering the target version.

If it is not proper, please tell me and I will find another way to make it.

Amorites:
Thanks a lot for your comment, they are extremely helpful.

ota-meshi (approved):

tsconfig.json
ota-meshi: Thanks for looking into what is causing the vuepress build not to work. I didn't know that setting affects vuepress.

tests/lib/rules/no-undef-components.js
```diff
@@ -661,6 +661,66 @@ tester.run('no-undef-components', rule, {
...
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup lang="ts">
+        import type Foo from './Foo.vue'
+        import type {HelloWorld1} from './components/HelloWorld'
+        import  { type HelloWorld2 } from './components/HelloWorld2'
+        import type {HelloWorld as HelloWorld3} from './components/HelloWorld3'
+        import  { type HelloWorld as HelloWorld4 } from './components/HelloWorld4';
+        import  { type default as HelloWorld5 } from './components/HelloWorld5';
```
FloEdelmann, line 673: Lint:

```suggestion
        import { type HelloWorld2 } from './components/HelloWorld2'
        import type {HelloWorld as HelloWorld3} from './components/HelloWorld3'
        import { type HelloWorld as HelloWorld4 } from './components/HelloWorld4';
        import { type default as HelloWorld5 } from './components/HelloWorld5';
```

FloEdelmann (approved):
LGTM!

tsconfig.json
Amorites: Sorry, I just lost my mind at that moment 😢. This doesn't affect `js` files, errors just disappear accidentally.

lib/rules/no-undef-components.js
ota-meshi: This comment of mine was wrong 😓

PR #1968, opened 2022-09-12
Add type-based declaration rule of `defineProps` and `defineEmits`

Amorites:
Implement #1960 #1959

PR #1972, opened 2022-09-15
Chore: ts-parser support in demo

ota-meshi:
This PR modifies the documentation site demo so that it parses well using typescript.

https://deploy-preview-1972--eslint-plugin-vue.netlify.app/rules/valid-define-emits.html
https://deploy-preview-1972--eslint-plugin-vue.netlify.app/rules/valid-define-props.html

We'll need to clear the cache in netlify for this to take effect.

Related to https://github.com/vuejs/eslint-plugin-vue/pull/1968#issuecomment-1244839400

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -132,7 +132,6 @@ module.exports = {
     'unicorn/prefer-optional-catch-binding': 'off', // not supported by current ESLint parser version
     'unicorn/prefer-module': 'off',
     'unicorn/prevent-abbreviations': 'off',
-    'unicorn/no-nested-ternary': 'off' // conflicts with Prettier
   },
   overrides: [
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -131,7 +131,7 @@ module.exports = {
     'unicorn/no-useless-undefined': 'off',
     'unicorn/prefer-optional-catch-binding': 'off', // not supported by current ESLint parser version
     'unicorn/prefer-module': 'off',
-    'unicorn/prevent-abbreviations': 'off',
+    'unicorn/prevent-abbreviations': 'off'
   },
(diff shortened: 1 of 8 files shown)


docs/.vuepress/components/eslint-code-block.vue
```diff
@@ -90,6 +96,14 @@ export default {
...
         parser: 'vue-eslint-parser',
         parserOptions: {
+          parser: this.typescript
+            ? this.tsEslintParser
+            : this.langTs
+            ? {
+                ts: this.tsEslintParser,
+                typescript: this.tsEslintParser
+              }
+            : null,
```
FloEdelmann, line 106: It's really hard to read this. Can we extract part of this to a separate variable to make it more readable? Or use `if`? Or configure Prettier to allow parentheses here?

Also, is the second check really needed? Or can we always set the TS parser for the `ts` and `typescript` blocks?

.eslintrc.js
```diff
@@ -131,7 +131,8 @@ module.exports = {
     'unicorn/no-useless-undefined': 'off',
     'unicorn/prefer-optional-catch-binding': 'off', // not supported by current ESLint parser version
     'unicorn/prefer-module': 'off',
-    'unicorn/prevent-abbreviations': 'off'
+    'unicorn/prevent-abbreviations': 'off',
+    'unicorn/no-nested-ternary': 'off' // conflicts with Prettier
```
FloEdelmann, line 135: But it makes the code more readable than Prettier :see_no_evil:

FloEdelmann (changes requested):
Nice, this looks mostly good. I have just one suggestion.

FloEdelmann (approved):
This is much better now. LGTM! :)

PR #1973, opened 2022-09-16
Add block-attributes-order rule

### LINKED ISSUE #1970: tag-attr-order `<script lang="ts" setup>` vs `<script setup lang="ts">`

I would love some rule that either enforces:
- `<script lang="ts" setup>` 
- OR
- `<script setup lang="ts">`

Same for:
- `<style lang="sass" scoped>`
- OR
- `<style scoped lang="sass">`

I don't mind which one should be the default, as long as we can enforce some kind of consistency.

I don't mind whatever the rule is called. I don't have a good idea of a good name myself XD
maybe something like `tag-attr-order` but it can be confused with other Vue attr stuff 

[ ] Other (please specify:)

Kingwl:
Fixes #1970


mesqueeb:
@Kingwl thanks for the PR!

is the default order going to be 
```vue
<script setup lang="ts"></script>
```
OR
```vue
<script lang="ts" setup></script>
```
it's hard to tell from the PR markdown file alone.

Kingwl:
Added the default orders into markdown.

AFAIK, there's no offical Vue recommended block attributes order. 

@sodatea Any suggestions?

mesqueeb:
@Kingwl https://vuejs.org/guide/typescript/composition-api.html

Official Vue docs use `<script setup lang="ts">`

So I think having the same default as that seems best : )

Kingwl:
Done

docs/rules/block-attributes-order.md
```diff
@@ -0,0 +1,92 @@
...
+
+> enforce order of block attributes
+
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule aims to enforce ordering of block attributes.
+
+### The default order
```
ota-meshi, line 18: I think we need to discuss option formats.

If we want to be more granular and configurable by the user, I suggest using an array of objects with elements and ordering, like this:

```js
{'vue/block-attributes-order': ['error',
  [
    { element: 'script', order: ['setup, 'lang', 'src'] },
    { element: 'template', order: ['functional', 'lang', 'src'] },
    { element: 'style', order: ['scoped', 'module', 'lang', 'src'] },
    { element: 'i18n', order: ['global', 'locale', 'lang', 'src'] },
  ]
]}
```

You can use CSS selectors for the `element` property. This is similar to the [vue/component-tags-order option](https://eslint.vuejs.org/rules/component-tags-order.html#options).
The order option specifies the order of the attributes. It can use attribute names, regular expressions and their arrays, and negation. This is similar to the [svelte/sort-attributes option](https://ota-meshi.github.io/eslint-plugin-svelte/rules/sort-attributes/#options).

If you want lang and src to always be at the back:

```js
{'vue/block-attributes-order': ['error',
  [
    { element: '*', order: [['/.*/', '!/^(?:src|lang)$/'], 'lang', 'src'] }
  ]
]}
```

Patterns can be used to correct some order even for unknown blocks and unknown attributes.

What do you think?

ota-meshi (changes requested):

xiaoxiangmoe:
Any update on this? @ota-meshi @Kingwl

ota-meshi:
I've been very busy lately with non-volunteer work. I haven't been able to check the latest changes for this PR yet.
Status is always as you can see in PR. Please refrain from comments asking about progress.

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+const defaultOptions = Object.freeze([
+  {
+    element: 'style',
+    order: [
+      WELL_KNOWN_STYLE_ATTRS.module,
+      WELL_KNOWN_STYLE_ATTRS.scoped,
+      WELL_KNOWN_STYLE_ATTRS.lang,
+      WELL_KNOWN_STYLE_ATTRS.src
+    ]
+  },
```
FloEdelmann, line 67: I think you can inline them here; since the constants are only used here. I.e.:

```ts
  {
    element: 'style',
    order: ['module', 'scoped', 'lang', 'src']
  },
```

docs/rules/block-attributes-order.md
```diff
@@ -0,0 +1,129 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/block-attributes-order
+description: enforce order of block attributes
+since: v4.3.0
```
FloEdelmann, line 6: This rule is not yet released. Please remove this line and re-run `npm run update`.

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+  function getMatchedOrderOption(element) {
+    const matchedOption = normalizedOrderOptions.find((option) =>
+      matchOrderOption(option, element)
+    )
+    if (matchedOption) {
+      return matchedOption
+    }
+    return normalizedDefaultOptions.find((option) =>
+      matchOrderOption(option, element)
+    )
```
FloEdelmann, line 199: I don't understand why there is an `if` and the same function call again otherwise. Could it not be simplified to this?

```suggestion
    return normalizedDefaultOptions.find((option) =>
      matchOrderOption(option, element)
    )
```

tests/lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,345 @@
...
+      code: '<template functional foo lang="pug" src="./template.html"></template>',
+      options: [
+        [
+          {
+            element: 'template',
+            order: ['functional', 'foo', 'lang', 'src']
+          }
+        ]
+      ]
+    },
```
FloEdelmann, line 99: The file name doesn't match the test here.

tests/lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,345 @@
...
+          {
+            element: 'style',
+            order: ['module', 'src', 'lang']
+          }
+        ]
+      ]
+    },
+    {
+      filename: 'test-attributes-options-not-specify-some-attribute-2.vue',
+      code: '<style module scoped src="./style.css" lang="less"></style>',
```
FloEdelmann, line 114: I think that unknown attributes should only be allowed at the end.

tests/lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,345 @@
...
+  ],
+  invalid: [
+    {
+      filename: 'test-template-attributes-with-invalid-order.vue',
+      code: '<template src="./template.html" lang="pug" functional></template>',
+      errors: [
+        {
+          type: 'VAttribute',
+          message: 'Attribute "lang" should go before "src".'
+        },
```
FloEdelmann, line 221: Please add `line` and `column` of the reported error to all the expected errors in this file. However, you can also remove the `type: 'VAttribute'` everywhere.

FloEdelmann (changes requested):

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+])
+
+/**
+ *
+ * @param {string | string[]} item
+ * @return {RegExp | RegExp[]}
+ */
+function ordersToRegexOrders(item) {
+  return Array.isArray(item) ? item.map((x) => new RegExp(x)) : new RegExp(item)
+}
```
ota-meshi, line 93: Could you change it to accept more complex options by doing the following?

```suggestion
const { toRegExp } = require('../utils/regexp')

/**
 * @param {string | string[]} item
 * @return {(str: string) => boolean}
 */
function ordersToMatcher(item) {
    return Array.isArray(item) ? compileMatcher(item) : compileMatcher([item])
}

/**
 * Compile matcher
 * @param {string[]} pattern
 * @return {(str: string) => boolean}
 */
function compileMatcher(pattern) {
  const rules = []
  for (const p of pattern) {
    let negative, patternStr
    if (p.startsWith("!")) {
      // If there is `!` at the beginning, it will be parsed with a negative pattern.
      negative = true
      patternStr = p.substring(1)
    } else {
      negative = false
      patternStr = p
    }
    const regex = toRegExp(patternStr)
    rules.push({ negative, match: (str) => regex.test(str) })
  }
  return (str) => {
    // If the first rule is a negative pattern, they are considered to match if they do not match that pattern.
    let result = Boolean(rules[0]?.negative)
    for (const { negative, match } of rules) {
      if (result === !negative) {
        // Even if it matches, the result does not change, so skip it.
        continue
      }
      if (match(str)) {
        result = !negative
      }
    }
    return result
  }
}
```

This change allows users to use pattern formats such as:

- `"foo"` … Matches only the `foo` attribute name.
- `"/foo/"` … Matches attribute names that match the `/foo/` regex. That is, it matches the attribute name including `foo`.
- `"!foo"` … Exclude `foo` attribute from the matched attribute names. When used first in the array or alone, matches other than the `foo` attribute name.
- `"!/foo/"` … Excludes attributes that match the `/foo/` regex from the matched attribute names. When used first in the array or alone, matches an attribute name that does not match the `/foo/` regex.
- `["foo", "/^bar/u"]` … Matches the `foo` attribute or the attribute name that matches the `/^bar/u` regex.
- `["/^foo/u", "!foo-bar", "/^baz/u"]` … Matches an attribute name that matches `/^foo/u` and other than `foo-bar`, or an attribute name that matches `/^baz/u`.

I think the rule of other plugin will also be helpful.
https://ota-meshi.github.io/eslint-plugin-svelte/rules/sort-attributes/

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+      {
+        type: 'array',
+        items: {
+          type: 'object',
+          properties: {
+            element: {
+              type: 'string'
+            },
+            order: {
+              type: 'array',
```
ota-meshi, line 296: Could you please define the schema so that at least one item is set?

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+      description: 'enforce order of block attributes',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/block-attributes-order.html'
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'array',
+        items: {
+          type: 'object',
```
ota-meshi, line 290: Could you define the schema so that the two properties are required?

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+      WELL_KNOWN_SCRIPT_ATTRS.lang,
+      WELL_KNOWN_SCRIPT_ATTRS.src
+    ]
+  },
+  {
+    element: 'template',
+    order: [
+      WELL_KNOWN_TEMPLATE_ATTRS.functional,
+      WELL_KNOWN_TEMPLATE_ATTRS.lang,
+      WELL_KNOWN_TEMPLATE_ATTRS.src
```
ota-meshi, line 81: After changing it to accept complex ordering, could you change it so that `lang` and `src` are always ordered last?

e.g.

```suggestion
      "functional",
      ["!lang", "!src"],
      "lang",
      "src"
```

Change other elements in the same way.

lib/rules/block-attributes-order.js
```diff
@@ -0,0 +1,320 @@
...
+    ]
+  }
+])
```
ota-meshi, line 84: Could you add the following item to force the order of the custom elements as well?

```js
  {
    element: '*',
    order: [
      ["!lang", "!src"],
      "lang",
      "src"
    ]
  },
```

ota-meshi (changes requested):

FloEdelmann:
Ping @Kingwl

mesqueeb:
@Kingwl friendly bump : )

paescuj:
Hey there, would be great to have this rule 👍 Happy to help! I can also come up with a new pull request if that is preferred.

FloEdelmann:
From our side, a new PR replacing this one would be fine.

FloEdelmann:
I'll close this for now as it seems abandoned and there are already some merge conflicts. Feel free to open another PR :slightly_smiling_face: 

@paescuj this is your chance! :wink:

Thomasan1999:
@FloEdelmann I was looking for this rule. Has it been implemented since?

FloEdelmann:
No

Issue #1978, opened 2022-09-17
`vue/padding-line-between-tags` with new 'consistent' option


mesqueeb:
**What rule do you want to change?**
`vue/padding-line-between-tags`
https://eslint.vuejs.org/rules/padding-line-between-tags.html#vue-padding-line-between-tags

**Does this change cause the rule to produce more or fewer warnings?**
depends on the config

**How will the change be implemented? (New option, new default behavior, etc.)?**
new option `"blankLine": "consistent"`

I don't want blank lines in between **all** siblings or after specific tags.

What I would love:
- no blank lines between siblings: GOOD
- blank lines between siblings: also GOOD
- under the _same parent_ blank lines between some siblings, but not others: BAD

Inconsistent spacing between **same parent siblings** is really inconsistent and drives me crazy. However, I do not want to force or disallow blank lines, I feel it's often a case-by-case choice depending how cluttered a parent with siblings looks.

My rule wish is:
```json
{
  "vue/padding-line-between-tags": ["error", [
    { "blankLine": "consistent", "prev": "*", "next": "*" }
  ]]
}
```

**Please provide some example code that this change will affect:**

GOOD
```vue
<template>
  <header>
    <div></div>
    <div></div>
    <div></div>
  </header>

  <div></div>
  
  <div />
  
  <footer></footer>
</template>
```

GOOD
```vue
<template>
  <header>
    <div></div>

    <div></div>
    
    <div></div>
  </header>
  <div></div>
  <div />
  <footer></footer>
</template>
```

BAD
```vue
<template>
  <header>
    <div></div>

    <div></div>
    <div></div>
  </header>
  <div></div>

  <div />
  <footer></footer>
</template>
```

FloEdelmann:
CC @dev1437 and @amiranagram, follow-up to #1832 and #1966.

dev1437:
https://github.com/vuejs/eslint-plugin-vue/pull/1982 @FloEdelmann, but we will definitely need more edge cases defined and tests for them. If @mesqueeb wants to try the current behaviour and see if there are any issues, that would be great.

mesqueeb:
@dev1437 sure! And thanks so much for this implementation! 

I think it's easiest for me to test it if you could release it to npm once. @FloEdelmann are you comfortable releasing it to npm once?

I can open new issues for any lint bugs for this rule i come across in the future.

FloEdelmann:
[...]

PR #1982, opened 2022-09-19
Add `consistent` option to `vue/padding-line-between-tags`

dev1437:
For #1978

### DIFF
--- a/lib/rules/padding-line-between-tags.js
+++ b/lib/rules/padding-line-between-tags.js
@@ -125,6 +125,14 @@ function checkNewline(context) {
     { blankLine: 'always', prev: '*', next: '*' }
   ]
 
+  const reverseConfigureList = [...configureList].reverse()
+
+  /**
+   * It has the style of the first `blankLine="consistent"`.
+   * @type {Map<VElement, "always" | "never">}
+   */
+  const firstConsistentBlankLines = new Map()
+
   /**
    * @param {VElement} block
    */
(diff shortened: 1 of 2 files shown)


lib/rules/padding-line-between-tags.js
```diff
@@ -109,40 +140,38 @@ function checkNewline(context) {
...
-            for (
-              let i = endTag.loc.start.line;
-              i < closestSibling.loc.start.line - 1 && hasOnlyTextBetween;
-              i++
-            ) {
-              hasOnlyTextBetween = !/^\s*$/.test(
-                context.getSourceCode().lines[i]
-              )
+          insertNewLine(context, block, closestSibling, lineDifference)
+        } else if (configure.blankLine === 'consistent') {
```
ota-meshi, line 144: I think we need to look more into how the `consistent` option works.
For example, I think the following case should not be warned.

```vue
<template>
  <div>
    <div />

    <div />

    <br/>
    <br/>
    <br/>
    <br/>
    <br/>
    <br/>

    <div />

    <div />
  </div>
</template>

<script>
/* eslint   "vue/padding-line-between-tags": ["error", [
  { "blankLine": "consistent", "prev": "*", "next": "*" },
  { "blankLine": "never", "prev": "br", "next": "br" },
]] */
</script>
```

What do you think?

ota-meshi (changes requested):

lib/rules/padding-line-between-tags.js
dev1437: Hi, I'm not sure what you mean by this, this template will not produce any errors

dev1437: Nevermind, I reran the test and it failed :)

lib/rules/padding-line-between-tags.js
```diff
@@ -109,40 +154,38 @@ function checkNewline(context) {
...
+              child.loc.start.line - prevChild.loc.end.line
+
+            if (siblingLineDifference > 1) {
+              newlineCount++
             }
+            prevChild = /** @type {VElement} */ (child)
           }
+          const ratio = newlineCount / (siblingElements.length - 1)
+
+          if (0.5 - ratio < 0.005) {
```
ota-meshi, line 182: I think it would be simpler to check if it matches the first line break style instead of the ratio. Users can switch styles using auto-fix after changing the first one to their preferred line break style.

ota-meshi (changes requested):

dev1437:
Updated @ota-meshi let me know if anything else

lib/rules/padding-line-between-tags.js
```diff
@@ -109,40 +157,23 @@ function checkNewline(context) {
...
+              hasConsistentConfiguration(element, configureList)
+          )
+
+          const siblingLineDifference =
+            siblingElements[1].loc.start.line - siblingElements[0].loc.end.line
+          if (siblingLineDifference > 1) {
+            insertNewLine(context, block, closestSibling, lineDifference)
+          } else {
+            removeExcessLines(context, endTag, closestSibling, lineDifference)
           }
```
ota-meshi, line 174: I think that branching can be simplified by rewriting the logic beyond L150 as follows.


```js
    const configure = reverseConfigureList.find(
      (configure) =>
        (configure.prev === '*' || block.name === configure.prev) &&
        (configure.next === '*' || closestSibling.name === configure.next)
    )

    if (!configure) {
      return
    }
    const lineDifference = closestSibling.loc.start.line - endTag.loc.end.line

    let blankLine = configure.blankLine
    if (blankLine === 'consistent') {
      const firstConsistentBlankLine = firstConsistentBlankLines.get(
        block.parent
      )
      if (firstConsistentBlankLine == null) {
        // If don't have a style for the first `blankLine="consistent"` yet, set the style and finish checking.
        firstConsistentBlankLines.set(
          block.parent,
          lineDifference > 1 ? 'always' : 'never'
        )
        return
      }
      blankLine = firstConsistentBlankLine
    }

    if (blankLine === 'always') {
      insertNewLine(context, block, closestSibling, lineDifference)
    } else {
      removeExcessLines(context, endTag, closestSibling, lineDifference)
    }
```

For this we need to add the following variable declarations to L127.

```js
  const reverseConfigureList = [...configureList].reverse()

  /**
   * It has the style of the first `blankLine="consistent"`.
   * @type {Map<VElement, "always" | "never">}
   */
  const firstConsistentBlankLines = new Map()
```

ota-meshi (changes requested):

dev1437:
Changes done @ota-meshi, nice job

ota-meshi (approved):
LGTM! Thank you!

FloEdelmann (approved):

PR #1985, opened 2022-09-20
docs: improve example to better clarify rule

mesqueeb:


FloEdelmann:
Thanks for these docs improvements!

PR #1988, opened 2022-09-22
Fix false positives for uninitialized vars in `vue/no-ref-as-operand` rule

### LINKED ISSUE #1980: `vue/no-ref-as-operand` should not autofix when variable is `Ref | NotARef`
**Checklist**

- **ESLint version:**  8.23.1
- **eslint-plugin-vue version:** 9.5.1
- **Node version:** 16.14.2
- **Operating System:** win

```js
module.exports = {
  extends: ['jkarczm/vuetify'],
  parserOptions: {
    project: 'tsconfig.json',
  },
}
```
(where `jkarczm/vuetify` is from https://github.com/jacekkarczmarczyk/eslint-config-jkarczm package, but probably including only `vue/no-ref-as-operand` matters)

**What did you do?**	

```ts
let foo: Ref<number> | undefined;

if (!foo) {
  foo = ref(5);
}

```

no error, no autofixing

**What actually happened?**

code was "fixed" to

```js
let foo: Ref<number> | undefined;

if (!foo.value) {
  foo.value = ref(5);
}
```

ota-meshi:
close #1980

### DIFF
--- a/lib/rules/no-ref-as-operand.js
+++ b/lib/rules/no-ref-as-operand.js
@@ -9,8 +9,21 @@ const utils = require('../utils')
 
 /**
  * @typedef {import('../utils/ref-object-references').RefObjectReferences} RefObjectReferences
+ * @typedef {import('../utils/ref-object-references').RefObjectReferenceForIdentifier} RefObjectReferenceForIdentifier
  */
 
+/**
+ * Checks whether the given identifier reference has been initialized with a ref object.
+ * @param {RefObjectReferenceForIdentifier | null} data
+ * @returns {data is RefObjectReferenceForIdentifier}
+ */
+function isRefInit(data) {
+  const init = data && data.variableDeclarator && data.variableDeclarator.init
+  if (!init) {
+    return false
+  }
+  return data.defineChain.includes(/** @type {any} */ (init))
+}
 module.exports = {
   meta: {
     type: 'suggestion',
@@ -37,7 +50,7 @@ module.exports = {
      */
     function reportIfRefWrapped(node) {
       const data = refReferences.get(node)
-      if (!data) {
+      if (!isRefInit(data)) {
         return
       }
       context.report({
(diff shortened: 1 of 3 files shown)


jacekkarczmarczyk:
Not sure about the PR implementation, but I think the title is wrong - it's not only about "uninitialized" variables, it's about variables that are `Ref | NotARef`, for example
```ts
let foo: Ref<number> | number = 5;

if (typeof foo === 'number') {
  foo = ref(3); 
}
```
is also "fixed" to 
```ts
const foo: Ref<number> | number = 5;

if (typeof foo.value === 'number') {
  foo.value = ref(3);
}
```
even though it's initialized

ota-meshi:
I am not familiar with English. Could you suggest a title?

The example you provided seems to work fine when I checked it on the demo site.

```vue
<script lang="ts">
import { ref } from 'vue'
let foo: Ref<number> | number = 5;

foo = ref(3);
</script>
```
https://deploy-preview-1988--eslint-plugin-vue.netlify.app/rules/no-ref-as-operand.html

jacekkarczmarczyk:
I'm not native english neither, but I think that issue title describes the problem fine and might be used as a PR title too

ota-meshi:
The rule will report it if: 

```vue
<script lang="ts">
import { ref } from 'vue'
let foo: Ref<number> | number = ref(3);

foo = 5;
</script>
```

(If you think this is a false positive, I recommend turning off the rule.)

So I think `Ref | NotARef` doesn't show the correct meaning.

jacekkarczmarczyk:
I see your point. I'm not really sure if your code should report the error, however suggestion to turn the rule off seems to make sense as the problem would be reported by TS anyway, at least in this case (and that's actually how I've found out the problem, if not TS I might have missed it and end up with incorrect code). 

On the other hand there are cases where TS wouldn't spot the error, e.g.
```ts
let t: Ref|number = ref(5);
function fn () {
  if (t > 3) t = 1;
}
```
I'm investigating this issue now

PR #1989, opened 2022-09-26
Add `globals` option to `component-name-in-template-casing`

lsdsjy:
Currently the `component-name-in-template-casing` can't check globally registered components. This PR adds an option to explicitly list global components to check.

### DIFF
--- a/lib/rules/component-name-in-template-casing.js
+++ b/lib/rules/component-name-in-template-casing.js
@@ -40,6 +40,11 @@ module.exports = {
       {
         type: 'object',
         properties: {
+          globals: {
+            type: 'array',
+            items: { type: 'string' },
+            uniqueItems: true
+          },
           ignores: {
             type: 'array',
             items: { type: 'string' },
@@ -63,13 +68,15 @@ module.exports = {
       : defaultCase
     /** @type {RegExp[]} */
     const ignores = (options.ignores || []).map(toRegExp)
+    /** @type {string[]} */
+    const globals = (options.globals || []).map(casing.pascalCase)
     const registeredComponentsOnly = options.registeredComponentsOnly !== false
     const tokens =
       context.parserServices.getTemplateBodyTokenStore &&
       context.parserServices.getTemplateBodyTokenStore()
 
     /** @type { Set<string> } */
-    const registeredComponents = new Set()
+    const registeredComponents = new Set(globals)
 
     if (utils.isScriptSetup(context)) {
       // For <script setup>
@@ -106,15 +113,8 @@ module.exports = {
         }
         return true
       }
-      // We only verify the components registered in the component.
-      if (
-        [...registeredComponents]
-          .filter((name) => casing.isPascalCase(name)) // When defining a component with PascalCase, you can use either case
-          .some(
-            (name) =>
-              node.rawName === name || casing.pascalCase(node.rawName) === name
-          )
-      ) {
+      // We only verify the registered components.
+      if (registeredComponents.has(casing.pascalCase(node.rawName))) {
         return true
       }
 
--- a/tests/lib/rules/component-name-in-template-casing.js
+++ b/tests/lib/rules/component-name-in-template-casing.js
@@ -901,7 +901,13 @@ tester.run('component-name-in-template-casing', rule, {
           <RouterView />
         </template>
       `,
-      errors: ['Component name "router-view" is not PascalCase.']
+      errors: [
+        {
+          message: 'Component name "router-view" is not PascalCase.',
+          line: 3,
+          column: 11
+        }
+      ]
     },
     {
       code: `
@@ -915,7 +921,13 @@ tester.run('component-name-in-template-casing', rule, {
           <router-view />
         </template>
       `,
-      errors: ['Component name "RouterView" is not kebab-case.']
+      errors: [
+        {
+          message: 'Component name "RouterView" is not kebab-case.',
+          line: 3,
+          column: 11
+        }
+      ]
     },
     {
       code: `
@@ -929,7 +941,13 @@ tester.run('component-name-in-template-casing', rule, {
           <router-view />
         </template>
       `,
-      errors: ['Component name "RouterView" is not kebab-case.']
+      errors: [
+        {
+          message: 'Component name "RouterView" is not kebab-case.',
+          line: 3,
+          column: 11
+        }
+      ]
     }
   ]
 })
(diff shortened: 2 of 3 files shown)


tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -168,6 +168,24 @@ tester.run('component-name-in-template-casing', rule, {
...
+      `,
+      options: ['PascalCase', { globals: ['RouterView'] }]
+    },
+    {
+      code: `
+        <template>
+          <RouterView />
+        </template>
+      `,
+      options: ['PascalCase', { globals: ['router-view'] }]
```
FloEdelmann, line 188: Please add two test cases for the `kebab-case` option as well.

tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -854,6 +872,39 @@ tester.run('component-name-in-template-casing', rule, {
...
+    },
+    {
+      code: `
+        <template>
+          <router-view />
+        </template>
+      `,
+      options: ['PascalCase', { globals: ['RouterView'] }],
+      output:
+        '\n        <template>\n          <RouterView />\n        </template>\n      ',
```
FloEdelmann, line 884: Please use `` `template literals` `` with actual newlines instead of normal `'strings'` with escaped newlines.

tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -854,6 +872,39 @@ tester.run('component-name-in-template-casing', rule, {
...
+      errors: ['Component name "router-view" is not PascalCase.']
```
FloEdelmann, line 885: Please also check line and column of the error.

FloEdelmann (changes requested):
The option looks good to me. But I have a few requests.

tests/lib/rules/component-name-in-template-casing.js
lsdsjy: Done.

tests/lib/rules/component-name-in-template-casing.js
lsdsjy: Done.

tests/lib/rules/component-name-in-template-casing.js
lsdsjy: Sorry I don't know what u mean by "line and column of the error" 😥

FloEdelmann: See e.g. https://github.com/vuejs/eslint-plugin-vue/blob/467e85fdf29139050d92e61bd7ea217a53af87cb/tests/lib/rules/no-restricted-component-options.js#L145-L149

tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -168,6 +168,40 @@ tester.run('component-name-in-template-casing', rule, {
...
+      options: ['kebab-case', { globals: ['RouterView'] }]
+    },
+    {
+      code: `
+        <template>
+          <router-view />
+        </template>
+      `,
+      options: ['kebab-case', { globals: ['router-view'] }]
     }
```
FloEdelmann, line 205: On a second thought, these could be combined:

```suggestion
    {
      code: `
        <template>
          <div>
            <RouterView />
            <RouterLink />
          </div>
        </template>
      `,
      options: ['PascalCase', { globals: ['RouterView', 'router-link'] }]
    },
    {
      code: `
        <template>
          <div>
            <router-view />
            <router-link />
          </div>
        </template>
      `,
      options: ['kebab-case', { globals: ['RouterView', 'router-link'] }]
    }
```

lsdsjy: Yes that's clever👍

lib/rules/component-name-in-template-casing.js
```diff
@@ -106,15 +113,8 @@ module.exports = {
...
-      if (
-        [...registeredComponents]
-          .filter((name) => casing.isPascalCase(name)) // When defining a component with PascalCase, you can use either case
-          .some(
-            (name) =>
-              node.rawName === name || casing.pascalCase(node.rawName) === name
-          )
-      ) {
+      // We only verify the registered components.
+      if (registeredComponents.has(casing.pascalCase(node.rawName))) {
```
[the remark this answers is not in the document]
lsdsjy: I've [tested](https://stackblitz.com/edit/vitejs-vite-x9voie?file=src%2FApp.vue,src%2Fmain.js) and found that it won't work. But the original logic is also wrong because [it filters out the non-pascal-case names](https://github.com/vuejs/eslint-plugin-vue/pull/1989/files/5eb9c6945173f9a4d065e27a86c7d002cc61386e#diff-ae2b61140fbfc71296afe34f070548020a61fdf6a98e2320e47143c776da7589L112). So the bahaviour is essentially the same here. 
I think the situation is a little complicated here: if a component is registered with kebab-case, but used as PascalCase, how should our rule behave here if case style is set to PascalCase? Seems like it shouldn't report it as an error but if it doesn't, the component in fact won't be resolved and rendered correctly.(But maybe it is out of the rule's scope?)

ota-meshi (changes requested):
Thank you for this PR! The rule implementation looks good! Could you please add a example code to the docs where we can try out the new option?

lsdsjy:
> Could you please add a example code to the docs where we can try out the new option?

Done.

ota-meshi (approved):
LGTM!

lib/rules/component-name-in-template-casing.js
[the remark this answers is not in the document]
FloEdelmann: @ota-meshi is this logic change fine for you?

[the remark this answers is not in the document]
ota-meshi: @FloEdelmann Yeah. I think it's fine.

lsdsjy:
> When registering a component with kebab-case, can it be used with PascalCase?

I've thought a little more about this problem. Consider this:
```js
// main.js
Vue.component('SomeComp', ...)

// App.js
export default {
  components: { 'some-comp': ... },
  template: '<SomeComp />'
}
```
It works fine in that `<SomeComp />` would resolve to the globally registered component, though there is a locally registered component `some-comp` and we would think it's a mistake at first glance. So if the options is `kebab-case` and `registeredComponentsOnly` is:
- `true`, then `<SomeComp />` is OK because currently we cannot determine whether it is a "(locally) registered component";
- `false`, then `<SomeComp />` would be an lint error as expected.

So the current logic is fine.

FloEdelmann (approved):

PR #1994, opened 2022-09-28
Add new `require-prop-comment` rule

### LINKED ISSUE #1993: require-prop-comment

require prop should have a comment

[X ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

```vue
      <template>
        <div>1</div>
      </template>
      <script setup>
      const props = defineProps({
        a: Number
      })
      </script>
```
```vue
      <template>
        <div>1</div>
      </template>
      <script>
      import { defineComponent } from '@vue/composition-api'

        props: {
          /**
           * a comment
           */
          a: Number,
          b: Number,
          // c
          // comment
          c: Number
        },
        setup() {}
      })
      </script>
```

**Additional context**

czb3279338858:
Fixes #1993

### DIFF
--- a/docs/rules/no-required-prop-with-default.md
+++ b/docs/rules/no-required-prop-with-default.md
@@ -53,31 +53,21 @@ This rule enforces all props with default values to be optional.
 ```vue
 <script>
   export default {
-    /* ✓ GOOD */
     props: {
-      name: {
+      /* ✓ GOOD */
+      foo: {
+        required: false,
+        default: 'Hello'
+      },
+      bar: {
+        required: true
+      },
+      /* ✗ BAD */
+      baz: {
         required: true,
         default: 'Hello'
-      }
-    }
-  }
-</script>
-```
-
-</eslint-code-block>
-
-<eslint-code-block fix :rules="{'vue/no-required-prop-with-default': ['error', { autofix: true }]}">
-
-```vue
-<script>
-  export default {
-    /* ✗ BAD */
-    props: {
-      name: {
-        required: true,
-        default: 'Hello'
-      }
-    }
+      },
+    },
   }
 </script>
 ```


lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,92 @@
...
+        context.report({
+          node: prop.node,
+          message: 'The "{{name}}" property should have a comment.',
+          data: {
+            name: prop.propName
+          }
+        })
+      }
+    }
+    // 合成选择器
```
FloEdelmann, line 61: Please only write English comments (even if it is bad English), then all maintainers have a chance to understand this.

lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,92 @@
...
+'use strict'
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const utils = require('../utils')
+
+// ------------------------------------------------------------------------------
+// Helpers
```
FloEdelmann, line 14: Please remove unneeded comment blocks.

lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * @author *****your name*****
```
FloEdelmann, line 2: Please add your name or username (in all files).

lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,92 @@
...
+      // utils.executeOnVueComponent(context, (obj) => {
+      //   verifyProps(utils.getComponentPropsFromOptions(obj))
+      // })
+
+      // 检查组件是否 vue 实例或组件
+      // 不适用于 <script setup>
+      // utils.executeOnVue(context, (obj) => {
+      //   // getComponentPropsFromOptions 从组件配置中获取所有 prop
+      //   verifyProps(utils.getComponentPropsFromOptions(obj))
+      // })
```
FloEdelmann, line 89: Please remove this commented code.

tests/lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,200 @@
...
+      import { defineComponent } from '@vue/composition-api'
+      export const ComponentName = defineComponent({
+        props: {
+          /**
+           * a comment
+           */
+          a: Number
+        },
+        render() {
+          return <div>1</div>
```
FloEdelmann, line 70: JSX requires additional test setup and thus errors (see the CI failures). Since it is not needed in the test cases for this rule, please just remove it.

tests/lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,200 @@
...
+          // c
+          // comment
+          c: Number
+        }
+      })
+      </script>
+      `,
+      errors: [
+        {
+          line: 10
```
FloEdelmann, line 99: Please also add the `message` and `column` to the checked errors.

lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,92 @@
...
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'require prop should have a comment',
```
FloEdelmann, line 27:
```suggestion
      description: 'require that props have a comment',
```

Please re-run the update script to update the description in all files.

docs/rules/require-prop-comment.md
```diff
@@ -0,0 +1,68 @@
...
+  props: {
+    /**
+     * a comment
+     */
+    a: Number,
+    // b comment
+    b: Number,
+    // c
+    // comment
+    c: Number
```
FloEdelmann, line 37: Should single-line comments really be considered "good"? When documenting props is the goal, I think JSDoc comments should be preferred over single-line comments.

Probably also documentation generator tools will ignore single-line comments. What's your opinion about this?

docs/rules/require-prop-comment.md
```diff
@@ -0,0 +1,68 @@
...
+---
+# vue/require-prop-comment
+
+> require prop should have a comment
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule requires that the prop should have a comment
```
FloEdelmann, line 15:
```suggestion
This rule enforces that every prop has a comment that documents it.
```

FloEdelmann (changes requested):

lib/rules/require-prop-comment.js
czb3279338858: The above problems will be solved in the near future

docs/rules/require-prop-comment.md
czb3279338858: It has been completed and will be merged in the next submission

lib/rules/require-prop-comment.js
czb3279338858: ok

docs/rules/require-prop-comment.md
czb3279338858: I think JSDoc comments should be preferred over single-line comments too.
The second parameter will provide a choice，line 、block or unlimited
I am thinking about the following questions
1.Can I mix them?
2.Need to limit the quantity?

FloEdelmann: I think the rule should only consider JSDoc comments (`/** */`) and ignore block comments (`/* */`) and line comments `//`. No need to limit the number then, since only the JSDoc comment immediately before the statement will be considered.

PR #1996, opened 2022-09-28
Create `vue/multiline-ternary` extension rule

dev1437:
Wrap multiline ternary rule for vue.

Not sure about the indenting, maybe it is not an issue and would be handled by other rules. Looks strange.

### DIFF
--- a/docs/rules/multiline-ternary.md
+++ b/docs/rules/multiline-ternary.md
@@ -2,16 +2,48 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/multiline-ternary
-description: Enforce newlines between operands of ternary expressions in `<template>` and `<style>`
+description: Enforce newlines between operands of ternary expressions in `<template>`
 ---
 # vue/multiline-ternary
 
-> Enforce newlines between operands of ternary expressions in `<template>` and `<style>`
+> Enforce newlines between operands of ternary expressions in `<template>`
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
 - :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 This rule is the same rule as core [multiline-ternary] rule but it applies to the expressions in `<template>` and `<style>`.
+
+## :book: Rule Details
+
+<eslint-code-block fix :rules="{'vue/multiline-ternary': ['error']}">
+
+```vue
+<template>
+  <div>
+    <!-- ✓ GOOD -->
+    <div :class="isEnabled
+      ? 'check'
+      : 'stop'" />
+
+    <!-- ✗ BAD -->
+    <div :class="isEnabled ? 'check' : 'stop'" />
+  </div>
+</template>
+
+<style>
+div {
+  /* ✓ GOOD */
+  color: v-bind('myFlag
+    ? foo
+    : bar');
+
+  /* ✗ BAD */
+  color: v-bind('myFlag ? foo : bar');
+}
+</style>
+```
+
+</eslint-code-block>
 
 ## :books: Further Reading
 
--- a/docs/rules/multiline-ternary.md
+++ b/docs/rules/multiline-ternary.md
@@ -2,29 +2,17 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/multiline-ternary
-description: xxx
+description: Enforce newlines between operands of ternary expressions in `<template>` 
 ---
 # vue/multiline-ternary
 
-> xxx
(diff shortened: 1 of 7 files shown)


FloEdelmann:
Could you please fix the lint issues? Also, there are some tests failing in ESLint v6, @ota-meshi maybe you can help how to skip those tests there?

dev1437:
Not sure why the test fail in eslint v6 but run fine locally, but lint is okay now

ota-meshi:
It seems that the multiline-ternary rule did not previously support autofix.

I think it is necessary to check the version of eslint and ignore the `output` check (set `null`).
I think the following test case will be helpful.
https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-loss-of-precision.js#L20

lib/rules/multiline-ternary.js
```diff
@@ -0,0 +1,16 @@
...
+
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const { wrapCoreRule } = require('../utils')
+
+// eslint-disable-next-line no-invalid-meta, no-invalid-meta-docs-categories
+module.exports = wrapCoreRule('multiline-ternary', {
+  skipDynamicArguments: true
```
ota-meshi, line 15: I think it's probably a rule that can be applied in CSS v-bind as well.

```suggestion
  skipDynamicArguments: true,
  applyDocument: true
```

Additionally, could you add test cases to check if it works?

ota-meshi (changes requested):

FloEdelmann:
Please also add the new rule to https://deploy-preview-1996--eslint-plugin-vue.netlify.app/rules/#extension-rules.

It seems that this doesn't work automatically when running `npm run update`.

tests/lib/rules/multiline-ternary.js
```diff
@@ -0,0 +1,276 @@
...
+          'test': someReallyLongCondition ? aVeryLongOutput : thisCantFitOnASingleLine
+        }">
+        </div>
+      </template>
+      `,
+      options: ['never']
+    }
+  ],
+  invalid: [
+    ...(semver.gte(ESLint.version, '7.1.0')
```
ota-meshi, line 48: I think it is better to branch at the output property instead of branching here.

```js
     output: semver.gte(ESLint.version, '7.0.0')
       ? ` .... `
       : null,
```

lib/rules/multiline-ternary.js
```diff
@@ -0,0 +1,17 @@
...
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const { wrapCoreRule } = require('../utils')
+
+// eslint-disable-next-line no-invalid-meta, no-invalid-meta-docs-categories
+module.exports = wrapCoreRule('multiline-ternary', {
+  skipDynamicArguments: true,
+  applyDocument: true
```
ota-meshi, line 16: Could you add test cases for CSS v-bind()?

ota-meshi (changes requested):

ota-meshi (approved):
LGTM! Thank you!

lib/rules/multiline-ternary.js
FloEdelmann: Will this also include ternary statements in the `<script>` block? I think those should be excluded, as the core `multiline-ternary` rule will catch those already.

Also, please update the docs and/or description to mention `v-bind` in the `<style>` block.

ota-meshi:
https://github.com/vuejs/eslint-plugin-vue/pull/1996#discussion_r985069682

tests/lib/rules/multiline-ternary.js
```diff
@@ -0,0 +1,296 @@
...
+      `
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <script>
+        let test = someReallyLongCondition ? aVeryLongOutput : thisCantFitOnASingleLine
+      </script>
+      `
+    },
```
FloEdelmann, line 41:
```suggestion
    {
      // doesn't check ternary statements in <script> block
      filename: 'test.vue',
      code: `
      <script>
        let test = someReallyLongCondition ? aVeryLongOutput : thisCantFitOnASingleLine
      </script>
      `
    },
```

docs/rules/multiline-ternary.md
```diff
@@ -2,16 +2,16 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/multiline-ternary
-description: Enforce newlines between operands of ternary expressions in `<template>`
+description: Enforce newlines between operands of ternary expressions in `<template>` and `<style>`
```
FloEdelmann, line 5: Won't this be overridden by `npm run update`?

dev1437: Yep, where do I need to update it?

dev1437: Done, sorry been very busy

FloEdelmann (approved):

PR #2006, opened 2022-10-08
WIP: Improve `vue/v-on-function-call` rule to check for functions

### LINKED ISSUE #2001: edit `v-on-function-call` to enforce `@click="() => fn()"` over `@click="fn()"` and `@click="fn"`
https://eslint.vuejs.org/rules/v-on-function-call.html#vue-v-on-function-call

I want to have a new option for `v-on-function-call` to enforce a 3rd way of writing that is not yet enforcable. My reasoning is very sound I believe bet let me elaborate:

There are 3 ways in Vue to pass an event listener in HTML:
```html
<button @click="fn" />
<button @click="fn()" />
<button @click="() => fn()" />    
```

To me there are clear PROs for using the last way and only CONs for using the first two. This is why I want to enforce this style with a new rule.

Here is my reasoning:

```html
<button @click="fn" />
```
- CON: it's unclear to see what payload comes out of the event and what payload goes into the `fn` if any.
- CON: sometimes `fn` doesn't accept a payload, but will still receive one because of this way of writing
- CON: sometimes `fn` is used somewhere else with a payload, but passed here as well, but it will end up receiving `PointerEvent | MouseEvent` implicitly which can cause bugs that are hard to find because it's implicit
- CON: a _lot_ of VSCode themes color this the same color as a _prop_, so it becomes hard to see that it's a function:
  image
 

```html
<button @click="fn()" />
```
- CON: this just looks like a function is being passed and _immidiately executed_...
- CON: if this were a _prop_, then this way of writing would immidiately execute the function. So it's really confusing

```html
<button @click="() => fn()" />    
```
- PRO: it doesn't look like a function is being immidiately executed
- PRO: it's clear what payload comes out of the click event and what payload goes into the `fn`, if any

[ ] Other (please specify:)

I believe the explanation above is clear enough but let me know if it doesn't

ota-meshi:
close #2001

### DIFF
--- a/lib/rules/v-on-function-call.js
+++ b/lib/rules/v-on-function-call.js
@@ -18,6 +18,24 @@ const utils = require('../utils')
 // ------------------------------------------------------------------------------
 
 /**
+ * @param {RuleContext} context
+ */
+function parseOptions(context) {
+  /** @type {Set<'always' | 'never' | 'function'>} */
+  const allows = new Set()
+  if (Array.isArray(context.options[0])) {
+    for (const o of context.options[0]) allows.add(o)
+  } else {
+    allows.add(context.options[0] || 'never')
+  }
+
+  const option = context.options[1] || {}
+  const ignoreIncludesComment = !!option.ignoreIncludesComment
+
+  return { allows, ignoreIncludesComment }
+}
+
+/**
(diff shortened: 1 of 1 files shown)


FloEdelmann:
Superseded by #2009.

PR #2008, opened 2022-10-11
Improve `new-rule` tool and templates

### LINKED ISSUE #1998: Meta: Improve tools and templates
- [ ] The [`new-rule` tool](https://github.com/vuejs/eslint-plugin-vue/blob/master/tools/new-rule.js) (`npm run new`) should take a second (required) command line parameter asking for the author name/username and fill that into the template automatically. Currently, many new contributors ignore that field and it needs to be raised in the review.
- [x] ~The [`update-docs-rules-index` tool](https://github.com/vuejs/eslint-plugin-vue/blob/master/tools/update-docs-rules-index.js) (included in `npm run update`) should also automatically update the list of extension rules in https://eslint.vuejs.org/rules/#extension-rules. Noticed in [#1996 (comment)](https://github.com/vuejs/eslint-plugin-vue/pull/1996#issuecomment-1263497587).~ actually seems to be already implemented.
- [ ] The `Requirements`, `Helpers` and `Rule Definition` [comment sections](https://github.com/vuejs/eslint-plugin-vue/blob/cae6d299618442d7bb05f21b14b32073b18d43f2/tools/new-rule.js#L31-L45) are sometimes hard to follow and their benefit is negligible. Also, most contributors ignore them completely, so it needs to be mentioned manually in the review. I suggest to drop them in the [`new-rule` tool](https://github.com/vuejs/eslint-plugin-vue/blob/master/tools/new-rule.js) and in every existing rule.

FloEdelmann:
Closes #1998.


ota-meshi (approved):
LGTM! Thank you.

PR #2009, opened 2022-10-12
Add `vue/v-on-handler-style` rule and deprecate `vue/v-on-function-call` rule

ota-meshi:
This PR adds `vue/v-on-handler-style` rule.
And deprecates `vue/v-on-function-call` rule.

close https://github.com/vuejs/eslint-plugin-vue/issues/2001


docs/rules/v-on-handler-style.md
```diff
@@ -0,0 +1,153 @@
...
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <button v-on:click="handler()">...</button>
+
+  <!-- ✗ BAD -->
+  <button v-on:click="handler">...</button>
+  <button v-on:click="() => handler()">...</button>
+  <button v-on:click="(foo) => handler(foo)">...</button>
+</template>
```
FloEdelmann, line 101: Please also add an example and test for ignored style here (and check if it is actually ignored):

```vue
<template>
  <!-- ✓ GOOD -->
  <button v-on:click="handler()">...</button>

  <!-- ✗ BAD -->
  <button v-on:click="handler">...</button>
  <button v-on:click="() => handler()">...</button>
  <button v-on:click="(foo) => handler(foo)">...</button>

  <!-- Ignore -->
  <button v-on:click="(foo, bar) => handler(foo, bar)">...</button>
  <button v-on:click="handler">...</button> <!-- if handler has multiple parameters and/or event has multiple payload parameters; don't know if we can check that -->
</template>
```

tests/lib/rules/v-on-handler-style.js
```diff
@@ -0,0 +1,913 @@
...
+    },
+    // inline -> method
+    {
+      filename: 'test.vue',
+      code: `<template>
+        <button @click="foo(123)" />
+        <button @click="foo.bar()" />
+        <button @click="foo?.()" />
+        <button @click="foo();foo();" />
+        <button @click="{}" />
```
FloEdelmann, line 59: Shouldn't these cases be reported (without fix) to suggest moving the code into methods?

docs/rules/v-on-handler-style.md
FloEdelmann: Or, on a second thought: Maybe we should rather simplify the a lot logic by not ignoring anything? If the user wants a specific pattern, better enable it in the rule? By default, everything can be achieved by either the method or the inline style, so the default is fine anyway.

tests/lib/rules/v-on-handler-style.js
ota-meshi: We need to check if the expression other than the identifier call expression can be replaced with method + method handler.
For example, if you're using a variable defined with v-for, we can't replace it with a method handler.

```vue
<template v-for="e in list">
  <button @click="handler(e)" />
  <button @click="handlers[e]()" />
  <button @click="handler(a(b), c(d), e + f)" />
</template>
```

I thought it would be difficult to check it, but now that I've looked into it, it seems I can check it by checking the variable associated with the VElement.

I would change this PR to report inline handlers that can be replaced with method handlers.

ota-meshi: Hmm. Method style is insufficient as you say.
Do you think the user should be able to specify an option which styles to allow next in case the conversion is not possible for some reason if only method styles are allowed?

ota-meshi: It's still under discussion, but I've made the changes mentioned in https://github.com/vuejs/eslint-plugin-vue/pull/2009#discussion_r993185072.

ota-meshi: What do you think about adding options like:

```json
{
  "vue/v-on-handler-style": ["error",
    [
      "method",
      "inline-function"
    ],
    {
      "priority": false
    }
  ]
}
```

- priority: true

```vue
<template>
  <!-- ✓ GOOD -->
  <button v-on:click="handler" /> <!-- method handler -->
  <template v-for="e in list">
    <button @click="e" /> <!-- method handler -->
    <button @click="()=>foo(e)" /> <!-- inline function: cannot use method handler -->
  </template>

  <!-- ✗ BAD -->
  <button v-on:click="() => handler()" /> <!-- inline function: can use high priority method handler instead -->
  <template v-for="e in list">
    <button @click="()=>e()" /> <!-- inline function: can use high priority method handler instead -->
  </template>

  <button v-on:click="handler()" /> <!-- inline handler: can use method handler, or inline function instead -->
  <template v-for="e in list">
    <button @click="foo(e)" /> <!-- inline handler: cannot use method handler, but can use inline function instead -->
  </template>
</template>
```

- priority: false (default)

```vue
<template>
  <!-- ✓ GOOD -->
  <button v-on:click="handler" /> <!-- method handler -->
  <button v-on:click="() => handler()" /> <!-- inline function -->
  <template v-for="e in list">
    <button @click="e" /> <!-- method handler -->
    <button @click="()=>foo(e)" /> <!-- inline function -->
    <button @click="()=>e()" /> <!-- inline function -->
  </template>

  <!-- ✗ BAD -->
  <button v-on:click="handler()" /> <!-- inline handler: can use method handler, or inline function instead -->
  <template v-for="e in list">
    <button @click="foo(e)" /> <!-- inline handler: cannot use method handler, but can use inline function -->
  </template>
</template>
```

FloEdelmann: I like the idea of using the order of the array. Maybe we don't need that option and the "priority mode" is always active?

So if we want to use the method style where possible and the inline style otherwise and the inline function style only where there are multiple event payload parameters: `["method", "inline", "inline-function"]`.

If we want to use the inline function style only: `["inline-function"]`

If we want to use the inline style only (and also report cases where there are multiple event payload parameters): `["inline"]`
Multiple event payload parameters can be merged into one object, so these cases are possible to fix manually.

Specifying only `["method"]` should then be forbidden by the schema because it would otherwise report also cases that are impossible to fix.

ota-meshi: Thank you for your opinion! I will change this rule without new options.

ota-meshi: @FloEdelmann I changed the rule. Could you please check again?

ota-meshi: Sorry, please wait a moment because I thought it should be refactored more.

ota-meshi: @FloEdelmann I finished refactoring. Please check it again 🙏

docs/rules/v-on-handler-style.md
```diff
@@ -106,50 +149,52 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
...
   <!-- ✓ GOOD -->
-  <button v-on:click="() => handler()" />
+  <button v-on:click="handler" />
+  <button v-on:click="() => handler() /* comment */" />
 
   <!-- ✗ BAD -->
-  <button v-on:click="handler" />
   <button v-on:click="handler()" />
+  <button v-on:click="() => handler()" />
+  <button v-on:click="handler() /* comment */" />
```
FloEdelmann, line 178: I think with `ignoreIncludesComment: true`, this should be also allowed?

docs/rules/v-on-handler-style.md
```diff
@@ -106,50 +149,52 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
...
 <template>
   <!-- ✓ GOOD -->
   <button v-on:click="handler" />
   <button v-on:click="handler() /* comment */" />
-  <button v-on:click="() => handler() /* comment */" />
 
   <!-- ✗ BAD -->
   <button v-on:click="handler()" />
   <button v-on:click="() => handler()" />
+  <button v-on:click="() => handler() /* comment */" />
```
FloEdelmann, line 197: Again

docs/rules/v-on-handler-style.md
```diff
@@ -46,57 +46,100 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
...
+  <button v-on:click="count++" />
   <button v-on:click="() => handler()" />
+  <button v-on:click="() => count++" />
   <button v-on:click="(a, b) => handler(a, b)" />
   <template v-for="e in list">
-    <button @click="e()" />
-    <button @click="() => e()" />
+    <button v-on:click="e()" />
+    <button v-on:click="() => e()" />
+    <button v-on:click="handler(e)" /> <!-- Can use inline function -->
```
FloEdelmann, line 87: This comment is confusing to me; the other lines also have better alternatives but no comment. So I'd remove the comment completely.

docs/rules/v-on-handler-style.md
```diff
@@ -46,57 +46,100 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
...
+  <button v-on:click="count++" />
+  <button v-on:click="() => handler()" />
+  <button v-on:click="() => count++" />
+  <button v-on:click="(a, b) => handler(a, b)" />
   <template v-for="e in list">
-    <button @click="handler(e)" />
-    <button @click="() => handler(e)" />
+    <button v-on:click="e()" />
+    <button v-on:click="() => e()" />
+    <button v-on:click="() => handler(e)" /> <!-- Can use inline handler -->
```
FloEdelmann, line 116: Again.

lib/rules/v-on-handler-style.js
```diff
@@ -0,0 +1,582 @@
...
+      preferInlineOverMethod:
+        'Prefer inline handler over method handler in v-on.',
+      preferInlineOverInlineFunction:
+        'Prefer inline handler over inline function in v-on.',
+      preferInlineOverInlineFunctionWithMultipleParams:
+        'Prefer inline handler over inline function in v-on. Prepare for it, the custom event should be changed to a single payload.',
+      preferInlineFunctionOverMethod:
+        'Prefer inline function over method handler in v-on.',
+      preferInlineFunctionOverInline:
+        'Prefer inline function over inline handler in v-on.'
```
FloEdelmann, line 160:
```suggestion
      preferMethodOverInline:
        'Prefer method handler over inline handler in v-on.',
      preferMethodOverInlineWithoutIdCall:
        'Prefer method handler over inline handler in v-on. Note that you may need to create a new method.',
      preferMethodOverInlineFunction:
        'Prefer method handler over inline function in v-on.',
      preferMethodOverInlineFunctionWithoutIdCall:
        'Prefer method handler over inline function in v-on. Note that you may need to create a new method.',
      preferInlineOverMethod:
        'Prefer inline handler over method handler in v-on.',
      preferInlineOverInlineFunction:
        'Prefer inline handler over inline function in v-on.',
      preferInlineOverInlineFunctionWithMultipleParams:
        'Prefer inline handler over inline function in v-on. Note that the custom event must be changed to a single payload.',
      preferInlineFunctionOverMethod:
        'Prefer inline function over method handler in v-on.',
      preferInlineFunctionOverInline:
        'Prefer inline function over inline handler in v-on.'
```

lib/rules/v-on-handler-style.js
```diff
@@ -0,0 +1,582 @@
...
+      /**
+       * Range without quotes.
+       * @type {Range}
+       */
+      const range = hasQuote
+        ? [firstToken.range[1], lastToken.range[0]]
+        : [firstToken.range[0], lastToken.range[1]]
+
+      return {
+        innerRange: range,
```
FloEdelmann, line 251:
```suggestion
      /** @type {Range} */
      const rangeWithoutQuotes = hasQuote
        ? [firstToken.range[1], lastToken.range[0]]
        : [firstToken.range[0], lastToken.range[1]]

      return {
        innerRange: rangeWithoutQuotes,
```

FloEdelmann (changes requested):

docs/rules/v-on-handler-style.md
ota-meshi: The `ignoreIncludesComment` option is only ignored when converting to method handlers. Because the user can't convert it to a method handler without removing the comment.
Comments can be kept between inline handler <-> inline function.
I wrote an option description but I can't explain it well 😓

ota-meshi:
Thank you for your review.
I made a change to this PR other than ignoreIncludesComment.

docs/rules/v-on-handler-style.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+- First option ... Specifies the name of an allowed style. Default is `["method", "inline-function"]`.
+  - `["method", "inline-function"]` ... Allow handlers by method binding. e.g. `v-on:click="handler"`. Allow inline functions where method handlers cannot be used. e.g. `v-on:click="() => handler(listItem)"`.
+  - `["method", "inline"]` ... Allow handlers by method binding. e.g. `v-on:click="handler"`. Allow inline handlers where method handlers cannot be used. e.g. `v-on:click="handler(listItem)"`.
+  - `"inline-function"` ... Allow inline functions. e.g. `v-on:click="() => handler()"`
+  - `"inline"` ... Allow inline handlers. e.g. `v-on:click="handler()"`
+- Second option
+  - `ignoreIncludesComment` ... If `true`, do not report expressions containing comments to be converted to method handlers. However, it will be reported if the second style is not used. This option takes effect if `"method"` is allowed. Default is `false`.
```
FloEdelmann, line 63: What about this explanation?

```suggestion
  - `ignoreIncludesComment` ... If `true`, do not report inline handlers or inline functions containing comments, even if the preferred style is `"method"`. Default is `false`.
```

docs/rules/v-on-handler-style.md
FloEdelmann: Ah, I see. Can you check if my explanation suggestion below is correct?

ota-meshi: Thank you for your suggestion! I have accepted it into this PR!

FloEdelmann (approved):
:heavy_check_mark: LGTM

PR #2013, opened 2022-10-19
Allow restricting classes by regexp in `vue/no-restricted-class`

### LINKED ISSUE #1877: Allow regex in `vue/no-restricted-class`
[vue/no-restricted-class](https://eslint.vuejs.org/rules/no-restricted-class.html)

No

Allow a regex pattern just like we did [here](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-restricted-custom-event.js#L30)

```vue
<div class="ml-2" />
```

Right now to restrict the classes which are not RTL friendly I have to write long list of them:
```js
'vue/no-restricted-class': ['error', 'ml-1', 'ml-2', /* And more... */],
```

Once this rule accept regex pattern we don't have to write long list, just below is enough:
```js
'vue/no-restricted-class': ['error', '/^(m|p)(l|r)-/'],
```

**Additional context**
None

kindoflew:
Closes #1877

This PR adds support for RegEx to the `no-restricted-class` rule and an update to the example in the docs.

I didn't add all the `buildMatcher` logic like in the example linked in the Issue (`no-restricted-custom-event`: https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-restricted-custom-event.js) because this rule seemed a little more straightforward.

Please let me know if you need anything else. Thanks!

### DIFF
--- a/lib/rules/no-restricted-class.js
+++ b/lib/rules/no-restricted-class.js
@@ -12,18 +12,20 @@ const regexp = require('../utils/regexp')
  * @param {string} className
  * @param {*} node
  * @param {RuleContext} context
- * @param {Array<string>} forbiddenClasses
+ * @param {Set<string>} forbiddenClasses
+ * @param {Array<RegExp>} forbiddenClassesRegexps
  */
-const reportForbiddenClass = (className, node, context, forbiddenClasses) => {
-  const isForbidden = forbiddenClasses.some((cl) => {
-    if (regexp.isRegExp(cl)) {
-      const re = regexp.toRegExp(cl)
-      return re.test(className)
-    } else {
-      return className.includes(cl)
-    }
-  })
-  if (isForbidden) {
+const reportForbiddenClass = (
+  className,
+  node,
+  context,
+  forbiddenClasses,
+  forbiddenClassesRegexps
+) => {
+  if (
+    forbiddenClasses.has(className) ||
+    forbiddenClassesRegexps.some((re) => re.test(className))
+  ) {
     const loc = node.value ? node.value.loc : node.loc
     context.report({
       node,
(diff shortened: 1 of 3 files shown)


lib/rules/no-restricted-class.js
```diff
@@ -5,16 +5,25 @@
...
- * @param {Set<string>} forbiddenClasses
+ * @param {Array<string>} forbiddenClasses
  */
 const reportForbiddenClass = (className, node, context, forbiddenClasses) => {
-  if (forbiddenClasses.has(className)) {
+  const isForbidden = forbiddenClasses.some((cl) => {
+    if (regexp.isRegExp(cl)) {
+      const re = regexp.toRegExp(cl)
+      return re.test(className)
+    } else {
```
FloEdelmann, line 22: Doing this in every iteration seems unnecessarily expensive. It should be done once when parsing the options (in the `create` function).

Probably `forbiddenClasses` should stay a `Set`, so `forbiddenClasses.has(className)` still is a constant-time lookup, and then a new `forbiddenClassesRegexps` can be an array that you can check additionally.

tests/lib/rules/no-restricted-class.js
```diff
@@ -30,6 +30,10 @@ ruleTester.run('no-restricted-class', rule, {
     {
       code: `<template><div :class="'' + {forbidden: true}">Content</div></template>`,
       options: ['forbidden']
+    },
+    {
+      code: `<template><div class="allowed"">Content</div></template>`,
```
FloEdelmann, line 35:
```suggestion
      code: `<template><div class="allowed">Content</div></template>`,
```

FloEdelmann (changes requested):
Thanks for your contribution! Please see the comments below.

FloEdelmann (approved):
Looks good to me. Thank you!

PR #2018, opened 2022-10-24
Add new `vue/require-prop-comment` rule

### LINKED ISSUE #1993: require-prop-comment

czb3279338858:
Fixes #1993

### DIFF
--- a/lib/rules/no-required-prop-with-default.js
+++ b/lib/rules/no-required-prop-with-default.js
@@ -17,7 +17,7 @@ module.exports = {
     hasSuggestions: true,
     type: 'problem',
     docs: {
-      description: 'enforce props with default values ​​to be optional',
+      description: 'enforce props with default values to be optional',
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/no-required-prop-with-default.html'
     },


czb3279338858:
npm run lint,Changed the description of many files，I'm not sure this is correct, but I still submitted the code

czb3279338858:
1 failing：The file causing the error is not modified by me. Should I repair it before submitting

docs/rules/README.md
```diff
@@ -276,42 +277,42 @@ The following rules extend the rules provided by ESLint itself and apply them to
 
 | Rule ID | Description |    |    |
 |:--------|:------------|:--:|:--:|
-| [vue/array-bracket-newline](./array-bracket-newline.md) | enforce linebreaks after opening and before closing array brackets in `<template>` | :wrench: | :lipstick: |
```
FloEdelmann, line 279: `git pull origin master && npm run update` should probably fix these unrelated changes.

lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @param {import('../utils').ComponentProp[]} props
+     */
+    function verifyProps(props) {
+      for (const prop of props) {
+        if (!prop.propName) continue
+
+        const beforeComments = sourceCode.getCommentsBefore(prop.node)
+        let message = ''
```
FloEdelmann, line 91: Using an empty string is a bit unclean. Better explicitly allow `undefined` and change the functions to `return undefined` instead of `return ''`.

```suggestion
        /** @type {string|undefined} */
        let message
```

docs/rules/require-prop-comment.md
```diff
@@ -0,0 +1,180 @@
...
+
+This rule enforces that every prop has a comment that documents it.
+
+<eslint-code-block :rules="{'vue/require-prop-comment': ['error']}">
+
+```vue
+<!-- ✓ GOOD -->
+<template>
+  <div>1</div>
+</template>
```
FloEdelmann, line 23: Not needed for this example:

```suggestion
```

docs/rules/require-prop-comment.md
```diff
@@ -0,0 +1,180 @@
...
+
+</eslint-code-block>
+
+<eslint-code-block :rules="{'vue/require-prop-comment': ['error']}">
+
+```vue
+<!-- ✗ BAD -->
+<template>
+  <div>1</div>
+</template>
```
FloEdelmann, line 51: Again:

```suggestion
```

docs/rules/require-prop-comment.md
```diff
@@ -0,0 +1,180 @@
...
+const props = defineProps({
+  /*
+   * a comment
+   */
+  /*
+   * a other comment
+   */
+  a: Number,
+})
+</script>
```
FloEdelmann, line 100: Why is this bad? There is a block comment before the `a` prop.

tests/lib/rules/require-prop-comment.js
```diff
@@ -0,0 +1,263 @@
...
+  }
+})
+
+tester.run('require-prop-comment', rule, {
+  valid: [
+    {
+      code: `
+      <template>
+        <div>1</div>
+      </template>
```
FloEdelmann, line 24: Please remove the template from all tests, it is superfluous in this rule.

FloEdelmann (changes requested):

docs/rules/README.md
czb3279338858: complete

docs/rules/require-prop-comment.md
czb3279338858: This test is to check block，default is JSDoc

docs/rules/require-prop-comment.md
czb3279338858: block and JSDoc
1、Multiple lines are supported, and it is unnecessary to have multiple lines at the same time
2、The document generator may have difficulty recognizing which is what it needs

tests/lib/rules/require-prop-comment.js
czb3279338858: complete

czb3279338858:
Where there is dispute, should I initiate a new merger request? @FloEdelmann

FloEdelmann:
You don't need to open a new PR, you can instead push new commits to this branch. Also, rebasing and force-pushing is not required, we will squash the whole branch into one commit on the `master` branch when merging the PR.

docs/rules/require-prop-comment.md
czb3279338858: I misunderstood your meaning. The template has been deleted

docs/rules/require-prop-comment.md
FloEdelmann: I'd say this is an edge case that doesn't need to be handled in this rule. If the rule logic can be simplified by looking only if there is *a* comment for a prop (and not if there is *exactly one* comment), then we should rather do that.

docs/rules/require-prop-comment.md
czb3279338858: Completed

docs/rules/require-prop-comment.md
czb3279338858: complete

czb3279338858:
@FloEdelmann I push new commits.

PR #2020, opened 2022-10-27
feat: add `attribute-name-validator` rule

youxingz:
Add rule: `attribute-name-validator`

Since many company have some cases which are not allowed to use the specific attribute name, e.g. `id`, `xx-id`, and some typos like `jd`, `refs` (instead of `ref`). Then we added this rule, try to use regex pattern to test these attribute names of custom components.

Some useful regex options:
  - allow all letters and numbers: `/^([a-zA-Z_$][a-zA-Z\d_$]*)$/`
  - disable attribute `ref`: `/[^(ref)]/`
  - disable attribute `jd`: `/[^(jd)]/`

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -230,7 +230,7 @@ For example:
 | [vue/no-multiple-objects-in-class](./no-multiple-objects-in-class.md) | disallow to pass multiple objects into array to class |  | :hammer: |
 | [vue/no-potential-component-option-typo](./no-potential-component-option-typo.md) | disallow a potential typo in your component property | :bulb: | :hammer: |
 | [vue/no-ref-object-destructure](./no-ref-object-destructure.md) | disallow destructuring of ref objects that can lead to loss of reactivity |  | :warning: |
-| [vue/no-required-prop-with-default](./no-required-prop-with-default.md) | enforce props with default values ​​to be optional | :wrench::bulb: | :warning: |
+| [vue/no-required-prop-with-default](./no-required-prop-with-default.md) | enforce props with default values to be optional | :wrench::bulb: | :warning: |
 | [vue/no-restricted-block](./no-restricted-block.md) | disallow specific block |  | :hammer: |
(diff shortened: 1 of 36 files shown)


FloEdelmann:
We already have the [`vue/valid-attribute-name`](https://eslint.vuejs.org/rules/valid-attribute-name.html) and [`vue/no-restricted-static-attribute`](https://eslint.vuejs.org/rules/no-restricted-static-attribute.html) rules that do similar things. Is this not sufficient?

youxingz:
> We already have the [`vue/valid-attribute-name`](https://eslint.vuejs.org/rules/valid-attribute-name.html), [`vue/no-restricted-static-attribute`](https://eslint.vuejs.org/rules/no-restricted-static-attribute.html) and [`vue/no-restricted-v-bind`](https://eslint.vuejs.org/rules/no-restricted-v-bind.html) rules that do similar things. Is this not sufficient?

The rule `valid-attribute-name` does not support custom regex pattern, and we hope this rule can be more open, allowing users to set specific naming rules by themselves. Especially when some teams have their own naming rules.

FloEdelmann:
But the [`vue/no-restricted-static-attribute`](https://eslint.vuejs.org/rules/no-restricted-static-attribute.html) and [`vue/no-restricted-v-bind`](https://eslint.vuejs.org/rules/no-restricted-v-bind.html) rules do support custom regex patterns.

youxingz:
@FloEdelmann Thank you for your mention, I did't notice that rule can accpet regex parameters before. And I then checked the source code of `[vue/no-restricted-static-attribute](https://eslint.vuejs.org/rules/no-restricted-static-attribute.html)`, there is no option for user to decide which kind of component can be tested by rules. Can I create a new pull request and add such option to `vue/no-restricted-static-attribute`? It will look like:
```javascript
// scope: 'custom-component' | 'all'
options: [{ key: 'id', message: 'Using `id` is not allowd in custom component.', scope: 'custom-component' }]
```

FloEdelmann:
Yes, that option exists already. Use the `element` property inside the configuration object.

Feel free to open a PR to improve the docs if you found that lacking!

PR #2021, opened 2022-10-27
Add example for rule `vue/no-restricted-static-attribute`

youxingz:
From the conversation of https://github.com/vuejs/eslint-plugin-vue/pull/2020, after some unnecessary work and the valuable comments from @FloEdelmann, I then added a example to the document to make it clearer.

The example of the `element` property about the rule `vue/no-restricted-static-attribute`:
```javascript
options = [
    {
        "key": "style",
        "element": /^([A-Z$][a-zA-Z_$]*)$/,
        "message": "Using \"style\" is not allowed in custom component. Use \"class\" instead."
    }
]
```

### DIFF
--- a/docs/rules/no-restricted-static-attribute.md
+++ b/docs/rules/no-restricted-static-attribute.md
@@ -46,7 +46,7 @@ Alternatively, the rule also accepts objects.
     },
     {
       "key": "style",
-      "element": /^([A-Z$][a-zA-Z_$]*)$/,
+      "element": "/^([A-Z][a-zA-Z_-]*)$/",
       "message": "Using \"style\" is not allowed in custom component. Use \"class\" instead."
     }
   ]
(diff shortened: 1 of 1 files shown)


docs/rules/no-restricted-static-attribute.md
```diff
@@ -43,6 +43,11 @@ Alternatively, the rule also accepts objects.
     {
       "key": "stlye",
       "message": "Using \"stlye\" is not allowed. Use \"style\" instead."
+    },
+    {
+      "key": "style",
+      "element": /^([A-Z$][a-zA-Z_$]*)$/,
```
FloEdelmann, line 49: Shouldn't it be a string? Also, the `$` signs don't make sense here:

```suggestion
      "element": "/^([A-Z][a-zA-Z_-]*)$/",
```

FloEdelmann (changes requested):

docs/rules/no-restricted-static-attribute.md
youxingz: oh sorry, I made a new commit which revised this typo. Thank you for your review!

FloEdelmann (approved):

PR #2024, opened 2022-10-31
Ignore `<Textarea>` components in `vue/no-textarea-mustache`

### LINKED ISSUE #2023: Rule `no-textarea-mustache` cannot difference `<Textarea>` component
**Checklist**

- **ESLint version:** `^8.24.0`
- **eslint-plugin-vue version:** `^9.5.1`
- **Node version:** `16.17.1`
- **Operating System:** `Windows 10`

```json5
{
  extends: [
    'plugin:vue/vue3-recommended',
    // 'standard',
    // 'plugin:jsonc/recommended-with-jsonc',
    // 'plugin:yml/standard',
    // 'plugin:markdown/recommended',
    '@vue/typescript/recommended'
  ],
  env: {
    es6: true,
    browser: true,
    node: true
  },
  parser: 'vue-eslint-parser',
  parserOptions: {
    ecmaVersion: 'latest'
  },
  reportUnusedDisableDirectives: true
}
```

**What did you do?**	

```vue
<template>
  <Textarea value="Almost" :max-length="10">
    <template #count="{ value }">
      {{ `${value.length}/10` }}
    </template>
  </Textarea>
</template>
```

No error occurred.

**What actually happened?**

Get a error:

```log
error  Unexpected mustache. Use 'v-model' instead  vue/no-textarea-mustache
```

You can add a valid test in [here](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-textarea-mustache.js#L25):

```json5
{
  filename: 'test.vue',
  code: '<template><div><Textarea>{{text}}</Textarea></div></template>'
}
```

Then test it and it will fail.

qmhc:
Fix #2023

### DIFF
--- a/lib/rules/no-textarea-mustache.js
+++ b/lib/rules/no-textarea-mustache.js
@@ -27,6 +27,13 @@ module.exports = {
           return
         }
 
+        const sourceCode = context.getSourceCode()
+        const parentStartTag = sourceCode.getText(node.parent.startTag)
+
+        if (parentStartTag.startsWith('<T')) {
+          return
+        }
+
         context.report({
           node,
           loc: node.loc,
--- a/lib/rules/no-textarea-mustache.js
+++ b/lib/rules/no-textarea-mustache.js
@@ -20,18 +20,13 @@ module.exports = {
   },
   /** @param {RuleContext} context */
   create(context) {
-    const sourceCode = context.getSourceCode()
-
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VExpressionContainer} node */
       "VElement[name='textarea'] VExpressionContainer"(node) {
-        if (node.parent.type !== 'VElement') {
-          return
-        }
-
-        const parentStartTag = sourceCode.getText(node.parent.startTag)
-
-        if (parentStartTag.startsWith('<T')) {
+        if (
+          node.parent.type !== 'VElement' ||
+          node.parent.rawName.startsWith('T')
+        ) {
           return
         }
 
--- a/lib/rules/no-textarea-mustache.js
+++ b/lib/rules/no-textarea-mustache.js
@@ -22,11 +22,8 @@ module.exports = {
   create(context) {
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VExpressionContainer} node */
-      "VElement[name='textarea'] VExpressionContainer"(node) {
-        if (
-          node.parent.type !== 'VElement' ||
-          node.parent.rawName.startsWith('T')
-        ) {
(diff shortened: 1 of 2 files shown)


lib/rules/no-textarea-mustache.js
```diff
@@ -20,13 +20,21 @@ module.exports = {
...
+    const sourceCode = context.getSourceCode()
+
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VExpressionContainer} node */
       "VElement[name='textarea'] VExpressionContainer"(node) {
         if (node.parent.type !== 'VElement') {
           return
         }
 
+        const parentStartTag = sourceCode.getText(node.parent.startTag)
```
FloEdelmann, line 32: No need for the whole `sourceCode` here. You can use `node.parent.rawName` instead. See https://astexplorer.net/#/gist/d568be011d22bbc7c364e1eb4e2de668/b7f5155c779ee53528caa1373035b4d7436d4146

```suggestion
        const parentStartTag = node.parent.rawName
```

FloEdelmann (changes requested):

lib/rules/no-textarea-mustache.js
```diff
@@ -23,7 +23,10 @@ module.exports = {
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VExpressionContainer} node */
       "VElement[name='textarea'] VExpressionContainer"(node) {
-        if (node.parent.type !== 'VElement') {
+        if (
+          node.parent.type !== 'VElement' ||
+          node.parent.rawName.startsWith('T')
```
FloEdelmann, line 28:
```suggestion
          node.parent.rawName[0] === 'T'
```

FloEdelmann (approved):
Looks good to me. Thanks for reporting and fixing it yourself!

I have just one suggestion, but I don't know if it's easier to read this way.

lib/rules/no-textarea-mustache.js
qmhc: I think using a named method should be easier than using notations to read.

What's your starting point about this?

ota-meshi: How about rewriting the selector?

```diff
-      "VElement[name='textarea'] VExpressionContainer"(node) {
+      "VElement[rawName ='textarea'] VExpressionContainer"(node) {`
```

qmhc: LGTM~

FloEdelmann: Ah, even better!

PR #2019, opened 2022-10-25
Add new `vue/require-prop-comment` rule

czb3279338858:
Fixes #1993

PR #2026, opened 2022-11-02
Extract JSDoc comment recognition into util functions

FloEdelmann:
and add tests for them.

Follow-up to #2019. This was the commit that @czb3279338858 removed accidentally.


ota-meshi (approved):
LGTM! Thank you!

PR #2031, opened 2022-11-06
Ignore well known tags in `component-name-in-template-casing`

Kaciras:
Vue does not treat built-in HTML tags as components, so them should be always excluded from verification.

e.g.

```html
<template>
    <div/>
<template>
<script setup>
    import Div from "component"
</script>
```

will be compiled to

```javascript
import Div from "component"

const __sfc__ = {
  __name: 'App',
  setup(__props) {

return (_ctx, _cache) => {
  // `Div` is not used.
  return (_openBlock(), _createElementBlock("div"))
}
}
```


ota-meshi (approved):
LGTM!

FloEdelmann (approved):

Issue #2033, opened 2022-11-10
Rule suggestion: add fixer to require-name-property rule


0xYootou:
**What rule do you want to change?**

`require-name-property`

**Does this change cause the rule to produce more or fewer warnings?**

`no`

**How will the change be implemented? (New option, new default behavior, etc.)?**

`add fixer`

**Please provide some example code that this change will affect:**

if the component code is like this, and the component filename is  `./DemoComponent.vue`
```vue
export default defineComponent({
    components: { },
    setup() {
      return {
      };
    },
});
```

**What does the rule currently do for this code?**

add name property to component.

**What will the rule do after it's changed?**

```vue
export default defineComponent({
    name:'DemoComponent',
    components: { },
    setup() {
      return {
      };
    },
});
```

PR #2034, opened 2022-11-10
Add suggestion for `vue/require-name-property` rule

0xYootou:
this is a pull request for issues: #2033

### DIFF
--- a/lib/rules/require-name-property.js
+++ b/lib/rules/require-name-property.js
@@ -4,9 +4,31 @@
  */
 'use strict'
 
+const path = require('path')
 const utils = require('../utils')
 const { getVueComponentDefinitionType } = require('../utils')
+/**
+ * Get the text of the empty indentation part of the line which the given token is on.
+ * @param {SourceCode} sourceCode the source code object
+ * @param {Token} token the token to get the indentation text of the line which the token is on
+ * @returns {string} The text of indentation part.
+ */
+function getLineEmptyIndent(sourceCode, token) {
+  const LT_CHAR = /[\n\r\u2028\u2029]/
+  const EMPTY_CHAR = /\s/
+  const text = sourceCode.text
+  let i = token.range[0] - 1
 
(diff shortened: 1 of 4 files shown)


lib/rules/require-name-property.js
```diff
@@ -27,7 +28,8 @@ module.exports = {
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/require-name-property.html'
     },
-    fixable: null,
+    fixable: 'whitespace',
+    hasSuggestions: true,
```
FloEdelmann, line 32: You added a fixer, not suggestions.

lib/rules/require-name-property.js
```diff
@@ -47,7 +49,23 @@ module.exports = {
...
-        message: 'Required name property is not set.'
+        message: 'Required name property is not set.',
+        fix: (fixer) => {
+          const extension = path.extname(context.getFilename())
+          const filename = path.basename(context.getFilename(), extension)
+          // fix only when property is not empty
+          if (component.properties.length > 0) {
+            const startColumn = component.properties[0].loc.start.column
+            // insert name property before the first property
+            return fixer.insertTextBefore(
```
FloEdelmann, line 60: Can you please add test cases for the fixer?

FloEdelmann (changes requested):
Thanks for the suggestion and the implementation!

Can we always auto-fix this to the filename or should we rather only add a suggestion for that, which the user can manually apply?

Also, please see the issues below.

0xYootou:
ok, i think add a suggestion for that is better too, i will rewrite the code and add test case.

0xYootou:
i push new code for this feature, and add some new test case for this rule and suggestions.

FloEdelmann (changes requested):
Thanks for the changes!

ota-meshi (approved):
LGTM! Thank you!

PR #2040, opened 2022-11-18
fix(define-macros-order): update rule to support interface/type expor…

Jungzl:
I'm using this workaround https://github.com/johnsoncodehk/volar/issues/1232#issuecomment-1165416439 , but `define-macros-order` has conflict with Volar,


ota-meshi (approved):
LGTM! Thank you.

FloEdelmann (approved):

PR #2041, opened 2022-11-20
fix(docs): update example code for watch()

JeraldVin:
The 1st argument for watch() should be a ref/reactive/getter function/array of ref, reactive, getter function. But in the example code, the watch callback has been used as the 1st argument


ota-meshi (approved):
Thank you!

PR #2042, opened 2022-11-21
Improve indent rule to support more ts syntax

ota-meshi:
This PR adds support for more typescript syntax to indent rules.

```ts
// Type parameters given to call expression, and new expressions
foo<T>(arg)
// ^^^
new Foo<T>(arg)
//     ^^^

// Instantiation expression
const ErrorMap = Map<string, Error>
//               ^^^^^^^^^^^^^^^^^^

// Satisfies operator
var a = {} satisfies Foo
//         ^^^^^^^^^

// Accessor property
class Foo {
  accessor foo;
//^^^^^^^^^^^^
}
```

### DIFF
--- a/lib/utils/indent-ts.js
+++ b/lib/utils/indent-ts.js
@@ -29,6 +29,7 @@ const { isTypeNode } = require('./ts-ast-utils')
  * @typedef {import('@typescript-eslint/types').TSESTree.TSImportEqualsDeclaration} TSImportEqualsDeclaration
  * @typedef {import('@typescript-eslint/types').TSESTree.TSAbstractMethodDefinition} TSAbstractMethodDefinition
  * @typedef {import('@typescript-eslint/types').TSESTree.TSAbstractPropertyDefinition} TSAbstractPropertyDefinition
+ * @typedef {import('@typescript-eslint/types').TSESTree.TSAbstractAccessorProperty} TSAbstractAccessorProperty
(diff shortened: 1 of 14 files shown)


lib/utils/indent-ts.js
```diff
@@ -1302,6 +1325,42 @@ function defineVisitor({
...
+      const keyNode = node.key
+      const valueNode = node.value
+      const firstToken = tokenStore.getFirstToken(node)
+      const keyTokens = getFirstAndLastTokens(keyNode)
+      const prefixTokens = tokenStore.getTokensBetween(
+        firstToken,
+        keyTokens.firstToken
+      )
+      if (node.computed) {
+        prefixTokens.pop() // pop [
```
FloEdelmann, line 1338: Leftover comment?

```suggestion
        prefixTokens.pop()
```

FloEdelmann (approved):

lib/utils/indent-ts.js
ota-meshi: That pop, pops the `[` token, so it's not a leftover.

FloEdelmann: Ah, so maybe better

```suggestion
        prefixTokens.pop() // pop opening bracket character (`[`)
```

ota-meshi: Thanks!

PR #2045, opened 2022-11-23
Disable autofix in `vue/attribute-hyphenation` for attributes with underscore

### LINKED ISSUE #924: vue/attribute-hyphenation changes original code which would break app
* **ESLint version:** 5.16.0
* **eslint-plugin-vue version:** 5.2.3
* **Node version:**	10.14.2

```js
// https://eslint.org/docs/user-guide/configuring

module.exports = {
  root: true,
  parserOptions: {
    parser: 'babel-eslint'
  },
  env: {
    browser: true
  },
  extends: ['plugin:vue/essential'],
  plugins: ['vue'],
  // add your custom rules here
  rules: {
    // 'vue/prop-name-casing': ['error', 'camelCase']
    'vue/attribute-hyphenation': 'error'
  }
}
```

**What did you do?**	

```vue
<!-- HelloWorld/index.vue-->
<template>
  <div class="HelloWorld">
    <TestPara
      :AttrEe="4"
      :attr_ff="5"
      :attr_Gg="6"
      :Attr_Hh="7"
      :_attr_ii="8"
      :_attr_Jj="9"
      :_attrKk="10"
      :_AttrLl="11"
    />
  </div>
</template>

<script>
import TestPara from './TestPara'
export default {
  name: 'HelloWorld',
  components: {
    TestPara
  }
}
</script>
```
```vue
<template>
  <div class="TestPara"></div>
</template>

<script>
<!-- HelloWorld/TestPara.vue-->
export default {
  name: 'TestPara',
  props: [
    'AttrEe',
    'attr_ff',
    'attr_Gg',
    'Attr_Hh',
    '_attr_ii',
    '_attr_Jj',
    '_attrKk',
    '_AttrLl'
  ],
  created () {
    window.vm = this
    console.log(vm.$props)
  }
}
</script>
```
run 

```bash
cd .\vue\vue-prop-parse\
npx eslint .\src\components\HelloWorld\index.vue --fix
```

I can still access all the `$props` in the `TestPara` with the **valid value** without changing the code in the `TestPara`.

**What actually happened?**

```vue
<!-- HelloWorld/index.vue-->
<template>
  <div class="HelloWorld">
    <TestPara
      :attr-ee="4"
      :attr_ff="5"
      :attr-gg="6"
      :attr-hh="7"
      :_attr_ii="8"
      :-attr-jj="9"
      :-attr-kk="10"
      :-attr-ll="11"
    />
  </div>
</template>

<script>
import TestPara from './TestPara'
export default {
  name: 'HelloWorld',
  components: {
    TestPara
  }
}
</script>

```
After lint, I got the error at the console with 

> '-attr-jj' is not a valid attribute name.

And `Attr_Hh`, `attr_Gg`, `_AttrLl`, `_attrKk`, `_attr_Jj` in $props are `undefined`. I check the [test cases](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/attribute-hyphenation.js) and find you didn't consider these situations like `_` in the attributes name.

In this case, I thought you shouldn't try to fix the original code? Because that would break the app down. 

What do you think?

kurageso:
fixes #924 

I have modified it in reference to https://github.com/vuejs/eslint-plugin-vue/issues/924#issuecomment-1263187134

Please let me know if you need anything else. Thanks!


tests/lib/rules/attribute-hyphenation.js
```diff
@@ -58,6 +58,16 @@ ruleTester.run('attribute-hyphenation', rule, {
...
+    {
+      filename: 'test.vue',
+      code: '<template><div><custom :attr_ff="prop"></custom></div></template>',
+      options: ['always']
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div><custom :_attr_ii="prop"></custom></div></template>',
+      options: ['always']
     }
```
FloEdelmann, line 71: What is the difference between these test cases? I think one of them can be dropped.

tests/lib/rules/attribute-hyphenation.js
```diff
@@ -280,6 +290,90 @@ ruleTester.run('attribute-hyphenation', rule, {
           line: 1
         }
       ]
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div><custom :AttrEe="prop"></custom></div></template>',
+      output:
+        '<template><div><custom :attr-ee="prop"></custom></div></template>',
```
FloEdelmann, line 298: When the output doesn't change, I think you can use `output: null`.

lib/rules/attribute-hyphenation.js
```diff
@@ -94,8 +94,16 @@ module.exports = {
         data: {
           text
         },
-        fix: (fixer) =>
-          fixer.replaceText(node.key, text.replace(name, caseConverter(name)))
+        fix: (fixer) => {
+          if (useHyphenated && text.includes('_')) {
```
FloEdelmann, line 98: Probably underscores are also problematic when changing to camelCase, right?

```suggestion
          if (text.includes('_')) {
```

Can you also add test cases for this?

FloEdelmann (approved):
Thanks for the fix and the quick implementation of the feedback!

ota-meshi (approved):
Thank you!

PR #2046, opened 2022-11-23
Chore: fix playground in rule docs

ota-meshi:
This PR is a follow up to #2044.
eslint in the browser wasn't working (I've confirmed it works, but I may have missed something 😓 ).
This PR fixes that issue.


ota-meshi:
It seems to work so I will merge this and check.
https://deploy-preview-2046--eslint-plugin-vue.netlify.app/rules/valid-define-emits.html

FloEdelmann:
Hmm I also checked it in the deployment, as I suspected that it could break. Good that you found the culprit so quickly!

PR #2047, opened 2022-11-23
Chore: add script to site that remove service-worker

ota-meshi:
I've noticed that the latest docs site doesn't display from some devices.
This is because previously we were using `@vuepress/plugin-pwa` and the cache is displayed by the service worker.
https://vuepress.vuejs.org/plugin/official/plugin-pwa.html

This PR adds a script to remove service workers.
See also below for how to remove.
https://github.com/NekR/self-destroying-sw


ota-meshi:
I will merge and check it.

ota-meshi:
For me it seems to have worked.

PR #2052, opened 2022-12-01
Add `padding-lines-in-component-definition` rule

### LINKED ISSUE #1817: Rule proposal: `padding-line-between-component-options`
Enforces empty lines between top-level options of Vue components.

- [X] Enforces code style (layout)
- [ ] Warns about a potential error (problem)
- [ ] Other (please specify:)

```js
//✓ GOOD
export default {
  name: 'a-button',

  data() {
    return {
      // ...
    }
  },

  computed: {
   // …
  }
}
```
```js
// ✗ BAD
export default {
  name: 'a-button',
  data() {
    return {
      // ...
    }
  },
  computed: {
   // …
  }
}
```

**Additional context**

* Consistent format makes it easier to read, find, select, and extend.
* The rule can enforce *no empty lines* between options (swap GOOD and BAD in the example).
* I'm not sure about the term "options" but decided so because of the "Options API" name.
* [StackOverflow Question requesting this (5 Upvotes)](https://stackoverflow.com/questions/59632023/eslint-rule-to-enforce-empty-lines-in-between-vue-script-keys?noredirect=1#comment124867146_59632023)
* [Github Comment requesting this (6 upvotes)](https://github.com/vuejs/eslint-plugin-vue/issues/855#issuecomment-579371070)

ItMaga:
This PR adds `padding-lines-in-component-definition` rule that requires or disallows padding lines in component definition.

close #1817

### DIFF
--- a/lib/rules/padding-lines-in-component-definition.js
+++ b/lib/rules/padding-lines-in-component-definition.js
@@ -202,7 +202,7 @@ module.exports = {
      * @param {any} option
      * @param {any} nextOption
      * */
-    function check(properties, option, nextOption) {
+    function verify(properties, option, nextOption) {
       const groupSingleLineProperties = parseBooleanOption(
         options,
         OptionKeys.GroupSingleLineProperties
@@ -287,7 +287,7 @@ module.exports = {
         const propertyOption = parseOption(nextOption, name)
         if (!propertyOption) continue
 
-        check(
+        verify(
           nestedProperties,
           parseOption(propertyOption, OptionKeys.BetweenItems),
           parseOption(propertyOption, OptionKeys.WithinEach)
@@ -302,7 +302,7 @@ module.exports = {
             node.properties.filter((property) => property.type === 'Property')
           )
 
-          check(
+          verify(
             propertiesFiltered,
             parseOption(options, OptionKeys.BetweenOptions),
             parseOption(options, OptionKeys.WithinOption)
@@ -321,7 +321,7 @@ module.exports = {
           const propsOption = parseOption(withinOption, 'props')
           if (!propsOption) return
 
-          check(
+          verify(
             propNodes,
             parseOption(propsOption, OptionKeys.BetweenItems),
             parseOption(propsOption, OptionKeys.WithinEach)
@@ -338,7 +338,7 @@ module.exports = {
           const emitsOption = parseOption(withinOption, 'emits')
           if (!emitsOption) return
 
-          check(
+          verify(
             emitNodes,
             parseOption(emitsOption, OptionKeys.BetweenItems),
             parseOption(emitsOption, OptionKeys.WithinEach)
--- a/lib/rules/padding-lines-in-component-definition.js
+++ b/lib/rules/padding-lines-in-component-definition.js
@@ -1,0 +1,350 @@
+/**
+ * @author ItMaga <https://github.com/ItMaga>
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+/**
+ * @typedef {import('../utils').ComponentProp} ComponentProp
+ * @typedef {import('../utils').GroupName} GroupName
+ */
+
+const utils = require('../utils')
+
+const AvailablePaddingOptions = {
+  Never: 'never',
+  Always: 'always',
+  Ignore: 'ignore'
+}
+const OptionKeys = {
+  BetweenOptions: 'betweenOptions',
+  WithinOption: 'withinOption',
+  BetweenItems: 'betweenItems',
+  WithinEach: 'withinEach',
+  GroupSingleLineProperties: 'groupSingleLineProperties'
+}
+
+/**
+ * @param {Token} node
+ */
+function isComma(node) {
+  return node.type === 'Punctuator' && node.value === ','
+}
+
+/**
+ * Split the source code into multiple lines based on the line delimiters.
+ * @param {string} text Source code as a string.
+ * @returns {string[]} Array of source code lines.
+ */
+function splitLines(text) {
+  return text.split(/\r\n|[\r\n\u2028\u2029]/gu)
+}
+
+/**
+ * @param {any} initialOption
+ * @param {string} optionKey
+ * @private/
+ * */
+function parseOption(initialOption, optionKey) {
+  return typeof initialOption === 'string'
+    ? initialOption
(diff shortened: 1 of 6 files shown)


ota-meshi:
Thank you for this PR! I would love to check this PR, but right now I am busy and don't have time to do it. So please wait 🙏

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,350 @@
...
+                return insertLines(
+                  currentProperty,
+                  nextProperty,
+                  fixer,
+                  betweenLinesRange
+                )
+              }
+            })
+          } else if (
+            betweenLinesRange > 1 &&
```
ota-meshi, line 256: I think we should support cases where comments and non property node are included. Same for `groupSingleLineProperties`.

In the current implementation, using 'never' can cause comments and nodes to be lost due to auto-fix.

```vue
<script >
/* eslint vue/padding-lines-in-component-definition: ["error", "never"] */
export default {
  props: {
    a:{},
    /** lost */
    b:{}
  },
}
</script>
```

```vue
<script>
/* eslint vue/padding-lines-in-component-definition: ["error", "never"] */
export default {
  a: {},
  ...lost,
  b: {},
}
</script>
```

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,350 @@
...
+      utils.defineScriptSetupVisitor(context, {
+        onDefinePropsEnter(_, props) {
+          const propNodes = /** @type {Property[]} */ (
+            props
+              .filter((prop) => prop.node && prop.node.type === 'Property')
+              .map((prop) => prop.node)
+          )
+
+          const withinOption = parseOption(options, OptionKeys.WithinOption)
+          const propsOption = parseOption(withinOption, 'props')
```
ota-meshi, line 321: `withinOption` may be undef. So we have to guard here.

```suggestion
          const propsOption = withinOption && parseOption(withinOption, 'props')
```

Same goes for onDefineEmitsEnter.

Test case:

```vue
<script setup>
/* eslint vue/padding-lines-in-component-definition: ["error", {}] */
  defineProps({
    a: {
      type: String,
    },
    b: {
      type: String,
    },
  })
</script>

```

ota-meshi (changes requested):
Thank you for this PR. Sorry for the delay in checking.
It looks pretty good to me, but I have a few requests.

lib/rules/padding-lines-in-component-definition.js
ItMaga: @ota-meshi Thank you for review! I have a question about comments. If we have comments, should we implement them as part of property or as own property? As an example, what is the correct variant?

```vue
<script>
/* eslint vue/padding-lines-in-component-definition: ["error", "always"] */
export default {
  props: {
    a:{},
    
    /** lost */
    
    b:{}
  },
}
</script>
```
or
```vue
<script>
/* eslint vue/padding-lines-in-component-definition: ["error", "always"] */
export default {
  props: {
    a:{},
    
    /** lost */
    b:{}
  },
}
</script>
```

Also, will the behavior differ if we use JSDoc or multiline comments?

ota-meshi: I think it should be styled like this: 

```js
export default {
  props: {
    a:{},
    
    /** lost */
    b:{}
  },
}
```

I think any form of comment should behave the same. It can also be a suppressing comment.

```js
export default {
  props: {
    a:{},
    
    // eslint-disable-next-line rule
    b:{}
  },
}
```

ItMaga: Thanks for quick response! 🙏
I added comments and non-property node support

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,357 @@
...
+ */
+'use strict'
+
+/**
+ * @typedef {import('../utils').ComponentProp} ComponentProp
+ * @typedef {import('../utils').GroupName} GroupName
+ */
+
+const utils = require('../utils')
+const { isComment } = require('../utils/indent-utils')
```
ota-meshi, line 13: Please use eslint-utils instead.

```suggestion
const { isCommentToken } = require('eslint-utils')
```

ota-meshi (changes requested):
Thank you for changing this PR! I have one request.

ota-meshi (approved):
LGTM! Thank you!

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,357 @@
...
+ * @returns {string[]} Array of source code lines.
+ */
+function splitLines(text) {
+  return text.split(/\r\n|[\r\n\u2028\u2029]/gu)
+}
+
+/**
+ * @param {any} initialOption
+ * @param {string} optionKey
+ * @private/
```
FloEdelmann, line 54:
```suggestion
 * @private
```

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,357 @@
...
+ * */
+function parseOption(initialOption, optionKey) {
+  return typeof initialOption === 'string'
+    ? initialOption
+    : initialOption[optionKey]
+}
+/**
+ * @param {any} initialOption
+ * @param {string} optionKey
+ * @private/
```
FloEdelmann, line 64:
```suggestion
 * @private
```

lib/rules/padding-lines-in-component-definition.js
```diff
@@ -0,0 +1,357 @@
...
+  }
+  return initialOption[optionKey]
+}
+
+/**
+ * @param {(Property | SpreadElement)} currentProperty
+ * @param {(Property | SpreadElement)} nextProperty
+ * @param {boolean} option
+ * @returns {boolean}
+ * @private/
```
FloEdelmann, line 79:
```suggestion
 * @private
```

FloEdelmann (changes requested):
Looks really good! I just have some very minor improvement requests, then this is good to go!

ItMaga:
@FloEdelmann Thank you for feedback! All conversations were resolved

FloEdelmann (approved):

PR #2066, opened 2022-12-22
Added array-element-newline rule from eslint

alshyra:
Hello,
I'm currently using `eslint-plugin-vue` and it works great !
Even though it's not advised we depends exclusively on eslint for code style and format.

There is one rule in template I'd like to see: https://eslint.org/docs/latest/rules/array-element-newline that why I propose to you this pull request.

I've read the https://eslint.vuejs.org/developer-guide/ but this is my first pull request so let me know if I miss something.


tests/lib/rules/array-element-newline.js
```diff
@@ -0,0 +1,110 @@
...
+    ecmaVersion: 2020,
+    sourceType: 'module'
+  }
+})
+
+tester.run('array-element-newline', rule, {
+  valid: [
+    '<template><div :attr="[]" /></template>',
+    '<template><div :attr="[a]" /></template>',
+    '<template><div :attr="[a,\nb,\nc]" /></template>',
```
FloEdelmann, line 22: Could you please use template strings (`` `foo` ``) instead of single quotes, and replace `\n` with real newlines? That would be more readable. Indentation shouldn't matter for this rule.

tests/lib/rules/array-element-newline.js
```diff
@@ -0,0 +1,110 @@
...
+    {
+      code: '<template><div :attr="[a,b]" /></template>',
+      options: [{ minItems: 3 }]
+    }
+  ],
+  invalid: [
+    {
+      code: '<template><div :attr="[a, b]" /></template>',
+      output: '<template><div :attr="[a,\nb]" /></template>',
+      errors: ['There should be a linebreak after this element.']
```
FloEdelmann, line 65: Please specify errors as objects with message, line and column numbers:

```diff
-     errors: ['There should be a linebreak after this element.']
+     errors: [{
+       message: 'There should be a linebreak after this element.',
+       line: 1,
+       column: 25,
+       lineEnd: 2,
+       columnEnd: 1
+     }]
```

(The line and column numbers in my example might be off.)

FloEdelmann (changes requested):
Thanks, looks good! I only got small remarks regarding the test cases.

tests/lib/rules/array-element-newline.js
alshyra: Is it better now ?

tests/lib/rules/array-element-newline.js
alshyra: Done

FloEdelmann (approved):
Looks fine now :)

Thanks!

ota-meshi (approved):
Thank you!

Issue #1728, opened 2021-11-26
`vue/attributes-order` allow to differentiate between static and dynamic attributes


drik98:
**What rule do you want to change?**

`vue/attributes-order`

**Does this change cause the rule to produce more or fewer warnings?**

More depending on the config. If the config does not change it should be equal.

**How will the change be implemented? (New option, new default behavior, etc.)?**

New option. 
Currently only `OTHER_ATTR` can be used for attributes, no matter if they are static or dynamic (using v-bind). I'd like to differentiate between those two types. We like to have our static attributes first and our dynamic attributes second. I'd imagine you could give the option to use something like `OTHER_ATTR_STATIC` and `OTHER_ATTR_DYNAMIC` where as `OTHER_ATTR` would become an alias for `[ 'OTHER_ATTR_STATIC', 'OTHER_ATTR_DYNAMIC' ]` so current configs do not break.

**Please provide some example code that this change will affect:**

```vue
  <div
    prop-one="prop"
    :prop-two="prop"
    prop-three="prop">
```

**What does the rule currently do for this code?**

```vue
  <div
    prop-one="prop"
    :prop-two="prop"
    prop-three="prop">
```

**What will the rule do after it's changed?**

using the config  

```json
{
  "vue/attributes-order": ["error", {
    "order": [
      "DEFINITION",
      "LIST_RENDERING",
      "CONDITIONALS",
      "RENDER_MODIFIERS",
      "GLOBAL",
      ["UNIQUE", "SLOT"],
      "TWO_WAY_BINDING",
      "OTHER_DIRECTIVES",
      "OTHER_ATTR_STATIC",
      "OTHER_ATTR_DYNAMIC",
      "EVENTS",
      "CONTENT"
    ],
    "alphabetical": false
  }]
}
```

it should change to

```vue
  <div
    prop-one="prop"
    prop-three="prop"
    :prop-two="prop">
```

CamilleDrapier:
Just an idea in the same vein; In case this is considered, maybe it could be nice to further allow differentiation for "boolean/true attributes" as well (`OTHER_ATTR_TRUE`), such as the following:

```html
 <div
    prop-a="prop"
    :prop-b="prop"
    prop-c="prop"
    prop-d
    :prop-e="prop"
>
```

would become (with alphabetical sorting enabled):

```
 <div
    prop-d
    prop-a="prop"
    prop-c="prop"
    :prop-b="prop"
    :prop-e="prop"
>
```

[...]

PR #2068, opened 2022-12-31
Add new attributes order for dynamic and static props

### LINKED ISSUE #1870: Prefer boolean prop usage at first
This rule is related to this [rule](https://eslint.vuejs.org/rules/prefer-prop-type-boolean-first.html#vue-prefer-prop-type-boolean-first). In the existing rule we are enforcing defining the boolean rules first but in this rule, we can enforce using boolean props first.

[ ] Other (please specify:)

```vue
   <!-- GOOD -->
    <v-btn
      flat
      color="success"
    >
      Success
    </v-btn>

   <!-- BAD -->
    <v-btn
      color="info"
      flat
    >
      Info
    </v-btn>

   <!-- GOOD -->
    <v-btn
      flat
      variant="outilned"
      color="success"
      rounded="pill"
    >
      Success
    </v-btn>
```

**Additional context**
Placing boolean prop usage on top makes it easy to read the component props. 👀

GreatGui:
This PR adds `[ 'ATTR_STATIC', 'ATTR_DYNAMIC', 'ATTR_SHORTHAND_BOOL' ] ` attribute to `vue/attributes-order`. And `OTHER_ATTR` become an alias for [ 'ATTR_STATIC', 'ATTR_DYNAMIC', 'ATTR_SHORTHAND_BOOL' ]

close https://github.com/vuejs/eslint-plugin-vue/issues/1728, close #1870

### DIFF
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -35,11 +35,13 @@ This rule aims to enforce ordering of component attributes. The default order is
 - `OTHER_DIRECTIVES`
   e.g. 'v-custom-directive'
 - `OTHER_ATTR`
-  alias for `[ATTR_DYNAMIC, ATTR_STATIC]`:
+  alias for `[ATTR_DYNAMIC, ATTR_STATIC, ATTR_SHORTHAND_BOOL]`:
   - `ATTR_DYNAMIC`
     e.g. 'v-bind:prop="foo"', ':prop="foo"'
   - `ATTR_STATIC`
     e.g. 'prop="foo"', 'custom-prop="foo"'
+  - `ATTR_SHORTHAND_BOOL`
+    e.g. 'boolean-prop'
 - `EVENTS`
   e.g. '@click="functionCall"', 'v-on="event"'
 - `CONTENT`
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -35,11 +35,11 @@ This rule aims to enforce ordering of component attributes. The default order is
 - `OTHER_DIRECTIVES`
   e.g. 'v-custom-directive'
 - `OTHER_ATTR`
-  alias for '[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]'
-- `OTHER_ATTR_DYNAMIC`
-  e.g. 'v-bind:prop="foo"', ':prop="foo"'
-- `OTHER_ATTR_STATIC`
-  e.g. 'prop="foo"', 'custom-prop="foo"'
+  alias for `[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]`:
+  - `OTHER_ATTR_DYNAMIC`
+    e.g. 'v-bind:prop="foo"', ':prop="foo"'
+  - `OTHER_ATTR_STATIC`
+    e.g. 'prop="foo"', 'custom-prop="foo"'
 - `EVENTS`
   e.g. '@click="functionCall"', 'v-on="event"'
 - `CONTENT`
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -35,10 +35,10 @@ This rule aims to enforce ordering of component attributes. The default order is
 - `OTHER_DIRECTIVES`
   e.g. 'v-custom-directive'
 - `OTHER_ATTR`
-  alias for `[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]`:
-  - `OTHER_ATTR_DYNAMIC`
+  alias for `[ATTR_DYNAMIC, ATTR_STATIC]`:
+  - `ATTR_DYNAMIC`
     e.g. 'v-bind:prop="foo"', ':prop="foo"'
-  - `OTHER_ATTR_STATIC`
+  - `ATTR_STATIC`
     e.g. 'prop="foo"', 'custom-prop="foo"'
 - `EVENTS`
   e.g. '@click="functionCall"', 'v-on="event"'
--- a/docs/rules/attributes-order.md
+++ b/docs/rules/attributes-order.md
@@ -35,7 +35,11 @@ This rule aims to enforce ordering of component attributes. The default order is
 - `OTHER_DIRECTIVES`
   e.g. 'v-custom-directive'
 - `OTHER_ATTR`
-  e.g. 'custom-prop="foo"', 'v-bind:prop="foo"', ':prop="foo"'
+  alias for '[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]'
+- `OTHER_ATTR_DYNAMIC`
+  e.g. 'v-bind:prop="foo"', ':prop="foo"'
+- `OTHER_ATTR_STATIC`
+  e.g. 'prop="foo"', 'custom-prop="foo"'
 - `EVENTS`
   e.g. '@click="functionCall"', 'v-on="event"'
 - `CONTENT`
--- a/lib/rules/attributes-order.js
+++ b/lib/rules/attributes-order.js
@@ -70,7 +70,7 @@ function isVBindObject(node) {
 }
 
 /**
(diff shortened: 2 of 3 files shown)


lib/rules/attributes-order.js
```diff
@@ -200,17 +206,27 @@ function create(context) {
...
   )
 
+  for (const [index, item] of attributeOrder.entries()) {
+    if (Array.isArray(item) && item.includes(ATTRS.OTHER_ATTR)) {
+      const attributes = item.filter((i) => i !== ATTRS.OTHER_ATTR)
+      attributes.push(...otherAttrs)
+      attributeOrder[index] = attributes
+    } else if (item === ATTRS.OTHER_ATTR) {
+      attributeOrder[index] = otherAttrs
+    }
```
FloEdelmann, line 227: Maybe that's more readable?

```suggestion
    if (item === ATTRS.OTHER_ATTR) {
      attributeOrder[index] = otherAttrs
    } else if (Array.isArray(item)) {
      const itemIndex = item.indexOf(ATTRS.OTHER_ATTR)
      if (itemIndex > -1) {
        attributeOrder[index].splice(itemIndex, 1, ...otherAttrs)
      }
    }
```

lib/rules/attributes-order.js
```diff
@@ -200,17 +206,27 @@ function create(context) {
...
   ]
   if (context.options[0] && context.options[0].order) {
-    attributeOrder = context.options[0].order
+    attributeOrder = [...context.options[0].order]
   }
   const alphabetical = Boolean(
     context.options[0] && context.options[0].alphabetical
   )
 
+  for (const [index, item] of attributeOrder.entries()) {
```
FloEdelmann, line 220: Please add an explanatory comment here:

```suggestion

  // expand `OTHER_ATTR` alias 
  for (const [index, item] of attributeOrder.entries()) {
```

docs/rules/attributes-order.md
```diff
@@ -35,7 +35,11 @@ This rule aims to enforce ordering of component attributes. The default order is
 - `OTHER_DIRECTIVES`
   e.g. 'v-custom-directive'
 - `OTHER_ATTR`
-  e.g. 'custom-prop="foo"', 'v-bind:prop="foo"', ':prop="foo"'
+  alias for '[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]'
+- `OTHER_ATTR_DYNAMIC`
+  e.g. 'v-bind:prop="foo"', ':prop="foo"'
+- `OTHER_ATTR_STATIC`
+  e.g. 'prop="foo"', 'custom-prop="foo"'
```
FloEdelmann, line 42: Note that in this list, the default order is documented. So it doesn't make sense to list `OTHER_ATTR`, `OTHER_ATTR_DYNAMIC` and `OTHER_ATTR_STATIC` in the same level. I'd change it like this:

```suggestion
  alias for `[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC]`:
  - `OTHER_ATTR_DYNAMIC`
    e.g. 'v-bind:prop="foo"', ':prop="foo"'
  - `OTHER_ATTR_STATIC`
    e.g. 'prop="foo"', 'custom-prop="foo"'
```

lib/rules/attributes-order.js
```diff
@@ -20,6 +20,8 @@ const ATTRS = {
   TWO_WAY_BINDING: 'TWO_WAY_BINDING',
   OTHER_DIRECTIVES: 'OTHER_DIRECTIVES',
   OTHER_ATTR: 'OTHER_ATTR',
+  OTHER_ATTR_STATIC: 'OTHER_ATTR_STATIC',
+  OTHER_ATTR_DYNAMIC: 'OTHER_ATTR_DYNAMIC',
```
FloEdelmann, line 24: I'm wondering if the new options need the `OTHER_` prefix, or if we can rename them to `ATTR_DYNAMIC` and `ATTR_STATIC`. What's your opinion?

lib/rules/attributes-order.js
```diff
@@ -200,17 +206,27 @@ function create(context) {
...
     ATTRS.TWO_WAY_BINDING,
     ATTRS.OTHER_DIRECTIVES,
-    ATTRS.OTHER_ATTR,
+    [ATTRS.OTHER_ATTR_DYNAMIC, ATTRS.OTHER_ATTR_STATIC],
     ATTRS.EVENTS,
     ATTRS.CONTENT
   ]
   if (context.options[0] && context.options[0].order) {
-    attributeOrder = context.options[0].order
+    attributeOrder = [...context.options[0].order]
```
FloEdelmann, line 214: Do you think it makes sense to add a validation check here, so that `OTHER_ATTR` should not be specified together with either `OTHER_ATTR_DYNAMIC` or `OTHER_ATTR_STATIC`? Or actually, that no category should appear twice in the options?

FloEdelmann (changes requested):
Very nice, thank you! I don't know why the Netlify deployment is failing, but it doesn't seem to be related to your changes. @ota-meshi, could you please have a look at that?

@GreatGui, I commented a few smaller issues. Also, have you seen #1870? It would be great to implement both options in one pull request to avoid a breaking change.

So, would you fancy also implementing the following to fix #1870?

1. `OTHER_ATTR_STATIC` should only match non-shorthand attributes (e.g. `foo="bar"` but not `foo` on its own).
2. Introduce new option `OTHER_ATTR_SHORTHAND_BOOL` that matches shorthand boolean attributes.
3. `OTHER_ATTR` should be an alias for `[OTHER_ATTR_DYNAMIC, OTHER_ATTR_STATIC, OTHER_ATTR_SHORTHAND_BOOL]`.

lib/rules/attributes-order.js
```diff
@@ -153,7 +155,10 @@ function getAttributeType(attribute) {
     case 'slot-scope':
       return ATTRS.SLOT
     default:
-      return ATTRS.OTHER_ATTR
+      if (attribute.directive) {
```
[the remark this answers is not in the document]
GreatGui: Make sense

lib/rules/attributes-order.js
```diff
@@ -200,17 +206,27 @@ function create(context) {
     [ATTRS.UNIQUE, ATTRS.SLOT],
     ATTRS.TWO_WAY_BINDING,
     ATTRS.OTHER_DIRECTIVES,
-    ATTRS.OTHER_ATTR,
+    [ATTRS.OTHER_ATTR_DYNAMIC, ATTRS.OTHER_ATTR_STATIC],
```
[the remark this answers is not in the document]
GreatGui: Adjusted

lib/rules/attributes-order.js
GreatGui: I tried to do it this way, but it doesn't work.
Whenever I try to change some item of attributeOrder, I get an error

lib/rules/attributes-order.js
GreatGui: Added

docs/rules/attributes-order.md
GreatGui: Fixed

lib/rules/attributes-order.js
GreatGui: I removed all `OTHER_` prefix, it really got better

lib/rules/attributes-order.js
GreatGui: Fixed, it really got better

GreatGui:
@FloEdelmann I corrected everything I could and added the `ATTR_SHORTHAND_BOOL`

FloEdelmann:
It would be good to get this into the next version, could you please check the open threads? If you don't have time, I'll do the rest.

GreatGui:
@FloEdelmann fixed the points and added the `OTHER_ATTR ` check

ota-meshi (approved):
Thank you for this PR! LGTM!

FloEdelmann (approved):

Issue #2048, opened 2022-11-24
[`vue/component-name-in-template-casing`] False positive for dynamic component in template


Ky6uk:
I've found a false positive scenario when TypeScript is enabled and type `Component` is imported from the `vue` library. Not pretty sure this is the problem with this plugin. 🤔 

**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** `v8.28.0`
- **eslint-plugin-vue version:** `v9.8.0`
- **`@typescript-eslint/parser` version**: `v5.44.0`
- **Node version:** `v18.12.0`
- **Operating System:** Mac OS

**Please show your full configuration:**

```json5
{
    "parserOptions": {
        "parser": "@typescript-eslint/parser",
        "extraFileExtensions": [".vue"]
    },

    "extends": ["plugin:vue/base"],

    "rules": {
        "vue/component-name-in-template-casing": "error"
    }
}
```

**What did you do?**	

```vue
<template>
  <component />
</template>

<script lang="ts" setup>
  import { type Component } from 'vue';
</script>
```

**What did you expect to happen?**

No any issues with code.

**What actually happened?**

```
App.vue
  2:3  error  Component name "component" is not PascalCase  vue/component-name-in-template-casing

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

**Repository to reproduce this issue**

https://github.com/Ky6uk/eslint-vue-component-name-in-template-casing-bug

PR #2069, opened 2023-01-01
fix: [vue/component-name-in-template-casing] False positive for dynamic component in template

horitaka:
Modified at Jan. 07.

## Issue
This pull request is a fix for issue #2048.

## How to fixed
Linter check is ignored if type-only import and registeredComponentsOnly=true.
The logic to check whether type-only import or not is the same as for [no-undef-components](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-undef-components.js#L127-L145).

### DIFF
--- a/lib/rules/component-name-in-template-casing.js
+++ b/lib/rules/component-name-in-template-casing.js
@@ -11,6 +11,30 @@ const { toRegExp } = require('../utils/regexp')
 const allowedCaseOptions = ['PascalCase', 'kebab-case']
 const defaultCase = 'PascalCase'
 
+/**
+ * Checks whether the given variable is the type-only import object.
+ * @param {Variable} variable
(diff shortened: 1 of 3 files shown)


FloEdelmann (changes requested):
I don't think that built-in components should always be skipped. Rather, type-only imports should always be ignored in `<script setup>` for this rule.

horitaka:
@FloEdelmann Thank you for your review. I understood the expected behavior of this issue. I will fix it so that type-only imports are ignored.

horitaka:
@FloEdelmann I fixed to skip the check if components is type-only import. Could you review it again?

FloEdelmann (approved):
Looks good to me now. Thanks for fixing!

ota-meshi (approved):
LGTM! Thank you!

PR #2070, opened 2023-01-07
fix (vue/component-api-style): add spacing around comma in report message

43081j:
in the component-api-style rule, the output used to join by `,` which would lead to something like: `you should use a,b or c`.

this simply adds a space such that it becomes `a, b or c`.

i also added a test to hit this branch of code since there wasn't already one.


FloEdelmann (approved):
Looks good, thanks!

ota-meshi (approved):

PR #2067, opened 2022-12-29
Add a few tests & fix the macros-order rule

43081j:
It looks like the macros ordering rule has a bug in it, in that this test fails in master:

```ts
      code: `
        <script setup>
          defineEmits(['update:test'])
          console.log('test1')
          defineProps({ test: Boolean })
        </script>
      `,
      output: `
        <script setup>
          defineEmits(['update:test'])
          defineProps({ test: Boolean })
          console.log('test1')
        </script>
      `,
      options: optionsEmitsFirst,
```

the output actually ends up as:

```html
        <script setup>
          defineProps({ test: Boolean })
          defineEmits(['update:test'])
          console.log('test1')
        </script>
```

which contradicts the options passed (emits _then_ props).

also added a few tests to cover some gaps elsewhere

PR #2071, opened 2023-01-07
fix (vue/define-macros-order): hoist secondary expressions correctly

43081j:
as per #2067 

```ts
      code: `
        <script setup>
          defineEmits(['update:test'])
          console.log('test1')
          defineProps({ test: Boolean })
        </script>
      `,
      output: `
        <script setup>
          defineEmits(['update:test'])
          defineProps({ test: Boolean })
          console.log('test1')
        </script>
      `,
      options: optionsEmitsFirst,
```

the output ends up as:

```html
        <script setup>
          defineProps({ test: Boolean })
          defineEmits(['update:test'])
          console.log('test1')
        </script>
```

which contradicts the options passed (emits _then_ props).

this PR fixes that and adds tests for some gaps we had.

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -137,7 +137,7 @@ function create(context) {
 
           // need move only second
           if (secondStatement !== shouldSecondNode) {
-            reportNotOnTop(order[1], shouldSecondNode, shouldFirstNode)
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #2072, opened 2023-01-07
test (vue/max-attributes-per-line): add test for max value as object

43081j:
passing a multiline max value as an object wasn't tested before, this simply adds such a test.


ota-meshi (approved):
Thank you!

PR #2085, opened 2023-02-12
Added `no-deprecated-vue-extend` rule.

madogiwa0124:
`Vue.extend` is removed in Vue.js 3.0.0+.

> Vue.extend Removed
> https://v3-migration.vuejs.org/breaking-changes/global-api.html#vue-extend-removed

This rule checks for disallow using deprecated `Vue.extend`.


ota-meshi:
Thank you for opening the PR.

I'm not sure people should check it. I think we no longer use the `Vue` identifier when building applications with Vue3. I think it can be checked by just checking the default exports.

```json
     "no-restricted-imports": [
       "error",
       {
         "paths": [
           {
             "name": "vue",
             "importNames": ["default"]
           },
         ],
       },
     ],
```

madogiwa0124:
@ota-meshi 

Indeed. I had not thought of that ! That seems like a good way to do it 👍🏻 

Thanks for the review 😄

Issue #1636, opened 2021-09-24
'Strict' option for vue/no-unused-properties


dsl101:
**What rule do you want to change?**
vue/no-unused-properties

**Does this change cause the rule to produce more or fewer warnings?**
More when `strict: true`

**How will the change be implemented? (New option, new default behavior, etc.)?**
I [have argued](https://github.com/vuejs/eslint-plugin-vue/issues/1462) that this should be the default behaviour, in line with externally-called methods being considered 'unused' unless explicitly labelled in JSDoc style. But, if this change is undesirable, a `strict: true` option to vue/no-unused-properties might be a compromise, with the possibility of making `strict: true` the default in future.

**Please provide some example code that this change will affect:**

```vue
<script>
  export default {
    data () {
      return {
        unusedData: 42
      }
    },
    computed: {
      notReferenced () {
        return 'never seen'
      },
    },
    mounted () {
      const handlerName = 'eventHandler'
      // Event registration code ommitted, but would end up calling via:
      this[handlerName]
    },
    methods: {
      eventHandler () {
        // Used
      },
      notused () {
        // should show as warning
      }
    }
  }
</script>
```

**What does the rule currently do for this code?**
The presence of `this[handlerName]` (as noted in the original ticket) makes it almost impossible for static analysis to determine whether any particular method or property is actually used. The current behaviour appears to assume that _all_ properties, methods, data, and computeds are therefore used, and the above code produces _no_ warnings at all. Commenting out `this[handlerName]` restores the (in my opinion, correct) behaviour that only explicitly referenced properties are 'used'.

**What will the rule do after it's changed?**
Warn about any properties not explicitly referenced, or which are labelled as referenced by the developer using either `// eslint-disable-next-line vue/no-unused-properties` or `/** @public */` annotations.

ota-meshi:
I don't like the option to change the way it works, but I don't reject to add the option to change the way it works.

However, the name of the option can be confusing.
I'd like to have an option name that makes it easy to understand how to handle properties that can't be tracked.

dsl101:
I'm not precious about the name of the option—`strict` came to mind simply because of the analogy with JS `use strict` and the fact it turns on a more strict definition of 'unused'.

`explicit` is an alternative? When `true`, only explicitly referenced properties are considered used? Did you have something in mind that's less confusing?

ota-meshi:
There are some patterns that this rule considers to have used all properties.

e.g. 

```vue
<script>
[...]

PR #2101, opened 2023-03-08
Add option to show warnings when props could be accessed via this

dsl101:
Ref #1636, this is my first stab at adding an option to not treat generic access to `this` as 'using' properties / data / computed, etc. If this is the right approach, I can try adding tests if needed. It's more general I think than what you'd proposed in that issue, but for my use case it works well.

### DIFF
--- a/lib/rules/no-unused-properties.js
+++ b/lib/rules/no-unused-properties.js
@@ -55,7 +55,8 @@ const GROUP_SETUP = 'setup'
 const GROUP_WATCHER = 'watch'
 const GROUP_EXPOSE = 'expose'
 
-const STOPREPORTING_THIS = 'ThisExpression'
+const STOPREPORTING_UNKNOWN = 'unknownPropertyAccess'
+const STOPREPORTING_RETURN = 'returnInstance'
 
 const PROPERTY_LABEL = {
   props: 'property',
@@ -212,7 +213,10 @@ module.exports = {
           stopReporting: {
             type: 'array',
             items: {
-              enum: [ STOPREPORTING_THIS ]
+              enum: [
+                STOPREPORTING_UNKNOWN,
+                STOPREPORTING_RETURN
+              ]
             },
             additionalItems: false,
             uniqueItems: true
@@ -231,7 +235,7 @@ module.exports = {
     const groups = new Set(options.groups || [GROUP_PROPERTY])
     const deepData = Boolean(options.deepData)
     const ignorePublicMembers = Boolean(options.ignorePublicMembers)
-    const stopReporting = new Set(options.stopReporting || [STOPREPORTING_THIS])
+    const stopReporting = new Set(options.stopReporting || [STOPREPORTING_UNKNOWN, STOPREPORTING_RETURN])
 
     const propertyReferenceExtractor = definePropertyReferenceExtractor(context)
 
@@ -571,9 +575,18 @@ module.exports = {
           if (!utils.isThis(node, context)) {
             return
           }
-          if (!stopReporting.has(STOPREPORTING_THIS)) {
+
+          // e.g. return this
+          if (node.parent.type === 'ReturnStatement' &&
+            !stopReporting.has(STOPREPORTING_RETURN)) {
             return
-          }
+          } 
+          // e.g. this[unknown]
+          if (node.parent.computed &&
+            !stopReporting.has(STOPREPORTING_UNKNOWN)) {
+            return
+          } 
+
           const container = getVueComponentPropertiesContainer(vueData.node)
           const propertyReferences =
             propertyReferenceExtractor.extractFromExpression(node, false)
--- a/lib/rules/no-unused-properties.js
+++ b/lib/rules/no-unused-properties.js
@@ -55,8 +55,8 @@ const GROUP_SETUP = 'setup'
 const GROUP_WATCHER = 'watch'
 const GROUP_EXPOSE = 'expose'
 
-const STOPREPORTING_UNKNOWN = 'unknownPropertyAccess'
-const STOPREPORTING_RETURN = 'returnInstance'
+const REPORT_DYNAMIC_THIS = 'reportDynamicThis'
+const REPORT_RETURNED_THIS = 'reportReturnedThis'
 
 const PROPERTY_LABEL = {
   props: 'property',
@@ -210,10 +210,10 @@ module.exports = {
           },
           deepData: { type: 'boolean' },
           ignorePublicMembers: { type: 'boolean' },
-          stopReporting: {
+          reportingOptions: {
             type: 'array',
             items: {
-              enum: [STOPREPORTING_UNKNOWN, STOPREPORTING_RETURN]
+              enum: [REPORT_DYNAMIC_THIS, REPORT_RETURNED_THIS]
             },
             additionalItems: false,
             uniqueItems: true
@@ -232,19 +232,15 @@ module.exports = {
     const groups = new Set(options.groups || [GROUP_PROPERTY])
     const deepData = Boolean(options.deepData)
     const ignorePublicMembers = Boolean(options.ignorePublicMembers)
-    const stopReporting = new Set(
-      options.stopReporting || [STOPREPORTING_UNKNOWN, STOPREPORTING_RETURN]
-    )
-    const stopReportingUnknown = Boolean(
-      stopReporting.has(STOPREPORTING_UNKNOWN)
-    )
-    const stopReportingReturn = Boolean(stopReporting.has(STOPREPORTING_RETURN))
+    const reportingOptions = new Set(options.reportingOptions || [])
+    const reportDynamicThis = Boolean(reportingOptions.has(REPORT_DYNAMIC_THIS))
+    const reportReturnedThis = Boolean(reportingOptions.has(REPORT_RETURNED_THIS))
 
     const propertyReferenceExtractor = definePropertyReferenceExtractor(
       context,
       {
-        stopReportingUnknown,
-        stopReportingReturn
+        reportDynamicThis,
+        reportReturnedThis
       }
     )
 
--- a/lib/rules/no-unused-properties.js
+++ b/lib/rules/no-unused-properties.js
@@ -55,8 +55,8 @@ const GROUP_SETUP = 'setup'
 const GROUP_WATCHER = 'watch'
 const GROUP_EXPOSE = 'expose'
 
-const REPORT_DYNAMIC_THIS = 'reportDynamicThis'
-const REPORT_RETURNED_THIS = 'reportReturnedThis'
+const UNREFERENCED_UNKNOWN_MEMBER = 'unknownMemberAsUnreferenced'
+const UNREFERENCED_RETURN = 'returnAsUnreferenced'
 
 const PROPERTY_LABEL = {
   props: 'property',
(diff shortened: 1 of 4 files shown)


ota-meshi:
Thank you for this PR.
But, I'm not sure what you mean by the option you added. Could you please add test cases and documentation?

dsl101:
OK, before I go too far, can I check the original issue makes sense to you (it was a while ago, but I just got some time to revisit)? What I currently see (and am still to be convinced this is not the wrong behaviour) is that the plugin considers all elements in this code as 'used' and shows no warnings:
```vue
<script>
  export default {
    computed: {
      life () {
        return 42
      }
    },
    methods: {
      getThis () {
        return this
      },
    }
  }
</script>
```
My understanding is that is because returning `this` from a method means 'something' _could_ access any prop, data, computed, method, etc. I think this is a big assumption, and as you mentioned, you can access most things via `this.$refs` anyway, so why warn about _anything_? I would prefer this to apply to plain `this` too, _unless_ the item is marked with the JSDoc `@public`. However, your argument was not to change the current default behaviour but introduce an option.

I tried to follow your example in #1636 whereby the option `stopReporting` would default to `ThisExpression`, and effectively stop the plugin from reporting warnings when it encountered an ambiguous access via `this`. For my configuration, I will set `stopReporting` to an empty array, and get the warnings back.

If that all makes sense, I will write this up in docs, and add a test case—but only if you still think this is the right way to go...

ota-meshi:
Thanks for the explanation.  But I'm still not sure.  I need the specific source code.  I don't think the current changes match what you want to do.

dsl101:
> Thanks for the explanation. But I'm still not sure. I need the specific source code. I don't think the current changes match what you want to do.

I'm not sure what you mean by this? I've set this up locally with my changes, and by putting `stopReporting: []` in my eslint config, all my unused props are warned about again. This is my .eslintrc.js:
```js
module.exports = {
  extends: [
    // add more generic rulesets here, such as:
    // 'eslint:recommended',
    // 'plugin:vue/vue3-recommended',
    'plugin:vue/recommended' // Use this if you are using Vue.js 2.x.
  ],
  rules: {
    // override/add rules settings here, such as:
    // 'vue/no-unused-vars': 'error'
    'vue/no-unused-properties': ['warn', {
      'groups': [ 'props', 'data', 'computed', 'methods' ],
      'deepData': true,
      'ignorePublicMembers': true,
      'stopReporting': []
    }],
  }
}
```
and in Sublimetext, I now see this:
image

dsl101:
OK, I think I see what you mean. The change is too general at the moment, and also still warns about explicit accesses such as `this.life`. I will keep looking at where the general `this` or `this[variable]` access is detected.

dsl101:
Right, I _think_ part of the solution is in `utils/property-references.js` in function `extractFromExpression()`. This section at (currently) line 313. I've marked up where I _think_ the case of `this.life` vs `const h = 'life'; this[h]` is handled:
```js
        case 'MemberExpression': {
          if (parent.object === node) {
            // `arg.foo`
            const name = utils.getStaticPropertyName(parent)

            return name
              ? new PropertyReferencesForMember(parent, name, withInTemplate)
              : ANY  // ←----- This line should return NEVER to avoid suppressing warnings?
          }
          return NEVER
        }

```
With `const h = 'life'; this[h], `utils.getStaticPropertyName()` returns `null` (as it is a `computed` node), and so `extractFromExpression()` returns `ANY`, and suppresses _all_ warnings.

However, I can't see how to access the configuration options within that utility function, to selectively return `NEVER`.

ota-meshi:
I'm still not quite sure what this PR change is trying to do. Could you add test cases and documentation to clarify what you want to do?

dsl101:
I will try, but I'm not sure just documenting the new option is enough—especially since it's just an implementation of your own suggestion... Perhaps this example is clearer? Take this code—which of the `computed` items would you expect to be warned about as 'unused'?
```vue
<script>
  export default {
    computed: {
      one () {
        return 1
      },
      two () {
        return 2
      },
      three () {
        return 3
      },
    },
    methods: {
      handler () {
        console.log(this.one)
        const i = 'two'  // Could be two or something else
        console.log(this[i])
        return this
      },
    }
  }
</script>
```
The current behaviour of the plugin is to warn about _none_ of them being unused, despite only `one` being explicitly used in the `handler()`. I would like an option to be able to warn about `two` and `three` being unused in this case. Now, the _ideal_ solution would only show `three` as unused, since `two` is actually logged to the console, but I understand this is extremely hard if not impossible to determine with static analysis. So the 2 ways in which `two` and `three` _might_ be accessed, from your code in #1636, are via `this[varName]` and via `return this`. So I've added options, as you suggested, to stop reporting when those access methods are used, and set them as the default. Then I can remove those options via `stopReporting: []` in my config, and the warnings show up again.

If that makes it clear what I'm trying to achieve, I'm happy to add test cases and try to document that. But if it's still not clear, I'm not sure how else to describe the issue...

dsl101:
Does this change make sense now? I will try to update the documentation to explain if so, but want to make sure you're ok with it.

ota-meshi:
I am very busy right now with my daughters having the flu.  So I can't check this change yet.

dsl101:
Sorry to hear that—no problem at all, just checking in.

ota-meshi:
Thank you for adding the test cases.

But I think the PR change still doesn't handle the new options well.
For example, I don't think the following case can be handled well.

```js
    {
      filename: 'test.vue',
      code: `
      <script>
        export default {
          data () {
            return {
              foo: {
                bar: 1
              },
              baz: 2
            }
          },
          methods: {
            handler () {
              console.log(this.baz)
              return this.foo
            },
          }
        }
      </script>
      `,
      options: [{
        groups: ['data'],
        deepData: true,
        stopReporting: []
      }],
      errors: [
        {
          message: "'foo.bar' of data found, but never used.",
          line: 7,
        }
      ]
    }
```

To handle them better, I think we should pass an option to `lib/utils/property-references.js`'s `definePropertyReferenceExtractor` to change it to return `NEVER` when a `ReturnStatement` is encountered. I think the `MemberExpression` should also be handled within the `definePropertyReferenceExtractor`.

dsl101:
Yes, that's what I concluded in [this comment](https://github.com/vuejs/eslint-plugin-vue/pull/2101#issuecomment-1462383353), but I wasn't sure if passing options down there was even possible let alone desirable. When I get a bit more time, I'll try to update the PR.

dsl101:
OK, so one clarification; the `deepData` option is handled in the rule, so why not the `stopReporting` option?

dsl101:
OK, I've had a go at passing the options in. I'm not sure it's the most elegant way, and the defaults (`true` for both) are now baked in to the utility function, which I don't think is ideal. But it has no concept of the `options` structure that I can see.

Also, that utility function is called from `no-undef-properties.js`, but I don't think that rule is affected by these new options.

lib/utils/property-references.js
```diff
@@ -92,7 +92,10 @@ module.exports = {
 /**
  * @param {RuleContext} context The rule context.
  */
-function definePropertyReferenceExtractor(context) {
+function definePropertyReferenceExtractor(
+  context,
+  { stopReportingUnknown = true, stopReportingReturn = true } = {}
```
ota-meshi, line 97: Your changes are almost identical to what I had in mind. Thank you.

However, I think the option names are a bit confusing for this function. (It's not about the option names of the rule.)

For example, how about naming it like this:

```suggestion
  { unknownMemberAsUnreferenced = false, returnStatementAsUnreferenced = false } = {}
```

(Note that I set the initial value to false.)

dsl101: First of all, great that we're aligned :). In terms of changing the names, I still find the actual config options a bit confusing (I don't like 'negatives' in the options, so it would be good to remove the 'stop' part for example). Perhaps we can work out a name that will work in both places?

Since both are specifically about access to `this`, it seems like a good idea to get that into the option. `this[variable]` is a dynamic access to `this`, and `return this` is a potential access via the returned value, so could we use `reportDynamicThis` and `reportReturnedThis` as both the config and argument names? That has the advantage of not needing to set anything for the default options (to avoid changing the current behaviour)—it would only change by _adding_ those options. It also means the tests for this PR can be simplified, as no new default options are being introduced, and only explicit tests for the new functionality are needed.

I will update with those changes—easy to revert if you have a strong argument against changing the names...

ota-meshi: Hmm... Sorry but the naming is still confusing to me.
I think the `report` is confusing because the `definePropertyReferenceExtractor` function does not call `context.report()`. I think it should mean whether all properties are considered referenced or all properties are considered unreferenced.
Also, I think `This` is confusing because there are cases other than `this`. We may be better to use `Expression` instead.

dsl101: Yes, that makes sense. So the lower level utility is concerned with if things are considered referenced or not, whereas the outer rule is concerned with what to report—I will rename those accordingly.

> Also, I think This is confusing because there are cases other than this. We may be better to use Expression instead.

Since my use case is specifically around `this`, it seems like there should be some other cases to consider then. Can you share any examples which could be used in the tests, where this option should apply?

ota-meshi: >  Can you share any examples which could be used in the tests

For example, the following test does not return `this`.

```js
    {
      filename: 'test.vue',
      code: `
      <script>
        export default {
          data () {
            return {
              foo: {
                bar: 1
              },
              baz: 2
            }
          },
          methods: {
            handler () {
              const vm = this
              console.log(vm.baz)
              return vm.foo
            },
          }
        }
      </script>
      `,
      options: [{
        groups: ['data'],
        deepData: true,
        stopReporting: []
      }],
      errors: [
        {
          message: "'foo.bar' of data found, but never used.",
          line: 7,
        }
      ]
    }
```

dsl101:
My latest changes do work as expected [here](https://github.com/vuejs/eslint-plugin-vue/pull/2101#discussion_r1147033780) I think now—I've added an additional test. The new options do not mention the word 'this' any more. If you're happy with the changes as they are, I'll write up in the documentation.

lib/utils/property-references.js
dsl101: Added this in tests

ota-meshi (approved):
LGTM!

lib/rules/no-unused-properties.js
```diff
@@ -221,8 +232,19 @@ module.exports = {
...
-    const propertyReferenceExtractor = definePropertyReferenceExtractor(context)
+    const propertyReferenceExtractor = definePropertyReferenceExtractor(
+      context,
+      {
+        unknownMemberAsUnreferenced: Boolean(
+          unreferencedOptions.has(UNREFERENCED_UNKNOWN_MEMBER)
+        ),
+        returnAsUnreferenced: Boolean(
+          unreferencedOptions.has(UNREFERENCED_RETURN)
+        )
```
FloEdelmann, line 245: Are the explicit Boolean casts really needed? Doesn't `Set.has(…)` return a boolean already?

FloEdelmann (approved):
Looks good to me, especially the documentation and relatively simple implementation 👍

I have just one question, otherwise this is good to go :)

lib/rules/no-unused-properties.js
dsl101: Belt and braces, plus following the style of other options initialisation of being explicit, but yes you're right it's not necessary. I've taken it out, but note the rather opinionated Prettier rules now make it less readable imo...

FloEdelmann (approved):
The tests are failing at an unrelated file, maybe they are just flaky @ota-meshi.

Other than that, this looks good to me :slightly_smiling_face:

PR #2106, opened 2023-03-20
docs: fixed grammar

wenfangdu:
Ref: [Is "all HTML elements are a single word" grammatically correct?](https://ell.stackexchange.com/questions/334216/is-all-html-elements-are-a-single-word-grammatically-correct)

### DIFF
--- a/docs/rules/multi-word-component-names.md
+++ b/docs/rules/multi-word-component-names.md
@@ -16,7 +16,7 @@ since: v7.20.0
 This rule require component names to be always multi-word, except for root `App`
 components, and built-in components provided by Vue, such as `<transition>` or
 `<component>`. This prevents conflicts with existing and future HTML elements,
-since all HTML elements are a single word.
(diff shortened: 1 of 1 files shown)


docs/rules/multi-word-component-names.md
```diff
@@ -16,7 +16,7 @@ since: v7.20.0
 This rule require component names to be always multi-word, except for root `App`
 components, and built-in components provided by Vue, such as `<transition>` or
 `<component>`. This prevents conflicts with existing and future HTML elements,
-since all HTML elements are a single word.
+since every HTML element is a single word.
```
FloEdelmann, line 19: I think this sounds odd and incorrect now. Let's make "single word" plural in the original sentence, then it should be fine, right?

```suggestion
since all HTML elements are single words.
```

FloEdelmann:
Thanks for raising this!

Issue #2109, opened 2023-03-21
Rule Proposal: `vue/enforce-style-attribute`


mussinbenarbia:
**Please describe what the rule should do:**

Many projects agree on using one specific way of writing CSS. For example, using CSS Modules, strictly using Scoped CSS, or a combination of the two, which may be rare but could happen when migrating from one style to another. 

This rule helps developers enforce one or either of the two attributes in SFCs.

**What category should the rule belong to?**

[x] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

#### Option `either`: No warning as long as one attribute is specified. Does not enforce one specific style.
```vue
<!-- ✓ GOOD -->
<style scoped></style>
<style lang="scss" src="../path/to/style.scss" scoped></style>

<!-- ✓ GOOD -->
<style module></style>

<!-- ✗ BAD -->
<style></style>
```

#### Option `scoped`: Style tags must have the `scoped` attribute.
```vue
<!-- ✓ GOOD -->
<style scoped></style>
<style lang="scss" src="../path/to/style.scss" scoped></style>

<!-- ✗ BAD -->
<style module></style>
<style></style>
```

#### Option `module`: Style tags must have the `module` attribute.
```vue
<!-- ✓ GOOD -->
<style module></style>


<!-- ✗ BAD -->
<style scoped></style>
<style lang="scss" src="../path/to/style.scss" scoped></style>
<style></style>
```

ota-meshi:
Thank you for the rule proposal.
Sounds good to me.
However, I think your proposed rule is a little too strict.

Also, I have a similar rule available in another plugin.
https://future-architect.github.io/eslint-plugin-vue-scoped-css/rules/enforce-style-type.html
`eslint-plugin-vue-scoped-css`'s rule can use arrays to allow multiple attribute types in your application. You can also force it to be `plain`.

I think it would be better if we changed your implementation rule to accept options similar to the `eslint-plugin-vue-scoped-css`'s rule.
What do you think?

mussinbenarbia:
Thank you for the feedback and for linking to the other plugin! I think that makes sense. I will modify the implementation to accept the same options.

PR #2110, opened 2023-03-21
Add new `vue/enforce-style-attribute` rule

mussinbenarbia:
This PR adds a new rule as proposed in [this issue](https://github.com/vuejs/eslint-plugin-vue/issues/2109).


ota-meshi (changes requested):
See https://github.com/vuejs/eslint-plugin-vue/issues/2109#issuecomment-1479055924

mussinbenarbia:
@ota-meshi Sorry for the very long absence 🙏 I made some modifications to the code to mimick how it's done in the `eslint-plugin-vue-scoped-css` plugin. Could you please take a look at the implementation? I have not updated the docs as I would like to make sure you agree with the implementation first.

Thank you 🙇

docs/rules/enforce-style-attribute.md
```diff
@@ -0,0 +1,83 @@
...
+
+> enfore either the `scoped` or `module` attribute in SFC top level style tags
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+## :wrench: Options
+
+```json
+{
+  "vue/attribute-hyphenation": ["error", "either" | "scoped" |  "module"]
```
mussinbenarbia, line 18: issue: wrong rule name, need to update once implementation is finalized

ota-meshi:
That rule sounds good to me. Could you undraft this PR when it is ready for review?

ota-meshi (changes requested):
That rule sounds good to me. Could you undraft this PR when it is ready for review?

mussinbenarbia:
@ota-meshi Thank you for taking a look! I will update the docs to reflect the new options and mark it as ready for review.

mussinbenarbia:
@ota-meshi (Looping in @FloEdelmann as well) I updated the docs but I have a question. When I run `docs:watch` or `docs:build`, the following piece of text gets automatically added to my doc file for the rule:

```
- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
```

I'm not sure why. This rule does not have a fixer at the moment, so that paragraph is not required but it keeps being added automatically. Is it ok to remove it and push the code? Is that not going to make the CI fail?

mussinbenarbia:
Also, I'm now wondering if the `allows` option should actually be singular form `allow` to keep some form of consistency with other rules such as [no-shadow](https://github.com/eslint/eslint/blob/main/docs/src/rules/no-shadow.md) 🤔 Do you agree?

lib/rules/enforce-style-attribute.js
```diff
@@ -0,0 +1,153 @@
...
+        'enforce or forbid the use of the `scoped` and `module` attributes in SFC top level style tags',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/enforce-style-attribute.html'
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          allows: {
```
FloEdelmann, line 60: I agree that it should be called `allow` instead, for consistency with other rules.

lib/rules/enforce-style-attribute.js
```diff
@@ -0,0 +1,153 @@
...
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'enforce or forbid the use of the `scoped` and `module` attributes in SFC top level style tags',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/enforce-style-attribute.html'
+    },
+    fixable: 'code',
```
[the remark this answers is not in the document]
FloEdelmann: Sorry for the double work; I just checked it out myself already. See

* https://github.com/vuejs/eslint-plugin-vue/pull/2343

Thanks for checking though :slightly_smiling_face:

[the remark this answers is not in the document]
mussinbenarbia: No problem! Then I'll go ahead and resolve this 👍

docs/rules/enforce-style-attribute.md
```diff
@@ -0,0 +1,85 @@
...
+
+</eslint-code-block>
+
+## :wrench: Options
+
+```json
+{
+  "vue/enforce-style-attribute": [
+    "error",
+    { "allow": ["scoped", "module", "no-attributes"] }
```
ota-meshi, line 75: Hmm. I think the name `no-attributes` is misleading. lang="scss" has attributes. What do you think?

ota-meshi (commented):
Sorry for the late review.

docs/rules/enforce-style-attribute.md
mussinbenarbia: Yeah... that could be confusing 🤔 Maybe using `plain` like in the other plugin is the best choice here. If you agree I'll rename it

ota-meshi: I think `plain` is better :+1:

mussinbenarbia:
@ota-meshi I renamed the option to `plain`. I pushed the code but it seems some tests for other unrelated rules are failing. I would appreciate if you could take another look 🙇

ota-meshi:
Yeah. CI error should be fixed by #2357. So ignore it for now.

ota-meshi (approved):
LGTM! Thank you!

PR #2131, opened 2023-04-13
Fixed false negatives in `vue/camelcase` when using ESLint>=v8.38

ota-meshi:
This PR fixes a bug found in https://github.com/vuejs/eslint-plugin-vue/pull/2101#pullrequestreview-1383007618.


FloEdelmann (approved):
I don't know exactly what this is doing, but it seems to work fine.

Issue #1371, opened 2020-12-09
vue/no-mutating-props "reference-only" option


cdrini:
**What rule do you want to change?**
vue/no-mutating-props

**Does this change cause the rule to produce more or fewer warnings?**
The new "reference-only" option would cause it produce fewer warnings.

**How will the change be implemented? (New option, new default behavior, etc.)?**
New option: "reference-only" (i.e. Only error when mutating prop references)

**Please provide some example code that this change will affect:**

```vue
<template>
  <div>
    <input type="range" v-model.number="point.x" >
    <input type="range" v-model.number="point.y" >
  </div>
</template>
<script>
  export default {
    props: {
      point: {
        type: Object,
        required: true
      }
    },
  }
</script>
```

**What does the rule currently do for this code?**
Trigger an error

**What will the rule do after it's changed?**
When the default option is set, the error will trigger. When the option is set to "reference-only", it will not trigger an error.

**Additional context**

(Note there are some issues in the past suggesting this rule be loosened (#1339 , #1314; cc @leosdad @pimlie @decademoon ); here I'm suggesting an option be added instead)

My main reason for wanting an option here, is that modifying the _reference_ of a prop (e.g. `this.myProp = 3`) will cause actual run-time warnings from Vue, cause unexpected behaviour, and is deprecated:

> Due to the new rendering mechanism, whenever the parent component re-renders, the child component’s local changes will be overwritten. - https://vuejs.org/v2/style-guide/#Implicit-parent-child-communication-use-with-caution

Here's the warning Vue throws:

`Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "number"`

Modifying the _value_ of a prop (but leaving the reference the same) (e.g. `this.myProp.x = 3`), does not throw runtime warnings, and (I believe) does not cause any unexpected behaviour—it's just a code style decision. As long as the actual reference to the object remains unaltered, Vue should not have any issues. 

Here's a sandbox with these two cases for reference: https://codesandbox.io/s/elegant-lederberg-p1ilc

(Note: This is the same way that `const` behaves in JS, so shouldn't be unexpected to most developers:

```js
const x = 3;
x = 7; // Error!

const y = {};
y.foo = 3; // This is fine; it's the _reference_ that's const, not the value.
y = 20; // Error
```
)

[...]

PR #2135, opened 2023-04-20
Add `shallowOnly` option to `vue/no-mutating-props`

FlareZh:
Ref https://github.com/vuejs/eslint-plugin-vue/issues/1371

### DIFF
--- a/docs/rules/no-mutating-props.md
+++ b/docs/rules/no-mutating-props.md
@@ -123,16 +123,16 @@ This rule reports mutation of component props.
 ```json
 {
   "vue/no-mutating-props": ["error", {
-    "propProps": true
+    "shallowOnly": false
   }]
 }
 ```
 
-- "propProps" (`boolean`) Avoid mutating the value of a prop but leaving the reference the same. Default is `true`.
+- "shallowOnly" (`boolean`) Enables mutating the value of a prop but leaving the reference the same. Default is `true`.
 
-### "propProps": false
+### "shallowOnly": true
 
-<eslint-code-block :rules="{'vue/no-mutating-props': ['error', {propProps: false}]}">
+<eslint-code-block :rules="{'vue/no-mutating-props': ['error', {shallowOnly: true}]}">
 
 ```vue
 <!-- ✓ GOOD -->
--- a/docs/rules/no-mutating-props.md
+++ b/docs/rules/no-mutating-props.md
@@ -128,7 +128,7 @@ This rule reports mutation of component props.
 }
 ```
 
-- "shallowOnly" (`boolean`) Enables mutating the value of a prop but leaving the reference the same. Default is `true`.
+- "shallowOnly" (`boolean`) Enables mutating the value of a prop but leaving the reference the same. Default is `false`.
 
 ### "shallowOnly": true
 
--- a/docs/rules/no-mutating-props.md
+++ b/docs/rules/no-mutating-props.md
@@ -22,6 +22,8 @@ This rule reports mutation of component props.
 <template>
   <div>
     <input v-model="value" @click="openModal">
+    <button @click="pushItem">Push Item</button>
+    <button @click="changeId">Change ID</button>
   </div>
 </template>
 <script>
@@ -30,11 +32,25 @@ This rule reports mutation of component props.
       value: {
         type: String,
         required: true
+      },
+      list: {
+        type: Array,
+        required: true
+      },
+      user: {
+        type: Object,
+        required: true
       }
     },
     methods: {
       openModal() {
         this.value = 'test'
+      },
+      pushItem() {
+        this.list.push(0)
+      },
+      changeId() {
+        this.user.id = 1
       }
     }
   }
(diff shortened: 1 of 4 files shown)


docs/rules/no-mutating-props.md
```diff
@@ -88,7 +120,45 @@ This rule reports mutation of component props.
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+  "vue/no-mutating-props": ["error", {
+    "propProps": true
```
ota-meshi, line 126: Did you see the comment below regarding option names? Could you please comment what you think about those names?
https://github.com/vuejs/eslint-plugin-vue/issues/1371#issuecomment-1064620364

lib/utils/index.js
```diff
@@ -1778,9 +1778,11 @@ module.exports = {
 
   /**
    * @param {MemberExpression|Identifier} props
+   * @param {boolean} propProps
+   * @param {boolean} isRootProps
    * @returns { { kind: 'assignment' | 'update' | 'call' , node: ESNode, pathNodes: MemberExpression[] } | null }
    */
-  findMutating(props) {
+  findMutating(props, propProps = true, isRootProps = false) {
```
ota-meshi, line 1785: The findMutating function is also used outside of rules related to props. So I don't want to add those options to this function.
By checking pathNodes in the rule implementation, I think you can determine if you are mutating properties.

ota-meshi (changes requested):
Thank you for this PR! I have change request.

docs/rules/no-mutating-props.md
FlareZh: Thank you for your reply! I think "shallowOnly" is more appropriate. It has the same naming style with vue3.

lib/utils/index.js
FlareZh: I can delete the param `isRootProps` and make it work, but it seems impossible to determine whether the options `shallowOnly(propProps)` is ture or false without modifying the `findMutating` function, because the return value of `target.type` `UpdateExpression` and `UnaryExpression` is totally same.

So, could I change the return value `kind` to `unary` when `target.type === 'UnaryExpression'`?
Or could I replace the param `propProps` with `options`?

ota-meshi: If UpdateExpression and UnaryExpression are of the same kind, I'm not sure what the problem is. Could you explain a little more?

docs/rules/no-mutating-props.md
```diff
@@ -88,7 +120,45 @@ This rule reports mutation of component props.
...
-Nothing.
+```json
+{
+  "vue/no-mutating-props": ["error", {
+    "shallowOnly": false
+  }]
+}
+```
+
+- "shallowOnly" (`boolean`) Enables mutating the value of a prop but leaving the reference the same. Default is `true`.
```
FloEdelmann, line 131: I think the default is `false`, right? If no, then I think it should be.

```suggestion
- "shallowOnly" (`boolean`) Enables mutating the value of a prop but leaving the reference the same. Default is `false`.
```

FlareZh: Yes! Thank you!

lib/utils/index.js
FlareZh: Oh, there is no problem, it's my fault. I have pushed a new commit to revert it.

FloEdelmann:
Please run `git pull` (as I have updated the branch with the latest changes from the `master` branch), `npm install` and `npm run update` to regenerate the docs once more.

lib/rules/no-mutating-props.js
```diff
@@ -423,12 +478,22 @@ module.exports = {
...
 
+          const propsInfo = /** @type {PropsInfo} */ (
+            propsMap.get(vueObjectData.object)
+          )
+
           const nodes = utils.getMemberChaining(node)
           const first = nodes[0]
           let name
-          if (isVmReference(first)) {
+          if (isVmReference(first) && first.name !== propsInfo.name) {
```
ota-meshi, line 488: It looks like the following cases are no longer reported (without new option):

```vue
<script setup>
  const props = defineProps()
</script>

<template>
  <input v-model="props[foo]" >
</template>
```

I think we should branch like this:

```js
          if (isVmReference(first)) {
            if (first.name === propsInfo.name) {
              // props variable
              if (shallowOnly && nodes.length > 2) {
                return
              }
              name =
                (nodes[1] && utils.getStaticPropertyName(nodes[1])) ||
                first.name
            } else {
              if (shallowOnly && nodes.length > 1) {
                return
              }
              name = first.name
              if (!name || !propsInfo.set.has(name)) {
                return
              }
            }
          } else if (first.type === 'ThisExpression') {
            if (shallowOnly && nodes.length > 2) {
              return
            }
            const mem = nodes[1]
            if (!mem) {
              return
            }
            name = utils.getStaticPropertyName(mem)
            if (!name || !propsInfo.set.has(name)) {
              return
            }
          } else {
            return
          }
          report(node, name)
```

Also, could you add this test case as well?

ota-meshi (changes requested):

lib/rules/no-mutating-props.js
FlareZh: Of course, I have fixed it and added related tests.

ota-meshi (approved):
Looks great to me! Thank you!

FloEdelmann (approved):
Looks good, thank you!

Issue #1467, opened 2021-03-26
rule: no v-if on template root


privatenumber:
**Please describe what the rule should do:**

Warn on using `v-if` on the template root node, because:

1. Whether a component should be rendered or not is a usage concern. Specifically, it's confusing when a declared component is not rendered:

    ```html
    /* Usage */
    <template>
        <div>
            ...
            <custom-component /> /* Readers expect this to be rendered */
        </div>
    </template>


    /* CustomComponent.vue */
    <template>
        <div v-if="false">
            ...
        </div>
    </template>
    ```

2. There are performance benefits:
    1. Instantiating a new stateful component only for it to not be rendered via root `v-if="false"` is wasteful. If possible, it should happen in the parent usage life-cycle.
    2. When moving the `v-if` to the parent, the component can be asynchronously loaded, so not only will it save on component registration, it will save on module declaration and asset size.
        ```vue
        <template>
            <div>
                ...
                <custom-component v-if="shouldShow" /> /* Won't be loaded till true */
            </div>
        </template>

        <script>
        export default {
        	components: {
                CustomComponent: () => import('./CustomComponent.vue')
            }
        }
        </script>
        ```

**What category should the rule belong to?**

[ ] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[x] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

**Bad**
```html
/* Usage */
<template>
    <div>
        <custom-component />
    </div>
</template>


/* CustomComponent.vue */
<template>
    <div v-if="shouldShow">
        ...
    </div>
</template>
```

**Good**
```html
/* Usage */
<template>
    <div>
        <custom-component v-if="shouldShow" />
    </div>
</template>


/* CustomComponent.vue */
<template>
    <div>
        ...
    </div>
</template>
```

**Additional context**

- There are likely cases where it's not possible to do this. I think this would be a warning in "recommended", or not even a pre-enabled rule.

ota-meshi:
Thank you for rule proposal!
I think it's good to make it rule available to users.

I think the following template needs to be valid.

```vue
<template>
    <div v-if="mode === 'a'">
        ...
    </div>
    <div v-if="mode === 'b'">
        ...
    </div>
</template>
```

```vue
<template>
    <div v-if="loading">
        ...
    </div>
    <div v-else>
        ...
    </div>
</template>
```

[...]

PR #2138, opened 2023-04-25
fix(valid-template-root): no v-if on template root

songpengyuan:
Add root element v-if validation

#1467


lib/rules/valid-template-root.js
```diff
@@ -39,13 +39,49 @@ module.exports = {
...
+                hasRootVIfLength += 1
+              }
+              if (hasRootVElse) {
+                hasRootVElseLength += 1
+              }
+              if (hasRootVElseIf) {
+                hasRootVElseIfLength += 1
+              }
+            }
+          }
```
FloEdelmann, line 73: Could be simplified:

```suggestion
          const hasRootVIfLength = rootElements.filter(element => element.type === 'VElement' && utils.hasDirective(element, 'if')).length
          const hasRootVElseLength = rootElements.filter(element => element.type === 'VElement' && utils.hasDirective(element, 'else')).length
          const hasRootVElseIfLength = rootElements.filter(element => element.type === 'VElement' && utils.hasDirective(element, 'else-if')).length
```

lib/rules/valid-template-root.js
```diff
@@ -39,13 +39,49 @@ module.exports = {
...
+            hasRootVIfLength === 1 &&
+            hasRootVElseLength === 0 &&
+            hasRootVElseIfLength === 0
+          ) {
             context.report({
               node: element,
               loc: element.loc,
               message:
-                "The template root with 'src' attribute is required to be empty."
+                'The v-if should not be used on root element without v-else.'
```
FloEdelmann, line 84:
```suggestion
                '`v-if` should not be used on root element without `v-else`.'
```

tests/lib/rules/valid-template-root.js
```diff
@@ -119,6 +123,11 @@ tester.run('valid-template-root', rule, {
       errors: [
         "The template root with 'src' attribute is required to be empty."
       ]
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-if="foo"></div></template>',
+      errors: ['The v-if should not be used on root element without v-else.']
     }
```
FloEdelmann, line 131: I think this new behavior should be configurable, since it is actually a valid template root (which this rule checks) and some users might want to allow this.

FloEdelmann (changes requested):

tests/lib/rules/valid-template-root.js
songpengyuan: Thank you, the option ignoreOnlyIfDirective has been added

tests/lib/rules/valid-template-root.js
```diff
@@ -119,6 +129,12 @@ tester.run('valid-template-root', rule, {
       errors: [
         "The template root with 'src' attribute is required to be empty."
       ]
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-if="foo"></div></template>',
+      options: [{ ignoreOnlyIfDirective: false }],
+      errors: ['`v-if` should not be used on root element without `v-else`.']
     }
```
FloEdelmann, line 138: Please also add this test case as valid without options.

FloEdelmann (changes requested):

tests/lib/rules/valid-template-root.js
```diff
@@ -99,6 +99,16 @@ tester.run('valid-template-root', rule, {
...
+    {
+      filename: 'test.vue',
+      options: [{ ignoreOnlyIfDirective: false }],
+      code: '<template> <div v-if="mode === \'a\'"></div><div v-if="mode === \'b\'"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      options: [{ ignoreOnlyIfDirective: false }],
+      code: '<template> <div v-if="loading"></div><div v-else></div></template>'
+    }
```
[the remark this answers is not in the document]
FloEdelmann: This is not yet resolved.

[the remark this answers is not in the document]
songpengyuan: ok

ota-meshi:
Thank you for this PR!
However, I think the new option is better implemented as a separate rule. Because the valid-template-root rule is aimed at catching problems early, I think the new check is aimed at checking for template definitions that don't fit your project's tastes.
What do you think?

CC @FloEdelmann

ota-meshi (changes requested):
https://github.com/vuejs/eslint-plugin-vue/pull/2138#issuecomment-1528950118

lib/rules/valid-v-if-template-root.js
```diff
@@ -0,0 +1,72 @@
...
+      element.type === 'VElement' && utils.hasDirective(element, directiveName)
+  ).length
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce valid `v-if` directives on root element',
+      categories: ['vue3-essential', 'essential'],
```
FloEdelmann, line 28: This would be a breaking change. Also, I don't think that everyone will like to forbid `v-if` on the root tag.

lib/rules/valid-v-if-template-root.js
```diff
@@ -0,0 +1,72 @@
...
+          hasRootVElseLength === 0 &&
+          hasRootVElseIfLength === 0
+        ) {
+          context.report({
+            node: element,
+            loc: element.loc,
+            message:
+              '`v-if` should not be used on root element without `v-else`.'
+          })
+        }
```
FloEdelmann, line 68: I think this logic can be simplified to

```suggestion
        if (rootElements.length === 1 && rootElements[0].type === 'VElement' && utils.hasDirective(rootElements[0], 'if')) {
          context.report({
            node: element,
            loc: element.loc,
            message:
              '`v-if` should not be used on root element without `v-else`.'
          })
        }
```

tests/lib/rules/valid-v-if-template-root.js
```diff
@@ -0,0 +1,111 @@
...
+    {
+      filename: 'test.vue',
+      code: '<template><custom-component v-if="foo"></custom-component></template>',
+      errors: ['`v-if` should not be used on root element without `v-else`.']
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-if="foo"></div></template>',
+      errors: ['`v-if` should not be used on root element without `v-else`.']
+    }
```
FloEdelmann, line 109: Please also add a tests for this case:

```
<template><div /><div v-if="foo" /></template>
```

FloEdelmann (changes requested):

lib/rules/valid-v-if-template-root.js
```diff
@@ -0,0 +1,72 @@
...
+        if (element == null) {
+          return
+        }
+
+        const rootElements = []
+        for (const child of element.children) {
+          if (sourceCode.getText(child).trim() !== '') {
+            rootElements.push(child)
+          }
+        }
```
ota-meshi, line 51: I think we can write it like this:

```suggestion
        const rootElements = element.children.filter(utils.isVElement)
```

ota-meshi (changes requested):

lib/rules/valid-v-if-template-root.js
songpengyuan: This simplification, the following case will report an error
```
<template><div /><div v-if="foo" /></template>
```

tests/lib/rules/no-root-v-if.js
```diff
@@ -0,0 +1,116 @@
...
+      errors: ['`v-if` should not be used on root element without `v-else`.']
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-if="foo"></div></template>',
+      errors: ['`v-if` should not be used on root element without `v-else`.']
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div /><div v-if="foo" /></template>',
```
ota-meshi, line 112: In my opinion this template is valid.
I think the rule should report `v-if` where all of the root element can be lost.
What do you think?

ota-meshi (changes requested):

docs/rules/no-root-v-if.md
```diff
@@ -0,0 +1,39 @@
...
+```vue
+<!--  `v-if` should not be used on root element without `v-else` -->
+<template>
+  <div v-if="foo"></div>
+</template>
+
+<template><custom-component v-if="shouldShow" /></template>
+```
+
+</eslint-code-block>
```
FloEdelmann, line 30: Two `<template>` blocks break the example. I think one should suffice to show the point:

````suggestion
<eslint-code-block :rules="{'vue/no-root-v-if': ['error']}">

```vue
<template>
  <div v-if="foo"></div>
</template>
```

</eslint-code-block>
````

File: tests/lib/rules/no-root-v-if.js (hunk 10 above)
Comment (FloEdelmann) [inline tests/lib/rules/no-root-v-if.js:112]: Hmm, should that not be valid (in Vue 3, where multiple template root tags are allowed)?

There is no way to move that check outside the component, which is the goal of this rule.

**Review by FloEdelmann — CHANGES_REQUESTED** [review]
Please run `npm run update` to regenerate the docs.

File: lib/rules/no-root-v-if.js [hunk 12]
```diff
@@ -0,0 +1,70 @@
...
+    elseLength,
+    elseIfLength
+  }
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce valid `v-if` directives on root element',
```
ota-meshi, line 38: I don't think this description is correct.
I think that rule just reports the v-if of the root element. `valid`/`invalid` is irrelevant.

docs/rules/no-root-v-if.md
```diff
@@ -0,0 +1,37 @@
...
+description: disallow `v-if` directives on root element
+---
+
+# vue/no-root-v-if
+
+> disallow `v-if` directives on root element
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
+
+This rule checks whether every template root is valid.
```
FloEdelmann, line 14:
```suggestion
This rule reports template roots with `v-if`. Rendering of the whole component could be made conditional in the parent component (with a `v-if` there) instead.
```

docs/rules/no-root-v-if.md
```diff
@@ -0,0 +1,37 @@
...
+sidebarDepth: 0
+title: vue/no-root-v-if
+description: disallow `v-if` directives on root element
+---
+
+# vue/no-root-v-if
+
+> disallow `v-if` directives on root element
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
```
FloEdelmann, line 12: To fix the lint warning:

```suggestion
- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
```

Why did you change the automatically generated (and correct) syntax?

FloEdelmann (approved):
Apart from two little suggestions, this looks fine by me. Thanks!

docs/rules/no-root-v-if.md
songpengyuan: Could be a problem with my editor plugin, fixed

ota-meshi (approved):
LGTM! Thank you!

PR #2146, opened 2023-05-06
Update eslint-plugin-unicorn to v47

TatsuyaYamamoto:
Related issues: nothing

currently, main branch of eslint-plugin-vue cannot run lint.
Due to a bug in eslint-plugin-unicorn, if eslint-plugin-vue installs ESLint v42.0, eslint-plugin-unicorn must be v46.0.1 or higher.

```
> eslint-plugin-vue@9.11.0 lint
> eslint . --rulesdir eslint-internal-rules && markdownlint "**/*.md"


Oops! Something went wrong! :(

ESLint: 8.40.0

TypeError: Cannot read properties of undefined (reading 'getAllComments')
```
ref: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2076

## What I did

1. update eslint-plugin-unicorn
2. ~~fix lint error~~ turn off rules

## ~~How to fix lint error~~

1. precondition => `254 problems (253 errors, 1 warning)`
1. `npm run lint:fix` => ` 4 problems (3 errors, 1 warning)`
1. set top-level await manually => `1 problem (0 errors, 1 warning)`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -128,7 +128,7 @@ module.exports = {
     'unicorn/filename-case': 'off',
     'unicorn/no-null': 'off',
     'unicorn/no-array-callback-reference': 'off', // doesn't work well with TypeScript's custom type guards
-    'unicorn/no-negated-condition': 'off', // turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/no-negated-condition': 'off', // remove when there are few pull requests (ref: #2146)
     'unicorn/no-useless-undefined': 'off',
     'unicorn/prefer-optional-catch-binding': 'off', // not supported by current ESLint parser version
     'unicorn/prefer-module': 'off',
@@ -136,9 +136,9 @@ module.exports = {
     'unicorn/prefer-at': 'off', //                 turn off to prevent make breaking changes (ref: #2146)
     'unicorn/prefer-node-protocol': 'off', //      turn off to prevent make breaking changes (ref: #2146)
     'unicorn/prefer-string-replace-all': 'off', // turn off to prevent make breaking changes (ref: #2146)
-    'unicorn/prefer-ternary': 'off', //            turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/prefer-ternary': 'off', //            remove when there are few pull requests (ref: #2146)
     'unicorn/prefer-top-level-await': 'off', //    turn off to prevent make breaking changes (ref: #2146)
-    'unicorn/switch-case-braces': 'off', //        turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/switch-case-braces': 'off', //        remove when there are few pull requests (ref: #2146)
 
     'require-eslint-community': ['error']
   },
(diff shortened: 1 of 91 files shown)


TatsuyaYamamoto:
It seems to have another problem of eslint-plugin-vue and ESLint.

If ESLint v8.**40** is installed, eslint-plugin-vue’s tests failed.
If ESLint v8.**39**.1 is installed, the test succeeds

eslint-internal-rules/no-invalid-meta-docs-categories.js
```diff
@@ -71,15 +71,15 @@ function checkMetaValidity(context, exportsNode) {
           // fixes.push(fixer.insertTextBefore(category.value, '['), fixer.insertTextAfter(category.value, ']'))
 
           // for vue3 migration
```
FloEdelmann, line 73: This comment belongs to the `!== 'base'` branch, so please move it inside.

FloEdelmann (commented):
Thanks for the dependency update and lint fixes! However, I'm not sure if they are all compatible with the Node version we support. @ota-meshi which Node versions should we support?

Also, I'm not sure what causes the other failing tests.

ota-meshi:
Thank you for this PR.
However, source code changes contain many breaking changes.
Could you change the eslintrc so that we don't have to change the source code?

TatsuyaYamamoto:
> Could you change the eslintrc so that we don't have to change the source code?

I turned off some rules

> Also, could you revert the changed source code?

I reverted 92305cc, adf77e9

.eslintrc.js
```diff
@@ -128,10 +128,17 @@ module.exports = {
...
     'unicorn/no-useless-undefined': 'off',
     'unicorn/prefer-optional-catch-binding': 'off', // not supported by current ESLint parser version
     'unicorn/prefer-module': 'off',
     'unicorn/prevent-abbreviations': 'off',
+    'unicorn/prefer-at': 'off', //                 turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/prefer-node-protocol': 'off', //      turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/prefer-string-replace-all': 'off', // turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/prefer-ternary': 'off', //            turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/prefer-top-level-await': 'off', //    turn off to prevent make breaking changes (ref: #2146)
+    'unicorn/switch-case-braces': 'off', //        turn off to prevent make breaking changes (ref: #2146)
```
FloEdelmann, line 141: `unicorn/no-negated-condition`, `unicorn/prefer-ternary` and `unicorn/switch-case-braces` don't cause breaking changes, so I think they should be enabled (and the respective lint errors be fixed).

ota-meshi: I don't want that rule enabled. because it conflicts with the current open pull request. I think it's better to enable the rule when there are few pull requests.

TatsuyaYamamoto: for now, should I change comments only?

`turn off to prevent make breaking changes (ref: #2146)`
=> `remove when there are few pull requests (ref: #2146)`

FloEdelmann: Yes, I think that will be a good compromise :+1:

TatsuyaYamamoto: @FloEdelmann 
Done!  6f4dca6

FloEdelmann (approved):
Thanks for the changes!

ota-meshi (approved):
Thank you!

PR #2145, opened 2023-05-06
feat(processor): export meta object for flat config

TatsuyaYamamoto:
Related issue: nothing
Related PR: https://github.com/vuejs/vue-eslint-parser/pull/190

This PR’s purpose is to support ESLint flat config.

Loading eslint-plugin-vue in flat config, ESLint check processor’s name and version in meta object.
https://github.com/eslint/eslint/blob/main/lib/config/flat-config-array.js#L220
https://github.com/eslint/eslint/blob/main/lib/config/flat-config-array.js#L45-L76

PR #2147, opened 2023-05-06
Fixed rules incompatible with ESLint v8.40.

ota-meshi:
I looked into #2145 and found that the CI #2145 test was failing due to rules not compatible with ESLint v8.40.
This PR fixes that issue.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -133,7 +133,11 @@ module.exports = {
     'unicorn/prefer-module': 'off',
     'unicorn/prevent-abbreviations': 'off',
 
-    'require-eslint-community': ['error']
+    'require-eslint-community': ['error'],
+
+    // FIXME: The version we are currently using is not compatible.
+    // May be removed in #2146. https://github.com/vuejs/eslint-plugin-vue/pull/2146
+    'unicorn/expiring-todo-comments': 'off'
   },
   overrides: [
     {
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -133,7 +133,11 @@ module.exports = {
     'unicorn/prefer-module': 'off',
     'unicorn/prevent-abbreviations': 'off',
 
-    'require-eslint-community': ['error']
+    'require-eslint-community': ['error'],
+
+    // FIXME: The version we are currently using is not compatible.
+    // May be removed in #2146. https://github.com/vuejs/eslint-plugin-vue/pull/2146
+    'unicorn/expiring-todo-comments': 'off'
```
FloEdelmann, line 140: I don't understand why this is necessary. That rule was enabled for a long time without causing issues.

FloEdelmann (commented):
The actual change looks fine, but I have one question.

.eslintrc.js
ota-meshi: The problem that #2146 fixes is still there, so the lint process now crashes.
We should remove this workaround when we merge #2146.

FloEdelmann: Ah, I see.

FloEdelmann (approved):

PR #2149, opened 2023-05-08
Add meta object to the plugin

ota-meshi:
This PR is similar to #2145, but this PR adds a meta object to the plugin itself.
It seems that the plugin also needs it.

https://eslint.org/docs/latest/extend/plugins#metadata-in-plugins

### DIFF
--- a/tests/lib/meta.js
+++ b/tests/lib/meta.js
@@ -13,7 +13,7 @@ describe('Test for meta object', () => {
   })
 
   for (const [name, processor] of Object.entries(plugin.processors)) {
-    it(`"${name}" processor  should have a meta object.`, () => {
+    it(`"${name}" processor should have a meta object.`, () => {
       assert.deepStrictEqual(processor.meta, expectedMeta)
     })
   }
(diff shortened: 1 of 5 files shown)


tests/lib/meta.js
```diff
@@ -0,0 +1,20 @@
...
+  version: require('../../package.json').version
+}
+
+describe('Test for meta object', () => {
+  it('A plugin should have a meta object.', () => {
+    assert.deepStrictEqual(plugin.meta, expectedMeta)
+  })
+
+  for (const [name, processor] of Object.entries(plugin.processors)) {
+    it(`"${name}" processor  should have a meta object.`, () => {
```
FloEdelmann, line 16:
```suggestion
    it(`"${name}" processor should have a meta object.`, () => {
```

FloEdelmann (approved):

tests/lib/meta.js
ota-meshi: Thank you!

PR #2152, opened 2023-05-12
Add support for new `defineEmits` type syntax to `vue/require-explicit-emits` rule

### LINKED ISSUE #2126: Support for new `defineEmits` type syntax

- **ESLint version:**  v8.36.0
- **eslint-plugin-vue version:** v9.10.0
- **Node version:** v18.12.1

Support for new `defineEmits` type syntax.

See https://github.com/vuejs/core/blob/main/CHANGELOG.md#330-alpha7-2023-04-03

Some rules are affected. However, we have not yet listed the affected rules.
I think that, first, we need to list the rules affected by new `defineEmits` type syntax.

**Additional context**

ota-meshi:
close #2126


FloEdelmann (approved):
:heavy_check_mark: LGTM!

Issue #2125, opened 2023-04-04
Support for `defineSlots` and `slots` option


ota-meshi:
**Tell us about your environment**

- **ESLint version:**  v8.36.0
- **eslint-plugin-vue version:** v9.10.0
- **Node version:** v18.12.1

**The problem you want to solve.**

Support for `defineSlots` macro and `slots` option.

See https://github.com/vuejs/core/blob/main/CHANGELOG.md#330-alpha7-2023-04-03

**Your take on the correct solution to problem.**

Parser and some rules are affected. However, we have not yet listed the affected rules.
I think that, first, we need to list the rules affected by defineSlots and `slots` option.

**Additional context**

Related to https://github.com/vuejs/vue-eslint-parser/issues/184

ota-meshi:
It probably doesn't affect anything other than `vue/order-in-components` (#2153) and `vue/define-macros-order` rules.
However, it may be useful to have new rules.

PR #2153, opened 2023-05-12
Add `slots`+`expose` to `vue/order-in-components` default order

### LINKED ISSUE #2399: Add `slots` to `vue/order-in-components` default order

`vue/order-in-components`

More.

Add [`slots`](https://vuejs.org/api/options-rendering.html#slots) to default order. Currently, the default order is:

https://github.com/vuejs/eslint-plugin-vue/blob/e936a1524c3bde167e68a365bd6fcd1b86a059b2/lib/rules/order-in-components.js#L14-L76

I'd propose to add the `slots` option after `emits`, in the `Interface` category. This is probably a breaking change, but we could argue that the `slots` option (which was introduced in Vue 3.3) is not in use for long, so intentionally breaking might be okay.

@ota-meshi what do you think?

ota-meshi:
* Related to #2125
* Fixes #2399

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -88,6 +88,8 @@ export default {
       "model",
       ["props", "propsData"],
       "emits",
+      "slots",
+      "expose",
       "setup",
       "asyncData",
       "data",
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):
:heavy_check_mark: LGTM!

But since this is a breaking change, maybe we should first release a last v9.x version without this PR merged, so we don't force people to upgrade to a major version to use the new features.

Also, we could think about other breaking changes that we may want to accumulate into one v10 release.

FloEdelmann:
@ota-meshi I changed the target branch to `v10` now. Could you please resolve the conflicts so we can merge this?

PR #2154, opened 2023-05-12
Add support for `defineOptions` and `defineSlots` to `vue/define-macros-order` rule

ota-meshi:
This PR adds support for `defineOptions` and `defineSlots` in `vue/define-macros-order` rule. (Default behavior is unchanged.)

This PR also includes changes that make `utils.defineScriptSetupVisitor` visit `defineSlots` and `defineOptions`.


FloEdelmann (approved):
:heavy_check_mark: LGTM!

PR #2163, opened 2023-05-13
Add support for `defineOptions` and `defineSlots` to `vue/no-unsupported-features` rule

ota-meshi:
This PR adds support for `defineOptions` and `defineSlots` in `vue/no-unsupported-features` rule.


FloEdelmann (approved):
Wow, you are on fire! :fire: :smile: 

This PR also looks good to me.

PR #2165, opened 2023-05-13
Add `vue/valid-define-options` rule

ota-meshi:
This PR adds `vue/valid-define-options` rule that checks whether `defineOptions` compiler macro is valid.

### DIFF
--- a/lib/rules/valid-define-options.js
+++ b/lib/rules/valid-define-options.js
@@ -99,7 +99,7 @@ module.exports = {
                 if (utils.withinTypeNode(node)) {
                   continue
                 }
-                //`defineOptions` are referencing locally declared variables.
+                //`defineOptions` is referencing locally declared variables.
                 context.report({
                   node,
                   messageId: 'referencingLocally'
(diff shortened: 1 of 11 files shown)


lib/rules/valid-define-options.js
```diff
@@ -0,0 +1,127 @@
...
+      // TODO Switch in the next major version
+      // categories: ['vue3-essential', 'essential'],
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/valid-define-options.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      referencingLocally:
+        '`defineOptions` are referencing locally declared variables.',
```
FloEdelmann, line 24:
```suggestion
        '`defineOptions` is referencing locally declared variables.',
```

FloEdelmann (approved):
Apart from one little wording issue, this looks good to me :)

lib/rules/valid-define-options.js
ota-meshi: Thank you! I also fixed the place where the same sentence was used.

PR #2167, opened 2023-05-13
Add `vue/prefer-define-options` rule

ota-meshi:
This PR adds `vue/prefer-define-options` rule that enforce use of `defineOptions` instead of default export.


FloEdelmann (approved):
LGTM!

PR #2168, opened 2023-05-13
Add support for `generic` attribute to `vue/no-unsupported-features` rule

ota-meshi:


lib/rules/no-unsupported-features.js
```diff
@@ -115,7 +117,10 @@ module.exports = {
       forbiddenVBindPropModifierShorthand:
         '`.prop` shorthand are not supported until Vue.js "3.2.0".',
       forbiddenVBindAttrModifier:
-        '`.attr` modifiers on `v-bind` are not supported until Vue.js "3.2.0".'
+        '`.attr` modifiers on `v-bind` are not supported until Vue.js "3.2.0".',
+      // Vue.js 3.3.0+
+      forbiddenGenericAttribute:
+        '`generic` attribute is not supported until Vue.js "3.3.0".'
```
FloEdelmann, line 123: Does it really need to be reported? Adding a `generic` attribute won't do anything on Vue < 3.3, but it also won't hurt.

ota-meshi: That's a good point! I will close this PR 😅

PR #2171, opened 2023-05-14
Add update-resources script

ota-meshi:
This PR adds a script to update the embedded HTML/SVG element list resource.

The HTML/SVG element list comes from <https://unpkg.com/@mdn/browser-compat-data/data.json> and <https://html.spec.whatwg.org/multipage/obsolete.html>


tests/lib/rules/no-reserved-component-names.js
```diff
@@ -374,8 +364,6 @@ const invalidElements = [
   'Blink',
   'center',
   'Center',
-  'command',
-  'Command',
```
FloEdelmann, line 378: Where did the list previously come from? Is it okay that the `content`, `element`, `shadow`, `discard`, `unknown` and `command` names are now allowed?

ota-meshi: It seems they were listed on MDN.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element#obsolete_and_deprecated_elements

https://github.com/vuejs/eslint-plugin-vue/pull/757#issuecomment-452968477

But they have been updated and some are not currently listed.

We don't know where the old ones came from, so I think it's hard to keep track of them. So I think it makes sense to use the typescript definition.

FloEdelmann: `content` and `shadow` are still listed there. Maybe it makes sense to have an additional list next to the one pulled from the TypeScript definition?

Or should we strive to get those elements included in TypeScript itself?

ota-meshi: > Apparently, the TypeScript types are generated from https://github.com/mdn/browser-compat-data.

Thank you for letting me know. I didn't know that.
Hmm. It might make more sense to generate resource files from `@mdn/browser-compat-data` than from `typescript`. What do you think?

https://www.npmjs.com/package/@mdn/browser-compat-data

FloEdelmann: Yeah, skipping the intermediate TypeScript step seems to make sense.

ota-meshi: I'll look into how to do that.

FloEdelmann: Hmm, so BCD+WhatWG include everything that TypeScript has, plus `portal` HTML element and `discard` SVG element. Sounds like this is the better data source then.

However, running inside the version script might add new restricted elements (which could be a breaking change) only right before a new release. That seems dangerous to me.

Maybe we can try to make the script faster by not reconstructing the whole DOM with `jsdom`, but rather using a faster parser like [`node-html-parser`](https://www.npmjs.com/package/node-html-parser)?

ota-meshi: > However, running inside the version script might add new restricted elements (which could be a breaking change) only right before a new release. That seems dangerous to me.

So, I will to separate it from the update script and change it to check periodically with GitHub Actions.

> Maybe we can try to make the script faster by not reconstructing the whole DOM with jsdom, but rather using a faster parser like [node-html-parser](https://www.npmjs.com/package/node-html-parser)?

Thanks for sharing. I will check that out.

ota-meshi: I tried node-html-parser but it doesn't seem to work.
I suspect that the reason is that HTML with omitted `</html>` and `</body>` cannot be parsed well 🤔

ota-meshi: FYI: The `discard` SVG element seems to have been removed from the latest data.
https://github.com/mdn/browser-compat-data/pull/19782

FloEdelmann (approved):

FloEdelmann:
@ota-meshi I changed the target branch to `v10` now. Could you please resolve the conflicts so we can merge this?

ota-meshi:
Thanks! I've updated this PR.
I've also modified GitHub Actions to automatically create a pull request when a resource is changed.

.github/workflows/check-for-resources-update.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install Packages
+        run: npm install
+      - name: Update
+        run: npm run update-resources
+      - uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: Updates resources with latest
+          branch: update-unicode-alias
+          branch-suffix: timestamp
+          title: Updates resources with latest
```
FloEdelmann, line 30:
```suggestion
          commit-message: Updates resources
          branch: update-resources
          branch-suffix: timestamp
          title: Updates resources
```

PR #2181, opened 2023-05-17
Document related rules for component name rules

### LINKED ISSUE #2180: Rule Proposal: `vue/ban-component-names`

Ban some component names, such as those that are the same to native html tag or built-in components.

I think this rule can include `vue/multi-word-component-names`.

For `vue/multi-word-component-names`, I have encountered many developers who do not find this rule helpful as it is turned on by default, and it goes against their intuition since using a single word as a component name is very common in development, finally they will close the rule 😂.

So I want this rule can enhance `vue/multi-word-component-names`, provides more fine-tuned, intuitive control and greater customizability.

[ ] Other (please specify:)

```js
/* ✓ GOOD */
Vue.component('foo', {
  // ...
})

/* ✗ BAD */
Vue.component('title', {
  // ...
})
```

```vue
<script>
<!-- ✓ GOOD-->
export default {
  name: 'foo'
}

<!-- ✗ BAD -->
export default {
  name: 'transition'
}
</script>
```

```vue
<!-- filename: body.vue -->
<script>
/* ✗ BAD */
export default {
  // ...
}
</script>
```

**Additional context**

We can collect html tag names from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element).

Some potentially useful options:

- ~~allowHtmlTag~~: allow those that are same to html tag
- banPattern: use to match which name is banned
- customBans: specify names to ban
- ignores: names to ignore
- ignorePattern: use to match which name is ignored

It you approve, I am willing to make a PR for this.

qmhc:
Fix #2180.

### DIFF
--- a/docs/rules/multi-word-component-names.md
+++ b/docs/rules/multi-word-component-names.md
@@ -169,6 +169,10 @@ export default {
 
 </eslint-code-block>
 
+## :couple: Related Rules
+
+- [vue/no-reserved-component-names](./no-reserved-component-names.md)
+
 ## :books: Further Reading
 
 - [Style guide - Multi-word component names](https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names)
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):
Nice, thank you!

PR #2185, opened 2023-05-19
Extend `vue/no-dupe-keys` to support `<script setup>`

### LINKED ISSUE #2096: Extend `vue/no-dupe-keys` to support `<script setup>`
Similar to ESLint's [`no-shadow`](https://eslint.org/docs/latest/rules/no-shadow) rule, it would be great to have something similar, like `vue/no-prop-shadow` where it would warn on local variable declaration that has the same name as a prop, which can lead to issues when referencing it in the template because props can be directly accessed from the template.

[ ] Other (please specify:)

WARN
```vue
<script setup lang="ts">
defineProps<{
  msg: string;
}>()

const msg = "override props"; // WARN: there's already a prop named `msg`
</script>

<template>
  <div>{{ msg }}</div> <!-- vague reference. Is this from props or local variable? -->
</template>
```

WARN
```vue
<script lang="ts">
import { defineComponent, computed } from "vue";

  props: {
    msg: String
  },
  setup: (props) => {
    const msg = computed(() => `Message: ${props.msg}`);
    
    return {
      msg, // WARN: there's already a prop named `msg`
    };
  }
})
</script>

<template>
    <div>{{ msg }}</div> <!-- vague reference. Is this from props or local variable? -->
</template>
```

**Additional context**

More context from the original issue: 
- https://github.com/vuejs/eslint-plugin-vue/issues/2095

ItMaga:
close #2096

### DIFF
--- a/lib/rules/no-dupe-keys.js
+++ b/lib/rules/no-dupe-keys.js
@@ -4,10 +4,13 @@
  */
 'use strict'
 
+const { findVariable } = require('@eslint-community/eslint-utils')
 const utils = require('../utils')
 
 /**
  * @typedef {import('../utils').GroupName} GroupName
+ * @typedef {import('eslint').Scope.Variable} Variable
+ * @typedef {import('../utils').ComponentProp} ComponentProp
  */
 
 /** @type {GroupName[]} */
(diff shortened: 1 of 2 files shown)


lib/rules/no-dupe-keys.js
```diff
@@ -39,24 +42,43 @@ module.exports = {
...
+      utils.defineScriptSetupVisitor(context, {
+        onDefinePropsEnter(node, props) {
+          for (const prop of props) {
+            if (!prop.propName) continue
 
-        usedNames.add(o.name)
-      }
-    })
+            const variable = findVariable(context.getScope(), prop.propName)
+            if (!variable) continue
```
ota-meshi, line 70: Thank you for opening this PR!
It looks mostly good to me, but I have one concern.
Presumably the current implementation warns against declaring a prop with the same name as a variable in the global scope. The global scope contains some commonly named variables, so I think it's better to ignore it. What do you think?
https://github.com/sindresorhus/globals/blob/64d801f42fd9a35617622b1b19010701352efe6d/globals.json#L683

```suggestion
            if (!variable || variable.defs.length === 0) continue
```

ota-meshi (changes requested):

lib/rules/no-dupe-keys.js
ItMaga: Thank you for the quick feedback!

Nice catch! I even didn't think about this.
Fixed

ota-meshi (approved):
LGTM! Thank you!

PR #2189, opened 2023-05-24
Fix false positives for toRef props in `vue/no-dupe-keys` rule

ota-meshi:


lib/rules/no-dupe-keys.js
```diff
@@ -16,6 +16,48 @@ const utils = require('../utils')
...
+ * @returns {Pattern|null}
+ */
+function getPropsPattern(node) {
+  let target = node
+  if (
+    target.parent &&
+    target.parent.type === 'CallExpression' &&
+    target.parent.arguments[0] === target &&
+    target.parent.callee.type === 'Identifier' &&
+    target.parent.callee.name === 'withDefaults'
```
FloEdelmann, line 31: Please add a test case for the `withDefaults` function.

lib/rules/no-dupe-keys.js
```diff
@@ -63,12 +105,23 @@ module.exports = {
...
+
           for (const prop of props) {
             if (!prop.propName) continue
 
             const variable = findVariable(context.getScope(), prop.propName)
             if (!variable || variable.defs.length === 0) continue
 
+            for (const def of variable.defs) {
+              if (def.type !== 'Variable') continue
+              if (isInsideInitializer(def.node, propReferences)) return
```
FloEdelmann, line 122: Is it really `return` in this case or should it also be `continue`?

ota-meshi: This is correct. Ignore the report if the props variable is included in the variable initialization.

ota-meshi: Sorry. This is incorrect. The outer loop should continue.

ota-meshi: I fixed that. Thanks for finding that mistake!

FloEdelmann: Thanks for adding a test case!

FloEdelmann (approved):
Looks good to me now :)

PR #2194, opened 2023-05-27
Add `vue/no-console` rule

### LINKED ISSUE #2155: [New Rule] `vue/no-console`

Since Vue ^3.3 now [supports console in the template](https://github.com/vuejs/core/issues/7939), a `vue/no-console` rule is needed to regulate its usage.

`vue/no-console` should work exactly the same as [`no-console`](https://eslint.org/docs/latest/rules/no-console) but in the template.

[ ] Other (please specify:)

With `'vue/no-console': 'error'`, the following code should throw:

```vue
<template>
  <button @click="console.log"></button>
</template>
```

ItMaga:
Closes #2155

### DIFF
--- a/docs/rules/no-console.md
+++ b/docs/rules/no-console.md
@@ -14,8 +14,6 @@ description: Disallow the use of `console` in `<template>`
 
 This rule is the same rule as core [no-console] rule but it applies to the expressions in `<template>`.
 
-</eslint-code-block>
-
 ## :books: Further Reading
 
 - [no-console]
--- a/docs/rules/no-console.md
+++ b/docs/rules/no-console.md
@@ -1,0 +1,30 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-console
+description: Disallow the use of `console` in `<template>`
+---
+# vue/no-console
+
+> Disallow the use of `console` in `<template>`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule is the same rule as core [no-console] rule but it applies to the expressions in `<template>`.
+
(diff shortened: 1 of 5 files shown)


ota-meshi (changes requested):
~~Thank you for this PR.
Could you please add test cases to check if the rule works?~~
Forget this comment.
I saw and commented on the wrong diff.

tests/lib/rules/no-console.js
```diff
@@ -0,0 +1,128 @@
...
+tester.run('no-console', rule, {
+  valid: [
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+        <button @click="Console.log">button</button>
+      </template>
+      `,
+      options: [{ allow: ['log'] }]
```
ota-meshi, line 27: I don't think this option is needed if the intent was to test when the identifiers don't match.

```suggestion
```

ota-meshi (changes requested):

tests/lib/rules/no-console.js
ItMaga: Done

docs/rules/no-console.md
```diff
@@ -0,0 +1,30 @@
...
+
+> Disallow the use of `console` in `<template>`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule is the same rule as core [no-console] rule but it applies to the expressions in `<template>`.
+
+</eslint-code-block>
```
ota-meshi, line 17: Remove unnecessary end tag.

```suggestion
```

ota-meshi (approved):
LGTM! Thank you!

FloEdelmann (approved):
Looks good to me too :slightly_smiling_face:

PR #2198, opened 2023-05-30
Add `vue/require-macro-variable-name` rule

### LINKED ISSUE #2188: New rule `vue/require-macro-variable-name`
New config fields:
- Require using a defined name for the `define*` variable (e.g. `{ "name": "props" }`)
- Require name prefix in html (not for `vue/define-emits-declaration` e.g. `{ "prefix": true }`)

More (depends on configuration).

The default behavior does not change the current behavior of the rule.

`"vue/define-props-declaration": ["error", "type-based", { "name": "props" } ]`
```ts
interface Props {
  active: boolean;
}
// GOOD
const props = defineProps<Props>();
// BAD
const myProps = defineProps<Props>();
```
`"vue/define-props-declaration": ["error", "type-based", { "prefix": true } ]`
```html
<template>
  <!-- GOOD-->
  {{ props.active }}
  <!-- BAD-->
  {{ active }}
  {{ $props.active }}
</template>
```
`"vue/define-slots-declaration": ["error", "always", { "name": "slots", "prefix": true } ]`
```html
<template>
  <!-- GOOD-->
  {{ slots }}
  {{ slots.default }}
  <!-- BAD-->
  {{ $slots }}
  {{ $slots.default }}
</template>
```
`"vue/define-attrs-declaration": ["error", "always", { "name": "attrs", "prefix": true } ]`
```html
<template>
  <!-- GOOD-->
  {{ attrs }}
  <!-- BAD-->
  {{ $attrs}}
</template>
```

https://eslint.vuejs.org/rules/define-props-declaration.html

ItMaga:
Closes #2188

### DIFF
--- a/tests/lib/rules/require-macro-variable-name.js
+++ b/tests/lib/rules/require-macro-variable-name.js
@@ -291,6 +291,34 @@ tester.run('require-macro-variable-name', rule, {
         }
       ],
       options: [customOptions]
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+        const slotsCustom = defineSlots({})
+        const attrsCustom = useAttrs({})
+      </script>
+      `,
+      errors: [
+        {
+          message: `The variable name of "useAttrs" must be "attrs".`,
+          line: 4,
+          column: 15,
+          suggestions: [
+            {
+              desc: `Change the variable name to "attrs".`,
+              output: `
+      <script setup>
+        const slotsCustom = defineSlots({})
+        const attrs = useAttrs({})
+      </script>
+      `
+            }
+          ]
+        }
+      ],
+      options: [{ defineSlots: 'slotsCustom' }]
     }
   ]
 })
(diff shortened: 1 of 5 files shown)


tests/lib/rules/require-macro-variable-name.js
```diff
@@ -0,0 +1,296 @@
...
+        const useSlots = useSlots({})
+        const ${customOptions.useAttrs} = useAttrs({})
+      </script>
+      `
+            }
+          ]
+        }
+      ],
+      options: [customOptions]
+    }
```
ota-meshi, line 294: Could you please add a test case for when omit some of the options?
For example, if we use an option other than `useAttrs`, what is the `useAttrs` variable name forced to?

ota-meshi (changes requested):
Mostly looks good to me, but could you add one test case and change it to work if it doesn't work?

ota-meshi (approved):
LGTM! Thank you!

PR #2203, opened 2023-06-01
Don't report user-defined `createApp` function in `vue/one-component-per-file`

### LINKED ISSUE #2201: [one-component-per-file] do not check createApp function written by the developer
**Checklist**

- **ESLint version:** 8
- **eslint-plugin-vue version:** 9.14.1
- **Node version:** node16
- **Operating System:** win10

```json5

```

**What did you do?**	

```js
import { createApp } from 'other.js'
test(() => {
    createApp({})
})

test(() => {
    createApp({})
})
```
I write a function name `createApp` and when i write the unit test it has eslint `one-component-per-file` warning.

There is no eslint warning.

**What actually happened?**

It show the `one-component-per-file` warning`

candy-Tong:


lib/utils/index.js
```diff
@@ -2527,6 +2528,23 @@ function getVueComponentDefinitionType(node) {
...
+
+        // only lint the createApp function that import from vue
+        if (variable !== null && variable.defs.length === 1) {
+          const def = variable.defs[0]
+          if (
+            def.type === 'ImportBinding' &&
+            def.node.type === 'ImportSpecifier' &&
+            def.node.imported.type === 'Identifier' &&
+            def.node.parent.type === 'ImportDeclaration' &&
+            def.node.parent.source.value !== 'vue'
```
FloEdelmann, line 2542: `@vue/composition-api` should also be allowed; please also add test cases for that if there aren't any yet.

FloEdelmann (changes requested):
Thanks for implementing a fix! I have one comment, see below :)

lib/utils/index.js
FloEdelmann: And maybe also add a test with `createApp` defined in the same file instead of imported from another file?

FloEdelmann (approved):
Looks good to me now, thanks :)

lib/utils/index.js
```diff
@@ -2536,6 +2537,31 @@ function getVueComponentDefinitionType(node) {
       if (callee.name === 'createApp') {
         // for Vue.js 3.x
         // createApp({})
+
+        const variable = findVariable(context.getScope(), callee)
```
ota-meshi, line 2541: Sorry for the late reply. I am busy with office work.

I think it can be written more simply using `tracker.iterateEsmReferences`.

https://eslint-community.github.io/eslint-utils/api/scope-utils.html#tracker-iterateesmreferences

e.g.

```js
          const tracker = new ReferenceTracker(context.getSourceCode().scopeManager.globalScope)
          const traceMap = utils.createCompositionApiTraceMap({
            [ReferenceTracker.ESM]: true,
            createApp: {
              [ReferenceTracker.CALL]: true
            }
          })
          if ([...tracker.iterateEsmReferences(traceMap)].every(({ node }) => node !== callee)) {
            return null;
          }
```

ota-meshi (changes requested):

lib/utils/index.js
FloEdelmann: @candy-Tong Would you like to give this a try?

FloEdelmann: @candy-Tong Would you like to give this a try?

FloEdelmann: @candy-Tong Would you like to give this a try?

PR #2204, opened 2023-06-02
Add `vue/require-typed-ref` rule

### LINKED ISSUE #2105: New rule `vue/require-typed-ref`: disallow ref() without type parameter or argument

Disallow calling `ref()` function without generic type parameter or an argument when using TypeScript.

With TypeScript it is easy to prevent usage of `any` by using `no-implicit-any`. Unfortunately this rule is easily bypassed with Vue `ref()` function. Calling `ref()` function without a generic parameter or an initial value leads to ref having `Ref<any>` type.

[ ] Other (please specify:)

Bad:
```vue
<script setup lang="ts">
const counter = ref() // Ref<any>

counter.value = 0
counter.value = '' // Would be caught if we had typed ref
</script>
```

Bad: (maybe as an option to disallow explicit any too)
```vue
<script setup lang="ts">
const counter = ref<any>() // Ref<any>
</script>
```

Good:
```vue
<script setup lang="ts">
const counter = ref(0) // Ref<number>

conter.value = '' // Will result in a typecheck error
</script>
```

Good:
```vue
<script setup lang="ts">
const counter = ref<number>()
</script>
```

Additional example by @FloEdelmann
Good:
```vue
<script setup lang="ts">
const counter: Ref<number | undefined> = ref()
</script>
```

**Additional context**

Demivan:
Closes #2105

### DIFF
--- a/docs/rules/require-typed-ref.md
+++ b/docs/rules/require-typed-ref.md
@@ -1,0 +1,43 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/require-typed-ref
+description: xxx
+---
+# vue/require-typed-ref
+
+> require `ref` and `shallowRef` functions to be strongly typed
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule disallows calling `ref()` or `shallowRef()` functions without generic type parameter or an argument when using TypeScript.
+
+With TypeScript it is easy to prevent usage of `any` by using `no-implicit-any`. Unfortunately this rule is easily bypassed with Vue `ref()` function. Calling `ref()` function without a generic parameter or an initial value leads to ref having `Ref<any>` type.
+
+<eslint-code-block :rules="{'vue/require-typed-ref': ['error']}">
+
+```vue
+<script setup lang="ts">
+import { ref, shallowRef, type Ref } from 'vue'
+
+/* ✗ BAD */
+const count = ref() // Returns Ref<any> that is not type checked
+count.value = '50' // Should be a type error, but it is not
+
+const count = shallowRef()
+
+/* ✗ GOOD */
+const count = ref<number>()
+const count = ref(0)
+const count: Ref<number | undefined> = ref()
+</script>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
--- a/docs/rules/require-typed-ref.md
+++ b/docs/rules/require-typed-ref.md
@@ -39,16 +39,7 @@ const count: Ref<number | undefined> = ref()
 
 ## :wrench: Options
 
-```json
-{
-  "vue/require-typed-ref": ["error", {
-    "requireExplicitType": false
-  }]
-}
-```
-
-- `requireExplicitType` ... If `true`, requires `ref` and `shallowRef` functions to have type specified, even if there is a argument. Requires to write `const count = ref<number>(0)`.
-    default `false`
+Nothing.
 
 ## :mag: Implementation
 
--- a/docs/rules/require-typed-ref.md
+++ b/docs/rules/require-typed-ref.md
@@ -2,7 +2,7 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/require-typed-ref
-description: xxx
+description: require `ref` and `shallowRef` functions to be strongly typed
 ---
 # vue/require-typed-ref
 
@@ -28,7 +28,7 @@ count.value = '50' // Should be a type error, but it is not
 
 const count = shallowRef()
 
-/* ✗ GOOD */
+/* ✓ GOOD */
 const count = ref<number>()
 const count = ref(0)
 const count: Ref<number | undefined> = ref()
@@ -41,3 +41,7 @@ const count: Ref<number | undefined> = ref()
 
 Nothing.
 
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/require-typed-ref.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/require-typed-ref.js)
--- a/docs/rules/require-typed-ref.md
+++ b/docs/rules/require-typed-ref.md
@@ -47,6 +47,9 @@ const count: Ref<number | undefined> = ref()
 }
 ```
 
+- `requireExplicitType` ... If `true`, requires `ref` and `shallowRef` functions to have type specified, even if there is a argument. Requires to write `const count = ref<number>(0)`.
+    default `false`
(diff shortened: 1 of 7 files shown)


docs/rules/require-typed-ref.md
```diff
@@ -0,0 +1,53 @@
...
+
+> require `ref` and `shallowRef` functions to be strongly typed
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule disallows calling `ref()` or `shallowRef()` functions without generic type parameter or an argument when using TypeScript.
+
+With TypeScript it is easy to prevent usage of `any` by using `no-implicit-any`. Unfortunately this rule is easily bypassed with Vue `ref()` function. Calling `ref()` function without a generic parameter or an initial value leads to ref having `Ref<any>` type.
```
FloEdelmann, line 17:
```suggestion
With TypeScript it is easy to prevent usage of `any` by using [`noImplicitAny`](https://www.typescriptlang.org/tsconfig#noImplicitAny). Unfortunately this rule is easily bypassed with Vue `ref()` function. Calling `ref()` function without a generic parameter or an initial value leads to ref having `Ref<any>` type.
```

lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+    ]
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    let requireExplicitType = false
+    const option = context.options[0]
+    if (option) {
+      requireExplicitType = option.requireExplicitType
+    }
```
FloEdelmann, line 61:
```suggestion
    const { requireExplicitType = false } = context.options[0] ?? {}
```

docs/rules/require-typed-ref.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+```json
+{
+  "vue/require-typed-ref": ["error", {
+    "requireExplicitType": true
```
FloEdelmann, line 45: Please add an explanation for this option. Also, the default value is `false`.

```suggestion
    "requireExplicitType": false
```

tests/lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,191 @@
...
+        </script>
+      `
+    },
+    {
+      filename: 'test.js',
+      code: `
+        import { ref } from 'vue'
+        const count = ref()
+      `
+    }
```
FloEdelmann, line 80: Please also add tests (and fix the rule if necessary) for the Options API usage (`defineComponent` with a `setup` function).

FloEdelmann (changes requested):
Looks good overall! Please see a few remarks below.

tests/lib/rules/require-typed-ref.js
Demivan: Done and fixed other comments

tests/lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,264 @@
...
+      code: `
+      import { ref, defineComponent } from 'vue'
+      defineComponent({
+        setup() {
+          const count = ref<number>()
+          // ...
+          return {
+            count,
+            // ...
+          }
```
FloEdelmann, line 75: Let's keep the test code as clean and concise as possible:

```suggestion
          const count = ref<number>()
          return { count }
```

tests/lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,264 @@
...
+      code: `
+        import { ref, defineComponent } from 'vue'
+        defineComponent({
+          setup() {
+            const count = ref()
+            // ...
+            return {
+              count,
+              // ...
+            }
```
FloEdelmann, line 249:
```suggestion
            const count = ref()
            return { count }
```

FloEdelmann (approved):
Thanks, looks good to me now :slightly_smiling_face:

tests/lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,256 @@
...
+      errors: [
+        {
+          messageId: 'noType',
+          line: 3,
+          column: 23,
+          endLine: 3,
+          endColumn: 29
+        }
+      ]
+    },
```
ota-meshi, line 159: Do you think the following cases should be reported?

```js
    {
      filename: 'test.ts',
      code: `
        import { ref } from 'vue'
        const count = ref(0 as number | string)
      `,
      options: [{ requireExplicitType: true }],
      errors: ?
    },
```

Could you please add the test case with or without reporting?

ota-meshi (changes requested):
Thank you for this PR! Sorry for the late reply. I've been busy with office work lately.

tests/lib/rules/require-typed-ref.js
Demivan: I would report it. The idea behind `requireExplicitType` was to force usage of generic parameter everywhere.

@FloEdelmann, what do you think about it?  `requireExplicitType` was your idea :slightly_smiling_face:

FloEdelmann: Maybe let's even completely drop the "explicit" option for now, and we'll see later if it's needed and if it should be an option for this rule or another rule.

Demivan: @ota-meshi, @FloEdelmann Done

FloEdelmann (approved):

lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,106 @@
...
+            continue
+          }
+
+          if (ref.node.typeParameters == null) {
+            if (
+              ref.node.parent.type === 'VariableDeclarator' &&
+              ref.node.parent.id.type === 'Identifier'
+            ) {
+              if (ref.node.parent.id.typeAnnotation == null)
+                report(ref.name, ref.node)
```
FloEdelmann, line 97:
```suggestion
              if (ref.node.parent.id.typeAnnotation == null) {
                report(ref.name, ref.node)
              }
```

lib/rules/require-typed-ref.js
```diff
@@ -0,0 +1,106 @@
...
+function isNullOrUndefined(node) {
+  if (node.type === 'Literal' && node.value === null) {
+    return true
+  }
+
+  if (node.type === 'Identifier' && node.name === 'undefined') {
+    return true
+  }
+
+  return false
```
FloEdelmann, line 22:
```suggestion
  return (
    (node.type === 'Literal' && node.value === null) ||
    (node.type === 'Identifier' && node.name === 'undefined')
  )
```

FloEdelmann (commented):
Some code style suggestions (might violate with our Prettier config, so please check before applying).

ota-meshi (approved):
LGTM! Thank you!

PR #2208, opened 2023-06-08
Mark default options correctly in `vue/define-macros-order` docs

wenfangdu:


wenfangdu:
@ota-meshi Please unblock me, otherwise, I can't receive your messages.

docs/rules/define-macros-order.md
```diff
@@ -27,85 +27,85 @@ This rule reports the `defineProps` and `defineEmits` compiler macros when they
 
 - `order` (`string[]`) ... The order of defineEmits and defineProps macros. You can also add `"defineOptions"` and `"defineSlots"`.
 
-### `{ "order": ["defineProps", "defineEmits"] }` (default)
+### `{ "order": ["defineOptions", "defineProps", "defineEmits", "defineSlots"] }` (default)
```
FloEdelmann, line 30: Why is this order the default? Can you link to examples in the official Vue docs that use this order? Or some explanation?

FloEdelmann, line 30: Also, can you please add or adjust test cases for the new default order?

wenfangdu: @FloEdelmann Currently, the `eslint-plugin-vue` doc has two defaults, which seems to be done by mistake:
image

FloEdelmann: It is indeed a mistake in the docs that the second example is marked as "default". That should be fixed in any case, ideally in another PR to fix it without changing the default. Would you fancy doing that?

However, changing the default should only be done if it fits most use cases and/or is the official recommendation from the Vue docs. So can you link to any of those?

wenfangdu: Ah, I see. I thought the second one is meant to be the new default, so I adjusted the default order to reflect that, I've pushed a new commit to remove the false default order, please take a look.

FloEdelmann (approved):
Thanks!

PR #2210, opened 2023-06-09
Add `vue/no-restricted-component-names` rule

### LINKED ISSUE #2182: Rule Proposal: `vue/no-restricted-component-names`

Restrict some component names via a list of pattern.

[ ] Other (please specify:)

Options:

```json
{
  "vue/no-restricted-component-names": ["error", "Foo", "/^Baz(One|Two)/"]
}
```

```vue
<script>
<!-- ✗ BAD -->
export default {
  name: 'Foo'
}

<!-- ✗ BAD -->
export default {
  name: 'BazTwo'
}

<!-- ✓ GOOD-->
export default {
  name: 'Allow'
}
</script>
```

**Additional context**

Related https://github.com/vuejs/eslint-plugin-vue/issues/2180#issuecomment-1551298495.

ItMaga:
closes #2182 

---

I slightly changed the schema format that was proposed in the issue, as in [no-restricted-props](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-restricted-props.js) rule.

### DIFF
--- a/lib/rules/no-restricted-component-names.js
+++ b/lib/rules/no-restricted-component-names.js
@@ -5,8 +5,8 @@
 'use strict'
 
 const utils = require('../utils')
-const { isRegExp } = require('../utils/regexp')
-const { toRegExp } = require('../utils/regexp')
+const casing = require('../utils/casing')
+const { isRegExp, toRegExp } = require('../utils/regexp')
 
 /**
  * @typedef {object} OptionParsed
@@ -25,7 +25,7 @@ function buildMatcher(str) {
     const regex = toRegExp(str)
     return (s) => regex.test(s)
   }
-  return (s) => s === str
+  return (s) => s === casing.pascalCase(str) || s === casing.kebabCase(str)
 }
 
 /**
@@ -51,7 +51,9 @@ function parseOption(option) {
  * @private
  * */
 function createSuggest(property, suggest) {
-  if (!suggest) return []
+  if (!suggest) {
+    return []
+  }
 
   return [
     {
@@ -112,7 +114,9 @@ module.exports = {
       const propertyName = utils.getStaticPropertyName(property)
       if (propertyName === 'name' && property.value.type === 'Literal') {
         const componentName = property.value.value?.toString()
-        if (!componentName) return
+        if (!componentName) {
+          return
+        }
 
         for (const option of options) {
           if (option.test(componentName)) {
--- a/lib/rules/no-restricted-component-names.js
+++ b/lib/rules/no-restricted-component-names.js
@@ -97,7 +97,7 @@ module.exports = {
     messages: {
       // eslint-disable-next-line eslint-plugin/report-message-format
       disallow: '{{message}}',
-      suggest: 'Instead, change to `{{suggest}}`'
+      suggest: 'Instead, change to `{{suggest}}`.'
     }
   },
   /** @param {RuleContext} context */
--- a/lib/rules/no-restricted-component-names.js
+++ b/lib/rules/no-restricted-component-names.js
@@ -108,9 +108,12 @@ module.exports = {
     const options = context.options.map(parseOption)
 
     /**
-     * @param {Property | AssignmentProperty} property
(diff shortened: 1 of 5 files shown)


lib/rules/no-restricted-component-names.js
```diff
@@ -0,0 +1,153 @@
+/**
+ * @author ItMaga <https://github.com/ItMaga>
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+const utils = require('../utils')
+const { isRegExp } = require('../utils/regexp')
+const { toRegExp } = require('../utils/regexp')
```
FloEdelmann, line 9:
```suggestion
const { isRegExp, toRegExp } = require('../utils/regexp')
```

lib/rules/no-restricted-component-names.js
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * @param {Property | AssignmentProperty} property
+ * @param {string | undefined} suggest
+ * @returns {Rule.SuggestionReportDescriptor[]}
+ * @private
+ * */
+function createSuggest(property, suggest) {
+  if (!suggest) return []
```
FloEdelmann, line 54:
```suggestion
  if (!suggest) {
    return []
  }
```

lib/rules/no-restricted-component-names.js
```diff
@@ -0,0 +1,153 @@
...
+    const options = context.options.map(parseOption)
+
+    /**
+     * @param {Property | AssignmentProperty} property
+     */
+    function verify(property) {
+      const propertyName = utils.getStaticPropertyName(property)
+      if (propertyName === 'name' && property.value.type === 'Literal') {
+        const componentName = property.value.value?.toString()
+        if (!componentName) return
```
FloEdelmann, line 115:
```suggestion
        if (!componentName) {
          return
        }
```

lib/rules/no-restricted-component-names.js
```diff
@@ -0,0 +1,153 @@
...
+    /** @type {OptionParsed[]} */
+    const options = context.options.map(parseOption)
+
+    /**
+     * @param {Property | AssignmentProperty} property
+     */
+    function verify(property) {
+      const propertyName = utils.getStaticPropertyName(property)
+      if (propertyName === 'name' && property.value.type === 'Literal') {
+        const componentName = property.value.value?.toString()
```
FloEdelmann, line 114: I think that conversion between kebab-case and PascalCase should also be supported, i.e. `DisallowedComponent` should also report for `disallowed-component` and vice-versa. Please also add tests for that.

FloEdelmann (changes requested):
Thanks for the contribution! I have some comments, see below.

lib/rules/no-restricted-component-names.js
ItMaga: Thank you! Resolved

FloEdelmann (approved):
Looks good to me, thanks! :slightly_smiling_face:

lib/rules/no-restricted-component-names.js
```diff
@@ -0,0 +1,157 @@
...
+              suggest: createSuggest(property, option.suggest)
+            })
+          }
+        }
+      }
+    }
+
+    return utils.compositingVisitors(
+      utils.defineVueVisitor(context, {
+        Property(node) {
```
ota-meshi, line 140: Could you change it to use the `onVueObjectEnter` listener and only check the properties of the Vue object?
The following code may give false positives.

```vue
<script>
export default {
  data () {
    return {
      name: 'Disallow'
    }
  }
}
</script>
```

ota-meshi (changes requested):

lib/rules/no-restricted-component-names.js
ItMaga: Thank you! Done.

ota-meshi (approved):
LGTM! Thank you!

PR #2221, opened 2023-06-19
Restrict the `vue/no-console` rule to the `<template>` block

### LINKED ISSUE #2220: [Bug] `vue/no-console` affects code in `<script>`
`vue/no-console` is supposed to only check code in the `<template>`.

Repro: https://github.com/wenfangdu/vite-vue-3/tree/vue/no-console
image

ItMaga:
fixes #2220

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -382,6 +382,7 @@ module.exports = {
    * @param {boolean} [options.skipDynamicArguments] If `true`, skip validation within dynamic arguments.
    * @param {boolean} [options.skipDynamicArgumentsReport] If `true`, skip report within dynamic arguments.
    * @param {boolean} [options.applyDocument] If `true`, apply check to document fragment.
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):
Looks good to me :)

lib/utils/index.js
```diff
@@ -418,6 +419,7 @@ module.exports = {
       categories,
       skipDynamicArguments,
       skipDynamicArgumentsReport,
+      skipCoreHandlers,
```
ota-meshi, line 422: Hmm. In my opinion we should replace `context.getSourceCode().getScope()` with `wrapContextToOverrideTokenMethods()` rather than skip core's handler.
But I think we can merge this once to work around the current bug. I think we'll have to refactor after that.

ota-meshi (approved):
Thank you!

PR #2222, opened 2023-06-22
Rename `vue/component-tags-order` to `vue/block-order`

### LINKED ISSUE #2214: Rename `vue/component-tags-order` to `vue/block-order`
We use the word "block" in rule names to refer to the top-level tags in Vue SFCs:

* [`vue/block-lang`](https://eslint.vuejs.org/rules/block-lang.html)
* [`vue/block-tag-newline`](https://eslint.vuejs.org/rules/block-tag-newline.html)
* [`vue/no-empty-component-block`](https://eslint.vuejs.org/rules/no-empty-component-block.html)
* [`vue/no-restricted-block`](https://eslint.vuejs.org/rules/no-restricted-block.html)
* [`vue/no-restricted-block`](https://eslint.vuejs.org/rules/padding-line-between-blocks.html)

However, the [`vue/component-tags-order`](https://eslint.vuejs.org/rules/component-tags-order.html) rule does not fit that schema. I suggest to rename the rule to `vue/block-order`:

1. In the next minor version, add `vue/block-order` as a copy of `vue/component-tags-order` and [deprecate](https://eslint.vuejs.org/rules/#deprecated) the latter.
2. In the next major version, remove `vue/component-tags-order` completely.

---

Are there other rules for which it would make sense to rename them for consistency and/or clarity?

FloEdelmann:
Fixes #2214.


ota-meshi (approved):
Sorry for the late reply. I am busy with office work.
The pull request change looks good to me! However, it seems to change the configuration, so I will merge it when we start working on v10.

FloEdelmann:
It should not be a breaking change though, I only renamed the rule and followed the deprecation procedure. The breaking change would be to remove the old rule completely.

ota-meshi:
In my opinion adding a rule to a shared configuration is marked as a breaking change. (It's not a breaking change if we reduce the rules)

https://github.com/eslint/eslint#semantic-versioning-policy:~:text=eslint%3Arecommended%20is%20updated%20and%20may%20result%20in%20new%20linting%20errors%20(e.g.%2C%20rule%20additions%2C%20most%20rule%20option%20updates)

Users using shared configuration and turning off the `vue/component-tags-order` rule will receive the new warning unless they turn off the `vue/block-order` rule.

FloEdelmann:
Ok, should I then add the (deprecated) `vue/component-tags-order` rule back to the `recommended` configs and remove `vue/block-order` from the configs for now? We should then change it when we remove the `vue/component-tags-order` rule completely.

ota-meshi:
> Ok, should I then add the (deprecated) `vue/component-tags-order` rule back to the recommended configs and remove `vue/block-order` from the configs for now?

Yes. I think we need to do that to release this PR with v9 👍

ota-meshi (approved):
LGTM! Thank you!

PR #2223, opened 2023-06-22
Ignore blocks in `vue/html-self-closing`

### LINKED ISSUE #2202: Ignore `<template>` in `vue/html-self-closing` rule
`vue/html-self-closing`

N/A

Either by default or add an option to ignore `<template>` the same way `<style>` and `<script>` tags are ignored. Empty `<template>` tags don't make sense and should either be fully-removed from the Vue template by the engineer or left alone (if that's the way the engineer prefers it).

```vue
<CustomComponent>
    <template #slotName></template>
</CustomComponent>
```

```vue
<template></template>
```

```vue
<CustomComponent>
    <template #slotName />
</CustomComponent>
```

```vue
<template />
```

It would leave the code unaffected.

**Additional context**
N/A

FloEdelmann:
Fixes #2202.

### DIFF
--- a/lib/rules/html-self-closing.js
+++ b/lib/rules/html-self-closing.js
@@ -131,7 +131,7 @@ module.exports = {
       context,
       {
         VElement(node) {
-          if (hasInvalidEOF) {
+          if (hasInvalidEOF || node.parent.type === 'VDocumentFragment') {
             return
           }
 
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

PR #2224, opened 2023-06-22
Add new `vue/no-use-v-else-with-v-for`

### LINKED ISSUE #1939: Rule suggestion: `vue/no-use-v-else-with-v-for`
~[`vue/no-use-v-if-with-v-for`](https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html)~ New rule `vue/no-use-v-else-with-v-for`

More

Not sure, but probably default behavior.

```vue
<!-- GOOD, this is caught be the vue/no-use-v-if-with-v-for rule already -->
<div v-if="foo" v-for="x in xs">{{ x }}</div>
```

```vue
<!-- BAD -->
<div v-if="foo">foo</div>
<div v-else v-for="x in xs">{{ x }}</div>

<!-- GOOD -->
<div v-if="foo">foo</div>
<template v-else>
  <div v-for="x in xs">{{ x }}</div>
</template>
```

```vue
<!-- BAD -->
<div v-if="foo">foo</div>
<div v-else-if="bar" v-for="x in xs">{{ x }}</div>

<!-- GOOD -->
<div v-if="foo">foo</div>
<template v-else-if="bar">
  <div v-for="x in xs">{{ x }}</div>
</template>
```

Only report the first case (`v-if`+`v-for`).

Also warn about the other cases (`v-else`+`v-for` and `v-else-if`+`v-for`), and maybe even autofix to add a wrapper `<template>` tag.

FloEdelmann:
Fixes #1939.

### DIFF
--- a/lib/rules/no-use-v-else-with-v-for.js
+++ b/lib/rules/no-use-v-else-with-v-for.js
@@ -1,0 +1,49 @@
+/**
+ * @author Yosuke Ota
+ *
+ * Style guide: https://vuejs.org/style-guide/rules-essential.html#avoid-v-if-with-v-for
+ */
+'use strict'
+
+const { defineTemplateBodyVisitor, hasDirective } = require('../utils')
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'disallow using `v-else-if`/`v-else` on the same element as `v-for`',
+      categories: null,
+      url: 'https://eslint.vuejs.org/rules/no-use-v-else-with-v-for.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      unexpectedDirectiveWithVFor:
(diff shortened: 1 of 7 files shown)


lib/rules/no-use-v-else-with-v-for.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @author Yosuke Ota
+ *
+ * Style guide: https://vuejs.org/style-guide/rules-essential.html#avoid-v-if-with-v-for
+ */
```
ota-meshi, line 5:
```suggestion
```

tests/lib/rules/no-use-v-else-with-v-for.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * @author Yosuke Ota
+ */
```
ota-meshi, line 3:
```suggestion
```

ota-meshi (commented):
Thank you for this PR!
Almost looks good to me, but some unnecessary comments should be removed.

PR #2226, opened 2023-06-29
Lint this repo using ESLint Flat Config

nzakas:
This PR switches the repo to use flat config for ESLint. The goal of this exercise is to test out flat config on popular ESLint plugins to make sure things will work. To do that, I:

- Added a `eslint.config.js` file
- Deleted `.eslintrc.js` and `.eslintignore`
- Modified existing files that referenced local rules
- Updated the `tools/update` scripts to use flat config

I did not make any changes to the plugin itself, this is just the linting for this repo that was updated. 

I've tested this to make sure that it outputs the same results as with the eslintrc config system, but please let me know if I missed anything.


eslint.config.js
```diff
@@ -0,0 +1,221 @@
...
+      'coverage',
+      'node_modules',
+      'tests/fixtures',
+      'tests/integrations/eslint-plugin-import',
+
+      '!.vitepress',
+      'docs/.vitepress/dist',
+      'docs/.vitepress/build-system/shim/eslint.mjs',
+      'docs/.vitepress/build-system/shim/assert.mjs',
+      'docs/.vitepress/build-system/shim/path.mjs',
```
nzakas, line 23: This line was not in `.eslintrc.js` but I was getting linting errors in this file with flat config, so I added this here. I'm unsure why flat config was showing errors on this file while eslintrc was not, but please double-check that this should be excluded.

FloEdelmann (approved):
Looks good from my side, thank you!

What would be the next steps? `FlatCompat` and `eslint/use-at-your-own-risk` look like they need to be changed at some point :wink:

ota-meshi (approved):
Thank you very much!

nzakas:
@FloEdelmann yes, they'll need to be changed, but at least through ESLint v9 this code will continue to work, so there will be plenty of time to adjust as needed.

PR #2238, opened 2023-07-17
Add `vue/no-deprecated-model-definition` rule

### LINKED ISSUE #2236: Rule proposal: `vue/no-deprecated-model-definition`

In Vue 3, `v-model` is always a shortcut for `:modelValue="…" @update:modelValue="…"`. In Vue 2, those prop and event names could be customized with the [`model` definition](https://v2.vuejs.org/v2/api/#model). A new ESLint rule would help with finding those deprecated `model` definitions in the codebase while upgrading to Vue 3. See also https://v3-migration.vuejs.org/breaking-changes/v-model.html

[ ] Other (please specify:)

```ts
export default {
  model: {
    prop: 'list',
  }
}
```

```ts
  model: {
    prop: 'list',
    event: 'update'
  }
})
```

```ts
export default Vue.extend({
  model: {
    event: 'update'
  }
})
```

**Additional context**
An additional `allowVue3Compat` boolean option (`false` by default) could help with transitioning to Vue 3:

When enabled, the rule would not report `model` definitions that match the Vue 3 usage:

```ts
  model: {
    prop: 'foo-bar',
    event: 'update:foo-bar'
  }
})
```

After the Vue 3 migration, the option could be disabled again, the `model` definition could be deleted, and the `v-model:foo="…"` shorthand syntax could then be used without further changes.

FloEdelmann:
Fixes #2236

### DIFF
--- a/docs/rules/index.md
+++ b/docs/rules/index.md
@@ -226,6 +226,7 @@ For example:
 | [vue/next-tick-style](./next-tick-style.md) | enforce Promise or callback style in `nextTick` | :wrench: | :hammer: |
 | [vue/no-bare-strings-in-template](./no-bare-strings-in-template.md) | disallow the use of bare strings in `<template>` |  | :hammer: |
 | [vue/no-boolean-default](./no-boolean-default.md) | disallow boolean defaults | :wrench: | :hammer: |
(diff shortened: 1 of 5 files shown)


ota-meshi (approved):
LGTM!

PR #2244, opened 2023-07-20
Improve the `vue/no-setup-props-destructure` rule

### LINKED ISSUE #1933: `vue/no-setup-props-destructure` should also warn when using `toRefs`
 vue/no-setup-props-destructure

More.

New default behavior.

```vue
const foo = toRefs(props.foo)
```

No error / warning.

Display error / warning.

**Additional context**
Looks like the pattern above makes the variable reactive - but it looses reactivity instead.

### LINKED ISSUE #2007: `vue/no-setup-props-destructure` false negatives in object or array initialization
**Checklist**

- **ESLint version:** 8.24.0
- **eslint-plugin-vue version:** 9.6.0
- **Node version:** 16.17.0
- **Operating System:** Kubuntu 22.04

```json5
{
  "rules": {
    "vue/no-setup-props-destructure": "error"
  }
}
```

**What did you do?**
In setup function:

```ts
const foo = props.foo          // ESLint correctly reports a `vue/no-setup-props-destructure` error
const bar = { foo: props.foo } // no `vue/no-setup-props-destructure` error
const baz = [props.foo]        // no `vue/no-setup-props-destructure` error
const complex = [{
  foo: [props.foo],            // no `vue/no-setup-props-destructure` error
}]
```

An error is thrown for all statements, because they all destructure the props object, causing them to lose reactivity.

**What actually happened?**
Only an error for the first statement.

None created yet, will do if required.

### LINKED ISSUE #2092: `vue/no-setup-props-destructure` should trigger when prop is passed to function
vue/no-setup-props-destructure
More.
Perhaps an option that defaults to false would be best for backwards compatibility, though arguably it's more correct for it to be the default behavior.

```vue
<script setup>
const props = defineProps(["foo"]);

function foo(bar) {};

foo(props.foo);
</script>
```

Nothing.
Flag it as a loss of reactivity. It should be `foo(toRef(props, "foo"))` or `foo(computed(() => props.foo))` instead.
**Additional context**

The rule already flags code like `const foo = props.foo`. Passing it to a function incurs the same loss of reactivity, so it seems consistent to flag that as well. Note also that vue/no-ref-object-destructure already flags something like `foo(bar.value)` and that seems like an analogous case.

davidmyersdev:
- closes #1933
- closes #2007
- closes #2092

### Summary

Previously, the rule did not catch some reactivity-breaking uses of `props`.

#### Examples

Accessing a prop in a function call.

```vue
<script setup>
  const props = defineProps({ count: Number })
  
  const count = ref(props.count)
</script>
```

Accessing a prop in an object.

```vue
<script setup>
  const props = defineProps({ count: Number })
  
  const counter = { count: props.count }
</script>
```

Accessing a prop in an array.

```vue
<script setup>
  const props = defineProps({ count: Number })
  
  const counts = [props.count]
</script>
```

Accessing a prop in a nested object structure.

```vue
<script setup>
  const props = defineProps({ count: Number })
  
  const counters = [{ counts: [props.count] }]
</script>
```

### DIFF
--- a/lib/rules/no-setup-props-destructure.js
+++ b/lib/rules/no-setup-props-destructure.js
@@ -23,7 +23,10 @@ module.exports = {
         'Getting a value from the `props` in root scope of `{{scopeName}}` will cause the value to lose reactivity.'
     }
   },
-  /** @param {RuleContext} context */
+  /**
+   * @param {RuleContext} context
+   * @returns {RuleListener}
+   **/
   create(context) {
     /**
      * @typedef {object} ScopePropsReferences
@@ -59,6 +62,30 @@ module.exports = {
       }
 
       const rightNode = utils.skipChainExpression(right)
+
+      if (rightNode.type === 'CallExpression') {
+        const propRefs = [...propsReferences.refs.values()]
+        const isPropsMemberAccessed = propRefs.some((props) => {
+          const isPropsInCallExpression = utils.inRange(rightNode.range, props)
+
+          if (isPropsInCallExpression) {
+            const isPropMemberExpression =
+              props.parent.type === 'MemberExpression' &&
+              props.parent.object === props
+
+            if (isPropMemberExpression) {
+              return true
+            }
+          }
+
+          return false
+        })
+
+        if (isPropsMemberAccessed) {
+          return report(left, 'getProperty', propsReferences.scopeName)
+        }
+      }
+
       if (
         left.type !== 'ArrayPattern' &&
         left.type !== 'ObjectPattern' &&
@@ -66,6 +93,7 @@ module.exports = {
       ) {
         return
       }
+
       /** @type {Expression | Super} */
       let rightId = rightNode
       while (rightId.type === 'MemberExpression') {
@@ -114,6 +142,11 @@ module.exports = {
       }
       const propsReferenceIds = new Set()
       for (const reference of variable.references) {
+        // If reference is in another scope, we can't check it.
+        if (reference.from !== context.getScope()) {
+          continue
+        }
+
         if (!reference.isRead()) {
           continue
         }
--- a/lib/rules/no-setup-props-destructure.js
+++ b/lib/rules/no-setup-props-destructure.js
@@ -35,9 +35,8 @@ module.exports = {
      */
     /** @type {Map<FunctionDeclaration | FunctionExpression | ArrowFunctionExpression | Program, ScopePropsReferences>} */
     const setupScopePropsReferenceIds = new Map()
-    const outerExpressionTypes = new Set([
+    const wrapperExpressionTypes = new Set([
       'ArrayExpression',
-      'CallExpression',
       'ObjectExpression'
     ])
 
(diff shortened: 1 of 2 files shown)


ota-meshi:
Thank you for this PR!

The example reported in #2092 still seems to have false negatives.

```vue
<script setup>
const props = defineProps(["foo"]);

function foo(bar) {};

foo(props.foo);
</script>
```

I think we probably need to add CallExpression listener to check besides VariableDeclarator and AssignmentExpression.

davidmyersdev:
Thanks, @ota-meshi! That makes sense to me. I should have time to update this within the next few days.

lib/rules/no-setup-props-destructure.js
```diff
@@ -75,6 +91,34 @@ module.exports = {
...
+        }
+
+        return false
+      })
+
+      if (isMemberAccessed) {
+        return true
+      }
+
+      return false
```
FloEdelmann, line 120:
```suggestion
      return propRefs.some((props) => {
        const isPropsInExpressionRange = utils.inRange(node.range, props)

        if (isPropsInExpressionRange) {
          const isPropsMemberExpression =
            props.parent.type === 'MemberExpression' &&
            props.parent.object === props

          if (isPropsMemberExpression) {
            return true
          }
        }

        return false
      })
```

davidmyersdev: Thanks for the suggestion. 💯 I cleaned it up a bit to a point where it's still readable but much more concise.

ota-meshi (approved):
LGTM! Thank you.

ddenev:
Can someone please explain why is this bad:
```vue
<script setup>
  const props = defineProps({ count: Number })
  
  const count = ref(props.count)
</script>
```

Vue docs use the above as perfectly fine example for initializing a local data property - https://vuejs.org/guide/components/props.html#one-way-data-flow (see 1.):
Quote: 'In this case, it's best to define a local data property that uses the prop as its initial value'.

After the introduction of this change my code stopped working because I'm using this type of local property initialization from props, which was fine for years. And yes, I care only for the initial props value, I do not want to maintain reactivity in the local property.

davidmyersdev:
I can't speak for the team, but I don't look at it from a perspective of good or bad. I look at this rule as a means to prevent reactivity-breaking code, which your code is technically doing. Just like many other rules, it sometimes makes sense to _intentionally_ break them. If you understand why you want to break reactivity (e.g. only using the value as an initial value), then it's absolutely fine to disable this rule for a given line or file. It's the case where someone does it accidentally that is an issue, and this rule previously did not catch those cases.

ddenev:
thanks for the reply, @davidmyersdev 

I still think that the updated rule is too restrictive. Here are my arguments:

- it is perfectly acceptable (again, Vue docs recommend this) to use `props.aProp` for initializing a local state
- the rule name contains 'destructure', indicating it was originally taking care of destructuring props, and I am not destructuring props in my code (`const count = ref(props.count)` is NOT a destructuring)
- since I am aware that I don't want reactivity when initializing local state, I don't want to add loads of 'disable the rule' comments to my code

IMHO, the rule should take care only about destructuring props and not every case where props are used. Currently, it is too intrusive - it forces us to disable it every time we consciously don't want reactivity.

Please reconsider.

Thank you.

ddenev:
For reference, from Vue documentation:
image

richardsimko:
I agree with @ddenev. This rule now does way more than the name and docs implies. I was confused as to why we started getting lint failures all of a sudden and looked up the docs for this rule, it made no sense since no destructuring was made in the code that generated the error. Only after I found this issue did I get an explanation as to why the code failed.

If there is a need for this rule (Reactivity breaking prop access) please make it a different rule or rename the existing rule.

And in the future when making breaking changes to rules, please consider respecting semver and releasing it as a major version.

l3d00m:
I've opened #2259 since I strongly agree with @ddenev and @richardsimko

FloEdelmann:
Let's please discuss this in #2259, I'll lock the discussion here now.

PR #2248, opened 2023-07-24
fix(require-explicit-emits): Add support for kebab-cased custom events

Robo-Rin:
The existing [`vue/require-explicit-emits`](https://eslint.vuejs.org/rules/require-explicit-emits.html) rule does not support the scenario where a custom event is emitted in kebab-case. This usage was recommended as a workaround by core Vue members here: https://github.com/vuejs/core/issues/5220#issuecomment-1007488240 .

Before this PR with `allowProps: true`, this works but is ugly and violates [`vue/prop-name-casing`](https://eslint.vuejs.org/rules/prop-name-casing.html):
```js
props: ['onCustom-event'],
// ...
this.$emit('custom-event')
```

After this PR you will be able to do:
```js
props: ['onCustomEvent'],
// ...
this.$emit('custom-event')
```

### DIFF
--- a/docs/rules/require-explicit-emits.md
+++ b/docs/rules/require-explicit-emits.md
@@ -94,11 +94,12 @@ export default {
 ```vue
 <script>
 export default {
-  props: ['onGood', 'bad'],
+  props: ['onGood', 'onCustomEvent', 'bad'],
   methods: {
     foo () {
       // ✓ GOOD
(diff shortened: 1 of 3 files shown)


Robo-Rin:
Having some issues with the recommended strategy from the Vue issue I linked, will re-open if I get it working.

PR #2256, opened 2023-07-29
Fix incompatibility with typescript-eslint v6 in `vue/script-indent` rule

### LINKED ISSUE #2227: FYI: typescript-eslint@v6 will remove parserServices.hasFullTypeInformation
👋 Hi! I'm one of the maintainers of [typescript-eslint](https://typescript-eslint.io). We're getting ready to release a new v6 major version of the tooling, which includes some small breaking changes.

Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/7124: I see this repo contains references to `hasFullTypeInformation`. That property is something we've never recommended using (see that issue & https://github.com/typescript-eslint/typescript-eslint/issues/7158). We recommend instead using `ESLintUtils.getParserServices` per https://typescript-eslint.io/developers/custom-rules#typed-rules.

If you _really_ want to "progressively enhance" lint rules, `parserServices.program` will be `null` if type information isn't available. But we definitely recommend against that kind of rule behavior changing.

Overall, I'd suggest trying out typescript-eslint@v6 and making sure internal builds & downstream repos still compile & lint nicely with all the changes. Let me know if there's anything we can do to help with the v6 upgrade! 💖

ota-meshi:
I updated the typescript-eslint we use in our tests to v6 and found a bug in the script-indent rule, so I fix it.

This PR also updates the typescript-eslint we use in our tests to v6 and adds a workflow to test using typescript-eslint@v5.

I think we can close #2227 as this PR adds compatibility testing with typescript-eslint v6.

close #2227


tests/lib/rules/no-required-prop-with-default.js
```diff
@@ -508,47 +508,6 @@ tester.run('no-required-prop-with-default', rule, {
         }
       ]
     },
-    {
-      filename: 'test.vue',
-      code: `
-        <script setup lang="ts">
-          interface TestPropType {
-            readonly foo(): void
```
ota-meshi, line 516: This statement can no longer be tested because it causes a parsing error.

tests/fixtures/script-indent/ts-class-declaration-04.vue
```diff
@@ -5,6 +5,11 @@ class
     <
       T
     >
+  extends
```
ota-meshi, line 8: I swapped the order to avoid parsing errors.

.github/workflows/CI.yml
```diff
@@ -28,7 +26,27 @@ jobs:
...
+      run: npm install --legacy-peer-deps
+    - name: Test
+      run: npm test
+
+  test-for-ts-eslint-v5:
+    name: Test
+    strategy:
+      matrix:
+        node: [18]
         os: [ubuntu-latest]
```
FloEdelmann, line 50: A matrix is not really needed here if it's only one combination.

ota-meshi: Thank you for checking this PR. I fixed that.

FloEdelmann (approved):
LGTM :slightly_smiling_face:

PR #2273, opened 2023-08-11
Fix no-ref-as-operand and allow lint when ref as right operator

### LINKED ISSUE #2271: vue/no-ref-as-operand only lint left identifier of a logical expression
**Checklist**

- [ ] I have tried restarting my IDE and the issue persists.
- [ ] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

- **ESLint version:** v8.27.0
- **eslint-plugin-vue version:** v9.7.0
- **Node version:** v16.17.0
- **Operating System:** MacOS

```json5
module.exports = {
  env: {
    browser: true,
    node: true,
    es6: true,
  },
  extends: [
    'plugin:vue/recommended',
    'eslint:recommended',
    '@vue/eslint-config-typescript/recommended',
  ],
  settings: {
    'import/resolver': {
      typescript: { directory: 'tsconfig.json' },
      alias: {
        extensions: ['.js', '.jsx', '.ts', '.tsx', '.vue'],
        map: [
          // for more custom alias config not included in tsconfig
        ],
      },
    },
  },
  plugins: [
    'import'
  ],
}
```

**What did you do?**	

```vue
const myRef = ref(0);
const a = true;
if (a && myRef) {
  // do something
}
```

myRef will throw `vue/no-ref-as-operand` error, `vue/no-ref-as-operand` only lint left identifier of a logicalExpression

**What actually happened?**

eslint don't throw `vue/no-ref-as-operand` error, and 

lakb248:
fix #2271

### DIFF
--- a/lib/rules/no-ref-as-operand.js
+++ b/lib/rules/no-ref-as-operand.js
@@ -104,9 +104,6 @@ module.exports = {
       // refValue || other, refValue && other. ignore: other || refValue
       /** @param {Identifier & {parent: LogicalExpression}} node */
       'LogicalExpression>Identifier'(node) {
-        if (node.parent.left !== node) {
-          return
-        }
         // Report only constants.
         const data = refReferences.get(node)
         if (
--- a/tests/lib/rules/no-ref-as-operand.js
+++ b/tests/lib/rules/no-ref-as-operand.js
@@ -68,6 +68,8 @@ tester.run('no-ref-as-operand', rule, {
     1 - count.value
     count.value || other
     count.value && other
+    other && count.value
+    other || count.value
     var foo = count.value ? x : y
     `,
     `
@@ -78,8 +80,6 @@ tester.run('no-ref-as-operand', rule, {
     `
     import { ref } from 'vue'
     const foo = ref(true)
-    var a = other || foo // ignore
-    var b = other && foo // ignore
 
     let bar = ref(true)
     var a = bar || other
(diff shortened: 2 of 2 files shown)


ota-meshi:
Thanks for this PR. 
However, I think that this PR change has caused false positives in the following usages.

```vue
<script setup lang="ts">
import { ref, Ref } from 'vue'
const refValue1: Ref<number>  = ref(0)
const maybeRefValue: Ref<number> | null = getMaybeRef()
const refValue2 = maybeRefValue || refValue1
console.log(refValue2.value)
</script>
```

I think we should keep track of how the result of the expression is used.
Also, I think the ternary operator could be improved if it would be possible to track the usage of the result of the expression.

ota-meshi (changes requested):
https://github.com/vuejs/eslint-plugin-vue/pull/2273#issuecomment-1676517901

grindpride:
@ota-meshi @lakb248 
Maybe put the logic for this rule in a separate `no-ref-as-operand-strict` rule, or customize with a parameter (`strictCheck`) of the current one?
In my code there are no cases of the following type
```ts
const refValue1: Ref<number>  = ref(0)
const maybeRefValue: Ref<number> | null = getMaybeRef()
const refValue2 = maybeRefValue || refValue1
```
But errors with such case use are often encountered

```ts
const refValue1: Ref<number>  = ref(0)
if (refValue1) {
  //incorrect condition
}
```

FloEdelmann:
It's been nearly two years now without any activity, so I'm going to close this. Feel free to open a new PR!

PR #2275, opened 2023-08-15
Insert `defineOptions` after import declarations in `vue/prefer-define-options`

### LINKED ISSUE #2263: `prefer-define-options` should insert `defineOptions` after import expressions

Imporve autofix.

Same.

N/A

```vue
<script setup>
import {ref} from 'vue'
</script>

<script>
export default {inheritAttrs: true}
</script>
```

Currently fix to:

```vue
<script setup>
defineOptions({inheritAttrs: true})

import {ref} from 'vue'
</script>
```

Expect fix to:

```vue
<script setup>
import {ref} from 'vue'

defineOptions({inheritAttrs: true})
</script>
```

**Additional context**

[Playground](https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJx1lMFu2zAMhl8l8GWXCgV2DIoBe49eaJm2ucqSRklNvCLvvt92bcdZd0r0kabIn6Q+Khsars7VS7IqMZ8S5xJ/vHoZYtB8+lBub6dWw3D69l7426t/eV484YPD9p+vs3/DLRWH78T3rJJ/5qzpfMpa+Hb4tnqqtDhO1fkDKfgED5uDmlQia3X+/lS1ODWibLMEP5OOc2Y1ihR1IT4YSqO3JiJDSWz4yrYgzmq1BNiwdaQ0xUmbwVFK+DhJt4WyYYikbDx35g9r2Llv/nFFyl8x8nn2ly3rxZI1OGTe8XWFDdel61j3s+PM5p12Ai0MaZcOYMl84KFmPVrYoVhpD+yNx3snJ5ZwySTLSnmIeTwcjO0hl520ni87GiNNXdhq4+uDDHzNSqYOwTH56aZ8tCQeZCUtOZd7DaXrN1TQzmPEzoWa3ANcBvQReo+kv+q3+Hdy0iwtiBtVnAuczaWXzCmS3XRxATKHFqPFVtJdNzFoKK0R3/1PqEGu3Jg5GmYMw5Op3oyeLyaNQx3cRoIPNqNCJC4DfjlZilsiof4FGZ3bIszO6wGTn0Me0eq6iMuyl6y8CME7QO2faa0ssWsfRExfrVnqqQkXQGyqoOLGeBruwmBx0jSsSvu85V6SqRmLzHd7DUPxeCf2AxrCc8gBT4c48bzblAki1+4OJWrZtOIhyfhAsbl0v3ifOMQJQliEEo/G7faScK+jmnd5PyEWcUeQiTEONdk3vIis7PdBWY3YLNs/wmMrL5IXD+XfBW+bGYVdM5My7W7ytOS+DOvU1rAIhcf3Oc5XY0pQ/FrUlOLt9heqG/0r)

ItMaga:
closes #2263

### DIFF
--- a/lib/rules/prefer-define-options.js
+++ b/lib/rules/prefer-define-options.js
@@ -34,13 +34,13 @@ module.exports = {
     let defineOptionsNode = null
     /** @type {ExportDefaultDeclaration | null} */
     let exportDefaultDeclaration = null
-    /** @type {Set<ImportDeclaration>} */
-    const importsInScriptSetup = new Set()
+    /** @type {ImportDeclaration|null} */
+    let lastImportDeclaration = null
 
     return utils.compositingVisitors(
       utils.defineScriptSetupVisitor(context, {
         ImportDeclaration(node) {
-          importsInScriptSetup.add(node)
+          lastImportDeclaration = node
         },
         onDefineOptionsEnter(node) {
           defineOptionsNode = node
@@ -115,14 +115,7 @@ module.exports = {
         }
 
         /** @type {VStartTag | ImportDeclaration} */
-        let insertAfterTag = scriptSetup.startTag
-
-        if (importsInScriptSetup.size > 0) {
-          const lastImportDeclaration = [...importsInScriptSetup].pop()
-          if (lastImportDeclaration) {
-            insertAfterTag = lastImportDeclaration
-          }
-        }
+        let insertAfterTag = lastImportDeclaration || scriptSetup.startTag
 
         return [
           fixer.removeRange(removeRange),
--- a/lib/rules/prefer-define-options.js
+++ b/lib/rules/prefer-define-options.js
@@ -117,7 +117,7 @@ module.exports = {
         /** @type {VStartTag | ImportDeclaration} */
         let insertAfterTag = scriptSetup.startTag
 
-        if (importsInScriptSetup.size) {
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):
Looks good to me, thanks for your contribution!

lib/rules/prefer-define-options.js
```diff
@@ -34,9 +34,14 @@ module.exports = {
     let defineOptionsNode = null
     /** @type {ExportDefaultDeclaration | null} */
     let exportDefaultDeclaration = null
+    /** @type {Set<ImportDeclaration>} */
+    const importsInScriptSetup = new Set()
```
ota-meshi, line 38: I think we can change it to hold one element instead of `Set` because the same node will never be visited twice and only the last visited node will be used. The logic around L117 also needs to be changed.

```suggestion
    /** @type {ImportDeclaration|null} */
    let lastImportDeclaration = null
```

```js
// L42
        ImportDeclaration(node) {
          lastImportDeclaration = node
        },

// L117
        /** @type {VStartTag | ImportDeclaration} */
        const insertAfterTag = lastImportDeclaration || scriptSetup.startTag
```

ota-meshi (changes requested):
Sorry for the late review. Thank you for this PR.

ota-meshi (approved):
LGTM! Thank you!

PR #2280, opened 2023-08-28
Require key for conditionally rendered repeated components

### LINKED ISSUE #255: Rule proposal: `v-if-else-key`

felipemelendez:
This PR resolves #255 

When Vue updates the DOM, it employs a diffing algorithm to enhance efficiency. If elements of the same type are added or removed, Vue attempts to optimize the process by recycling the previous element for the new one. This recycling strategy is an integral part of Vue's optimization efforts to minimize rendering overhead.

However, this clever recycling strategy can sometimes lead to unexpected issues, primarily because it may not properly update the components. By assigning unique keys to each repeated component within the scope of the component, Vue can quickly identify which items have changed, been added, or been removed.

This lint rule is designed to guard against such recycling issues by requiring a unique key attribute on repeated components that are part of a conditional family (v-if, v-else-if, v-else) within the same scope level. The rule not only checks for the presence of a key attribute but also ensures that repeated components within these conditional families have unique keys.

If a component is found to violate this rule, the lint tool will suggest a fix by proposing a unique key based on the component's name and count (e.g., "some-component-2").

By recognizing the unique key attribute, Vue will abandon its recycling strategy and perform a full removal and addition of the new element, thereby ensuring that the component is rendered correctly.


felipemelendez:
Hi, this PR is now ready for review. As the person who previously attempted to address this issue very well said, "Better late than never" :grinning:

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
```diff
@@ -0,0 +1,371 @@
...
+const reservedNamesInVue2 = require('../utils/vue2-builtin-components')
+
+// =============================================================================
+// Constants and Variables
+// =============================================================================
+
+/**
+ * Curated list of standard HTML elements that serve as containers.
+ * We're not using an existing library util because we are only concerned about HTML elements that
+ * have the capacity to wrap other components.
```
FloEdelmann, line 22: Why? Shouldn't this also require a key with this rule?

```html
<span v-if="foo">Foo <strong>foo</strong></span>
<span v-else>Bar <strong>bar</strong></span>
```

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
```diff
@@ -0,0 +1,371 @@
...
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'require key attribute for conditionally rendered repeated components',
+      categories: null,
+      recommended: false,
+      url: 'https://eslint.vuejs.org/rules/require-key-for-conditionally-rendered-repeated-components.html'
```
FloEdelmann, line 203: The rule name is *really* long, even when compared to our longest rule name `vue/no-spaces-around-equal-signs-in-attribute`. Why not `v-if-else-key` like suggested in the issue?

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
felipemelendez: Hi! Thank you for taking the time to review my PR. I appreciate your attention to detail.

The key attribute becomes crucial when we deal with components or elements that hold internal state or have side-effects. Providing a unique key for each condition ensures that Vue treats each instance separately, preventing unintended state persistence or side-effects between condition changes. However, in straightforward cases like span elements, which don't have internal state or side-effects, the key attribute is not necessary for correct rendering behavior.

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
felipemelendez: ahh, naming can be difficult and sometimes I try too hard, I will change it to the suggested name

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
FloEdelmann: But shouldn't we restrict the rule to actual components then and not bother with `div`s etc. at all?

felipemelendez: yep, great point, the rule does not target standard HTML tags. It specifically targets custom Vue components that are used within conditional rendering contexts.
The rule defines a list of native HTML tags and reserved Vue component names as container elements, and there is a function `isContainerElement` that checks if a given node (representing a Vue component or element) is a container element. If the current node represents a container element, the function returns early, effectively skipping any further processing for that node. This ensures that the rule's logic for checking and enforcing the key attribute is only applied to custom Vue components, not native HTML tags or reserved Vue names.
I created a [test suite for this rule](https://gist.github.com/felipemelendez/981d58a4ae38782323b0df5d820ecfea), perhaps that will bring further clarity as to what is expected of the rule.

felipemelendez: ahh, oops, good catch, I created this rule for our app at work and it appears that I changed it so that it targets divs, it's been a while, and I can't remember why I did that but you are absolutely correct, my bad, I will fix it as soon as I get a chance. Thank you for making that observation

felipemelendez: ahh, that is a great observation, good catch, I will make that update, and refactor tests accordingly

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
felipemelendez: [c4871c6](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/c4871c603e145253ffb00b4b2f7ecc887fa99da1)

lib/rules/require-key-for-conditionally-rendered-repeated-components.js
felipemelendez: [0daeb91](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/0daeb91f54d1781a733f3de87f615e3f575b2cd0)

docs/rules/index.md
```diff
@@ -267,6 +267,7 @@ For example:
 | [vue/require-direct-export](./require-direct-export.md) | require the component to be directly exported |  | :hammer: |
 | [vue/require-emit-validator](./require-emit-validator.md) | require type definitions in emits | :bulb: | :hammer: |
 | [vue/require-expose](./require-expose.md) | require declare public properties using `expose` | :bulb: | :hammer: |
+| [vue/require-key-for-conditionally-rendered-repeated-components](./require-key-for-conditionally-rendered-repeated-components.md) | require key attribute for conditionally rendered repeated components | :wrench: | :warning: |
```
FloEdelmann, line 270: Please run `npm run update` to update the rule name everywhere.

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,378 @@
...
+// =============================================================================
+
+/**
+ * Curated list of standard HTML elements that serve as containers.
+ * We're not using an existing library util because we are only concerned about HTML elements that
+ * have the capacity to wrap other components.
+ *
+ * @type {Array<string>}
+ */
+const htmlContainerElements = [
```
FloEdelmann, line 26: Please remove this list (and everything that belongs to it); it should be unnecessary now with the `isCustomComponent` check.

FloEdelmann (changes requested):

lib/rules/v-if-else-key.js
felipemelendez: hmm, I don't think we can do that, the rule maintains a component usage stack (componentUsageStack), which is a stack of maps where each map represents a scope level, tracking the usage of custom Vue components at that level. When the rule encounters a container element (determined by isContainerElement), it pushes a new map onto the componentUsageStack, essentially entering a new scope level. This is done because container elements can have their own scope of children, and the rule needs to track component usage separately for each scope.

felipemelendez: This level of scope tracking enables us to discern that the following structure is acceptable:
```
<template>
    <div>
        <div v-if="some-condition" >
            <CustomComponent />
        </div>
        <CustomComponent v-else />
    </div>
</template>
```

And it allows us to identify that this structure, conversely, is problematic:
```
<template>
    <div>
        <CustomComponent v-if="someCondition" />
        <CustomComponent v-else-if="otherCondition" />
        <CustomComponent v-else />
    </div>
</template>
```

FloEdelmann: But again, it doesn't matter if the tag in your first example is a `<div>` or a `<span>`. The only tag that would need to be handled specially would be `<template>`.

felipemelendez: Sorry Flo, are we on the same page? I'm not sure what you mean (why are you singling out only template, component, and slot?). Perhaps I should not have used divs to illustrate my point?

I'm talking about differentiating between scope levels to ensure that we only add keys to conditionally rendered repeated components that are at the same scope level. 

Here is a more generic illustration:
```
<some-container>
    <ScopeOne v-if="conditionOne"> 
        <ScopeTwo v-if="conditionTwo" /> <--- this should have a key
        <ScopeTwo v-else /> <--- this should have a key
    </ScopeOne>
    <ScopeTwo v-else /> <--- this should not have a key
</some-container>
```

FloEdelmann: I agree, but what do you need the `htmlContainerElements` variable for then?

felipemelendez: ah, thank you for being patient, I was relying on memory and variable names to construct my mental model of the rules behavior, and both threw me off (several months have passed since I built this thing). Now I have done what I should have done from the beginning and studied my own code, and you are absolutely right, I have refactored. Once again thank you for your attention to detail. 
I have refactored and added a test to specifically ensure that what we have discussed here is covered by tests:
[c659339](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/c659339aca07875daedea6014c14d24d122be087)

[c89efd8](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/c89efd80cc10ff3d67a2570155141ea283bfb83b)

felipemelendez: by the way, I had already captured that idea in a previous test, but this new one makes the concept more quickly evident, I can remove it if you think it's unnecessary

FloEdelmann: Thanks for checking it again and adding a test!
This definitely looks easier now :slightly_smiling_face:

docs/rules/v-if-else-key.md
```diff
@@ -0,0 +1,49 @@
...
+</template>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :books: Further Reading
```
FloEdelmann, line 42: Can you add `Related rules` section linking to https://eslint.vuejs.org/rules/require-v-for-key.html (and vice-versa) please`

tests/lib/rules/require-key-for-conditionally-rendered-repeated-components.js
```diff
@@ -0,0 +1,313 @@
...
+const tester = new RuleTester({
+  parser: require.resolve('vue-eslint-parser'),
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module'
+  }
+})
+
+tester.run('v-if-else-key', rule, {
+  valid: [
```
FloEdelmann, line 19: Please add the following valid test case:

```html
<div v-if="foo" />
<span v-else />
```

FloEdelmann (changes requested):

tests/lib/rules/require-key-for-conditionally-rendered-repeated-components.js
felipemelendez: [d1b5039](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/d1b5039d8213b6a3e91159b53ad0f84b85cb6f13)

docs/rules/v-if-else-key.md
felipemelendez: [fdd59a5](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/fdd59a52b035d620455e82ccd38590495db64316)

docs/rules/v-if-else-key.md
```diff
@@ -4,11 +4,12 @@ sidebarDepth: 0
 title: vue/v-if-else-key
 description: require key attribute for conditionally rendered repeated components
 ---
+
 # vue/v-if-else-key
 
 > require key attribute for conditionally rendered repeated components
 
-- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> **_This rule has not been released yet._** </badge>
```
felipemelendez, line 12: I am not adding that underscore, for some reason when I hit save the asterisk is converting to an underscore 🙄

felipemelendez: that is a weird formatting setting, I did not add it (for what?!), so I wonder why this was not an issue the first time that I submitted 🤷
Anyhow, solved it

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,311 @@
...
+ * the nodes for each part of the conditional chain:
+ *
+ *  - 'if': Represents the node associated with the 'v-if' directive.
+ *  - 'elseIf': An array representing nodes associated with any 'v-else-if' directives.
+ *  - 'else': Represents the node associated with the 'v-else' directive.
+ *
+ * This structure helps in understanding the relationship between conditional directives
+ * within a given scope and aids in linting scenarios where these relationships matter.
+ *
+ * @type {Map<VElement, { if: VElement | null, elseIf: VElement[], else: VElement | null }>}
```
FloEdelmann, line 37: Since you are reusing this type, you can define it once using `@typedef` and then use its name everywhere:

```js
/**
 * @typedef ConditionalFamily
 * @property {VElement | null} if
 * @property {VElement[]} elseIf
 * @property {VElement | null} else
 */

/** @type {Map<VElement, ConditionalFamily>} */

By the way, does the `if` property have to be nullable or can we make it required?

Comment (felipemelendez) [reply lib/rules/v-if-else-key.js]: defined the type, refactored to made `if` required, and fine-tuned the docstrings, they were too verbose: [1206bbd](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/1206bbd1a8b1238c18a129c1e6d1228baa7d1067)

**Comment by felipemelendez** [comment]
Esteemed reviewers, this PR has been sitting here for months and Flo has already reviewed it, so I have a few questions that I hope will help push this through:

1. What are the next steps for this PR?
2. Why does it still show that changes are requested after I have addressed all the requested changes?
3. Why does it say that some checks were not successful? Is there something that I need to do to address those?

Thanks

**Comment by FloEdelmann** [comment]
1. I want to review the latest changes more thoroughly again. (That's why I self-requested a review.) After that, I'll either request/suggest more changes, or approve it and then assign it to the main maintainer @ota-meshi. I just haven't found the time to do so.
2. That's a quirk of GitHub's UI that you can ignore.
3. I actually don't know. Only the Netlify checks are failing; the GitHub Actions and CircleCI checks are passing. I remember we had this some time ago. @ota-meshi do you know how to resolve this? But anyway, @felipemelendez it's not caused by your changes (it also fails in other PRs), and you can ignore it for now.

File: lib/rules/v-if-else-key.js [hunk 9]
```diff
@@ -0,0 +1,299 @@
...
+      messageId: 'requireKey',
+      data: {
+        componentName
+      },
+      fix(fixer) {
+        const afterComponentNamePosition =
+          node.startTag.range[0] + componentName.length + 1
+        return fixer.insertTextBeforeRange(
+          [afterComponentNamePosition, afterComponentNamePosition],
+          ` key='${uniqueKey}'`
```
FloEdelmann, line 76: I think the most common style is to use double quotes for attributes, so please change the autofix accordingly. In codebases where single quotes are preferred, the [`vue/quote-props`](https://eslint.vuejs.org/rules/quote-props.html) autofix can change the double quotes to single quotes.

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,299 @@
...
+  utils.hasDirective(node, 'else')
+
+/**
+ * Checks whether a conditional family is complete.
+ * A conditional family is considered complete if:
+ * 1. An 'if' condition is present.
+ * 2. All elements of the family (if, else, else-if) share the same rawName.
+ * 3. There exists either an 'else' condition or at least one 'else-if' condition.
+ *
+ * @param {{ if: VElement, elseIf: VElement[], else: VElement | null }} family - The conditional family object with 'if', 'else-if', and 'else' properties.
```
FloEdelmann, line 102:
```suggestion
 * @param {ConditionalFamily} family - The conditional family object with 'if', 'else-if', and 'else' properties.
```

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,299 @@
...
+ * @returns {boolean} True if the conditional family is complete based on the above criteria, false otherwise.
+ */
+const isConditionalFamilyComplete = (family) => {
+  if (!family || !family.if) {
```
FloEdelmann, line 106: `if` cannot be falsy anymore, right?

```suggestion
  if (!family) {
```

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,299 @@
...
+const hasConditionalDirective = (node) =>
+  utils.hasDirective(node, 'if') ||
+  utils.hasDirective(node, 'else-if') ||
+  utils.hasDirective(node, 'else')
+
+/**
+ * Checks whether a conditional family is complete.
+ * A conditional family is considered complete if:
+ * 1. An 'if' condition is present.
+ * 2. All elements of the family (if, else, else-if) share the same rawName.
```
FloEdelmann, line 99: Why should the rule not trigger for (at least) the first two lines in this case?

```html
<component-a v-if="foo" />
<component-a v-else-if="bar" />
<component-b v-else />
```

Or (at least) the first and the last line in this case:

```html
<component-a v-if="foo" />
<component-b v-else-if="bar" />
<component-a v-else />
```

lib/rules/v-if-else-key.js
felipemelendez: thanks 👍🏼 
[8da18db](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/8da18db699dbac5eb21bfa62aae45143493da656)

lib/rules/v-if-else-key.js
felipemelendez: [a8709d7](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/a8709d716d6c9bfd3af247ea594d38e8f68b0183)

lib/rules/v-if-else-key.js
felipemelendez: f039fb7

lib/rules/v-if-else-key.js
felipemelendez: Hi Flo, sorry for the delayed response, things got a little hectic. Anyhow, the rule detects the first case, I'm not sure about the second one, but I'll examine this in detail as soon as I get a chance. Thanks again for making time to review this.

felipemelendez: Well, I am too curious and stubborn to leave this sort of question unanswered - I could not focus on anything else with this on my mind so I went back and studied the code, and you were correct to point this out, the rule was omitting the second scenario that you illustrated. 

I have fixed that problem and added a test for that scenario: 
2ff3a5f

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,296 @@
...
+/**
+ * Checks whether a conditional family is complete.
+ * A conditional family is considered complete if:
+ * 1. An 'if' condition is present.
+ * 2. There exists either an 'else' condition or at least one 'else-if' condition.
+ *
+ * @param {ConditionalFamily} family - The conditional family object with 'if', 'else-if', and 'else' properties.
+ * @returns {boolean} True if the conditional family is complete based on the above criteria, false otherwise.
+ */
+const isConditionalFamilyComplete = (family) => {
```
FloEdelmann, line 104: I think this function is not needed at all:

1. It's only used once
2. It's never called with a falsy parameter, so the `!family` check is not needed
3. `family.if` should never be undefined, so the second check is not needed

So let's inline the `(family.else || family.elseIf.length > 0)` check to the call site.

felipemelendez: never mind, that logic is simple and self-evident: 
42500ce

FloEdelmann (approved):
LGTM now, thanks!

tests/lib/rules/require-key-for-conditionally-rendered-repeated-components.js
```diff
@@ -0,0 +1,429 @@
...
+
+const tester = new RuleTester({
+  parser: require.resolve('vue-eslint-parser'),
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module'
+  }
+})
+
+tester.run('v-if-else-key', rule, {
```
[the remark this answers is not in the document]
felipemelendez: [51cef47](https://github.com/vuejs/eslint-plugin-vue/pull/2280/commits/51cef4700b81a4b78c504d0ffc18bd814dc6e8c8)

lib/rules/v-if-else-key.js
```diff
@@ -0,0 +1,275 @@
...
+ * @property {VElement[]} elseIf - An array of nodes associated with 'v-else-if' directives.
+ * @property {VElement | null} else - The node associated with the 'v-else' directive, or null if there isn't one.
+ */
+
+/**
+ * Map to store conditionally rendered components and their respective conditional directives.
+ *
+ * @type {Map<VElement, ConditionalFamily>}
+ */
+const conditionalFamilies = new Map()
```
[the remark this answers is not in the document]
felipemelendez: good catch, and thank you for the review :grin:
e6d5c92

ota-meshi (approved):
LGTM! Thank you!

PR #2292, opened 2023-10-06
fix: replace typeParameters with typeArguments

KaelWD:
https://github.com/typescript-eslint/typescript-eslint/pull/5384

#2256 missed a few

### DIFF
--- a/lib/rules/define-emits-declaration.js
+++ b/lib/rules/define-emits-declaration.js
@@ -47,7 +47,8 @@ module.exports = {
           }
 
           case 'runtime': {
-            if (node.typeParameters && node.typeParameters.params.length > 0) {
+            const typeArguments = node.typeArguments || node.typeParameters
+            if (typeArguments && typeArguments.params.length > 0) {
               context.report({
                 node,
                 messageId: 'hasTypeArg'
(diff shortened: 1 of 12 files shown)


FloEdelmann (approved):
Thanks! Looks good from my side. I think we should release this as a patch version as soon as possible.

ota-meshi (approved):
Thank you! I didn't notice it.

ota-meshi:
I looked into why the test is not currently failing. It appears that the deprecated typeParameters property is still left as an alias on those nodes. Therefore, there is no need to rush patch releases.
However, I haven't released this plugin recently, so I would like to check the remaining PRs and recent issues and release it soon.

KaelWD:
It floods the terminal with depreciation notices in my project, idk why it doesn't here.

ota-meshi:
Isn't that a warning about ESLint v9? If so, we support the old ESLint and cannot simply replace it. I have a number of plugins that have similar issues, so I'm considering whether to create a util package for them.

KaelWD:
No it's from typescript-eslint 6

ota-meshi:
Oh really. I've never seen it so I don't know.

PR #2297, opened 2023-10-19
Add `externalIgnores` option for `vue/singleline-html-element-content-newline`

qmhc:
## What to be solved

When users want to customize `ignores` option for `vue/singleline-html-element-content-newline` rule and don't want to lose the ignored elements which are specified by default, it's a hard work.

A long list is defined in `utils/inline-non-void-elements.json`, and user need to define it again in their config.


FloEdelmann (approved):
I guess that makes sense. Thanks for your contribution!

ota-meshi (approved):
LGTM! Sorry for the late review.

PR #2301, opened 2023-10-24
fix(component-name-in-template-casing): ignore vue template syntax

### LINKED ISSUE #2299: `vue/component-name-in-template-casing`: ignore `<component>` built-in element

- `vue/component-name-in-template-casing`

It will provide one fewer warning/error, because it will handle an edge-case.

The change will treat `<component>` as an element, not a component.

The `vue/component-name-in-template-casing` rule enforces pascal-cased component names. This will also enforce `<component>` to be written as `<Component>`, which is not correct. Technically, `<component>` is not a component, but an element.

The following code will be valid:
```vue
<template>
  <component is="div" />
</template>
```

It will enforce `<component>` to be pascal-cased as `<Component>`:

```vue
<template>
  <Component is="div" />
</template>
```

- Keep `<component>` as-is.

**Additional context**

As the [Vue docs state](https://vuejs.org/api/built-in-special-elements.html):

> [!NOTE]
> `<component>`, `<slot>` and `<template>` are component-like features and part of the template syntax. They are not true components and are compiled away during template compilation. As such, they are conventionally written with lowercase in templates.

The issue came up working with [`@antfu/eslint-config`](https://github.com/antfu/eslint-config). 
Related PR:
- https://github.com/antfu/eslint-config/pull/291

antfu:
fix #2299


ota-meshi (approved):
LGTM! Thank you!

Issue #2302, opened 2023-10-24
vue/require-toggle-inside-transition false positives


719media:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **eslint-plugin-vue version: 9.18.0

**What did you do?**	

The following all result in linting errors (`vue/require-toggle-inside-transition`), but should not be, as a result of https://github.com/vuejs/eslint-plugin-vue/issues/2290

```vue
<transition>
  <div key="hello" />
</transition>
```

```vue
<transition>
  <slot />
</transition>
```

```vue
<transition appear>
  <div>Something</div>
</transition>
```

**What did you expect to happen?**
No lint errors

**What actually happened?**
lint errors

ota-meshi:
Thank you for posting the issue.
They look like bugs. Please consider opening a PR if you want a quick fix.

Shyam-Chen:
```sh
  3:5  error  The element inside `<transition>` is expected to have a `v-if` or `v-show` directive  vue/require-toggle-inside-transition
```

```vue
<template>
  <Transition name="fade">
    <slot></slot>
  </Transition>
</template>
```

DevilTea:
> Thank you for posting the issue. They look like bugs. Please consider opening a PR if you want a quick fix.

Hello Ota Meshi,

I've opened a PR for it 🫡

PR #2303, opened 2023-10-27
Fix require-toggle-inside-transition should respect "<slot>", "v-bind:key", attribute "appear"

DevilTea:
fix: #2302 

Also, the following case that tansition with attribute `appear` should be valid.
```
<template>
  <Transition appear>
    <div />
  </Transition>
</template>
```

### DIFF
--- a/lib/rules/require-toggle-inside-transition.js
+++ b/lib/rules/require-toggle-inside-transition.js
@@ -32,6 +32,13 @@ module.exports = {
       if (utils.isCustomComponent(element)) {
         return
       }
+
+      /** @type VElement */ // @ts-expect-error
+      const parent = element.parent
+      if (utils.hasAttribute(parent, 'appear')) {
+        return
+      }
+
       if (
         element.name !== 'slot' &&
         !utils.hasDirective(element, 'if') &&
@@ -50,7 +57,7 @@ module.exports = {
       /** @param {VElement} node */
       "VElement[name='transition'] > VElement"(node) {
         verifyInsideElement(node)
-      },
+      }
     })
   }
 }
--- a/lib/rules/require-toggle-inside-transition.js
+++ b/lib/rules/require-toggle-inside-transition.js
@@ -56,6 +56,11 @@ module.exports = {
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VElement} node */
       "VElement[name='transition'] > VElement"(node) {
+        const child = node.parent.children.find(utils.isVElement)
+        if (child !== node) {
+          return
+        }
+
         verifyInsideElement(node)
       }
     })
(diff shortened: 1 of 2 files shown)


lib/rules/require-toggle-inside-transition.js
```diff
@@ -47,10 +56,6 @@ module.exports = {
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VElement} node */
       "VElement[name='transition'] > VElement"(node) {
-        const child = node.parent.children.find(utils.isVElement)
-        if (child !== node) {
-          return
-        }
```
ota-meshi, line 53: I don't think this change is necessary. Could you please revert?

ota-meshi (changes requested):
Thank you for this PR!

lib/rules/require-toggle-inside-transition.js
DevilTea: If this condition is kept, the following case would be failed.
```html
<template>
  <Transition appear>
    <div v-if="show" />                           // valid
    <div />                                                 // should be marked as error, but not
  </Transition>
</template>
```

ota-meshi: I think that second `<div>` is invalid for another reason.
If you want to check whether the elements within that `<transition>` are valid, I think we need to add another rule.

ota-meshi: [demo](https://play.vuejs.org/#eNp9kb1uwjAUhV/lygutRMnAhgJSWzG0Q1u1jF5McgMGx7b8E5BQ3r3XpoUMCE/2+c6xzrVP7NnaSReRzVjpKydtAI8h2gXXsrXGBTiBwwZ6aJxpYUTWEdeV0Z6MW3OAecIPjVAeH7kui/MlFKdDwNYqEZBOAOXKCe1lkEaDsBaFyzKBWnbQPclmzlm6kjMoFgnkVRTQCSXrgXdAh4ucFI+qhjVCK9weaxAe0DnjxrCOAbQJuUlxrUIdyuLSk41Z8DRcIzeTnTeaXuWUEpxVprVSofu0KeU5m0EmiQmlzOE9a8FFHP/r1Rar/Q19549J4+zLoUfXIWcXFoTbYDjj5c8HHml/ga2poyL3HfiN3qiYOp5tL1HXVHvgy23f8t9KvVn55TEgPcbfUKlocvbZzxn99+ud0a91p5NpznHds/4XUzm/5w==)

DevilTea: Okay, that make sense 🤔

ota-meshi (approved):
LGTM! Thank you!

Issue #1841, opened 2022-04-07
order-in-components does not autofix 


BoYuanSu:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 
   ^8.12.0
- **eslint-plugin-vue version:** 
   ^8.6.0
- **Node version:** 
   v14.17.0
- **Operating System:** 
   macOS 10.15.7

**Please show your full configuration:**

```json5
module.exports = {
    extends: [
      // add more generic rulesets here, such as:
      // 'eslint:recommended',
      'plugin:vue/recommended',
      // 'plugin:vue/recommended' // Use this if you are using Vue.js 2.x.
    ],
    parser:'vue-eslint-parser',
    rules: {
      // override/add rules settings here, such as:
      // 'vue/no-unused-vars': 'error'
    }
  }

```

**What did you do?**	

// src/components/Test.vue
```vue
import { mapState } from "vuex";

export default {
  name: 'Test',
  computed:{
    ...mapState(['count']),
    aaa(){
      return '123'
    }
  },
  data:()=>({
    data:'132'
  })
}
```

**What did you expect to happen?**
save file and auto fix eslint vue/order-in-components warning

**What actually happened?**

vue/order-in-components does not autofix 

```
yarn run v1.22.15
$ eslint --print-config path::String
{
  "env": {
    "browser": true,
    "es6": true
  },
  "globals": {},
  "parser": "/Users/ohbot-frontend/Projects/ohbot/test-eslint/node_modules/vue-eslint-parser/index.js",
  "parserOptions": {
    "ecmaVersion": 2020,
    "sourceType": "module"
  },
  "plugins": [
    "vue"
  ],
  "rules": {
    "vue/attributes-order": [
      "warn"
    ],
    "vue/component-tags-order": [
      "warn"
    ],
    "vue/no-lone-template": [
      "warn"
    ],
    "vue/no-multiple-slot-args": [
      "warn"
    ],
    "vue/no-v-html": [
      "warn"
    ],
    "vue/order-in-components": [
      "warn"
    ],
    "vue/this-in-template": [
      "warn"
    ],
    "vue/attribute-hyphenation": [
      "warn"
    ],
    "vue/component-definition-name-casing": [
      "warn"
    ],
    "vue/first-attribute-linebreak": [
[...]

Issue #724, opened 2018-12-14
vue/order-in-components doesn't fix warnings


jean-moldovan:
```
105:5  warning  The "computed" property should be above the "created" property on line 73  vue/order-in-components
```

Expected: vue/order-in-components is supposed to be fixable
(https://github.com/vuejs/eslint-plugin-vue/issues/299)
Current: not fixable

eslint-plugin-vue v5.0.0
node v9.11.1

Thanks!

michalsnik:
Hi @jean-moldovan This rule is fixable actually, but not in all cases. We explicitly disable auto-fix, when we detect that some of the nodes might contain side-effects. We want you to be safe and not break your app in hard to debug way - so we suggest you to consciously fix order by yourself.

jean-moldovan:
Thanks, @michalsnik. Mind sharing an example of a possible side-effect when "computed" goes above "created"?

Thanks again!

armano2:
we are not auto sorting it because there spread operator. [live demo](https://mysticatea.github.io/vue-eslint-demo/#eJx1UstOwzAQ/JXFQkor2anEsaq4coILSBwoUqNkU4zih/wILVH+nXWcNnDAB1uenZ2dXXtgtWmQbdnO107acL/XAFJZ4wIMoCr7gCGg8zBC64yCoo94KvY60fA00Rpsq9gRPWEAykQdsIHV+oIA1EZ702EpdWtW4UP68gu72ih89EcOE9BTEWk0AeucNfJ8Es1GEtwucmVZLs5Wb0WiVPr8VCksOFnMUsX7mmejaS0F/zgDcBii03B4zQQIBm6HydIv2fEGDpeUizGAxfM/mv2sNBPHRWPuMR207Tbz9BlnLnbo2XZgNOkNWVCog2ikwzrInl7qjk+RT38S0aMXfeX8FdVG0Kzb6KU+il60xtEu22vcuAadkFqk3owm6ZQ7ciZ1Q5dn+Z0r5OvL2aav4W1VIzmzVAkdAeitQ2TjD1nOvPw=)

deka:
Why this issue is closed ?

> We explicitly disable auto-fix, when we detect that some of the nodes might contain side-effects. 

Can you explain more ? update doc ? or fix side effects ?

kaantureyyen:
[...]

PR #2309, opened 2023-11-10
Update order-in-components.md

kazuma0129:
Hi team, I have added modifications to the document.

Even though this rule does not support auto-fix, the documentation explained that it does. So I have removed it.
Honestly, I'm not sure why it cannot be fixed automatically, but it was mentioned in this comment (https://github.com/vuejs/eslint-plugin-vue/issues/1841#issuecomment-1091610974).

related:
- https://github.com/vuejs/eslint-plugin-vue/issues/1841
- https://github.com/vuejs/eslint-plugin-vue/issues/724

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -10,7 +10,6 @@ since: v3.2.0
 > enforce order of properties in components
 
 - :gear: This rule is included in `"plugin:vue/vue3-recommended"` and `"plugin:vue/recommended"`.
-- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 ## :book: Rule Details
 


FloEdelmann:
That line is auto-generated from the rule:

FloEdelmann:
That line is auto-generated from the rule:

https://github.com/vuejs/eslint-plugin-vue/blob/b3129f9f3551f1ac189a0ee84b43561b3d0e4cc2/lib/rules/order-in-components.js#L220

And indeed, *some* problems can be auto-fixed, so the message in the documentation is not wrong.

kazuma0129:
@FloEdelmann Okay, It’s my misunderstanding.
By the way, there are definitely cases where this rule does not work, as actually reported in some of the issues, right?
Then I think it would be clearer to specify such cases in the documentation. What do you think?

FloEdelmann:
Sure, explaining those cases in the documentation would be a good idea.

Issue #2313, opened 2023-11-14
Rule suggestion: Add new option to specify tags to ignore rules `vue/no-deprecated-slot-attribute`


kazuma0129:
**What rule do you want to change?**
vue/no-deprecated-slot-attribute

**Does this change cause the rule to produce more or fewer warnings?**
fewer

**How will the change be implemented? (New option, new default behavior, etc.)?**
New option

**Please provide some example code that this change will affect:**

Suppose we have a code like this.
```vue
<template>
  <!-- bad -->
  <div slot="one"></div>
  <a slot="two"></div>
  <my-component slot="three"></div>
</template>
```

What I would like is an option like this.
Specifically, an option that allows the tag name to be passed as an array.

```
rules: {
    "vue/no-deprecated-slot-attribute": [
      "error",
      { ignore: ["a", "my-component"] }
    ],
}
```

As a result, this rule will be excluded for slots with tags specified as "ignore".
```vue
<template>
  <!-- still bad -->
  <div slot="one"></div>
  <!-- good -->
  <a slot="two"></div>
  <my-component slot="three"></div>
</template>
```

**What does the rule currently do for this code?**
There is no such option. 
It’s just applying or not applying the rule, which does not seem flexible to me.

**What will the rule do after it's changed?**
This option allows explicit and centralized control of where the rule is being ignored.

This is useful for custom elements that have slots as attributes.
Currently, the rule must be turned off or disabled each time it is used.

**Additional context**

PR #2314, opened 2023-11-14
Add new option to specify tags to ignore `vue/no-deprecated-slot-attribute`

kazuma0129:
closes: https://github.com/vuejs/eslint-plugin-vue/issues/2313

I have not yet updated the documentation for this rule. If the content of this PR and issue is accepted, I will start working on those.

### DIFF
--- a/lib/rules/no-deprecated-slot-attribute.js
+++ b/lib/rules/no-deprecated-slot-attribute.js
@@ -16,7 +16,17 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/no-deprecated-slot-attribute.html'
     },
     fixable: 'code',
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          ignore: {
+            type: 'array'
+          }
+        },
+        additionalProperties: false
+      }
+    ],
     messages: {
       forbiddenSlotAttribute: '`slot` attributes are deprecated.'
     }
--- a/lib/rules/syntaxes/slot-attribute.js
+++ b/lib/rules/syntaxes/slot-attribute.js
@@ -11,6 +11,12 @@ module.exports = {
   supported: '<3.0.0',
   /** @param {RuleContext} context @returns {TemplateListener} */
   createTemplateBodyVisitor(context) {
+    const options = context.options[0] || {}
+    /** @type {string[]} */
+    const ignore = Array.isArray(options.ignore)
+      ? options.ignore
+      : [options.ignore]
+
     const sourceCode = context.getSourceCode()
     const tokenStore =
       context.parserServices.getTemplateBodyTokenStore &&
@@ -95,6 +101,9 @@ module.exports = {
      * @returns {void}
      */
     function reportSlot(slotAttr) {
+      if (ignore.includes(slotAttr.parent.parent.name)) {
+        return
+      }
       context.report({
         node: slotAttr.key,
         messageId: 'forbiddenSlotAttribute',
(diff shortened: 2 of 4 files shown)


lib/rules/no-deprecated-slot-attribute.js
```diff
@@ -16,7 +16,17 @@ module.exports = {
...
     },
     fixable: 'code',
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          ignore: {
+            type: 'array'
+          }
```
FloEdelmann, line 25:
```suggestion
          ignore: {
            type: 'array',
            items: { type: 'string' },
            uniqueItems: true
          }
```

lib/rules/syntaxes/slot-attribute.js
```diff
@@ -11,6 +11,12 @@ module.exports = {
   supported: '<3.0.0',
   /** @param {RuleContext} context @returns {TemplateListener} */
   createTemplateBodyVisitor(context) {
+    const options = context.options[0] || {}
+    /** @type {string[]} */
+    const ignore = Array.isArray(options.ignore)
+      ? options.ignore
+      : [options.ignore]
```
FloEdelmann, line 18: Please make this a `Set`; it can be queried with `set.has()` in constant time (instead of `array.includes()` which takes linear time).

lib/rules/syntaxes/slot-attribute.js
```diff
@@ -95,6 +101,9 @@ module.exports = {
      * @returns {void}
      */
     function reportSlot(slotAttr) {
+      if (ignore.includes(slotAttr.parent.parent.name)) {
```
FloEdelmann, line 104: Please also ignore other casings:

```ts
const casing = require('../utils/casing')

const componentName = slotAttr.parent.parent.name
if (
  ignore.has(componentName) ||
  ignore.has(casing.pascalCase(componentName)) ||
  ignore.has(casing.kebabCase(componentName))
) { … }
```

And please add test cases for that.

FloEdelmann (changes requested):
Yeah, I guess this makes sense. I have some comments though. Thanks for your contribution!

lib/rules/syntaxes/slot-attribute.js
kazuma0129: Fixed: [847a098](https://github.com/vuejs/eslint-plugin-vue/pull/2314/commits/847a0986d974e78ed66f21ce03857365a261b329)

lib/rules/syntaxes/slot-attribute.js
kazuma0129: Fixed: [847a098](https://github.com/vuejs/eslint-plugin-vue/pull/2314/commits/847a0986d974e78ed66f21ce03857365a261b329)

FloEdelmann (approved):
LGTM :slightly_smiling_face:

kazuma0129:
@FloEdelmann @ota-meshi 
How about this one? Let me know if there is any work that needs to be done.

ota-meshi:
I need to review, merge and release this PR, but I don't have time to do that now 😓

kazuma0129:
@ota-meshi Thanks for the reply. If so, it's fine when you have time.🙆‍♂️ Just wanted to check on the status.

ota-meshi (changes requested):
The implementation looks good to me. Could you add documentation for the new option?

kazuma0129:
@ota-meshi 
I have updated the documentation. Please check it. [f3879f4](https://github.com/vuejs/eslint-plugin-vue/pull/2314/commits/f3879f40c3aa73635e0266973c8e053d363af738)

ota-meshi (approved):
LGTM! Thank you!

Issue #2242, opened 2023-07-19
Add `type-literal` as an option for `vue/define-emits-declaration`


studnitz:
**What rule do you want to change?**

`vue/define-emits-declaration`

**Does this change cause the rule to produce more or fewer warnings?**

It's just a stylistic rule for enforcing the [more ergonomic `defineEmits` types](https://blog.vuejs.org/posts/vue-3-3#more-ergonomic-defineemits) introduced in Vue v3.3.

**How will the change be implemented? (New option, new default behavior, etc.)?**

Could be a new option `type-literal` alongside `type-based` or could be a configuration option when using `type-based`. One could also think about enforcing the usage of [labeled tuple elements](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements) as shown in the example docs.

**Please provide some example code that this change will affect:**

```ts
// Valid with `type-literal`
const emit = defineEmits<{
  foo: [id: number]
  bar: [name: string, ...rest: any[]]
}>()

// Invalid with `type-literal`
const emit = defineEmits<{
  (e: 'foo', id: number): void
  (e: 'bar', name: string, ...rest: any[]): void
}>()
```

**What does the rule currently do for this code?**

Enforce either type-based or runtime based `defineEmits` declaration.

**What will the rule do after it's changed?**

Add also an option to enforce the new type syntax for `defineEmits` introduced in v3.3.

**Additional context**

FloEdelmann:
A new `type-literal` option makes sense :+1: 

Is there any reason to not use labeled tuple elements? Otherwise, I'd suggest to always require them in the `type-literal` case. See also

* https://github.com/typescript-eslint/typescript-eslint/issues/2457

Happy to accept a PR for the new option!

preetsahil:
Can i work on this issue ? or it is solved.

FloEdelmann:
Feel free to work on it 👍

PR #2315, opened 2023-11-14
Add support for stricter `type-literal` option in `vue/define-emits-declaration`

Meesayen:
Fixes: #2242 

This PR adds support for a stricter check to always use the "type literal" type based `defineEmits` signature introduced in Vue 3.3.

i.e. 

```ts
defineEmits<{
  // this line is good
  change: [value: string]

  // this line is bad
  (e: 'change', value: string): void
}>()
```

Not included in this PR is a strict check on the use of named tuples. Maybe it can be added as an extra sub-option to the `type-literal` option in the future, if possible.

### DIFF
--- a/docs/rules/define-emits-declaration.md
+++ b/docs/rules/define-emits-declaration.md
@@ -5,13 +5,15 @@ title: vue/define-emits-declaration
 description: enforce declaration style of `defineEmits`
 since: v9.5.0
 ---
+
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):
This looks great, thank you!

ota-meshi (approved):
LGTM! Sorry for the late review.

Issue #1291, opened 2020-08-28
Support for "Flat Config"


ota-meshi:
**The problem you want to solve.**

We need to check for compatibility with  ESLint's new Flat Config.

**Your take on the correct solution to problem.**

When `@eslint/eslintrc` is ready to use, check compatibility.
Modify the sharable configuration as needed.

**Additional context**

* #2226
* #2319
* https://github.com/eslint/eslint/issues/13481
* https://github.com/eslint/eslintrc
* https://eslint.org/blog/2022/08/new-config-system-part-2/
* https://eslint.org/docs/latest/use/configure/configuration-files-new

FloEdelmann:
Related:

* #2226

nzakas:
Now that this repo is linting itself using flat config, what would be the next steps? Do you have a plan for exporting flat configs from this plugin?

ota-meshi:
I think the next step is to exporting the flat config from this plugin 👍

Havrin:
just to make sure because I just ran into an error while setting this up with flat config from eslint: Flatconfig or `eslint.config.js` is currently not supported and will (maybe) be working with the next major version of `eslint-plugin-vue`, correct? There is no workaround?

FloEdelmann:
I haven't tried it yet, but you should be able to use `eslint-plugin-vue` in `eslint.config.js` like this:

```js
const { FlatCompat } = require('@eslint/eslintrc')

const eslintrc = new FlatCompat({
  baseDirectory: __dirname
})

module.exports = [
  // …
  ...eslintrc.plugins('vue'),
  ...eslintrc.extends('plugin:vue/recommended'),
  // …
]
```

(similar to how plugins/configs in our own ESLint config were migrated in #2226)

Havrin:
yeah, true that works, thanks! On [npmjs](https://www.npmjs.com/package/@eslint/eslintrc) it says

> This package is not intended for use outside of the ESLint ecosystem.

which I thought meant to not use it in my project.

Btw I am using `"type": "module"`  so my working config looks like this:

```js
import { FlatCompat } from '@eslint/eslintrc';
import globals from 'globals';
import path from 'path';
import { fileURLToPath } from 'url';
const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);

const eslintrc = new FlatCompat({
    baseDirectory: __dirname
})

export default [
    {
[...]

PR #2319, opened 2023-11-18
Add support for generating ESLint Flat Config

conradhale:
Will fix https://github.com/vuejs/eslint-plugin-vue/issues/1291

Needs more work, haven't had the time to work on this lately.

I was thinking it would be better to replace the generated config files with JS functions that dynamically return the correct configuration. Should decrease the size of the library and make the code a bit cleaner with the addition of flat configs and probably wouldn't impact runtime performance much. Most other eslint plugins seem to be doing this as well.


conradhale:
It might be a while before I have time to work on this, if any maintainers want to take over this PR, feel free to make any edits.

FloEdelmann:
@conradhale Thanks! I added some more commits, could you please have a look at them? I am not sure what is still missing to have full Flat config support, so I am handing this back to you again :slightly_smiling_face:

lib/configs/flat/base.js
```diff
@@ -0,0 +1,24 @@
...
+  languageOptions: {
+    parser: vueEslintParser,
+    globals: {
+      ...globals.browser,
+      ...globals.es2015
+    },
+    parserOptions: {
+      ecmaVersion: 2020,
+      sourceType: 'module'
+    }
```
aladdin-add, line 18: In most cases, you don't need to specify `parserOptions`. just use:

```js
  languageOptions: {ecmaVersion: 2020, sourceType: 'module'}
```
eslint will pass it to the parser. So you don't need to specify global.es2015 either.

aladdin-add (changes requested):

lib/configs/flat/base.js
aladdin-add: https://eslint.org/docs/next/use/configure/language-options#specifying-javascript-options

lib/index.js
```diff
@@ -258,7 +258,35 @@ module.exports = {
     'strongly-recommended': require('./configs/strongly-recommended'),
     'vue3-essential': require('./configs/vue3-essential'),
     'vue3-recommended': require('./configs/vue3-recommended'),
-    'vue3-strongly-recommended': require('./configs/vue3-strongly-recommended')
+    'vue3-strongly-recommended': require('./configs/vue3-strongly-recommended'),
+    'flat/base': {
+      plugins: { vue: module },
+      ...require('./configs/flat/base')
+    },
+    'flat/essential': {
```
ota-meshi, line 266: I would like to change `flat/essential`, `flat/recommended`, and `flat/strongly-recommended` to the config for Vue3. Use the `vue2-` prefix to config for Vue2 instead.

aladdin-add:
@ota-meshi Is there anything blocking this PR from being merged now? I have been implementing `npm init @eslint/config` for the flat configs, so it would be great see it's officially supported in the plugin. 😄

aladdin-add:
@conradhale I can pick this up! (if you don't mind :)

ota-meshi:
I would like the configuration to be renamed.
Correspond as follows:
- (Legacy -> Flat)
- essential -> flat/vue2-essential
- strongly-recommended -> flat/vue2-strongly-recommended
- recommended -> flat/vue2-recommended
- vue3-essential -> flat/essential
- vue3-strongly-recommended -> flat/strongly-recommended
- vue3-recommended -> flat/recommended
https://github.com/vuejs/eslint-plugin-vue/pull/2319/files#r1465649137

We also need test code to test the flat configuration.

Issue #2294, opened 2023-10-11
Rule suggestion: `vue/require-explicit-slots` to require all slots to be documented


Amixx:
**Please describe what the rule should do:**
I would like a rule that requires `defineSlots` to be present if there are any slots in the component - could be called `vue/no-undeclated-slots` or something similar. I did not find a rule that does this in the rule list, and did not find a relevant issue.

**What category should the rule belong to?**
[X] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<template>
<slot></slot>
</template>

<script setup lang="ts">
// any code, except `defineSlots`
</script>
```

**Additional context**
I am sorry if this issue is not submitted correctly, and will fix any errors. I am also willing to try to implement the rule myself if anyone can give me a starting point (maybe a rule that is somewhat similar)

FloEdelmann:
Sounds good!

For reference, here is the documentation of `defineSlots`: https://vuejs.org/api/sfc-script-setup.html#defineslots
That should also be included in the rule documentation.

Similar rules:
* https://eslint.vuejs.org/rules/require-explicit-emits.html
* https://eslint.vuejs.org/rules/no-undef-components.html

As a name for the new rule, I'd suggest `vue/require-explicit-slots` to be consistent with `vue/require-explicit-emits`.

Amixx:
Thanks for the quick comment, I will try to implement it over the weekend.

mussinbenarbia:
@Amixx @FloEdelmann I noticed that there was no PR for this yet so I gave this a try [here](https://github.com/vuejs/eslint-plugin-vue/pull/2325) 🙇 I hope it's ok

Issue #2166, opened 2023-05-13
Planning for v10


ota-meshi:
This issue lists the changes we plan to release in eslint-plugin-vue v10.

## 💥 Breaking Changes

- #2153 
- #2311
- #2669
- Drop support for typescript-eslint v5, v6
- #2630
  - Allow minor versions of this plugin to enhance the checks for new features in Vue and Nuxt.
  - Allow updating deprecated/reserved HTML+SVG element tag names in minor versions, see #2171
- Change presets ~or #1951~
  - #2640
  - #2668
  - use `ecmaVersion: 'latest'` (#1991)
  - #2674
  - #2627
  - #2628
  - #2629
  - #2653
  - #2648
  - #2652
  - And more...
- Rule changes:
  - #2626
  - #2655
  - maybe #2232?
- https://github.com/vuejs/vue-eslint-parser/pull/195
- #2645
- #2675

## ✨ Enhancements

- #2153 
- Support for Nuxt v3
  - Add support for `defineNuxtComponent()`
    - (Related to #2057)
    - (Related to #2191)
  - Add support for auto imports, and `"#imports"`
    - (Related to #1969)

## Already released earlier

- #1291
- #1969

ishowman:
from changelog, `vue/no-dupe-keys` only support in `    "eslint-plugin-vue": "^9.14.0",`, but I install 9.13.0, and vscode still show the error with `Duplicated key`. I don't know why...
![image](https://github.com/vuejs/eslint-plugin-vue/assets/16897283/028d7c34-4f99-4209-9b5c-6f97d4b00e6e)

here are my eslint config:
```
/* eslint-env node */
require('@rushstack/eslint-patch/modern-module-resolution');

module.exports = {
  root: true,
  extends: [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript',
    // fileURLToPath(new URL('./.eslintrc-auto-import.json', import.meta.url)),
    '@vue/eslint-config-prettier/skip-formatting',
    './.eslintrc-auto-import.json',
  ],
  parserOptions: {
    ecmaVersion: 'latest',
  },
  rules: {
    '@typescript-eslint/ban-ts-comment': 'off',
    '@typescript-eslint/no-unused-vars': [
      'error',
      {
        argsIgnorePattern: '^_',
      },
    ],
    '@typescript-eslint/no-explicit-any': 'off',
    curly: 'error',
[...]

PR #2325, opened 2023-11-25
feat: implement require-explicit-slots

mussinbenarbia:
This is an attempt at solving https://github.com/vuejs/eslint-plugin-vue/issues/2294

I would like to hear your opinion on the handling of slots declared twice. As you can see from the implementation, I'm reporting this as well but I'm not sure if such a thing should be responsibility of this rule.

What do you think?

Also, I did not implement a fixer for now as we do not know how the slot function is typed. I think this is good enough for now but any opinions?

Thank you.


mussinbenarbia:
Looks like the tests are failing 🤔 They all pass locally, let me see if there is some version mismatch somewhere.

FloEdelmann:
We plan to drop support for Node v14 and v16 in eslint-plugin-vue v10 (see #2166), so new rules intended to be merged before that major version should be compatible with Node v14 as well. But we can also mark this PR as a breaking change and only release it with or after v10.

mussinbenarbia:
@FloEdelmann I think this makes sense to introduce in v10 then since consumers of it wouldn't be on node 14 anyway due to the reasons mentioned above 🙏

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    const slotsDefined = new Set()
+
+    return utils.compositingVisitors(
+      utils.defineScriptSetupVisitor(context, {
+        onDefineSlotsEnter(node) {
+          const typeArgument = node.typeArguments.params[0]
```
ota-meshi, line 32: Could you please add test cases that does not use `@typescript-eslint/parser`? Probably the `typeArguments ` property does not exist.

ota-meshi (changes requested):

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,73 @@
...
+    schema: [],
+    messages: {
+      requireExplicitSlots:
+        'Slots must be explicitly defined with the defineSlots macro.',
+      alreadyDefinedSlot: 'Slot {{slotName}} is already defined.'
+    }
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    const slotsDefined = new Set()
```
ota-meshi, line 27: I think the rule should probably be disabled if it doesn't have `lang="ts"`.

I think we probably need a guard like this:

```suggestion
    const df =
      context.parserServices.getDocumentFragment &&
      context.parserServices.getDocumentFragment()
    if (!df) {
      return {}
    }
    const scripts = df.children
      .filter(utils.isVElement)
      .filter((e) => e.name === 'script')
    if (scripts.every((script) => !utils.hasAttribute(script, 'lang', 'ts'))) {
      return {}
    }
    const slotsDefined = new Set()
```

lib/rules/require-explicit-slots.js
mussinbenarbia: I see! This is why some other rules actually define the `@typescript-eslint/parser` option in a separate object and then spread it into some tests. I was considering TS by default but that's a wrong assumption to make. Thank you for the suggestion, I'll add the test 👍

docs/rules/require-explicit-slots.md
```diff
@@ -0,0 +1,71 @@
...
+# vue/require-explicit-slots
+
+> require slots to be explicitly defined with defineSlots
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule enforces all slots used in the template to be defined once
+in the `script setup` block with the [`defineSlots`](https://vuejs.org/api/sfc-script-setup.html) macro.
```
ota-meshi, line 17: Can this rule be changed to also check the options API to check if the slots option is set?

https://vuejs.org/api/options-rendering.html#slots

mussinbenarbia: I'll look into adding this today, thank you!

lib/rules/require-explicit-slots.js
mussinbenarbia: Unfortunately I seem to get a parsing error if I don't use `@typescript-eslint/parser` 🤔 

```sh
1) require-explicit-slots
       valid

      <template>
        <div>
          <slot />
        </div>
      </template>
      <script setup lang="ts">
      defineSlots<{
        default(props: { msg: string }): any
      }>()
      </script>:

      AssertionError [ERR_ASSERTION]: A fatal parsing error occurred: Parsing error: Unexpected token :
      + expected - actual

      -false
      +true
```

All test cases fail with the same error, a colon `:` seems to get appended at the end of the script closing tag. Since the syntax we are looking for is TS, I thought we must require `@typescript-eslint/parser`, am I misunderstanding?

cc @FloEdelmann

FloEdelmann: When not using the TS parser, you also can't use TS syntax, i.e. `lang="ts"` and type annotations like `: string` need to be removed.

mussinbenarbia: Makes sense, the thing I can't understand though is whether testing without TS is required in this case since both [defineSlots](https://vuejs.org/api/sfc-script-setup.html#defineslots) and the option api [slots](https://vuejs.org/api/options-rendering.html#slots) require TS in the first place.

If we disable the rule when the script does not have `lang="ts"`, as per @ota-meshi's suggestion below, then the tests can just assume that TS is being used, correct?

FloEdelmann: Hmm true, `defineSlots`/`slots` don't make much sense without TypeScript. But I think they would not break the build without it, right? So I think the rule should still run in a JavaScript-only component.

If we agree that the rule doesn't run in a non-TS component, then a test should verify that this rule doesn't report any warnings for a non-TS component not defining `defineSlots`/`slots` but still using `<slot>`.

mussinbenarbia:
@ota-meshi @FloEdelmann I made some changes to the implementation based on the comments. Could you please take another look when you get the chance? 🙏

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,128 @@
...
+    )
+    if (scripts.every((script) => !utils.hasAttribute(script, 'lang', 'ts'))) {
+      return {}
+    }
+    const slotsDefined = new Set()
+
+    return utils.compositingVisitors(
+      utils.defineScriptSetupVisitor(context, {
+        onDefineSlotsEnter(node) {
+          const typeArguments = node.typeArguments
```
ota-meshi, line 48: I think it will work even with older versions of dependencies by doing the following:

```suggestion
          const typeArguments =  'typeArguments' in node ? node.typeArguments : node.typeParameters
```

docs/rules/require-explicit-slots.md
```diff
@@ -0,0 +1,71 @@
...
+    <slot />
+  </div>
+</template>
+<script setup lang="ts">
+defineSlots<{
+  default(props: { msg: string }): any
+}>()
+</script>
+
+<!-- ✓ GOOD -->
```
ota-meshi, line 34: Could you please split the code block into one component so we can see it in the demo?

https://deploy-preview-2325--eslint-plugin-vue.netlify.app/rules/require-explicit-slots.html

````suggestion
```

</eslint-code-block>

<eslint-code-block :rules="{'vue/require-explicit-slots': ['error']}">

```vue
<!-- ✓ GOOD -->
````

File: lib/rules/require-explicit-slots.js [hunk 6]
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * @typedef {import('@typescript-eslint/types').TSESTree.TypeNode} TypeNode
+ */
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'require slots to be explicitly defined with defineSlots',
+      categories: ['vue3-strongly-recommended'],
```
ota-meshi, line 18: I don't think it is necessary to change the configuration, so could you please change it to undefined?

```suggestion
      categories: undefined,
```

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }),
+      utils.executeOnVue(context, (obj) => {
+        const slotsProperty = obj.properties.find(
+          (property) => property.key.name === 'slots'
+        )
+        if (!slotsProperty) return
+
+        const slotsTypeHelper =
+          slotsProperty.value.typeAnnotation.typeName.name === 'SlotsType' &&
```
ota-meshi, line 79: If we are not using TypeScript, `typeAnnotation` is not defined.

```suggestion
          slotsProperty.value.typeAnnotation?.typeName.name === 'SlotsType' &&
```

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,128 @@
...
+          }
+        }
+      }),
+      utils.defineTemplateBodyVisitor(context, {
+        "VElement[name='slot']"(node) {
+          let slotName = 'default'
+
+          const slotNameAttr = node.startTag.attributes.find(
+            (attribute) => attribute.key.name === 'name'
+          )
```
ota-meshi, line 112: We can use `utils.getAttribute()`.

```suggestion
          const slotNameAttr = utils.getAttribute(node, 'name')
```

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,128 @@
...
+                slotsDefined.add(slotName)
+              }
+            }
+          }
+        }
+      }),
+      utils.executeOnVue(context, (obj) => {
+        const slotsProperty = obj.properties.find(
+          (property) => property.key.name === 'slots'
+        )
```
ota-meshi, line 75: We can use `utils.findProperty()`.

```suggestion
        const slotsProperty = utils.findProperty(obj, 'slots')
```

lib/rules/require-explicit-slots.js
```diff
@@ -0,0 +1,128 @@
...
+        if (!slotsProperty) return
+
+        const slotsTypeHelper =
+          slotsProperty.value.typeAnnotation.typeName.name === 'SlotsType' &&
+          slotsProperty.value.typeAnnotation
+        if (!slotsTypeHelper) return
+
+        const typeArguments = slotsTypeHelper.typeArguments
```
ota-meshi, line 83:
```suggestion
        const typeArguments = 'typeArguments' in slotsTypeHelper ? slotsTypeHelper.typeArguments : slotsTypeHelper.typeParameters
```

ota-meshi (changes requested):

lib/rules/require-explicit-slots.js
mussinbenarbia: Thank you for the suggestion! I made this change but just to re-confirm my understanding, is this needed even though we are checking that the script has `lang="ts"`? We return `{}` if doesn't have the attribute.

ota-meshi: Even if lang="ts", if you do not use `as` as shown below, an undefined error will occur.

```js
      defineComponent({
        slots: Object,
      });
```

lib/rules/require-explicit-slots.js
mussinbenarbia: Indeed, all checks passed!

mussinbenarbia:
@ota-meshi I addressed the comments. I'm not sure if I understood the comment about the docs correctly. I'll make the change if it's still wrong

ota-meshi (approved):
LGTM! Thank you!

docs/rules/require-explicit-slots.md
```diff
@@ -0,0 +1,109 @@
...
+  </div>
+</template>
+<script setup lang="ts">
+defineSlots<{
+  bar(props: { msg: string }): any
+}>()
+</script>
+```
+
+</eslint-code-block>
```
FloEdelmann, line 105: I think we don't need to show *all* cases as an example, and I think it's easier to understand if there is only one code block. Maybe we want to add a second code block using the Options API, but 5 is too much.

````suggestion
<eslint-code-block :rules="{'vue/require-explicit-slots': ['error']}">

```vue
<template>
  <div>
    <!-- ✓ GOOD -->
    <slot />
    <slot name="foo" />

    <!-- ✗ BAD -->
    <slot name="bar" />
  </div>
</template>

<script setup lang="ts">
defineSlots<{
  default(props: { msg: string }): any
  foo(props: { msg: string }): any
}>()
</script>
```

</eslint-code-block>
````

docs/rules/require-explicit-slots.md
FloEdelmann, line 1: Can you please add a "Related rules" section and link to the [`vue/require-explicit-emits`](https://eslint.vuejs.org/rules/require-explicit-emits.html) rule, and also link from there to this new rule?

FloEdelmann (changes requested):

docs/rules/require-explicit-slots.md
mussinbenarbia: I reduced the number of examples to two 👍

FloEdelmann (approved):
Thanks for implementing the feedback! This looks good to me now :slightly_smiling_face:

PR #2327, opened 2023-11-27
fix: avoid calling typeParameters getter

KaelWD:
#2292

typeParameters was still being accessed when typeArguments is undefined


ota-meshi (approved):
LGTM! Thank you!

PR #2329, opened 2023-11-29
update vitepress

ota-meshi:
This PR updates VitePress to the latest version and changes the configuration so that the document works with the latest VitePress.


FloEdelmann (approved):

PR #2331, opened 2023-11-29
Improve compatibility with ESLint v9

ota-meshi:
This PR changes to not use APIs scheduled for removal in ESLint v9.

See https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/

### DIFF
--- a/docs/.vitepress/build-system/shim/eslint/use-at-your-own-risk.mjs
+++ b/docs/.vitepress/build-system/shim/eslint/use-at-your-own-risk.mjs
@@ -1,0 +1,3 @@
+export default {
+  /* empty */
+}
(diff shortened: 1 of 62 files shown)


FloEdelmann:
Are the replacements available in all ESLint versions we support? See also https://github.com/vuejs/eslint-plugin-vue/issues/2166 where we can drop support for ESLint v7.

ota-meshi:
Yes, I think we can support up to ESLint v6.2 for now.

FloEdelmann (approved):
Looks good to me then :slightly_smiling_face: 

One small (non-blocking) wish: Could you add a GitHub CI job with ESLint v9?

ota-meshi:
Unfortunately, ESLint v9 has not even been released in alpha version yet :sweat:  We can add tests once it is released.

Issue #2332, opened 2023-11-30
9.19.0 bug with vue/v-on-event-hyphenation


ElectricS01:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version: 8.54.0**  
- **eslint-plugin-vue version: 9.19.0** 
- **Node version: 18** 
- **Operating System: macOS** 

**Please show your full configuration:**

```json5
module.exports = {
  root: true,
  env: {
    node: true
  },
  plugins: ["prettier"],
  extends: [
    "plugin:vue/vue3-recommended",
    "eslint:recommended",
    "@vue/eslint-config-typescript",
    "prettier"
  ],
  rules: {
    "vue/multi-word-component-names": "off",
    "vue/no-reserved-component-names": "off"
  }
};
```

**What did you do?**	
Breaks on running `eslint . --fix --ignore-path .gitignore` when the option `plugin:vue/vue3-recommended` or higher is in the .eslintrc.cjs

```
Oops! Something went wrong! :(

ESLint: 8.54.0

Error: .eslintrc.cjs » plugin:vue/vue3-recommended » /Users/electrics01/GitHub/PrivateUploader/frontend/node_modules/eslint-plugin-vue/lib/configs/vue3-strongly-recommended.js:
        Configuration for rule "vue/v-on-event-hyphenation" is invalid:
        Value {"autofix":true} should be equal to one of the allowed values.
```
**Repository to reproduce this issue**
Use any with 9.19.0 and `plugin:vue/vue3-recommended`

kazuma0129:
We face the same problem.

ElectricS01:
I have a fix, will create a pr now

ElectricS01:
Fixed in #2333

ota-meshi:
Thank you for letting me know about the issue!
I fixed the problem and released it.
https://github.com/vuejs/eslint-plugin-vue/releases/tag/v9.19.1

PR #2333, opened 2023-11-30
Fix 'vue/v-on-event-hyphenation'

ElectricS01:


ElectricS01:
This PR fixes https://github.com/vuejs/eslint-plugin-vue/issues/2332

ota-meshi:
Oh, thank you!
However, since we need to add tests, etc., I close this PR in favor of #2334.

PR #2338, opened 2023-12-01
Improve compatibility with ESLint v9 (take2)

ota-meshi:
This PR is a follow-up PR for #2331.
I checked again and found that it still uses APIs that will be removed in ESLint v9. This PR aims to refactor those source codes to improve compatibility with ESLint v9.

### DIFF
--- a/lib/rules/jsx-uses-vars.js
+++ b/lib/rules/jsx-uses-vars.js
@@ -63,7 +63,12 @@ module.exports = {
           return
         }
 
-        context.markVariableAsUsed(name)
+        const sourceCode = context.getSourceCode()
+        if (sourceCode.markVariableAsUsed) {
+          sourceCode.markVariableAsUsed(name, node)
+        } else {
+          context.markVariableAsUsed?.(name)
+        }
       }
     }
   }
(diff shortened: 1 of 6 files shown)


lib/rules/jsx-uses-vars.js
```diff
@@ -63,7 +63,12 @@ module.exports = {
           return
         }
 
-        context.markVariableAsUsed(name)
+        const sourceCode = context.getSourceCode()
+        if (sourceCode.markVariableAsUsed) {
+          sourceCode.markVariableAsUsed(name, node)
+        } else {
+          context.markVariableAsUsed?.(name)
+        }
```
FloEdelmann, line 71: Can we extract this into a small util function?

typings/eslint/index.d.ts
```diff
@@ -18,7 +18,8 @@ export namespace Scope {
     scopes: Scope[]
     globalScope: Scope | null
     acquire(node: VAST.ESNode | VAST.Program, inner?: boolean): Scope | null
-    getDeclaredVariables(node: VAST.ESNode): Variable[]
+    /** Added in ESLint v8.38.0 */
```
FloEdelmann, line 21:
```suggestion
    /** @since ESLint v8.38.0 */
```

(also for the other occurrences)

ota-meshi:
I have changed this PR. Could you check it again?

FloEdelmann (approved):
Looks good to me, thanks!

PR #2343, opened 2023-12-04
Document non-fixable rules correctly

FloEdelmann:
This PR enables the `catchNoFixerButFixableProperty` rule option of the [`eslint-plugin/require-meta-fixable` rule](https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-fixable.md), which increases the number of false-positives, but also catches two non-fixable rules that were incorrectly described.


ota-meshi (approved):
Thank you!

Issue #2342, opened 2023-12-04
`vue/v-if-else-key` false-positives and false-negatives


FloEdelmann:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 8.55.0
- **eslint-plugin-vue version:** 9.19.2
- **Node version:** 18.18.2
- **Operating System:** Ubuntu

**Please show your full configuration:**
```json5
{
    "rules": {
        "vue/v-if-else-key": "error"
    }
}
```

**What did you do?**
**What did you expect to happen?**
**What actually happened?**
**Repository to reproduce this issue**

→ please see the following demo links, where each issue is highlighted:

* https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJylUUsOgjAQvco4GzY2JC4J5SRsKgxIhELaghpC4mk8mCexBDRVjMbYVd+b9n3aHpM6JQwwNFQ1pTAUxRLsCk1R3vcTnufQsSLjMWozAs45eFldezE6hx8CTFNekTTR9XwJ/ScGwhVjQMeGEkMpCDgIJQuZw44UAWNv5ayTIeWozYST038psshiC1CpadFiK5RtAXs6WV7YPl8zy/pD6N89t395TtcscF4A16jakjQGPXYt+ePXTUGsIwabYbgBPbajsw==
* https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJxtj8EKgzAQRH9lu+cugR4l9Uu8WB3bUE0kRq2I/97USpHinnYHZt7szIUrwQnrgKat84A0sxRHl2bY1oNzlZpJCte0zsIGGsRU14wr5zImlZI+iRBeLYqAkqyjMffW2Ds94EEiKf2lRUAMQd0h2ndg9SXvlCPuLfcrdyu/mbT6fcVn9n2NjpOZhx7q41tp8sTEyWVZ3uUzTjQ=
[...]

PR #2348, opened 2023-12-07
Fix siblings and interleaving issues reported in #2342

felipemelendez:
This PR addresses the bugs reported in #2342 

And added tests


felipemelendez:
Hi @FloEdelmann, thank you for reporting #2342, I have hopefully addressed those here, please let me know if I missed something.

FloEdelmann (commented):
I have not yet looked into your implementation, but I noticed one wrong test case.

tests/lib/rules/v-if-else-key.js
```diff
@@ -424,6 +508,62 @@ tester.run('v-if-else-key', rule, {
...
+        </script>
+        `,
+      output: `
+        <template>
+          <div>
+            <ComponentA key="component-a-1" v-if="foo" />
+
+            <ComponentA v-if="bar" />
+            <ComponentA key="component-a-3" v-else-if="baz" />
+            <ComponentA key="component-a-4" v-else />
```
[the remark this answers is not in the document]
felipemelendez: done, and thanks for pointing that out; I figure that the main benefit of this change is that it will make more sense to the end user?
[0923acc](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/0923acc43384a8665b382a56febeeed021520549)
[e978c9d](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/e978c9d0e43448cb599dedf4a5168602f80c697c)

lib/rules/v-if-else-key.js
```diff
@@ -25,6 +25,33 @@ const casing = require('../utils/casing')
...
+const hasConditionalRenderedSiblings = (node, componentName) => {
+  if (node.parent && node.parent.type === 'VElement') {
+    const siblings = node.parent.children.filter(
+      (child) => child.type === 'VElement'
+    )
+
+    return siblings.some(
+      (sibling) =>
+        sibling !== node &&
+        sibling.type === 'VElement' &&
```
FloEdelmann, line 46: If we need to check the element's type anyway in the `some` predicate, then we can avoid `filter`ing the list before.

```suggestion
    return node.parent.children.some(
      (sibling) =>
        sibling !== node &&
        sibling.type === 'VElement' &&
```

lib/rules/v-if-else-key.js
```diff
@@ -44,35 +71,39 @@ const checkForKey = (
...
-          node.startTag.range[0] + componentName.length + 1
-        return fixer.insertTextBeforeRange(
-          [afterComponentNamePosition, afterComponentNamePosition],
-          ` key="${uniqueKey}"`
-        )
+      if (node === conditionalFamily.if || node === conditionalFamily.else) {
+        needsKey = true
+      } else if (conditionalFamily.elseIf.includes(node)) {
+        needsKey = true
       }
```
FloEdelmann, line 88: Please replace with a single const:

```ts
const needsKey =
  conditionalFamily.if === node ||
  conditionalFamily.else === node ||
  conditionalFamily.elseIf.includes(node)
```

lib/rules/v-if-else-key.js
```diff
@@ -222,44 +253,55 @@ module.exports = {
...
-            // If this is the second occurrence, also apply a fix to the first occurrence
-            if (usageInfo.count === 1) {
-              const uniqueKeyForFirstInstance = `${casing.kebabCase(
-                componentName
-              )}-1`
+        if (isConditional) {
+          if (isIfDirective && usageInfo.hasIf) {
+            // Reset if family already has an 'if' directive
+            usageInfo.count = 1
+            usageInfo.firstNode = node
```
FloEdelmann, line 270: Will this work for the following case? Please also add this as a test case:

```vue
<ComponentA v-if="foo" />
<ComponentA v-else />

<ComponentA v-if="bar" />
<ComponentA v-else-if="baz" />
<ComponentA v-else />
```

An error should be reported for all of them.

tests/lib/rules/v-if-else-key.js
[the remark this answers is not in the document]
FloEdelmann: > does it make a difference which one of those two gets the key?

Probably not; while all but one have a key, the `undefined` key is still unique. But it makes it much more readable.

FloEdelmann:
@felipemelendez Are you still interested in fixing this?

felipemelendez:
> @felipemelendez Are you still interested in fixing this?

Yes! Sorry for the delay, I had a family emergency at the start of the year, so I've been distracted to say the least; I'll get back on this soon!

FloEdelmann:
No worries, there's more important stuff in life than coding :slightly_smiling_face:

tests/lib/rules/v-if-else-key.js
```diff
@@ -424,6 +508,69 @@ tester.run('v-if-else-key', rule, {
...
+        }
+        </script>
+        `,
+      output: `
+        <template>
+          <div>
+            <ComponentA key="component-a-1" v-if="foo" />
+            <ComponentA key="component-a-2" v-else />
+
+            <ComponentA v-if="bar" />
```
felipemelendez, line 539: I believe this should also have a key, but we're currently limiting the rule to one `if` per family. This seems like a rare case, and I'm unsure how to address it. Should we address it? Any ideas or suggestions?

lib/rules/v-if-else-key.js
felipemelendez: [2bc5596](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/2bc55964057a57279c1321d5d1198d3502f9ea26)

lib/rules/v-if-else-key.js
felipemelendez: Currently, an error is reported for all of them except the second `if`

lib/rules/v-if-else-key.js
```diff
@@ -25,6 +25,29 @@ const casing = require('../utils/casing')
...
+    return node.parent.children.some(
+      (sibling) =>
+        sibling !== node &&
+        sibling.type === 'VElement' &&
+        sibling.rawName === componentName &&
+        hasConditionalDirective(sibling)
+    )
+  }
+
+  return false
```
FloEdelmann, line 48: Better return early, that saves one indentation level and makes the code more readable:

```suggestion
  if (!node.parent || node.parent.type !== 'VElement') {
    return false
  }

  return node.parent.children.some(
    (sibling) =>
      sibling !== node &&
      sibling.type === 'VElement' &&
      sibling.rawName === componentName &&
      hasConditionalDirective(sibling)
  )
```

lib/rules/v-if-else-key.js
```diff
@@ -44,35 +67,36 @@ const checkForKey = (
   uniqueKey,
   conditionalFamilies
 ) => {
-  if (node.parent && node.parent.type === 'VElement') {
+  if (
```
FloEdelmann, line 70: Can you rewrite this into an early return again please?

lib/rules/v-if-else-key.js
```diff
@@ -44,35 +67,36 @@ const checkForKey = (
...
-    if (
-      conditionalFamily &&
-      (utils.hasDirective(node, 'bind', 'key') ||
-        utils.hasAttribute(node, 'key') ||
-        !hasConditionalDirective(node) ||
-        !(conditionalFamily.else || conditionalFamily.elseIf.length > 0))
-    ) {
-      return
-    }
+    if (conditionalFamily && !utils.hasAttribute(node, 'key')) {
```
FloEdelmann, line 77: Can you rewrite this into an early return again please?

tests/lib/rules/v-if-else-key.js
FloEdelmann: I also think that this should be reported. I don't think it's that rare (especially with `<div>`s), so let's better handle it.

I think all consecutive elements with the same element name should go into the same conditional family, i.e. `v-if` should not always create a new conditional family.

lib/rules/v-if-else-key.js
felipemelendez: [acb6be2](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/acb6be2f9d804e435aa200ef5a52a032bf95e85a)

lib/rules/v-if-else-key.js
felipemelendez: [f9600d5](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/f9600d5aa2714a9a2a6dd9621bbbd7d97e0af44d)

tests/lib/rules/v-if-else-key.js
felipemelendez: got it!
Logic refactor:
[3f4d687](https://github.com/vuejs/eslint-plugin-vue/pull/2348/commits/3f4d6877311b2be4185a4b89a680fd1d39cb2f24)

Updated test:
4cf8e87

FloEdelmann: Nice!

FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thank you!

Issue #2235, opened 2023-07-12
Force defineExpose at the end of <script>


Thomasan1999:
**Please describe what the rule should do:**

The rule would force to have `defineExpose` at the end of the `<script>` tag. When `<script setup>` is used, `defineExpose` is used as the alternative to `return` and is often expected to be at the end the same way `return` was. It also uses the data used in `<script>` and should therefore not be placed before the data itself. It is similar to `vue/define-macros-order` where define macros are forced at the beginning of `<script>`.

**What category should the rule belong to?**

[X] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<script setup>
const counter = ref(0);
const title = ref('');

defineExpose({
    counter,
    title,
});

const itemsText = computed(() => `${ counter.value } items`);
</script>

<script setup>
const counter = ref(0);
const title = ref('');

defineExpose({
    counter,
    title,
});

const shown = ref(false);

function show() {
    show.value = true;
}

watch(counter, show);
</script>
```

**Additional context**

FloEdelmann:
I think it could even be part of `vue/define-macros-order` with a new option. What do you think?

Eternal-Rise:
It would be great to have `vue/order-in-components` rule, but for script setup

Eternal-Rise:
@FloEdelmann, I'm not fully agree with that point. If your component have many different logic - probably it is a bad component, that should be decomposed.

FloEdelmann:
@Eternal-Rise True, but please don't discuss that in this issue.

J-Michalek:
> I think it could even be part of `vue/define-macros-order` with a new option. What do you think?

I agree and would welcome `defineExpose` to be added to the `vue/define-macros-order` rule as I ran into issues when I tried to use it right away and it reported invalid rule value.

waynzh:
@FloEdelmann @Thomasan1999 @J-Michalek I can give this a try. 
I recently met this issue as well and put up a [draft PR](https://github.com/vuejs/eslint-plugin-vue/pull/2349).

It differs from the original `order` property, so I added a new Boolean type property called  `defineExposeLast`. However, there are still some implementation details that I may need some help.

- open to suggestions for a better option name…
- Should we include `defineExpose` within the `order` schema? There could potentially have a conflict in the settings, like:
[...]

PR #2349, opened 2023-12-10
feat(define-macros-order): add `defineExposeLast` option

waynzh:
will close https://github.com/vuejs/eslint-plugin-vue/issues/2235

I'll be thankful for any suggestions on these TODOs.

- open to suggestions for a better option name…
- should we include `defineExpose` within the `order` schema? There is a potential conflict in the options, like:
`{ order: ['defineExpose', 'defineProps'], defineExposeLast: true }`.
- any other test cases should we add to verify the fixer?
- update the docs.

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -196,9 +196,14 @@ function create(context) {
       node,
       loc: node.loc,
       messageId: 'defineExposeNotTheLast',
-      fix(fixer) {
-        return moveNodeToLast(fixer, node, lastNode)
-      }
+      suggest: [
+        {
+          messageId: 'putExposeAtTheLast',
+          fix(fixer) {
+            return moveNodeToLast(fixer, node, lastNode)
+          }
+        }
+      ]
     })
   }
 
@@ -228,13 +233,8 @@ function create(context) {
       node.range[0] - beforeNodeToken.range[1]
     )
 
-    // insert position:  after target and comments (if any)
-    const afterTargetComment = sourceCode.getTokenAfter(target, {
-      includeComments: true
-    })
-
     return [
-      fixer.insertTextAfter(afterTargetComment, textNode),
+      fixer.insertTextAfter(target, textNode),
       fixer.removeRange([cutStart, cutEnd])
     ]
   }
@@ -322,6 +322,7 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/define-macros-order.html'
     },
     fixable: 'code',
+    hasSuggestions: true,
     schema: [
       {
         type: 'object',
@@ -345,7 +346,9 @@ module.exports = {
       macrosNotOnTop:
         '{{macro}} should be the first statement in `<script setup>` (after any potential import statements or type definitions).',
       defineExposeNotTheLast:
-        '`defineExpose` should be the last statement in `<script setup>`.'
+        '`defineExpose` should be the last statement in `<script setup>`.',
+      putExposeAtTheLast:
+        'Put `defineExpose` as the last statement in `<script setup>`.'
     }
   },
   create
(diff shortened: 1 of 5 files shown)


FloEdelmann (commented):
Thanks, this looks really good to me! :slightly_smiling_face: 

I couldn't come up with a better option name. I shortly thought about extending it to an array (something like `orderLast`), but I'm not sure if that's worth the effort or if `defineExpose` is the only macro for which it makes sense to put it at the end.

waynzh:
> I shortly thought about extending it to an array (something like `orderLast`)

Introducing a new `orderLast` array property or placing `defineExpose` into the `order` schema might potentially lead to configuration conflicts, like `{ order: ['defineExpose', 'defineProps'], anotherPropertyMarksExposeLast }`. I was at a loss as to how to deal with these conflicts, so I chose to add a new boolean property specifically for the 'defineExpose' macro.

> if `defineExpose` is the only macro for which it makes sense to put it at the end

In my personal practice, yes.
If it makes sense, I'll go ahead and update the test case and docs ASAP🤖

FloEdelmann:
> Introducing a new `orderLast` array property […] might potentially lead to configuration conflicts

If we'd introduce that `orderLast` option, we should validate the options so that a property can only be specified in either `order` or `orderLast, so that wouldn't be a problem.

But if `defineExpose` is the only macro for which it makes sense to put it at the end anyway, I think `defineExposeLast` is fine.

waynzh:
Putting `defineExpose` or other macros at the front can be understood as it aids in better understanding the current component. But if it's mandatory to place it at the end, it feels like only `defineExpose` is reasonable. 

Just curious, are there similar references that need to be restricted in the options? I'm not sure whether to enforce this in the JSON schema or to trigger an error in the code logic.🤔

FloEdelmann:
> are there similar references that need to be restricted in the options?

It's not possible in JSON schema. And it looks like we don't already do the same in another rule. But we could either throw an error, or just ignore values in `orderLast` that are already part of `order`.

tests/lib/rules/define-macros-order.js
```diff
@@ -170,6 +179,21 @@ tester.run('define-macros-order', rule, {
...
+          defineProps({
+            test: Boolean
+          })
+          console.log('test')
+          defineExpose({
+            a: 1
+          })
+        </script>
+      `,
+      options: optionsExposeLast
```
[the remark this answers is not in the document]
waynzh: Done 🤖

FloEdelmann (approved):
LGTM, thanks! :slightly_smiling_face:

lib/rules/define-macros-order.js
```diff
@@ -172,6 +187,58 @@ function create(context) {
...
+   * @param {ASTNode} node
+   * @param {ASTNode} lastNode
+   */
+  function reportExposeNotOnBottom(node, lastNode) {
+    context.report({
+      node,
+      loc: node.loc,
+      messageId: 'defineExposeNotTheLast',
+      fix(fixer) {
+        return moveNodeToLast(fixer, node, lastNode)
```
ota-meshi, line 200: I think that probably shouldn't support auto-fix. It may break user's code.

```
<script setup>
let a = 1
defineExpose({a})
a = 2
</script>
```

What do you think about supporting suggestions instead?

ota-meshi (changes requested):

lib/rules/define-macros-order.js
waynzh: Sorry for the late response. It seems reasonable to me not to support auto-fix, thanks for pointing out!
I'll add to `suggestion` and refine the test cases.

ota-meshi (approved):
LGTM! Thank you!

Issue #2337, opened 2023-11-30
vue/no-unused-emit-declarations doesn't detect emits in templates


henribru:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** v8.54.0
- **eslint-plugin-vue version:** v9.19.2
- **Node version:** v18.17.1
- **Operating System:** Ubuntu 23.04

**Please show your full configuration:**

```json5
/* eslint-env node */
require('@rushstack/eslint-patch/modern-module-resolution')

module.exports = {
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript',
    '@vue/eslint-config-prettier/skip-formatting'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  },
  rules: {
    "vue/no-unused-emit-declarations": "error",
  }
}
```

**What did you do?**	

```vue
<script setup lang="ts">
const emit = defineEmits({"foo": () => true})
</script>

<template>
  <button @click="emit('foo')">Foo</button>
</template>
</style>

```

**What did you expect to happen?**
The emit shouldn't be flagged as unused since it's used in the template

**What actually happened?**

The emit is flagged as unused:
```
6:27  error  `foo` is defined as emit but never used  vue/no-unused-emit-declarations
```
**Repository to reproduce this issue**

https://github.com/henribru/eslint-unused-emit-repro

PR #2340, opened 2023-12-02
fix(no-unused-emit-declarations): detect emits in templates

waynzh:
try to fix https://github.com/vuejs/eslint-plugin-vue/issues/2337

PR #2350, opened 2023-12-10
fix(require-explicit-emits): detect template emits

waynzh:
This rule has a similar issue to https://github.com/vuejs/eslint-plugin-vue/issues/2337.
The unused emits in the template were not detected, so this PR fixes that.

### DIFF
--- a/lib/rules/require-explicit-emits.js
+++ b/lib/rules/require-explicit-emits.js
@@ -271,7 +271,7 @@ module.exports = {
             return
           }
 
-          // e.g. $emit() / emits() in template
+          // e.g. $emit() / emit() in template
           if (
             callee.type === 'Identifier' &&
             (callee.name === '$emit' ||
--- a/lib/rules/require-explicit-emits.js
+++ b/lib/rules/require-explicit-emits.js
@@ -71,6 +71,25 @@ function getNameParamNode(node) {
   return null
 }
 
+/**
+ * Check if the given name matches defineEmitsNode variable name
+ * @param {string} name
+ * @param {CallExpression | undefined} defineEmitsNode
+ * @returns {boolean}
+ */
+function verifyEmitVariableName(name, defineEmitsNode) {
+  const node = defineEmitsNode?.parent
+  if (!node) return false
+
+  /** @type {string | undefined} */
+  let defineEmitVariable
+  if (node.type === 'VariableDeclarator' && node.id.type === 'Identifier') {
+    defineEmitVariable = node.id?.name
+  }
+
+  return name === defineEmitVariable
+}
+
 module.exports = {
   meta: {
     type: 'suggestion',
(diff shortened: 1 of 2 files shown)


lib/rules/require-explicit-emits.js
```diff
@@ -71,6 +71,25 @@ function getNameParamNode(node) {
   return null
 }
 
+/**
+ * Check if the given name matches defineEmitsNode variable name
+ * @param {string} name
+ * @param {CallExpression | undefined} defineEmitsNode
+ * @returns {boolean}
+ */
+function verifyEmitVariableName(name, defineEmitsNode) {
```
FloEdelmann, line 80: Please rename this function to `isEmitVariableName`, see https://github.com/vuejs/eslint-plugin-vue/pull/2340#discussion_r1425643428.

lib/rules/require-explicit-emits.js
```diff
@@ -71,6 +71,25 @@ function getNameParamNode(node) {
...
+  const node = defineEmitsNode?.parent
+  if (!node) return false
+
+  /** @type {string | undefined} */
+  let defineEmitVariable
+  if (node.type === 'VariableDeclarator' && node.id.type === 'Identifier') {
+    defineEmitVariable = node.id?.name
+  }
+
+  return name === defineEmitVariable
```
FloEdelmann, line 90:
```suggestion

  if (
    node?.type === 'VariableDeclarator' &&
    node.id.type === 'Identifier'
  ) {
    return name === node.id.name
  }

  return false
```

lib/rules/require-explicit-emits.js
```diff
@@ -251,7 +270,16 @@ module.exports = {
           if (!vueTemplateDefineData) {
             return
           }
-          if (callee.type === 'Identifier' && callee.name === '$emit') {
+
+          // e.g. $emit() / emits() in template
```
FloEdelmann, line 274:
```suggestion
          // e.g. $emit() / emit() in template
```

tests/lib/rules/require-explicit-emits.js
```diff
@@ -1983,6 +1983,25 @@ emits: {'foo': null}
...
       ...getTypeScriptFixtureTestOptions()
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+        <div @click="emits('bar')"/>
+      </template>
+      <script setup lang="ts">
+      const emits = defineEmits<(e: 'foo') => void>()
```
FloEdelmann, line 1994: The convention is to call the return value of `defineEmits` `emit`, since it is a function that will emit the given event.

```suggestion
        <div @click="emit('bar')"/>
      </template>
      <script setup lang="ts">
      const emit = defineEmits<(e: 'foo') => void>()
```

tests/lib/rules/require-explicit-emits.js
FloEdelmann, line 1: Can you please also add a new `valid` test?

FloEdelmann (changes requested):
Thanks for finding and fixing this! I have a few comments, but they should be all fairly small.

tests/lib/rules/require-explicit-emits.js
waynzh: Thanks for pointing that out!

waynzh:
@FloEdelmann  Thanks for the suggestions! I've made the updates. Please take a look when you have a moment.

FloEdelmann (approved):
Looks good to me, thanks!

ota-meshi (commented):
LGTM! Thank you!

PR #2358, opened 2024-01-09
Remove Vue 3 syntax status notice from docs

FloEdelmann:
The described situation has changed/improved since the notice was added, so I think the notice can be dropped.


ota-meshi (approved):
Thank you!

PR #2360, opened 2024-01-10
Add support for defineModel

### LINKED ISSUE #2130: Support for `defineModel`

- **ESLint version:**  v8.36.0
- **eslint-plugin-vue version:** v9.10.0
- **Node version:** v18.12.1

Add support for `defineModel` macro.

See https://github.com/vuejs/core/blob/main/CHANGELOG.md#330-alpha9-2023-04-08

Parser and some rules are affected. However, we have not yet listed the affected rules.
I think that, first, we need to list the rules affected by defineModel.

**Additional context**

Related to https://github.com/vuejs/vue-eslint-parser/issues/186

### LINKED ISSUE #2326: Support for Vue.js 3.4
- [x] #2355
- [ ] #2130

ota-meshi:
close #2326 
close #2130

### DIFF
--- a/lib/rules/no-unused-emit-declarations.js
+++ b/lib/rules/no-unused-emit-declarations.js
@@ -338,8 +338,8 @@ module.exports = {
             node.parent.type === 'VariableDeclarator' &&
             node.parent.init === node
           ) {
-            // If store the return of defineModel() in a variable, we can mark the 'update:modelName' event as used if use that variable.
-            // If that variable is unused it will already be reported by `no-unused-var` rule.
+            // If the return value of defineModel() is stored in a variable, we can mark the 'update:modelName' event as used if that that variable is used.
+            // If that variable is unused, it will already be reported by `no-unused-var` rule.
             emitCalls.set(`update:${model.name.modelName}`, node)
           }
         },
--- a/lib/rules/no-unused-emit-declarations.js
+++ b/lib/rules/no-unused-emit-declarations.js
@@ -332,6 +332,17 @@ module.exports = {
             emitReferenceIds
           })
         },
+        onDefineModelEnter(node, model) {
+          if (
+            node.parent &&
+            node.parent.type === 'VariableDeclarator' &&
+            node.parent.init === node
+          ) {
+            // If store the return of defineModel() in a variable, we can mark the 'update:modelName' event as used if use that variable.
+            // If that variable is unused it will already be reported by `no-unused-var` rule.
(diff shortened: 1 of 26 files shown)


ota-meshi:
I realized that we also needed to modify rules related to refs, such as `no-ref-as-operand`.
So I'll draft this PR.

lib/rules/no-unused-emit-declarations.js
```diff
@@ -332,6 +332,17 @@ module.exports = {
...
           })
         },
+        onDefineModelEnter(node, model) {
+          if (
+            node.parent &&
+            node.parent.type === 'VariableDeclarator' &&
+            node.parent.init === node
+          ) {
+            // If store the return of defineModel() in a variable, we can mark the 'update:modelName' event as used if use that variable.
+            // If that variable is unused it will already be reported by `no-unused-var` rule.
```
FloEdelmann, line 342:
```suggestion
            // If the return value of defineModel() is stored in a variable, we can mark the 'update:modelName' event as used if that that variable is used.
            // If that variable is unused, it will already be reported by `no-unused-var` rule.
```

lib/rules/no-unused-properties.js
```diff
@@ -452,6 +454,35 @@ module.exports = {
...
+          const container = getVueComponentPropertiesContainer(
+            context.getSourceCode().ast
+          )
+          if (
+            node.parent &&
+            node.parent.type === 'VariableDeclarator' &&
+            node.parent.init === node
+          ) {
+            // If store the return of defineModel() in a variable, we can mark the model prop as used if use that variable.
+            // If that variable is unused it will already be reported by `no-unused-var` rule.
```
FloEdelmann, line 471:
```suggestion
            // If the return value of defineModel() is stored in a variable, we can mark the model prop as used if that that variable is used.
            // If that variable is unused, it will already be reported by `no-unused-var` rule.
```

lib/rules/require-prop-types.js
```diff
@@ -44,37 +67,20 @@ module.exports = {
...
-            break
-          }
-          case 'FunctionExpression':
-          case 'ArrowFunctionExpression': {
-            hasType = false
-            break
-          }
-        }
+        hasType = optionHasType(value) ?? true
       }
```
FloEdelmann, line 84: Please change this into a `const` with a ternary operator:

```js
const hasType = prop.type === 'array' ? false : (optionHasType(prop.value) ?? true)
```

FloEdelmann (approved):
LGTM :slightly_smiling_face:

PR #2276, opened 2023-08-16
Add Vuex support to `vue/no-unused-properties`

fuale:
Changes are aimed at improving Vuex support.

Following Vuex methods will be checked: `mapGetters`, `mapState`, `mapActions`, `mapMutations`, when `no-unused-properties` will be enabled

example:
```vue
<script>
  export default {
    computed: {
      ...mapGetters(["bar"]),
    },
    methods: {
      baz() {
        return this.bar // no error
      }
    }
  }
</script>
<template>
  {{ baz() }}
</template>
```

Changes will affect #1675 (Vuex and Pinia support)

PR #2361, opened 2024-01-10
Fix type errors caused by TypeScript

ota-meshi:
This PR fixes a type error when running `npm run tsc`.

It also fixes false positive edge cases for Vuex support added in #2276.


FloEdelmann (approved):
Nice!

Issue #2243, opened 2023-07-19
Rule suggestion: `vue/no-restricted-v-on`


ericuldall:
**What rule do you want to change?**
no-restricted-static-attribute

**Does this change cause the rule to produce more or fewer warnings?**
More warnings

**How will the change be implemented? (New option, new default behavior, etc.)?**
It would be nice to optionally allow directive matching

**Please provide some example code that this change will affect:**

```vue
<div @click="foo">Why you clicking a div?</div>
```

**What does the rule currently do for this code?**
Nothing

**What will the rule do after it's changed?**
Allow matching key to be `@click`

**Additional context**
I'm trying to implement it like so:
```
{
  "vue/no-restricted-static-attribute": ["error",
    {
      "key": "@click",
      "element": "div",
      "message": "Using \"@click\" is not allowed on div's. Use interactive elements instead."
    }
  ]
}
```

Perhaps this makes sense as a new rule like `vue/no-handlers-on-non-interactive-elements`
but for now the flexibility of just grouping the directives as an attribute would work as well.

FloEdelmann:
I think it should be a new rule `vue/no-restricted-v-on` that works similarly to the [`vue/no-restricted-v-bind`](https://eslint.vuejs.org/rules/no-restricted-v-bind.html) rule and other `vue/no-restricted-*` rules.

thesheppard:
@FloEdelmann @ericuldall  I've taken a shot at this and have attached a PR

PR #2367, opened 2024-01-12
Add new `vue/no-restricted-v-on` rule

thesheppard:
This is an attempt at solving https://github.com/vuejs/eslint-plugin-vue/issues/2243

Keen for feedback 🫡

### DIFF
--- a/docs/rules/no-restricted-v-on.md
+++ b/docs/rules/no-restricted-v-on.md
@@ -56,7 +56,7 @@ Alternatively, the rule also accepts objects.
 
 The following properties can be specified for the object.
 
-- `argument` ... Specify the argument name or pattern.
+- `argument` ... Specify the argument name or pattern or `null`. If `null` is specified, it matches `v-on=`.
 - `modifiers` ... Specifies an array of the modifier names. If specified, it will only be reported if the specified modifier is used.
 - `element` ... Specify the element name or pattern. If specified, it will only be reported if used on the specified element.
 - `message` ... Specify an optional custom message.
--- a/docs/rules/no-restricted-v-on.md
+++ b/docs/rules/no-restricted-v-on.md
@@ -24,10 +24,14 @@ This rule takes a list of strings, where each string is a argument name or patte
 }
 ```
 
-<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', 'bar']}">
+<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', 'bar', '/^bar/']}">
 
 ```vue
 <template>
+  <!-- ✓ GOOD -->
+  <div v-on:click="x" />
+  <div @tap="x" />
+
   <!-- ✗ BAD -->
   <div v-on:foo="x" />
   <div @bar="x" />
--- a/docs/rules/no-restricted-v-on.md
+++ b/docs/rules/no-restricted-v-on.md
@@ -20,11 +20,11 @@ This rule takes a list of strings, where each string is a argument name or patte
 
 ```json
 {
-  "vue/no-restricted-v-on": ["error", "/^v-/", "foo", "bar"]
+  "vue/no-restricted-v-on": ["error", "foo", "bar"]
 }
 ```
 
-<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', '/^v-/', 'foo', 'bar']}">
+<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', 'bar']}">
 
 ```vue
 <template>
@@ -36,30 +36,12 @@ This rule takes a list of strings, where each string is a argument name or patte
 
 </eslint-code-block>
 
-By default, `'/^v-/'` is set. This prevents mistakes intended to be directives.
-
-<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error']}">
-
-```vue
-<template>
-  <!-- ✗ BAD -->
-  <MyInput @v-model="x" />
-  <div @v-if="x" />
-</template>
-```
-
-</eslint-code-block>
-
 Alternatively, the rule also accepts objects.
 
 ```json
 {
   "vue/no-restricted-v-on": [
     "error",
-    {
-      "argument": "/^v-/",
-      "message": "Using `@v-xxx` is not allowed. Instead, remove `@` and use it as directive."
-    },
     {
       "argument": "foo",
       "message": "Use \"v-on:x\" instead."
@@ -74,7 +56,7 @@ Alternatively, the rule also accepts objects.
 
 The following properties can be specified for the object.
 
-- `argument` ... Specify the argument name or pattern or `null`. If `null` is specified, it matches `v-on=`.
+- `argument` ... Specify the argument name or pattern.
 - `modifiers` ... Specifies an array of the modifier names. If specified, it will only be reported if the specified modifier is used.
 - `element` ... Specify the element name or pattern. If specified, it will only be reported if used on the specified element.
 - `message` ... Specify an optional custom message.
--- a/docs/rules/no-restricted-v-on.md
+++ b/docs/rules/no-restricted-v-on.md
@@ -20,11 +20,11 @@ This rule takes a list of strings, where each string is a argument name or patte
 
 ```json
 {
-  "vue/no-restricted-v-on": ["error", "foo", "bar"]
+  "vue/no-restricted-v-on": ["error", "foo", "/^bar/"]
 }
 ```
 
-<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', 'bar', '/^bar/']}">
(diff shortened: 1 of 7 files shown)


docs/rules/no-restricted-v-on.md
```diff
@@ -0,0 +1,127 @@
...
+---
+# vue/no-restricted-v-on
+
+> disallow specific argument in `v-on`
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> ***This rule has not been released yet.*** </badge>
+
+## :book: Rule Details
+
+This rule allows you to specify `v-on` argument names that you don't wantt to use in your application.
```
FloEdelmann, line 15:
```suggestion
This rule allows you to specify `v-on` argument names that you don't want to use in your application.
```

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,192 @@
...
+'use strict'
+
+const utils = require('../utils')
+const regexp = require('../utils/regexp')
+
+const DEFAULT_OPTIONS = [
+  {
+    argument: '/^v-/',
+    message:
+      'Using `@v-xxx` is not allowed. Instead, remove `@` and use it as directive.'
```
FloEdelmann, line 14: I don't think this is a common use case, and certainly not a valid suggestion. Please remove the default option altogether (also from the docs). So by default, the rule should do nothing.

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,192 @@
...
+                uniqueItems: true
+              }
+            },
+            required: ['argument'],
+            additionalProperties: false
+          }
+        ]
+      },
+      uniqueItems: true,
+      minItems: 0
```
FloEdelmann, line 156: Without a default, it makes no sense to enable the rule without at least one item:

```suggestion
      minItems: 1
```

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,192 @@
...
+      description: 'disallow specific argument in `v-on`',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-restricted-v-on.html'
+    },
+    fixable: null,
+    schema: {
+      type: 'array',
+      items: {
+        oneOf: [
+          { type: ['string', 'null'] },
```
FloEdelmann, line 127: Why is `null` allowed?

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,192 @@
...
+                  enum: [
+                    'prevent',
+                    'stop',
+                    'capture',
+                    'self',
+                    'once',
+                    'passive'
+                  ]
+                },
+                uniqueItems: true
```
FloEdelmann, line 147: If the `modifiers` array is set, it should contain at least one item:

```suggestion
                uniqueItems: true,
                minItems: 1
```

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,192 @@
...
+  /** @param {RuleContext} context */
+  create(context) {
+    /** @type {ParsedOption[]} */
+    const options = (
+      context.options.length === 0 ? DEFAULT_OPTIONS : context.options
+    ).map(parseOption)
+
+    // const options = DEFAULT_OPTIONS.map(parseOption)
+
+    return utils.defineTemplateBodyVisitor(context, {
```
FloEdelmann, line 173:
```suggestion
    return utils.defineTemplateBodyVisitor(context, {
```

tests/lib/rules/no-restricted-v-on.js
FloEdelmann, line 1: Very nice test cases!

FloEdelmann (changes requested):

lib/rules/no-restricted-v-on.js
thesheppard: When you added null, it would use the `DEFAULT_OPTIONS` options  as an additional argument, I've removed this as it's no longer necessary

thesheppard:
Have addressed PR comments, it's ready for another look 💪🏾

lib/rules/no-restricted-v-on.js
FloEdelmann: Ah, I remember we also allow `null` for the `no-restricted-v-bind` rule to disallow `v-bind` without an argument. So `null` should indeed be a possible option here that can be set to disallow `v-on` without an argument (e.g. `v-on="listeners"`). Sorry for not checking you test cases earlier.

thesheppard: Cool no problem, I've addressed this @FloEdelmann  🦦:
- Updated docs
- Added another test case for this condition

docs/rules/no-restricted-v-on.md
```diff
@@ -0,0 +1,109 @@
...
+{
+  "vue/no-restricted-v-on": ["error", "foo", "bar"]
+}
+```
+
+<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', 'bar']}">
+
+```vue
+<template>
+  <!-- ✗ BAD -->
```
FloEdelmann, line 31: Can you please

1. add a "GOOD" example (just in this code block, above the "BAD" lines)
2. add back a pattern example (e.g. `"/^bar/"`)

FloEdelmann (changes requested):
Really cool, thanks for the implementing the feedback so quickly! I have just one final remark regarding the docs. Other than that, it looks good already :slightly_smiling_face:

docs/rules/no-restricted-v-on.md
thesheppard: Sorted 💪🏾

docs/rules/no-restricted-v-on.md
```diff
@@ -0,0 +1,113 @@
...
+  <div v-on:click="x" />
+  <div @tap="x" />
+
+  <!-- ✗ BAD -->
+  <div v-on:foo="x" />
+  <div @bar="x" />
+</template>
+```
+
+</eslint-code-block>
```
FloEdelmann, line 41:
````suggestion
```json
{
  "vue/no-restricted-v-on": ["error", "foo", "/^bar/"]
}
```

<eslint-code-block :rules="{'vue/no-restricted-v-on': ['error', 'foo', '/^bar/']}">

```vue
<template>
  <!-- ✓ GOOD -->
  <div v-on:click="x" />
  <div @tap="x" />

  <!-- ✗ BAD -->
  <div v-on:foo="x" />
  <div @bar-baz="x" />
</template>
```

</eslint-code-block>
````

File: docs/rules/no-restricted-v-on.md [hunk 9]
```diff
@@ -0,0 +1,113 @@
...
+
+## :couple: Related Rules
+
+- [vue/no-restricted-static-attribute]
+
+[vue/no-restricted-static-attribute]: ./no-restricted-static-attribute.md
+
+- [vue/no-restricted-v-bind]
+
+[vue/no-restricted-v-bind]: ./no-restricted-v-bind.md
```
FloEdelmann, line 108:
```suggestion
- [vue/no-restricted-static-attribute]
- [vue/no-restricted-v-bind]

[vue/no-restricted-static-attribute]: ./no-restricted-static-attribute.md
[vue/no-restricted-v-bind]: ./no-restricted-v-bind.md
```

Can you also link to this new rule from those rules, please?

FloEdelmann (approved):
Some more minor docs improvement suggestions. Thanks!

thesheppard:
I've added the links to the related docs 🙏🏾

lib/rules/no-restricted-v-on.js
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * @param {VDirectiveKey} key
+ * @param {ParsedOption} option
+ */
+function defaultMessage(key, option) {
+  const von = key.name.rawName === '@' ? '' : 'v-on'
+  const arg =
+    key.argument != null && key.argument.type === 'VIdentifier'
+      ? `@${key.argument.rawName}`
+      : ''
```
ota-meshi, line 95: Could you please add a test case using longform `v-on`?
It builds a broken error message.

image

https://deploy-preview-2367--eslint-plugin-vue.netlify.app/rules/no-restricted-v-on.html

I think we should probably do something like this:

```suggestion
  const von = key.name.rawName === '@' ? '@' : 'v-on'
  const arg =
    key.argument != null && key.argument.type === 'VIdentifier'
      ? `${von === '@' ? '' : ':'}${key.argument.rawName}`
      : ''
```

ota-meshi (changes requested):

lib/rules/no-restricted-v-on.js
thesheppard: Have fixed

ota-meshi (approved):
LGTM! Thank you!

PR #2357, opened 2024-01-09
Upgrade vue-eslint-parser and add support for v-bind same-name shorthand

ota-meshi:
close #2355

related to #2326

PR #2373, opened 2024-01-23
Remove v-bind shorthand syntax from list of examples rule will catch.

earlAchromatic:
This syntax gets removed at the parser level and so will never be enforced, even in Vue 2.

See https://github.com/vuejs/eslint-plugin-vue/pull/2357.


ota-meshi (approved):
Thank you!

Issue #2359, opened 2024-01-09
Enforce/disallow `v-bind` same name shorthand style


FloEdelmann:
**Please describe what the rule should do:**

`v-bind` same name shorthand was [added in Vue 3.4](https://blog.vuejs.org/posts/vue-3-4#v-bind-same-name-shorthand) and is supported since #2355/#2357. A new rule (e.g. `vue/v-bind-same-name-style`) or a new option of the existing [`vue/v-bind-style`](https://eslint.vuejs.org/rules/v-bind-style.html) rule should be added that allows always enforcing that shorthand where possible, or always disallow it.

**What category should the rule belong to?**
[x] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

### `"sameNameShorthand": "always"`

```vue
<template>
  <!-- ✓ GOOD -->
  <div :foo />
  <div v-bind:foo />

  <div :foo="bar" />
  <div :foo='bar' />
  <div v-bind:foo="bar" />
  <div v-bind:foo='bar' />

  <!-- ✗ BAD -->
  <div :foo="foo" />
  <div :foo='foo' />
  <div v-bind:foo="foo" />
  <div v-bind:foo='foo' />
</template>
```

### `"sameNameShorthand": "never"`

```vue
<template>
  <!-- ✓ GOOD -->
  <div :foo="foo" />
  <div :foo='foo' />
  <div v-bind:foo="foo" />
  <div v-bind:foo='foo' />

  <div :foo="bar" />
  <div :foo='bar' />
  <div v-bind:foo="bar" />
  <div v-bind:foo='bar' />

  <!-- ✗ BAD -->
  <div :foo />
  <div v-bind:foo />
</template>
```

**Additional context:**

* #2355 
* #2357
* #2326

waynzh:
I’d be happy to help out with this one!
And I've noticed that the AST structure generated by both ways seems to be the same, as per the [eslint-vue-parser](https://github.com/vuejs/vue-eslint-parser/blob/master/src/template/index.ts#L731). It looks like we can't differentiate them just by checking the `directive.value`.

Would love to get some insights on how we might tell them apart and get it working. Any thoughts?

waynzh:
> We already have a function that checks if v-bind same-name shorthand.

[...]

PR #2374, opened 2024-01-23
feat: add new `v-bind-same-name-style` rule

waynzh:
close https://github.com/vuejs/eslint-plugin-vue/issues/2359.

Still figuring out how to categorize and set default value for this new rule. For now, I've set `never` as the default to avoid causing any breaking changes.


ota-meshi:
Thank you for working on this change.
However, I would rather add new options to existing [vue/v-bind-style] rule than add new rule. What do you think?
(Sorry, I should have commented on this earlier.)

[vue/v-bind-style]: https://eslint.vuejs.org/rules/v-bind-style.html

waynzh:
> I would rather add new options

Thanks for your input.

I've been contemplating whether to integrate this feature within the existing rule or to implement it as a separate one. I'm later leaning towards a separate rule as this rule particularly targets for cases where the `key` and `value` are the same name, one could use in `longform` and still omit the `value`.

Before next move, I'd like to gather opinions on the best way to incorporate this feature in [vue/v-bind-style](https://eslint.vuejs.org/rules/v-bind-style.html). Here are some options I'm considering:

1. Introduce a `sameNameShorthand` enum option that keeps the rule in shorthand form (e.g., `:foo=”bar”`) and omits the value only when it has same name `:foo="foo"` → `:foo` .
2. Modify the configuration to accept an array, allowing for combinations like `['longform', 'sameNameShorthand']`.
3. I'm open to other suggestions as well.

What do you think? I'm looking forward to your insights.

ota-meshi:
My idea is to add a third object option.
This is probably what @FloEdelmann wrote in #2359.

```jsonc
{
   "vue/v-bind-style": [
     "error",
     "shorthand", // "shorthand", or "longform"
     {
       "sameNameShorthand": "ignore" // "always", "never", or "ignore"
     }
   ]
}
```

waynzh:
Yeah, that makes sense to me. I'll close this PR for now, and when I find some time, I'll complete it and then open a new one.

Issue #2371, opened 2024-01-19
vue/define-macros-order does not support multiple defineModel macro calls


hendrikheil:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** v8.56.0
- **eslint-plugin-vue version:** v9.20.1
- **Vue version:** v3.4.15
- **Node version:** v18.18.2 
- **Operating System:** MacOS 14.1

**Please show your full configuration:**

```json5
module.exports = {
  extends: [
    'plugin:vue/base',
  ],

  parserOptions: {
    parser: {
      js: 'espree',
      jsx: 'espree',

      ts: '@typescript-eslint/parser',
      tsx: '@typescript-eslint/parser',
    },
  },
  rules: {
    "vue/define-macros-order": ["error", {
      order: ['defineOptions', 'defineModel', 'defineProps', 'defineEmits', 'defineSlots'],
      defineExposeLast: true,
    }]
  },
};
```

**What did you do?**	
I defined two defineModel calls as is allowed in vue 3.4.

**What did you expect to happen?**
Both defineModel calls are treated the first macro "group" and they are sorted alphabetically.

**What actually happened?**

```
npm run lint:ts


> vue-eslint-multiple-define-model@1.0.0 lint:ts
> eslint --ext .vue .


/Users/hendrik/projects/stafftastic/vue-eslint-multiple-define-model/reproduction.vue
  9:1  error  defineModel should be the first statement in `<script setup>` (after any potential import statements or type definitions)  vue/define-macros-order

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
  ```

**Repository to reproduce this issue**

https://github.com/hendrikheil/vue-eslint-multiple-define-model

PR #2386, opened 2024-01-30
fix: support multiple macro calls in vue/define-macros-order

hendrikheil:
Not a pro when it comes to AST stuff, but I think this should do it.

Should fix https://github.com/vuejs/eslint-plugin-vue/issues/2371

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -104,7 +104,7 @@ function create(context) {
   const order = (options[0] && options[0].order) || DEFAULT_ORDER
   /** @type {boolean} */
   const defineExposeLast = (options[0] && options[0].defineExposeLast) || false
-  /** @type {Map<string, ASTNode>} */
+  /** @type {Map<string, ASTNode[]>} */
   const macrosNodes = new Map()
   /** @type {ASTNode} */
   let defineExposeNode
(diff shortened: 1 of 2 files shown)


tests/lib/rules/define-macros-order.js
```diff
@@ -238,6 +238,22 @@ tester.run('define-macros-order', rule, {
...
+          const second = defineModel('second')
+
+          const slots = defineSlots()
+        </script>
+      `,
+      options: [
+        {
+          order: ['defineModel', 'defineSlots']
+        }
+      ]
```
ota-meshi, line 256: Thank you for this PR!
Could you also add invalid test cases with multiple defineModels?

ota-meshi (changes requested):

tests/lib/rules/define-macros-order.js
hendrikheil: Added a couple I could think of. Also one that tests that order of occurrence is maintained and not re-ordered like I suggested in the initial issue

Can you think of any other interesting cases to check?

lib/rules/define-macros-order.js
```diff
@@ -142,7 +144,10 @@ function create(context) {
           program
         )
         const orderedList = order
-          .map((name) => ({ name, node: macrosNodes.get(name) }))
+          .flatMap((name) => {
+            const nodes = macrosNodes.get(name) ?? []
+            return nodes?.map((node) => ({ name, node }))
```
FloEdelmann, line 149: This can't be undefined, since you are using `??` above.

```suggestion
            return nodes.map((node) => ({ name, node }))
```

hendrikheil: Good catch ;)

FloEdelmann (approved):
LGTM now :)

ota-meshi (approved):
LGTM! Thank you!

PR #2390, opened 2024-02-02
feat(no-deprecated-slot-attribute): improve autofix to wrap `<template v-slot>`

privatenumber:
Currently doing a Vue 2.7 migration in an old codebase...

I was thinking it should be possible to autofix cases of the deprecated `slot` attribute where the element is not `<template>`, by wrapping it in a `<template v-slot>`.

This PR demonstrates it, and the auto-fixes worked fine in my codebase. Want to get feedback to see if I'm missing any edge-cases.


FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thank you!

Issue #2262, opened 2023-08-01
Rule "vue/valid-v-model" is conflicted with typed value in template


chengzhuo5:
```html
<template>
  <!-- Error: 'v-model' directives require the attribute value which is valid as LHS. -->
  <input v-model="a as string" />
</template>
<script setup lang="ts">
import { ref} from "vue"
const a = ref<string|number>()
</script>
```

chengzhuo5:
Demo: https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJxNjsEKgzAQRH9l2VMLFe8SpR+SS2pXCSSbkN0UivXfG7GHnt/Mm9lwTk/CAY1SzMEpTZYBjOdcFV5dbDSMFh04AdHiebUIfQuZ/q9hZC4+KwhpzRAcr62jYrExH3MqChsUWmCHpaQIFl+VLFqeE4uCg/Gg5hz4cI0PKtPleqyc5glvmF0RKjhszdwe3/Wd6aQdSfCs/S+x71+QDEvf

PR #2393, opened 2024-02-03
feat(valid-v-model): add support for type assertions and non-null expressions

auvred:
Fixes https://github.com/vuejs/eslint-plugin-vue/issues/2262

Both `TSAsExpression` and `TSNonNullExpression` are valid `v-model` args: [vue playground](https://play.vuejs.org/#eNqFUk1PwzAM/Ssmlw1ptEK7TV0lQJPgwIcAiUsuVeuWTmkS5WNMKv3vOFk3dmDs1NrPjp/fc89utE42HtmCZbY0rXZg0XkNopDNkjNnOcu5bDutjIMeDNYwQG1UBxNqmxygO9XpMZ+kIQivEsxlqaR10NkGlqE9s860soFvkF6IfDq5RyEUfCgjqovJJZdZuiNCYylw2GlROKQIIPu8zvs+vjUMWUpRzMbZm6tOVSiIc4ALC7s5nEF6qupiBLP0aAqb0dLEuW6bZG2VJGX60M9ZSQ+0As2zdi3txNkCIhKwMKx0IeWMx9k+vbbbkOPsxaBFs0HODpgrTIOxhbPV2xNu6f8AEksvqPof8BWtEj5Q2ZXdelkRu6O6goT9eoj+kBLvdrV1KO2eeyAaKodYzxn5FSQ6teEv3Xkyj31cDiTW3uszFwQw3kGQny6hwrqV+Bii8SLy6RnzW6m9O3YwfP+0cPgBZS7zuw==)

[`TSNonNullExpression` spec](https://github.com/typescript-eslint/typescript-eslint/blob/9883ee275e79006f048813da851f00ba34f09b92/packages/ast-spec/src/expression/TSNonNullExpression/spec.ts#L5)

### DIFF
--- a/lib/rules/valid-v-model.js
+++ b/lib/rules/valid-v-model.js
@@ -49,7 +49,12 @@ function isOptionalChainingMemberExpression(node) {
  * @returns {boolean} `true` if the node can be LHS.
  */
 function isLhs(node) {
-  return node.type === 'Identifier' || node.type === 'MemberExpression'
+  return (
+    node.type === 'Identifier' ||
+    node.type === 'MemberExpression' ||
+    node.type === 'TSAsExpression' ||
+    node.type === 'TSNonNullExpression'
+  )
 }
 
 /**
--- a/tests/lib/rules/valid-v-model.js
+++ b/tests/lib/rules/valid-v-model.js
@@ -143,6 +143,24 @@ tester.run('valid-v-model', rule, {
     {
       filename: 'comment-value.vue',
       code: '<template><MyComponent v-model="/**/" /></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><MyComponent v-model="a as string"></MyComponent></template>',
+      languageOptions: {
+        parserOptions: {
+          parser: require.resolve('@typescript-eslint/parser')
+        }
+      }
+    },
+    {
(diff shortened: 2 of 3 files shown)


lib/rules/valid-v-model.js
```diff
@@ -49,7 +49,12 @@ function isOptionalChainingMemberExpression(node) {
  * @returns {boolean} `true` if the node can be LHS.
  */
 function isLhs(node) {
-  return node.type === 'Identifier' || node.type === 'MemberExpression'
+  return (
+    node.type === 'Identifier' ||
+    node.type === 'MemberExpression' ||
+    node.type === 'TSAsExpression' ||
+    node.type === 'TSNonNullExpression'
```
ota-meshi, line 56: Hmm. I think we need to check if the actual expression in LHS. For example, `foo()!` cannot be used in `v-model`, but I think it will be allowed with this change.

ota-meshi (changes requested):

lib/rules/valid-v-model.js
auvred: Thanks for catching this! I missed it somehow...

ota-meshi (approved):
LGTM! Thank you!

PR #2407, opened 2024-02-19
feat: support flat configs

aladdin-add:
the main changes:
* flat configs are exported as `flat/*`.
* ~~flat configs are exported as arrays.~~
* configs naming: 
  * `flat/recommended`is for vue 3. 
  * `flat/vue2-recommended` is for vue 2.

resubmit https://github.com/vuejs/eslint-plugin-vue/pull/2319.
fixes https://github.com/vuejs/eslint-plugin-vue/issues/1291


lib/index.js
FloEdelmann, line 1: Would it make sense to only rename the flat configs for now? Then it wouldn't be a breaking change and could be tested before releasing a major version.

We could also introduce (additional) `vue2-` prefixed (non-flat) configs, so migrating for people using the non-prefixed Vue 2 presets would already be possible.

aladdin-add: > Would it make sense to only rename the flat configs for now? Then it wouldn't be a breaking change and could be tested before releasing a major version.

can do. but the difficulties are:
* The configs were auto generated, so This inconsistency leads to complexity in implementing.
* Intuitively users will think `flat/xxx` => `xxx` is 1 to 1. so, it might cause confusion.

re the releasing, we (maybe) can push a pre-release for testing? 🤔

FloEdelmann: @ota-meshi what do you think?

ganondev:
Hi, just wanted to share some of the work challenges I had getting this to work in the context of my vue3+typescript project.
The work I did was from the original @conradhale fork, but I checked out @aladdin-add as well and it seems to be missing one thing I had to add:

* Setting `plugins: { vue: module },` for every flat config option wasn't working out. I ended up changing the generator to set a recursive reference for each flat config at the bottom of `index.js` after declaring the export object. Here is an example:

```js
module.exports.configs['flat/base'].plugins.vue = module.exports;
```

https://github.com/ganondev/eslint-plugin-vue/blob/master/tools/update-lib-index.js#L62

This is what the `eslint.config.js` for the project using this looks like with that update:

```js
import eslint from '@eslint/js';
import tseslint from 'typescript-eslint';

import vue from 'eslint-plugin-vue';


export default [
  {
    ignores: [
      'api/',
      'node_modules/',
      'dist/',
      '.yarn/',
      '.pnp.cjs',
      '.pnp.loader.mjs'
    ],
  },
  {
    ...eslint.configs.recommended,
    languageOptions: {
      ...eslint.configs.recommended.languageOptions,
      globals: {
        require: 'readonly',
        console: 'readonly',
      }
    },
  },
  ...tseslint.configs.recommended,
  {
    files: ['**/*.vue'],
    ...vue.configs["flat/vue3-recommended"],
    languageOptions: {
      ...vue.configs["flat/vue3-recommended"].languageOptions,
      parserOptions: {
        ...vue.configs["flat/vue3-recommended"].languageOptions.parserOptions,
        parser: '@typescript-eslint/parser'
      }
    },
  },
  {
    rules: {
      indent: ['error', 2],
    }
  },

];
```

Please let me know if this is how this is intended to look. It would be nice if I could parameterize the typescript parser option in, instead of flatting out down to the shared `parserOptions`.
Also, it's worth noting that with the other shared configs, the vue one **must** be the last one, otherwise the typescript or eslint config will attempt to process `.vue` and throw a fit. I think this should be documented, not sure how much control the vue plugin would have to be able to soften that.

Unfortunately, I can't put export change up against the active fork here since it's a different upstream, but I wanted to make sure this at least gets shared.

aladdin-add:
@ganondev thanks for trying it - it seems a bug, I've pushed a new commit, should be fixed now.

lib/index.js
ota-meshi: I think it's a good to only rename the flat configs to avoid/reduce breaking changes.
We can consider renaming legacy configs in the next major version.

> it might cause confusion

It may be confusing as you say, but now that Vue2 has reached EOL, I would like to drop the `vue3-` prefix. So I think not using `flat/vue3-` could reduce confusion in the next major version.

aladdin-add: well, I'll change it, and skip the flat config being auto-generated.

lib/configs/flat/base.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @fileoverview flat config - unlike eslintrc configs, it's not auto generated.
```
FloEdelmann, line 2: Hmm, why are flat configs no autogenerated? I think they have to be autogenerated, so we don't forget to add rules to all relevant configs.

aladdin-add: not familiar with the generating logic, it seems hard to share categories for eslintrc and flat configs?

https://github.com/vuejs/eslint-plugin-vue/blob/024742478ca726d6b649b5102379e1ed34e72889/tools/lib/categories.js#L10-L38

aladdin-add: you can always add it back when having consistent naming with eslintrc.(maybe next major version). thoughts?

FloEdelmann: Hmm, the logic is not that complicated:

```js
function getFlatConfigName(categoryTitle) {
  if (categoryTitle === 'base') {
    return 'flat/base'
  }
  if (categoryTitle.startsWith('vue3-') {
    return `flat/${categoryTitle.replace('vue3-', '')}`
  }
  return `flat/vue2-${categoryTitle}`
}
```

I'd prefer having it autogenerated right now.

aladdin-add: well, I will give it a try.

aladdin-add: added it in ca03026e2b1078791e7bb5f9e77207dd74c458c8.

aladdin-add:
friendly ping @ota-meshi

docs/rules/index.md
```diff
@@ -35,18 +35,17 @@ Rules in this category are enabled for all presets provided by eslint-plugin-vue
 ## Priority A: Essential (Error Prevention)
 
 - :three: Indicates that the rule is for Vue 3 and is included in all of `"plugin:vue/vue3-essential"`, `"plugin:vue/vue3-strongly-recommended"` and `"plugin:vue/vue3-recommended"` presets.
-- :two: Indicates that the rule is for Vue 2 and is included in all of `"plugin:vue/essential"`, `"plugin:vue/strongly-recommended"` and `"plugin:vue/recommended"` presets.
+- :two: Indicates that the rule is for Vue 2 and is included in `"plugin:vue/essential"` preset.
```
FloEdelmann, line 38: All changes in this file are unexpected. Did you forget to run `npm run update` or did the configs actually change? If so, please make sure that the existing configs stay the same.

lib/utils/config-helpers.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @fileoverview eslint config helpers
+ * @author 唯然<weiran.zsd@outlook.com>
+ */
+
+exports.extendsRules = function (config, rules) {
```
FloEdelmann, line 6: I think this function should be called `extendRules`, not `extendsRules`. Also, please use `module.exports` instead of only `exports` to stay consistent with other files.

tools/update-docs.js
```diff
@@ -123,7 +123,9 @@ class DocFile {
     }
     if (meta.docs?.categories) {
       const presets = getPresetIds(meta.docs.categories).map(
-        (categoryId) => `\`"plugin:vue/${categoryId}"\``
+        // non-prefixed configs are for vue2
+        // e.g. 'recommended' -> 'vue2-recommended'
```
FloEdelmann, line 127:
```suggestion
        // e.g. 'vue2-recommended' -> 'recommended'
```

tools/update-lib-index.js
```diff
@@ -13,7 +13,7 @@ const fs = require('fs')
 const path = require('path')
 const { FlatESLint } = require('eslint/use-at-your-own-risk')
 const rules = require('./lib/rules')
-const configs = require('./lib/configs')
+// const configs = require('./lib/configs')
```
FloEdelmann, line 16: If this is no longer needed, please remove this line

docs/rules/index.md
aladdin-add: yes, `npm run update` got executed. maybe it was affected by changing the `rule.meta.docs.categories: xxx -> vue2-xxx`. will take a look later.

lib/utils/config-helpers.js
aladdin-add: will do! 👍  (I thought there would be more than one function to export,)

aladdin-add: sidenote: you may want to enable `exports-style` to enforce it.

https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/exports-style.md

lib/index.js
```diff
@@ -276,3 +266,48 @@ module.exports = {
...
+  essential: require('./configs/vue2-essential'),
+  'no-layout-rules': require('./configs/no-layout-rules'),
+  recommended: require('./configs/vue2-recommended'),
+  'strongly-recommended': require('./configs/vue2-strongly-recommended'),
+  'vue3-essential': require('./configs/vue3-essential'),
+  'vue3-recommended': require('./configs/vue3-recommended'),
+  'vue3-strongly-recommended': require('./configs/vue3-strongly-recommended'),
+
+  // flat configs
+  'flat/base': Object.assign(baseConfig, require('./configs/flat/base.js')),
```
ota-meshi, line 284: Wouldn't this result in them sharing the same object? It might be safe to use spread instead.

```suggestion
  'flat/base': {...baseConfig, ...require('./configs/flat/base.js')},
```

ota-meshi (changes requested):

lib/index.js
aladdin-add: good catch! 👍  
should be addressed in 95ddc6f4c65555c3e8fa0b027b22f0d4fbbc04fb.

tests/lib/configs/configs.js
```diff
@@ -3,8 +3,13 @@
 const { ESLint } = require('../../eslint-compat')
 const plugin = require('../../../lib/index')
 
-describe('configs', () => {
+// TODO: test flat configs
```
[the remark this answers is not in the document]
aladdin-add: added in https://github.com/vuejs/eslint-plugin-vue/pull/2407/commits/56dac4476630908450b7a7fa48a7456ca25fdb7a.

tools/update-lib-index.js
```diff
@@ -13,7 +13,6 @@ const fs = require('fs')
 const path = require('path')
 const { FlatESLint } = require('eslint/use-at-your-own-risk')
 const rules = require('./lib/rules')
-const configs = require('./lib/configs')
```
FloEdelmann, line 16: Is that file still needed with your changes? Then please remove it.

tools/update-lib-index.js
```diff
@@ -52,6 +46,33 @@ module.exports = {
     }
   }
 }
+
+const baseConfig = {plugins: {vue: mod}}
+
+mod.configs = {
```
FloEdelmann, line 52: I'm not sure why you first define `mod` and then set its `configs` property. Can we inline it again?

tests/lib/configs/flat.js
```diff
@@ -0,0 +1,98 @@
...
+
+describe('flat configs', () => {
+  it('should export base config', () => {
+    const base = plugin.configs['flat/base']
+    assert.ok(base)
+    assert.equal(typeof base, 'object')
+    assert.strictEqual(base.plugins.vue, plugin)
+    assert.strictEqual(base.rules['vue/comment-directive'], 'error')
+  })
+  it('should export essential config', () => {
```
FloEdelmann, line 19:
```suggestion
  })

  it('should export essential config', () => {
```

FloEdelmann (commented):
Thanks for adding the tests. I think this looks quite good already! I have just a few more minor nits, then this is ready to go from my side :slightly_smiling_face: 

If you still don't have enough, an improvement (for a follow-up PR) would be adding types for the configs, see [eslint/eslint#18093 (comment)](https://togithub.com/eslint/eslint/issues/18093#issuecomment-1951489312). But I can totally understand if you don't want to put more effort into this, so everyone else reading this, feel free to go ahead! I think that adding non-typed flat config support is the step with the most impact though, so thanks a lot for putting this forward!

tools/update-lib-index.js
aladdin-add: it's to ref the mod: `plugins: {vue: mod}`.  
you can see it in the eslint migration docs: https://eslint.org/docs/head/extend/plugin-migration-flat-config#migrating-configs-for-flat-config :)

tools/update-lib-index.js
aladdin-add: addressed in b88edc82238703965e11633c5b58e268e16bd238.

tools/update-lib-index.js
FloEdelmann: Ah, I see. Can we rename the variable to `plugin` then, please?

aladdin-add: https://github.com/vuejs/eslint-plugin-vue/pull/2407/commits/ae60a4af72728d2ed45ff610977a3c55f49f0693

FloEdelmann (approved):
Thank you!

ota-meshi (approved):
LGTM! Thank you!

Issue #2409, opened 2024-02-26
`sameNameShorthand` from `vue/v-bind-style` confuses hyphens with underscores


adamsol:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 8.41.0
- **eslint-plugin-vue version:** 9.22.0
- **Vue version:** 3.4.20
- **Node version:** 14.19.3
- **Operating System:** Windows 10

**Please show your full configuration:**

```json5
module.exports = {
  extends: ['plugin:vue/vue3-recommended'],
  rules: {
    'vue/v-bind-style': [1, 'shorthand', { sameNameShorthand: 'always' }],
  },
  ignorePatterns: ['node_modules/'],
};
```

**What did you do?**	

```vue
<template>
  <div :test-value="test_value" />
</template>
```

```npm run lint```

**What did you expect to happen?**
No warning.

**What actually happened?**

```
/home/projects/vitejs-vite-uc4css/src/App.vue
  2:8  warning  Expected same-name shorthand  vue/v-bind-style

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

**Repository to reproduce this issue**

https://stackblitz.com/edit/vitejs-vite-uc4css?file=src%2FApp.vue

FloEdelmann:
Thanks for the bug report. PRs welcome!

PR #2410, opened 2024-02-27
fix(v-bind-style): only change kebabCase to camelCase

waynzh:
fix https://github.com/vuejs/eslint-plugin-vue/issues/2409

### DIFF
--- a/tests/lib/rules/v-bind-style.js
+++ b/tests/lib/rules/v-bind-style.js
@@ -259,6 +259,13 @@ tester.run('v-bind-style', rule, {
     },
     {
       filename: 'test.vue',
+      code: '<template><div :foo_bar="foo_bar" /></template>',
+      output: '<template><div :foo_bar /></template>',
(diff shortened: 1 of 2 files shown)


tests/lib/rules/v-bind-style.js
FloEdelmann, line 1: Could you please add another test case `<div :foo_bar="foo_bar" />` that reports an error suggesting to fix this to `<div :foo_bar />` if the `sameNameShorthand` option is set to `always`?

FloEdelmann (commented):
Thanks for the fix! I have one suggestion for an additional test case, otherwise this looks good to me already :slightly_smiling_face:

tests/lib/rules/v-bind-style.js
waynzh: Update ✅

FloEdelmann (approved):
Nice, thank you!

ota-meshi (approved):
LGTM! Thank you!

PR #2411, opened 2024-02-27
Remove dead link from docs

travis5491811:
Removed further reading section. The link doesn't provide reader more information about the subject. It just redirects to a datadog marketing page.


FloEdelmann (approved):
Thanks for finding and fixing!

PR #2416, opened 2024-03-05
fix(v-for-delimiter-style): ignore Punctuator token

waynzh:
fixes https://github.com/vuejs/vue-eslint-parser/issues/226 (`vue-eslint-parser` issue!)

Maybe we can simply ignore all 'Punctuator' type token, without the need to specifically declare `token.type !== 'Punctuator' || (token.value !== ')' && token.value !== ',')`?

### DIFF
--- a/lib/rules/v-for-delimiter-style.js
+++ b/lib/rules/v-for-delimiter-style.js
@@ -42,7 +42,7 @@ module.exports = {
             node.left.length > 0
               ? node.left[node.left.length - 1]
               : tokenStore.getFirstToken(node),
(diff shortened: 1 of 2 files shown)


FloEdelmann (commented):
Can you please add some test cases with `v-for="(key, value) in items"` and check that the fix does not break them?

FloEdelmann (approved):
Looks good to me, thanks! :slightly_smiling_face:

ota-meshi (approved):
LGTM! Thank you!

PR #2417, opened 2024-03-05
perf: remove unnecessary second filter for type narrowing

waynzh:
resolves this https://github.com/vuejs/eslint-plugin-vue/pull/2397#discussion_r1479773243 as a perf improvement


FloEdelmann (approved):
Thanks! This looks like a nice easy yet effective improvement 🙂

PR #2423, opened 2024-03-11
feat(order-in-components): add side effects suggestions

### LINKED ISSUE #2418: Offer suggestions in `vue/order-in-components` when autofix is not available
In [`vue/order-in-components`](https://eslint.vuejs.org/rules/order-in-components.html), autofix is only enabled when we are sure that the things we are reordering are free of side effects, to not break the code. However, the automatic reordering is really useful.

I suggest to add the (unsafe) reordering as a suggestion, so the user can decide for themselves whether to automatically reorder the rule in this case.

*(Originially reported in #2272)*

waynzh:
closes #2418

### DIFF
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -259,6 +259,30 @@ module.exports = {
     }
 
     /**
+     * @param {Property} propertyNode
+     * @param {Property} unorderedPropertyNode
+     */
+    function* handleFix(fixer, propertyNode, unorderedPropertyNode) {
+      const afterComma = sourceCode.getTokenAfter(propertyNode)
+      const hasAfterComma = isComma(afterComma)
+
+      const beforeComma = sourceCode.getTokenBefore(propertyNode)
+      const codeStart = beforeComma.range[1] // to include comments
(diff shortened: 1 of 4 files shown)


lib/rules/order-in-components.js
```diff
@@ -231,7 +232,9 @@ module.exports = {
     ],
     messages: {
       order:
-        'The "{{name}}" property should be above the "{{firstUnorderedPropertyName}}" property on line {{line}}.'
+        'The "{{name}}" property should be above the "{{firstUnorderedPropertyName}}" property on line {{line}}.',
+      hasSideEffect:
+        'There may be some side effects, the "{{name}}" property should be manually moved above the "{{firstUnorderedPropertyName}}" property on line {{line}}.'
```
FloEdelmann, line 237: Please rename this message to `reorderWithSideEffects` and change its wording to `Manually move "{{name}}" property above "{{firstUnorderedPropertyName}}" property on line {{line}} (might break side effects).`

FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thank you!

PR #2427, opened 2024-03-14
Fix checking for script setup in vue/require-typed-ref

kovalchukq:


tests/lib/rules/require-typed-ref.js
waynzh, line 1: Perhaps it also needs a invalid test case for this `export default setup`? Something like:
```
<script lang="ts">
  import { ref } from 'vue'
  export default {
    setup() {
      const count = ref()
    }
  }
</script>
```

kovalchukq: Fixed

FloEdelmann (commented):
Thanks! Could you please fix the lint error (run `npm run lint:fix`)? Otherwise, this looks good to me.

ota-meshi:
Hmm... What is this PR fixing? Could you please explain the details?

kovalchukq:
> Hmm... What is this PR fixing? Could you please explain the details?
This PR skip checking `require-typed-ref ` rule for vue files that doesn't have lang="ts".

ota-meshi (approved):
Sorry for the slow response.
Thank you for your explanation. 
LGTM! Thank you!

SamyDgz (approved):

Issue #2429, opened 2024-03-15
Rule Proposal: Maximum number of props 


kevsommer:
**Please describe what the rule should do:**
The rule should allow developers to set a maximum number of props for Vue components in a project. 
In my personal experience a high of number of props correlates strongly with complexity of a component. 
Therfore, I think it would be a good addition to the vue eslint rule set, to give developers the option to enforce this rule.

**What category should the rule belong to?**

[x] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<!-- BAD with maxProps set to 5 -->
<script setup>
defineProps({ 
  prop1: String,
  prop2: String,
  prop3: String,
  prop4: String,
  prop5: String,
  prop6: String,
  prop7: String,
  prop8: String,
})
</script>
```

```vue
<!-- GOOD with maxProps set to 5 -->
<script setup>
defineProps({ 
  prop1: String,
  prop2: String,
  prop3: String,
})
</script>
```

**Additional context**
So far I haven't found a similar rule proposal.
But given the ruleset includes the vue/max-len rule, which similarly helps control the complexity of components, I think it would be a good addition.

PR #2430, opened 2024-03-15
implements proposed max-props rule

kevsommer:
This is a WIP PR that addresses issue #2429. 

It implements the rule for both, the Composition and Options API.

However, I haven't been able to get the case for `defineProps` with type-only declaration to work.

```vue
<script setup lang="ts">
defineProps<{ prop1: string, prop2: string, }>();
</script>
```


kevsommer:
Should work for `defineProps` with type-only declaration now.

docs/rules/max-props.md
```diff
@@ -0,0 +1,64 @@
...
+  prop1: String
+})
+</script>
+```
+
+</eslint-code-block>
+
+Nothing.
+
+## :mag: Implementation
```
FloEdelmann, line 61: "Nothing." is only the placeholder for when there are no rule options.

```suggestion
</eslint-code-block>

## :mag: Implementation
```

lib/rules/max-props.js
```diff
@@ -0,0 +1,69 @@
...
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          maxProps: {
+            type: 'integer',
+            minimum: 1
+          }
+        },
+        additionalProperties: false
```
FloEdelmann, line 26:
```suggestion
        additionalProperties: false,
        minProperties: 1
```

lib/rules/max-props.js
```diff
@@ -0,0 +1,69 @@
...
+    messages: {
+      tooManyProps:
+        'Component has too many props ({{propCount}}). Maximum allowed is {{limit}}.'
+    }
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    /**
+     * @type {Record<string, number>}
+     */
```
FloEdelmann, line 38:
```suggestion
    /** @type {Record<string, number>} */
```

lib/rules/max-props.js
```diff
@@ -0,0 +1,69 @@
...
+    const option = context.options[0] || {}
+
+    /**
+     *
+     * @param {import('../utils').ComponentProp[]} props
+     */
```
FloEdelmann, line 44:
```suggestion
    /**
     * @param {import('../utils').ComponentProp[]} props
     */
```

tests/lib/rules/max-props.js
```diff
@@ -0,0 +1,189 @@
...
+})
+
+tester.run('max-props', rule, {
+  valid: [
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+
+      </template>
```
FloEdelmann, line 25: Please remove the empty `<template>` blocks from the test cases, they don't add any value.

FloEdelmann (changes requested):
Thanks, this looks quite good! :slightly_smiling_face: 
I only have some minor nitpicks.

kevsommer:
Thank you for the review! I have implemented the requested changes :)

FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thanks!

PR #2432, opened 2024-03-21
feat: no-props-shadow

### LINKED ISSUE #2287: Rule Proposal: vue/no-props-shadow
This is similar to #2172 but I would like to allow the use of implicit props and would like to warn/error when a variable shadowing a prop is defined. 

```vue
<script setup lang="ts">
import {computed} from 'vue';

const props = defineProps<{ isDisabled: boolean }>();

// Creating variables with the same name as props 
// creates an ambiguity in the template
const isDisabled = computed(() => props.isDisabled ? 'yes' : 'no'));
</script>

<template>
  <!-- Is this `isDisabled.value` or `props.isDisabled` ? -->
  <h1>Disabled: {{ isDisabled }}</h1>
</template>
```

- [X] Warns about a potential error (problem)
- [ ] Other (please specify:)

```vue
<script setup lang="ts">
import {ref} from 'vue';
const props = defineProps<{foo: boolean}>();

// BAD
const foo = ref(false);
</script>
```
```vue
<script setup lang="ts">
import {computed} from 'vue';
const props = defineProps<{foo: boolean}>();

// BAD
const foo = computed(() => {
  return props.foo ? 'yes' : 'no';
});
</script>
```
```vue
<script setup lang="ts">
defineProps<{ foo: 'yes' | 'no' }>();

// BAD
const foo = false;
</script>
```

**Additional context**
It might be worthwhile to abstract #2172 and this proposal into a rule that can support either with options as they do seem fairly closely related.

RiESAEX:
closes #2287


lib/rules/no-props-shadow.js
```diff
@@ -0,0 +1,76 @@
...
+      onDefinePropsEnter(_, props) {
+        propSet = new Set(
+          props
+            .map((p) => p.propName)
+            .filter(
+              /**
+               * @returns {propName is string}
+               */
+              (propName) =>
+                utils.isDef(propName) && !GLOBALS_WHITE_LISTED.has(propName)
```
RiESAEX, line 44: Maybe there should be another rule disallow people defining props in GLOBALS_WHITE_LISTED?
BTW, I found an obsolete GLOBALS_WHITE_LISTED in "no-mutating-props" but don't know what to do  https://github.com/vuejs/eslint-plugin-vue/blob/f84b8ee5b9e3b54a09209faff744ef2849a6cc3e/lib/rules/no-mutating-props.js#L14-L20

RiESAEX:
@FloEdelmann @ota-meshi  hey, Is there anything I can do to make this be merged?

PR #2436, opened 2024-03-25
Change flat config to an array to accept `*.vue`

ota-meshi:
This PR changes the flat config to an array to accept `*.vue` extensions.

I noticed that using the following flat config I receive the following message:

> File ignored because of a matching ignore pattern. Use "--no-ignore" to disable file ignore settings or use "--no-warn-ignored" to suppress this warning.

```js
const plugin = require('eslint-plugin-vue')
module.exports = [
  plugin.configs['flat/base'],
  {
    rules: {
      'vue/no-duplicate-attributes': 'error'
    }
  }
]
```

I think this is because the flat config does not include the `*.vue` extension in `files`.

I was thinking of adding `files: ['*.vue']` to all configs, but that doesn't work because some of the rules of this plugin can also be used in `*.js` files. Therefore, we added `files: ['*.vue']` only to the config that has a parser and modified it to provide the config as an array.

@aladdin-add  I would like to hear your opinion if possible.


lib/configs/flat/vue2-essential.js
```diff
@@ -5,72 +5,74 @@
...
+      'vue/valid-v-for': 'error',
+      'vue/valid-v-html': 'error',
+      'vue/valid-v-if': 'error',
+      'vue/valid-v-model': 'error',
+      'vue/valid-v-on': 'error',
+      'vue/valid-v-once': 'error',
+      'vue/valid-v-pre': 'error',
+      'vue/valid-v-show': 'error',
+      'vue/valid-v-slot': 'error',
+      'vue/valid-v-text': 'error'
```
aladdin-add, line 75: If no `files` is specified, `(m|c)?js` will be used by default. 

question: is all these rules only applied to js files (no vue files)?

aladdin-add:
I always like to export an object (if possible). the usage can be like:

```js
const plugin = require('eslint-plugin-vue');

module.exports = [
  {files: ["**/*.js", "**/*.vue"], plugin.configs['flat/base']},
];
```

If you export the array format it will be difficult for the user to combine them, e.g. to add "*.ts".

lib/configs/flat/vue2-essential.js
ota-meshi: I have added test cases and they work fine in `*.vue` as well.
Am I misunderstanding something?

https://github.com/vuejs/eslint-plugin-vue/pull/2436/commits/1438a8dfcb591df64e7a58503c7b959dc9b2fddd

ota-meshi:
I think that the only way to properly specify the extension to apply the configuration provided in the array is as follows.

```js
const plugin = require('eslint-plugin-vue');
module.exports = [
  ...plugin.configs['flat/base'].map((config) => ({
    ...config,
    files: ["**/*.js", "**/*.vue"]
  })),
]
```

I agree that objects are easier to handle, but I don't think it's possible in this case 🤔

lib/configs/flat/vue2-essential.js
aladdin-add: seems it's working fine with the preprocesser.

tests/integrations/flat-config.js
```diff
@@ -0,0 +1,43 @@
...
+    ) {
+      return
+    }
+
+    const result = JSON.parse(
+      cp.execSync(`${ESLINT} a.vue --format=json`, {
+        encoding: 'utf8'
+      })
+    )
+    // console.log(JSON.stringify(result, null, 2))
```
aladdin-add, line 39: should be removed. :)

aladdin-add (approved):
LGTM.

ota-meshi:
Thank you for the review! I have fixed the documentation.

PR #2441, opened 2024-04-04
Support Pinia methods in `vue/no-unused-properties`

### LINKED ISSUE #1675: Vuex and Pinia support

- **ESLint version:** 7.x.x
- **eslint-plugin-vue version:** 7.20.0
- **Node version:** 14.x.x

When mapping props with Vuex's `mapState` & `mapGetters` and methods with `mapMutations` & `mapActions`, `vue/no-unused-properties` doesn't report those props/methods if they're unused. Also, `vue/no-undef-properties` reports those props as undefined if they're used. 

Considering Vuex is a core part of the Vue ecosystem & is widely used, the aforementioned rules should take mapped props & methods into account.

**Additional context**
image

luizotcarvalho:
Changes are aimed at improving Pinia support.

Following Pinia methods will be checked: `mapGetters`, `mapState`, `mapWritableState` and `mapActions`, when no-unused-properties will be enabled.

Its not possible to support `mapStores` easily because de result variable is the store id + Store ([See more](https://pinia.vuejs.org/cookbook/options-api.html#Giving-access-to-the-whole-store)) and would fail always.

Fixes #1675.

### DIFF
--- a/lib/rules/no-unused-properties.js
+++ b/lib/rules/no-unused-properties.js
@@ -496,12 +496,16 @@ module.exports = {
           let groupName = null
           if (/^mapMutations|mapActions$/u.test(node.callee.name)) {
             groupName = 'methods'
-          } else if (/^mapState|mapGetters$/u.test(node.callee.name)) {
+          } else if (
+            /^mapState|mapGetters|mapWritableState$/u.test(node.callee.name)
+          ) {
             groupName = 'computed'
           }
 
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):
Oh wow, that was fast! And the code looks good to me. Thanks :slightly_smiling_face:

ota-meshi (approved):
LGTM! Thank you!

PR #2442, opened 2024-04-06
feat: update default configs from `component-tags-order` to `block-order`

RayGuo-ergou:
### Description

This pr updated recommend rule sets from `component-tags-order` to `block-order` as `component-tags-order` is replaced by `block-order`.


ota-meshi:
Changing the config (swapping rules) is a breaking change, so we don't do that yet. 
Thank you.

PR #2450, opened 2024-04-15
fix(configs): remove deprecated component-tags-order rule

ferferga:
This rules is deprecated according to the documentation, so it should be removed from the config.


FloEdelmann:
Please see https://github.com/vuejs/eslint-plugin-vue/pull/2442.

PR #2456, opened 2024-04-26
fix: Prevent spread operator from disabling the rule

GuillaumeGuerinPP:
The rule `vue/require-default-prop` return an error for all optional props when using `withDefaults` and a spread operator as second argument

image

The rule seems to return an empty object if something different from `Property` is found. Ignoring it instead will allow the check for at least props defined in the file

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -3121,7 +3121,7 @@ function getWithDefaultsProps(node) {
 
   for (const prop of param.properties) {
     if (prop.type !== 'Property') {
-      return {}
+      continue
     }
     const name = getStaticPropertyName(prop)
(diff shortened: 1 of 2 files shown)


GuillaumeGuerinPP:
> Thanks for the fix! Could you please add one `valid` and one `invalid` test case to to the rule tests?

Thanks for you quick answer 🙌 Test added here : https://github.com/vuejs/eslint-plugin-vue/pull/2456/commits/edceab2fbb15548570312c3b54572afc713e2377

FloEdelmann (approved):
Nice, looks good to me!

ota-meshi (approved):
LGTM! Thank you!

PR #2457, opened 2024-04-29
Support MathML elements

mordae:
Add support for MathML elements same way SVG elements are supported.

For context, Vue now supports MathML out-of-the-box as well: https://github.com/vuejs/core/pull/7836


FloEdelmann (approved):
Looks good to me, thanks!

ota-meshi (approved):
LGTM! Thank you!

PR #2466, opened 2024-05-27
implements autofix in define-props-declaration (#2465)

### LINKED ISSUE #2465: Add autofix to `define-props-declaration`: transform runtime syntax to type-based syntax
`define-props-declaration`
Does not change the number of warnings.
Implementing a `fix` method in the `context.report` `descriptor` parameter.

The rule will support `separateInterface` option that will opt-in to define type passed to the `defineProps` as an interface above `defineProps` call.

The rule will be able to autofix the following code:
```vue
      <script setup lang="ts">
      const props = defineProps({
        kind: String
      })
      </script>
```
transforming it to:
```vue
      <script setup lang="ts">
      const props = defineProps<{ kind: string }>()
      </script>
```

It des not fix it.

The rule will have the autofix feature  transforming runtime syntax to type-based syntax
**Additional context**

mpiniarski:
Fixes #2465


mpiniarski:
@FloEdelmann can I get some attention to this PR?

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,168 @@
...
+      return 'boolean'
+    }
+    case 'Object': {
+      return 'Record<string, any>'
+    }
+    case 'Array': {
+      return 'any[]'
+    }
+    case 'Function': {
+      return '() => void'
```
FloEdelmann, line 27: I think all functions should be allowed by default, so we have to allow any number of arguments and any return type:

```suggestion
      return '(...args: any[]) => any'
```

tests/lib/rules/define-props-declaration.js
```diff
@@ -117,13 +121,42 @@ tester.run('define-props-declaration', rule, {
       })
       </script>
       `,
+      output: `
+      <script setup lang="ts">
+      const props = defineProps<{ kind: string }>()
```
FloEdelmann, line 126: If `required` is not set (or for shorthand prop types), shouldn't this result in optional props (i.e. `kind?: string`)?

lib/rules/define-props-declaration.js
```diff
@@ -14,10 +176,19 @@ module.exports = {
...
-    fixable: null,
+    fixable: 'code',
     schema: [
       {
         enum: ['type-based', 'runtime']
+      },
+      {
+        type: 'object',
+        properties: {
+          separateInterface: {
```
FloEdelmann, line 187: Please rename to `autoFixToSeparateInterface` or similar.

lib/rules/define-props-declaration.js
```diff
@@ -27,20 +198,87 @@ module.exports = {
...
-      onDefinePropsEnter(node) {
+      onDefinePropsEnter(node, props) {
         switch (defineType) {
           case 'type-based': {
             if (node.arguments.length > 0) {
               context.report({
                 node,
-                messageId: 'hasArg'
+                messageId: 'hasArg',
+                *fix(fixer) {
```
FloEdelmann, line 219: Please extract this function to the top scope to reduce indentation and thus make it easier to read.

lib/rules/define-props-declaration.js
```diff
@@ -27,20 +198,87 @@ module.exports = {
...
+                    )
+
+                    const definePropsType = `{ ${propTypes
+                      .map(
+                        ({ name, type, required, defaultValue }) =>
+                          `${name}${
+                            required === false || defaultValue ? '?' : ''
+                          }: ${type}`
+                      )
+                      .join(', ')} }`
```
FloEdelmann, line 232: Can you extract some variables here to make the template strings easier to read?

FloEdelmann (changes requested):
Thanks for the ping! I had an initial look now.

lib/rules/define-props-declaration.js
FloEdelmann: Also, please describe this option in the rule docs.

tests/lib/rules/define-props-declaration.js
mpiniarski: You're right. Fixed.

lib/rules/define-props-declaration.js
mpiniarski: done

mpiniarski:
@FloEdelmann applied changes, please review again

lib/rules/define-props-declaration.js
FloEdelmann: You refactored some code here, but it is still quite unreadable because the template expressions span multiple lines. I think something like this is much more readable:

```js
const definePropTypes = propTypes.map(
  ({ name, type, required, defaultValue }) => {
    const optionalSuffix = required === false || defaultValue ? '?' : ''
    return `${name}${optionalSuffix}: ${type}`
  }
)
const definePropTypesString = `{ ${definePropTypes.join(', ')} }`
```

mpiniarski: @FloEdelmann How about now?

docs/rules/define-props-declaration.md
```diff
@@ -39,15 +41,23 @@ const props = defineProps({
 ## :wrench: Options
 
 ```json
-  "vue/define-props-declaration": ["error", "type-based" | "runtime"]
+{
+  "vue/define-props-declaration": ["error",
+    "type-based" | "runtime",
+    {
+      "autoFixToSeparateInterface": false
```
ota-meshi, line 48: I don't think this is a good option, if there is more than one way to fix it, I think it should be implemented as a suggested feature.
Could you please use the suggestions API instead of auto-fix?

https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,234 @@
...
+          ({ name, defaultValue }) =>
+            `${name}: ${sourceCode.getText(defaultValue)}`
+        )
+        .join(PROPS_SEPARATOR)
+
+      yield fixer.insertTextBefore(node, `withDefaults(`)
+      yield fixer.insertTextAfter(node, `, { ${defaultsCode} })`)
+    }
+    return null
+  } catch (error) {
```
ota-meshi, line 71: Why do we need to catch errors? Could you please implement it to avoid that error?

ota-meshi (changes requested):
Thank you for working on that implementation!

It seems to be causing an error in CI. Could you please fix that as well?

lib/rules/define-props-declaration.js
mpiniarski: Sure. We don't need to catch afaik, as all the cases I've predicted in the tests are handled without an error. I'll remove that line.

docs/rules/define-props-declaration.md
mpiniarski: I'll just remove this option in that case. I don't want to overcomplicate it. 
Also, preferring inline types vs define separate one can be probably handled by a different linter rulse.

tests/lib/rules/define-props-declaration.js
```diff
@@ -10,8 +10,11 @@ const rule = require('../../../lib/rules/define-props-declaration')
 const tester = new RuleTester({
   languageOptions: {
     parser: require('vue-eslint-parser'),
-    ecmaVersion: 2020,
-    sourceType: 'module'
+    ecmaVersion: '2020',
```
FloEdelmann, line 13: To fix the failing tests, this has to be reverted:

```suggestion
    ecmaVersion: 2020,
```

docs/rules/define-props-declaration.md
FloEdelmann: Please also remove it from the docs again.

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,215 @@
...
+      const typeProperty = utils.findProperty(node, 'type')
+      if (typeProperty == null) {
+        return sourceCode.getText(node)
+      }
+      return optionGetType(typeProperty.value, sourceCode)
+    }
+    case 'ArrayExpression': {
+      return node.elements
+        .map((element) => {
+          // TODO handle SpreadElement
```
FloEdelmann, line 128: Is this TODO still open? If so, please resolve it and add a test case.

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,215 @@
...
+
+    return defaultProperty.value
+  }
+
+  // Unknown
+  return undefined
+}
+
+/**
+ * @typedef {import('../utils').ComponentProp} ComponentProp
```
FloEdelmann, line 215: Please move this to the top of the file, before it is used.

FloEdelmann:
@ota-meshi Could you please clear the Netlify cache and restart the deployment? That would make the rule (and docs) changes easier to test.

ota-meshi:
I cleared the cache and restart. The PR status remains failed, but it seems to have worked.
https://deploy-preview-2466--eslint-plugin-vue.netlify.app/rules/define-props-declaration.html

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,214 @@
...
+    }
+    default: {
+      return sourceCode.getText(node)
+    }
+  }
+}
+
+/**
+ * @param {Expression} node
+ * @returns {boolean | undefined }
```
FloEdelmann, line 181:
```suggestion
 * @returns {boolean}
```

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,214 @@
...
+    }
+  }
+
+  // Unknown
+  return false
+}
+
+/**
+ * @param {Expression} node
+ * @returns {Expression | undefined }
```
FloEdelmann, line 201:
```suggestion
 * @returns {Expression | undefined}
```

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,214 @@
...
+          `${name}: ${sourceCode.getText(defaultValue)}`
+      )
+      .join(PROPS_SEPARATOR)
+
+    yield fixer.insertTextBefore(node, `withDefaults(`)
+    yield fixer.insertTextAfter(node, `, { ${defaultsCode} })`)
+  }
+  return null
+}
+const mapNativeType = (/** @type {string} */ nativeType) => {
```
FloEdelmann, line 60:
```suggestion

/**
 * @param {string} nativeType
 * @returns {string}
 */
const mapNativeType = (nativeType) => {
```

FloEdelmann (approved):
Some more small JSDoc suggestions, then this is good to go from my side! Thanks for the contribution!

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,217 @@
...
+          `${name}: ${sourceCode.getText(defaultValue)}`
+      )
+      .join(PROPS_SEPARATOR)
+
+    yield fixer.insertTextBefore(node, `withDefaults(`)
+    yield fixer.insertTextAfter(node, `, { ${defaultsCode} })`)
+  }
+  return null
+}
+  /**
```
FloEdelmann, line 60: Lint failure:

```suggestion
}

/**
```

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,217 @@
...
+ * @param {string} nativeType
+ * @returns {string}
+ */
+const mapNativeType = (nativeType) => {switch (nativeType) {
```
FloEdelmann, line 64: Lint failure:

```suggestion
const mapNativeType = (nativeType) => {
  switch (nativeType) {
```

so1ve:
Should we fix it to multi-line interface, then make the separator configurable?

FloEdelmann:
No, let's keep it single-line and leave the rest to other ESLint rules or formatters like Prettier.

tests/lib/rules/define-props-declaration.js
```diff
@@ -108,6 +111,7 @@ tester.run('define-props-declaration', rule, {
     }
   ],
   invalid: [
+    // default
```
ota-meshi, line 114: Could you please add a test case to check if an error occurs when an array is used for props?

For example: 

```vue
      <script setup lang="ts">
      const props = defineProps(['kind'])
      </script>
```

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,219 @@
...
+  }
+}
+
+/**
+ * @param {ComponentProp} prop
+ * @param {SourceCode} sourceCode
+ */
+function getComponentPropData(prop, sourceCode) {
+  if (prop.propName === null) {
+    throw new Error('Unexpected prop with null name.')
```
ota-meshi, line 100: Why does it need to throw an error here?
Can't we make it guard against those before processing the autofix?

ota-meshi (changes requested):

tests/lib/rules/define-props-declaration.js
mpiniarski: I've implemented handling of such case - the kind is a not required of type string.

lib/rules/define-props-declaration.js
mpiniarski: What do you mean?
I thinks it's more clear if we throw an error here.

ota-meshi: I think that if we throw an error, the user's eslint command will fail.
Did you throw the error with that intention? If so, why is that?
If auto-fix is not possible, I think it should be handled so that auto-fix is not performed, rather than throwing an error.

tests/lib/rules/define-props-declaration.js
```diff
@@ -162,6 +633,26 @@ tester.run('define-props-declaration', rule, {
           line: 3
         }
       ]
+    },
+    // array
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup lang="ts">
+      const props = defineProps([kind])
```
ota-meshi, line 642: The array elements must be strings. The `kind` variable shouldn't auto-fix because we don't know what's actually in it.
Could you leave this test as a test case for when not to auto-fix and add a new test case using a string literal?

https://vuejs.org/guide/components/props.html#props-declaration

ota-meshi (changes requested):

lib/rules/define-props-declaration.js
mpiniarski: I see. Let me catch errors and ignore them.

I prefer to leave a sign of an error somewhere, maybe in `console.debug()`, but it is up to you.

tests/lib/rules/define-props-declaration.js
mpiniarski: My bad, you're right. It does not default to a `string`. Leaving it unhandled then.

mpiniarski:
@ota-meshi can you take a look?

lib/rules/define-props-declaration.js
```diff
@@ -6,6 +6,224 @@
...
+  }
+}
+
+/**
+ * @param {ComponentProp} prop
+ * @param {SourceCode} sourceCode
+ */
+function getComponentPropData(prop, sourceCode) {
+  if (prop.type === 'array') {
+    throw new Error(`Unable to resolve types based on array prop declaration.`)
```
ota-meshi, line 102: Could you please implement it so that the autofix stops without throwing an error?

ota-meshi (changes requested):

tests/lib/rules/define-props-declaration.js
ota-meshi: Could you please add the test case you requested earlier?

```vue
      <script setup lang="ts">
      const props = defineProps(['kind'])
      </script>
```

https://github.com/vuejs/eslint-plugin-vue/pull/2466#discussion_r1676641695

FloEdelmann:
It's been two years now without any activity, so I'm going to close this. Feel free to open a new PR!

Issue #2005, opened 2022-10-06
Rule suggestion: `vue/prefer-single-payload-in-event`


FloEdelmann:
**Please describe what the rule should do:**
When emitting an event, one can pass an event payload. This can consist of multiple parameters. However, when using the `always` style in the event handler, one can only receive the first event payload parameter.

```vue
<my-button @my-event="myHandler($event)" />
```

See also the [`vue/v-on-function-call` rule](https://eslint.vuejs.org/rules/v-on-function-call.html) and https://github.com/vuejs/eslint-plugin-vue/issues/2001#issuecomment-1269475268.

**What category should the rule belong to?**
[ ] Enforces code style (layout)
[x] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```ts
// BAD:
this.$emit('my-event', foo, bar)
emit('my-event', foo, bar)
defineEmits({ 'my-event': (foo, bar) => true })
defineEmits({ 'my-event': (foo: string, bar: string) => true })
defineEmits<{ (e: 'my-event', foo: string, bar: string): void }>()
{ emits: { 'my-event': (foo, bar) => true } }
{ emits: { 'my-event': (foo: string, bar: string) => true } }

// GOOD:
this.$emit('my-event', foo)
emit('my-event', foo)
defineEmits(['my-event'])
defineEmits({ 'my-event': null })
defineEmits({ 'my-event': (foo) => true })
defineEmits({ 'my-event': (foo: string) => true })
defineEmits<{ (e: 'my-event', foo: string): void }>()
{ emits: ['my-event'] }
{ emits: { 'my-event': null }
{ emits: { 'my-event': (foo) => true } }
{ emits: { 'my-event': (foo: string) => true } }
```

PR #2471, opened 2024-06-06
feat(v-on-handler-style): allow `["inline", "inline-function"]` option

### LINKED ISSUE #2460: Suggestion: allow `["inline", "inline-function"]` option for `v-on-handler-style`

Allow [`v-on-handler-style`](https://eslint.vuejs.org/rules/v-on-handler-style.html) rule to accept `["inline", "inline-function"]` as the first option.

I find method binding syntax `@event="handler"` unsafe as it's sometimes not clear what and how many arguments will be passed to the handler (for example, if a component is not typed). That said, only allowing inline event handlers (`inline` option) does the job very well in 95% cases, but in the other 5% cases it's sometimes easier to use an inline function as a handler. And it wouldn't be an issue as well if we had an option to disable a specific ESLint rule for a specific attribute (like in Prettier with `prettier-ignore-attribute`), but since it's *seemingly* not possible, I'm suggesting to allow `["inline", "inline-function"]` option for `v-on-handler-style` rule.

N/A

New option value

```vue
<!-- ✅ Allow both syntaxes ... -->
<some-component @input="handler($event)" @change="(payload) => { ... }"></some-component>
<!-- ❌ ... but never the method binding one -->
<some-component @input="handler1" @change="handler2"></some-component>
```

Does not allow these two styles to be used simultaneously.

Will allow those two styles.

**Additional context**

ByScripts:
Fixes #2460 

Allow using `["inline", "inline-function"]` option for `v-on-handler-style` rule.

In this case:
- `method` style will be forbidden
  - ❌ `@click="myHandler"`
- `inline-function` style will only be accepted with at least 1 argument:
  - ❌ `@click="() => myHandler()"`
  - ✅ `@click="myHandler()"`
  - ✅ `@click="(evt) => myHandler(evt)"`
  - ✅ `@click="(arg1, arg2) => myHandler(arg1, arg2)"`
- `inline` style will be required in other cases
  - ❌ `@click="() => myHandler($event)"`
  - ✅ `@click="myHandler($event)"`
  - ❌ `@click="() => count++"` 
  - ✅ `@click="count++"`

### DIFF
--- a/docs/rules/v-on-handler-style.md
+++ b/docs/rules/v-on-handler-style.md
@@ -47,7 +47,7 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
 ```json
 {
   "vue/v-on-handler-style": ["error",
-    ["method", "inline-function"], // ["method", "inline-function"] | ["method", "inline"] | "inline-function" | "inline"
+    ["method", "inline-function"], // ["method", "inline-function"] | ["method", "inline"] | ["inline", "inline-function"] | "inline-function" | "inline"
     {
       "ignoreIncludesComment": false
     }
@@ -58,6 +58,7 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
 - First option ... Specifies the name of an allowed style. Default is `["method", "inline-function"]`.
   - `["method", "inline-function"]` ... Allow handlers by method binding. e.g. `v-on:click="handler"`. Allow inline functions where method handlers cannot be used. e.g. `v-on:click="() => handler(listItem)"`.
   - `["method", "inline"]` ... Allow handlers by method binding. e.g. `v-on:click="handler"`. Allow inline handlers where method handlers cannot be used. e.g. `v-on:click="handler(listItem)"`.
+  - `["inline", "inline-function"]` ... Allow inline handlers. e.g. `v-on:click="handler()"`. Allow inline functions if they have at least 1 argument. e.g. `v-on:click="(arg1, arg2) => handler(arg1, arg2)"`.
   - `"inline-function"` ... Allow inline functions. e.g. `v-on:click="() => handler()"`
   - `"inline"` ... Allow inline handlers. e.g. `v-on:click="handler()"`
 - Second option
@@ -116,6 +117,33 @@ This rule aims to enforce a consistent style in `v-on` event handlers:
     <button v-on:click="() => e()" />
     <button v-on:click="() => handler(e)" />
   </template>
+</template>
+```
+
+</eslint-code-block>
+
+### `["inline", "inline-function"]`
+
+<eslint-code-block fix :rules="{'vue/v-on-handler-style': ['error', ['inline', 'inline-function']]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <button v-on:click="count++" />
+  <button v-on:click="handler()" />
+  <button v-on:click="handler($event)" />
+  <button v-on:click="(arg) => handler(arg)" />
+  <template v-for="e in list">
+    <button v-on:click="handler(e)" />
+    <button v-on:click="handler($event, e)" />
+    <button v-on:click="(arg) => handler(arg, e)" />
+  </template>
+
+  <!-- ✗ BAD -->
+  <button v-on:click="() => count++" />
+  <button v-on:click="handler" />
+  <button v-on:click="() => handler()" />
+  <button v-on:click="() => handler($event)" />
 </template>
 ```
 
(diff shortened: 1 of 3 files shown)


FloEdelmann:
Cool, thank you! Should the `inline-function` style really already be accepted with one argument (where `inline` would also work with `$event`), or only with two or more?

ByScripts:
This would prevent the interdiction to use `$event` in favor of explicit naming:

```vue
<template>
  <!-- Using an inline function with a good argument name is better -->
  <MyComp @increase="(count) => myHandler(count)" />`

  <!-- It's why we don't allow $event, as it is meaningless (and is absolutely not an Event here) -->
  <MyComp @increase="myHandler($event)" />`
</template>
```

FloEdelmann:
I think if you want to force explicit parameter naming, then you should only allow `inline-function`. It's 5 extra characters for functions without parameters (`() => `), but it's a simple and effective rule.

If you prefer shortness over explicitness however, then `["inline", "inline-function"]` is fine, but IMO it should only fall back to `inline-function` when `inline` does not work at all, i.e. only for functions with more than one parameter.

ByScripts:
I see your point, but it's really common to only need an inline handler.

I don't think we should force users to use `@event="() => handler()"` where `@event="handler()"` is enough.

The primary goal of this PR is to forbid the usage of method handler, as it's error prone, and allow the two others.

I updated my proposal to add an `allowInlineFuncSingleArg?: boolean` option that can only be used in conjunction with `['method', 'inline-function']` or `['inline', 'inline-function']`.

- When `allowInlineFuncSingleArg` is `true`:

  ❌ `() => handleFilter()`
      => `handleFilter()` if `['inline', 'inline-function']`
      => `handleFilter` if `['method', 'inline-function']`

  ✅ `(filter) => handleFilter(filter)`

  ✅ `(filter, negate) => handleFilter(filter, negate)`

- When `allowInlineFuncSingleArg` is `false` (by default)

  ❌ `() => handleFilter()`
      => `handleFilter()` if `['inline', 'inline-function']`
      => `handleFilter` if `['method', 'inline-function']`

  ❌ `(filter) => handleFilter(filter)` => `handleFilter($event)`
      => `handleFilter($event)` if `['inline', 'inline-function']`
      => `handleFilter` if `['method', 'inline-function']`

  ✅ `(filter, negate) => handleFilter(filter, negate)`

FloEdelmann:
Hmm, I don't like the extra option. The logic is quite convoluted already anyway, and the extra option makes it even harder to understand.

But I think we can change the rule like this:

* For cases like `@click="(arg1, arg2) => handler(arg1, arg2)"`, always allow the `inline-function` style, because there is no other way to represent those. Multiple event payloads should rather be reported at the emitting side, see https://github.com/vuejs/eslint-plugin-vue/issues/2005
* Add a new option `["inline", "inline-function"]` that allows `inline` but disallows using `$event`, and allows using `inline-function` in that case.

ByScripts:
Thank you for your reply.

After some thought, I wonder if this rule isn't trying to handle too many things at once.

I agree on your first point (although I think `vue/event-prefer-single-payload` would be a better name. At first I read "Prefer Single-Event Payload" ^^)

Perhaps it would be easier to disallow these different behaviors with new, independent rules?

For example:

1. `@click="count++"` ➡️ Disallow with `vue/event-no-inline-expr`
2. `@click="increase"` ➡️ Disallow with `vue/event-no-method-handler`
3. `@click="increase()"` ➡️ Disallow with `vue/event-no-inline-handler`
4. `@click="() => increase()"` ➡️ Disallow with `vue/event-no-arg-less-inline-function`
5. `@click="increase($event)"` ➡️ Disallow with `vue/no-dollar-event`
6. `@click="(count) => increase(count)"` ➡️ No reason to disallow?
7. `@click="(count, multiply) => increase(count, multiply)"` ➡️ No reason to disallow?

This is just a quick overview, and I may be overlooking certain situations.

ota-meshi:
Thanks for working on new option and for the suggestions.

However, for now I don't think it's a good idea to split up what the `v-on-handler-style` rule is doing, because I think it would be hard to design the options in such a way that those rules don't conflict.

FloEdelmann:
@ByScripts are you interested in changing the rule like I described in my [comment above](https://github.com/vuejs/eslint-plugin-vue/pull/2471#issuecomment-2179404896)?

jiikoosol:
I just stumbled on this one when going through a larger codebase which has been using the method-style when defining v-on event handlers. Shouldn't the vue/v-on-handler-style rule by default forbid using the method style as in Vue 3 the onClick event is not cached when the function is typed without parentheses?
Here's an example:

Not cached:
Screenshot 2025-02-27 at 7 54 21

Cached:
Screenshot 2025-02-27 at 7 54 12

PR #2476, opened 2024-06-19
Add name to flat configs.

ota-meshi:
This PR adds names to each element of the flat config.
Adding names to configs makes them easier to check with `@eslint/config-inspector`.


ota-meshi:
I think the CI error will be fixed in #2477.

FloEdelmann (approved):

PR #2487, opened 2024-06-27
feat(require-toggle-inside-transition): check directive

### LINKED ISSUE #2467: vue/require-toggle-inside-transition - question: should it still be an error with a conditional appear attribute?

- vue/require-toggle-inside-transition

- fewer

- new default behaviour - don't give an error when there is a conditional appear attribute on a node

```vue
<transition :appear="appIsLoaded">
  <div>Something</div>
</transition>
```

- shows an error on the div - "The element inside `<transition>` is expected to have a `v-if` or `v-show` directive"

- not show an error

**Additional context**

We are using a conditional `appear` attribute so that the animation is not shown on the first load of the app, only when navigating between components.

I'm not 100% sure this is a valid use of appear and transition, so I would like to confirm this too.

Thanks!

waynzh:
fix #2467

### DIFF
--- a/lib/rules/require-toggle-inside-transition.js
+++ b/lib/rules/require-toggle-inside-transition.js
@@ -6,6 +6,20 @@
 
 const utils = require('../utils')
 
+/**
+ * @param {VDirective} vBindAppear
+ */
+function isValidateBindAppear(vBindAppear) {
+  if (
+    vBindAppear.value?.expression &&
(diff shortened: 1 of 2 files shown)


ota-meshi (approved):
LGTM! Thank you!

FloEdelmann (approved):
Looks good to me, apart from a minor naming suggestion :slightly_smiling_face:

Issue #2448, opened 2024-04-14
Full example for flat config with object


Narretz:
**Tell us about your environment**

- **ESLint version:** 9
- **eslint-plugin-vue version:**  9
- **Vue version:** 2
- **Node version:** 20

**The problem you want to solve.**
I converted my eslint config to flat config. I used a lot of overrides, and tried to recreated that with specific objects that contain plugins etc.

So for eample previously my vue config looked like this: 

```
overrides: [
    {
      files: ['packages/client/src/components/**/*.vue', 'stories/**/*.vue'],
      extends: ['plugin:vue/strongly-recommended'],
      parser: 'vue-eslint-parser',
      parserOptions: {
        parser: '@typescript-eslint/parser',
      },
      rules: {
        'max-lines': ['error', { max: 900 }],
      }
    }
]
```

The only way I managed to recreate this is by exactly spreading the arrays in the flat vue config. This involved looking at the plugin to see how many arrays it includes. Is there no easier way? I don't even know if I have included everything.
```
  {
    files: ['packages/client/src/components/**/*.vue', 'stories/**/*.vue'],
    languageOptions: {
      ...vue.configs['flat/vue2-strongly-recommended'][1].languageOptions,
      parserOptions: {
        parser: '@typescript-eslint/parser',
      },
    },
    plugins: {
      vue,
    },
    rules: {
      ...vue.configs['flat/vue2-strongly-recommended'][2].rules,
      ...vue.configs['flat/vue2-strongly-recommended'][3].rules,
      'max-lines': ['error', { max: 900 }],
    },
  },
```

**Your take on the correct solution to problem.**

I don't have a good idea, but it would definitely help if the rules and the language options were accessible in a merged object.

higherorderfunctor:
This is an example repo I use for performance testing, but this is the general approach I use in my repos: https://github.com/higherorderfunctor/effect/blob/chore/eslint-testing/eslint.config.js

I took the inverse approach of copying the common stuff into each plugins rule sets with a mapping function instead of trying to copy rules into a single object with the common stuff.

```ts
const overrideWith = (overrides, configs) =>
  configs.map((config) =>
    { ...config, ...overrides, plugins: { ...overrides.plugins, ...(config.plugins ?? {}) } }
  )

const tsOverrides = {
  ... // common settings
}

const tsRules = overrideWith(tsOverrides, [
  eslint.configs.recommended, // no array
  tseslint.configs.eslintRecommended, // no array
[...]

PR #2490, opened 2024-06-29
docs: add an example on how to use custom parser with flat config

Lehoczky:
Related to: https://github.com/vuejs/eslint-plugin-vue/issues/2448

📚 Description

This PR adds an example on how to use this plugin with TypeScript, which is a common use case.

📜 Notes

The getting started guide is getting a bit hard to follow with a lot of warnings, duplicated text for each configuration style. It might worth extracting the `.eslintrc` format to a different subpage like it is done in [the official eslint docs](https://eslint.org/docs/latest/use/configure/configuration-files).

### DIFF
--- a/docs/user-guide/index.md
+++ b/docs/user-guide/index.md
@@ -186,6 +186,8 @@ export default ts.config(
 )
 ```
 
+:::
+
 The `parserOptions.parser` option can also specify an object to specify multiple parsers. See [vue-eslint-parser README](https://github.com/vuejs/vue-eslint-parser#readme) for more details.
 
 ### How does ESLint detect components?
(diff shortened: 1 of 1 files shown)


docs/user-guide/index.md
```diff
@@ -146,6 +146,48 @@ If you want to use custom parsers such as [@babel/eslint-parser](https://www.npm
   }
 ```
 
+Full example:
+
+::: code-group
+
+```json [Legacy Config]
```
ota-meshi, line 153: Thanks for the PR!
I think we'll avoid using the terms "Legacy Config" and "Flat Config" in our documentation.
Could you please change it to use `.eslintrc` and `eslint.config.js` instead?

ota-meshi (changes requested):

docs/user-guide/index.md
Lehoczky: Sure thing! I saw the same naming convention elsewhere in the docs, I changed those too^^

ota-meshi (approved):
LGTM! Thank you!

PR #2493, opened 2024-06-30
docs: fix build on windows

### LINKED ISSUE #2492: `TypeError: require is not a function` when running `npm run docs:watch` locally
I've tried to contribute to the documentation, but when I run `npm run docs:watch`, I only get a blank page with the following error:

```
Uncaught TypeError: require is not a function
    at index.js:9:9
```

image

What I did:
- installed dependencies with `npm install`
- ran `npm run docs:watch`

I've also tried to build the library with `npm run tsc`, but got the following error:

```
lib/rules/order-in-components.js:265:25 - error TS7006: Parameter 'fixer' implicitly has an 'any' type.

265     function* handleFix(fixer, propertyNode, unorderedPropertyNode) {
                            ~~~~~


Found 1 error in lib/rules/order-in-components.js:265
```

I'm on:
- Windows 10
- Node v20.15.0

Lehoczky:
Resolves #2492 

This PR uses [pathe](https://www.npmjs.com/package/pathe) instead of the built-in path module from Node.js to handle the different file separators between Windows and other operating systems.

I tested it, and works, it's only a draft because I'm not sure you want to add another dependency to the project, even if it is only a dev dependency.

### DIFF
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -259,6 +259,7 @@ module.exports = {
     }
 
     /**
+     * @param {RuleFixer} fixer
      * @param {Property} propertyNode
      * @param {Property} unorderedPropertyNode
      */
--- a/docs/.vitepress/build-system/build.mts
+++ b/docs/.vitepress/build-system/build.mts
@@ -2,7 +2,7 @@
  * Pre-build cjs packages that cannot be bundled well.
  */
 import esbuild from 'esbuild'
-import path from 'path'
+import path from 'pathe'
 import fs from 'fs'
 import { fileURLToPath } from 'url'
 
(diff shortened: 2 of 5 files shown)


lib/rules/order-in-components.js
```diff
@@ -259,6 +259,7 @@ module.exports = {
     }
 
     /**
+     * @param {RuleFixer} fixer
```
Lehoczky, line 262: This is not needed, It just fixes the tsc error on the main branch. This was implicitly any before.

ota-meshi (approved):
Thanks for this PR!
Looks good to me. If you think it's OK to merge, please remove the draft mark.

PR #2481, opened 2024-06-21
Show nested outline in docs

FloEdelmann:
| Before | After |
| ------ | ----- |
| grafik | grafik |

See documentation: https://vitepress.dev/reference/default-theme-config#outline

PR #2497, opened 2024-07-03
Improve docs

FloEdelmann:
### 1. Remove `outline: deep` frontmatter which is now unnecessary

* Follows #2481.

### 2. Bump eslint-plugin-jsonc to latest version

* I somehow missed this in #2478, that's probably causing errors like [this](https://togithub.com/vuejs/eslint-plugin-vue/pull/2481#issuecomment-2182473055).

### 3. Also expose `SourceCode` in `eslint` shim

Fixes this issue in https://eslint.vuejs.org/rules/max-lines-per-block.html:

> Screenshot_20240703_145453

because `SourceCode` is needed in

https://github.com/vuejs/eslint-plugin-vue/blob/c64bf9448099bd7ca3512dca705c81730afec4ab/lib/rules/max-lines-per-block.js#L7

### DIFF
--- a/docs/.vitepress/build-system/src/eslint.mjs
+++ b/docs/.vitepress/build-system/src/eslint.mjs
@@ -1,5 +1,8 @@
 // @ts-nocheck
-import * as all from '../../../../node_modules/eslint/lib/linter/linter.js'
-const Linter = all.Linter
-export { Linter }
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
Thank you!

PR #2498, opened 2024-07-06
fix: change flat config name separator to `/`

ntnyq:
This PR changes the all flat configs' name separator from `:` to `/`.

As ESLint recommended use `/` as the separator.

Refs:

- [Configuration Naming Conventions](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-naming-conventions)


FloEdelmann (approved):
Thanks!

ota-meshi (approved):

Issue #2499, opened 2024-07-07
Allow specifying custom macros in `vue/define-macros-order`


Ericlm:
**What rule do you want to change?**
I would like to update the `define-macros-order` rule.

**Does this change cause the rule to produce more or fewer warnings?**
More.

**How will the change be implemented? (New option, new default behavior, etc.)?**
With the unplugging-vue-router which allows routes to be typesafe, there is now a new [definePage macro](https://uvr.esm.is/guide/extending-routes.html#definepage) that allows to customize the generated route.

I think this macro should be ordered just like defineProps, emits, slots etc. I think it could be at the top of the component, even before defineModel.

**Please provide some example code that this change will affect:**

```vue
<script setup lang="ts">
import { useRouter } from 'vue-router'

const router = useRouter()

definePage({}) // Can be placed anywhere

function hello(): void {
  router.push({ name: '/' })
}
</script>
```

**What does the rule currently do for this code?**
The rule does not enforce a location for `definePage`, so it can be placed anywhere in the script.

**What will the rule do after it's changed?**
It will enforce the `definePage` macro to be at the top of the script (or elsewhere).

FloEdelmann:
`unplugging-vue-router` will not be officially supported in eslint-plugin-vue, so this feature request is out of scope.

But you can already adjust the order manually. I think it should be possible to allow custom values there, so that you could add `definePage` there manually.

PR #2501, opened 2024-07-08
Update rule `define-macros-order` for custom macros

Ericlm:
**⚠️ Newbie alert ⚠️**
_This is my very first contribution to this project so I hope I did the things well; but don't hesitate to tell me what is not right._ 😄

This PR adds the ability to use the [define macro order rule](https://eslint.vuejs.org/rules/define-macros-order.html#defineexposelast-true) for any macro specified in the `order` array. This allows users to specify `definePage` for example or any macro wanted.

Basically, I added a `CallExpression` that checks whether the Identifier is included into the `order` array but **not** in the `ORDER_SCHEMA` as this is already checked with earlier events (`onDefineExposeExit` etc.)
I removed the `enum` from the `properties` of the rule as it's no longer limited to `ORDER_SCHEMA`. And finally, I've added new tests.

I also have some questions:
- If `defineExpose` is declared inside the `order`, I guess it should be ignored? Otherwise it would conflict with `defineExposeLast` I guess.
- Maybe the check could be more performant if there was no check against `ORDER_SCHEMA` in the condition? (removing the callbacks and the call to `has()`).

See #2499

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -140,7 +140,10 @@ function create(context) {
         defineExposeNode = getDefineMacrosStatement(node)
       },
 
-      'CallExpression:exit'(node) {
+      /** @param {CallExpression} node */
+      'Program > ExpressionStatement > CallExpression, Program > VariableDeclaration > VariableDeclarator > CallExpression'(
+        node
+      ) {
         if (
           node.callee &&
           node.callee.type === 'Identifier' &&
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -113,7 +113,7 @@ function create(context) {
 
   if (order.includes(MACROS_EXPOSE) && defineExposeLast) {
     throw new Error(
-      'defineExposeLast cannot be used if defineExpose macro is in order array.'
+      "`defineExpose` macro can't be in the `order` array if `defineExposeLast` is true."
     )
   }
 
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -11,12 +11,14 @@ const MACROS_PROPS = 'defineProps'
 const MACROS_OPTIONS = 'defineOptions'
 const MACROS_SLOTS = 'defineSlots'
 const MACROS_MODEL = 'defineModel'
-const ORDER_SCHEMA = new Set([
+const MACROS_EXPOSE = 'defineExpose'
+const KNOWN_MACROS = new Set([
   MACROS_EMITS,
   MACROS_PROPS,
   MACROS_OPTIONS,
   MACROS_SLOTS,
-  MACROS_MODEL
+  MACROS_MODEL,
+  MACROS_EXPOSE
 ])
 const DEFAULT_ORDER = [MACROS_PROPS, MACROS_EMITS]
 
@@ -133,12 +135,11 @@ function create(context) {
       },
 
       'CallExpression:exit'(node) {
-        // check if the node is a macro in the order
         if (
           node.callee &&
           node.callee.type === 'Identifier' &&
           order.includes(node.callee.name) &&
-          !ORDER_SCHEMA.has(node.callee.name)
+          !KNOWN_MACROS.has(node.callee.name)
         ) {
           macrosNodes.set(node.callee.name, [getDefineMacrosStatement(node)])
         }
@@ -344,8 +345,7 @@ module.exports = {
   meta: {
     type: 'layout',
     docs: {
-      description:
-        'enforce order of `defineEmits` and `defineProps` compiler macros',
+      description: 'enforce order of specified compiler macros.',
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/define-macros-order.html'
     },
@@ -358,8 +358,8 @@ module.exports = {
           order: {
             type: 'array',
             items: {
-               type: string,
-               minLength: 1
+              type: 'string',
+              minLength: 1
             },
             uniqueItems: true,
             additionalItems: false
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -11,13 +11,13 @@ const MACROS_PROPS = 'defineProps'
 const MACROS_OPTIONS = 'defineOptions'
 const MACROS_SLOTS = 'defineSlots'
 const MACROS_MODEL = 'defineModel'
-const ORDER_SCHEMA = [
+const ORDER_SCHEMA = new Set([
   MACROS_EMITS,
   MACROS_PROPS,
   MACROS_OPTIONS,
   MACROS_SLOTS,
   MACROS_MODEL
-]
+])
 const DEFAULT_ORDER = [MACROS_PROPS, MACROS_EMITS]
 
 /**
@@ -130,6 +130,18 @@ function create(context) {
       },
       onDefineExposeExit(node) {
         defineExposeNode = getDefineMacrosStatement(node)
+      },
+
+      'CallExpression:exit'(node) {
+        // check if the node is a macro in the order
+        if (
+          node.callee &&
+          node.callee.type === 'Identifier' &&
+          order.includes(node.callee.name) &&
+          !ORDER_SCHEMA.has(node.callee.name)
+        ) {
+          macrosNodes.set(node.callee.name, [getDefineMacrosStatement(node)])
+        }
       }
     }),
     {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "start": "npm run test:base -- --watch --growl",
     "test:base": "mocha \"tests/lib/**/*.js\" --reporter dot",
     "test": "nyc npm run test:base -- \"tests/integrations/*.js\" --timeout 60000",
+    "test:define-macros-order": "mocha \"tests/lib/rules/define-macros-order.js\" --timeout 60000",
```
FloEdelmann, line 11: This is fine for testing locally, but please avoid committing and pushing debugging changes like this. I'll revert this change.

tests/lib/rules/define-macros-order.js
```diff
@@ -254,6 +273,22 @@ tester.run('define-macros-order', rule, {
...
       ]
+    },
+    {
+      filename: 'test.vue',
+      code: `
+        <script setup>
+        const page = definePage()
+
+        const first = defineModel('first')
+          const second = defineModel('second')
```
FloEdelmann, line 284: Even if it's just test code, let's format it properly:

```suggestion
        const first = defineModel('first')
        const second = defineModel('second')
```

lib/rules/define-macros-order.js
```diff
@@ -130,6 +130,18 @@ function create(context) {
       },
       onDefineExposeExit(node) {
         defineExposeNode = getDefineMacrosStatement(node)
+      },
+
+      'CallExpression:exit'(node) {
+        // check if the node is a macro in the order
```
FloEdelmann, line 136: I think this comment is not necessary, the condition is quite readable on its own.

```suggestion
```

lib/rules/define-macros-order.js
```diff
@@ -11,13 +11,13 @@ const MACROS_PROPS = 'defineProps'
 const MACROS_OPTIONS = 'defineOptions'
 const MACROS_SLOTS = 'defineSlots'
 const MACROS_MODEL = 'defineModel'
-const ORDER_SCHEMA = [
+const ORDER_SCHEMA = new Set([
```
FloEdelmann, line 14: Maybe rename to `KNOWN_MACROS` or similar? It's not used in the schema anymore.

Also, please add `defineExpose` (via a new constant `MACROS_EXPOSE`) to the set.

lib/rules/define-macros-order.js
```diff
@@ -130,6 +130,18 @@ function create(context) {
...
         defineExposeNode = getDefineMacrosStatement(node)
+      },
+
+      'CallExpression:exit'(node) {
+        // check if the node is a macro in the order
+        if (
+          node.callee &&
+          node.callee.type === 'Identifier' &&
+          order.includes(node.callee.name) &&
+          !ORDER_SCHEMA.has(node.callee.name)
```
FloEdelmann, line 141: > Maybe the check could be more performant if there was no check against `ORDER_SCHEMA` in the condition? (removing the callbacks and the call to `has()`).

`Set.has()` is constant-time, so this is irrelevant to performance. But of course, there are multiple callbacks for the same element now, which adds some overhead. But since some of the handlers have some special logic (both in this file and [where they are defined](https://github.com/vuejs/eslint-plugin-vue/blob/f5cb93db8fd9ff40288406b6119e3c171b1ce66b/lib/utils/index.js#L1400-L1442)), we can't really change that.

FloEdelmann (changes requested):
> If `defineExpose` is declared inside the `order`, I guess it should be ignored? Otherwise it would conflict with `defineExposeLast` I guess.

Yeah, it would conflict. However, it's better to not fail silently,  but to validate the options: If `defineExposeLast` is true and `defineExpose` is declared inside the `order`, throw an error.

Anyway, thanks for picking this up so fast!

package.json
FloEdelmann: Reverted in [`efd97d7`](https://github.com/vuejs/eslint-plugin-vue/pull/2501/commits/efd97d7d564f8d8c1a1c0eba38d175fd5c49cbda).

lib/rules/define-macros-order.js
Ericlm: Okay, I got the point. I may be interested to do that after.

lib/rules/define-macros-order.js
```diff
@@ -109,6 +111,12 @@ function create(context) {
   /** @type {ASTNode} */
   let defineExposeNode
 
+  if (order.includes(MACROS_EXPOSE) && defineExposeLast) {
+    throw new Error(
+      'defineExposeLast cannot be used if defineExpose macro is in order array.'
```
FloEdelmann, line 116: I'd rather reverse the error message:

```suggestion
      "`defineExpose` macro can't be in the `order` array if `defineExposeLast` is true."
```

FloEdelmann (changes requested):

lib/rules/define-macros-order.js
```diff
@@ -109,6 +111,12 @@ function create(context) {
   /** @type {ASTNode} */
   let defineExposeNode
 
+  if (order.includes(MACROS_EXPOSE) && defineExposeLast) {
+    throw new Error(
```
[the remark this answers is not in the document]
Ericlm: I tried to find a way for testing errors but I couldn't find any. So I created a temporary test on my machine with the bad options that crashed as expected.

FloEdelmann (approved):
Looks good to me now. Thank you for your first contribution!

Ericlm:
You're welcome! Really happy to add this 😁

lib/rules/define-macros-order.js
```diff
@@ -130,6 +138,17 @@ function create(context) {
...
+      },
+
+      'CallExpression:exit'(node) {
+        if (
+          node.callee &&
+          node.callee.type === 'Identifier' &&
+          order.includes(node.callee.name) &&
+          !KNOWN_MACROS.has(node.callee.name)
+        ) {
+          macrosNodes.set(node.callee.name, [getDefineMacrosStatement(node)])
```
ota-meshi, line 150: I think we need to check whether the CallExpression is located in a place that is accepted as a macro.
For example, I don't think the following definePage is a macro.

```vue
<script setup>
function fn() {
  definePage()
}
</script>
```

ota-meshi (changes requested):

lib/rules/define-macros-order.js
Ericlm: I'm wondering how to do that 🤔 I guess macros are only located at the "root" of the script?

ota-meshi: Even if macros could be used in places other than the root, I don't think this rule would allow we to change their order.

Ericlm: I made a simple rule where the node needs to be an `ExpressionStatement` or a `VariableDeclaration`.
```js
 const macro = getDefineMacrosStatement(node)
 if (
  macro.type === 'ExpressionStatement' ||
  macro.type === 'VariableDeclaration'
 )
   macrosNodes.set(node.callee.name, [macro])
```
But I don't know if it's a good approach. I added a test so that 
```js
<script setup>
function fn() {
    definePage()
}
</script>
```
is considered valid (`definePage` is ignored), but I have an issue with this test : 
```vue
<script setup>
import Foo from 'foo'
/** model */
const model = defineModel()
/** emits */
defineEmits(['update:foo'])

const val = () => definePage()
</script>
```
This one keeps failing because I don't know how to ignore the "macro".

---
There is a `getDefineMacrosStatement` that is simply a function that searches recursively for the `Program` node; maybe it could be more "precise"? 
Because macros (I believe):
- Have the `Program` as parent
- Are expressions that are called
- Can be assigned like `const props = defineProps<...>()`
- Are located directly on the first column, or just after an assignment

Maybe it is more precise, but these are some ideas to target more precisely what "macros" are within the AST 🙂

ota-meshi: Hmm. I've never used custom macros, so I'm not familiar with them.
If you want to prioritize support for just `definePage`, I think it's reasonable for now to only support `Program > ExpressionStatement > CallExpression (custom macro)`.

Ericlm: Instead of `CallExpression`, I added:
- `Program > ExpressionStatement > CallExpression` to target things like `defineProps()`, `definePage()`, `defineSomething()`...
- `Program > VariableDeclaration > VariableDeclarator > CallExpression` to target `const props = defineProps()`, `const emit = defineEmits()`, `const stuff = defineStuff()`...

ota-meshi (approved):
LGTM! Thank you!

PR #2506, opened 2024-07-10
feat: export parser

Shyam-Chen:
```js
// eslint.config.js
import typescriptEslint from 'typescript-eslint';
import eslintPluginVue from 'eslint-plugin-vue';

export default [
  {
    languageOptions: {
      parser: eslintPluginVue.parser, // <-
      parserOptions: {
        parser: typescriptEslint.parser,
        sourceType: 'module',
      },
    },
  },
  // ...
];
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,6 +7,7 @@
 
 const plugin = {
   meta: require('./meta'),
+  parser: require('vue-eslint-parser'),
   configs: {
     // eslintrc configs
     base: require('./configs/base'),


ota-meshi:
This plugin will not work if you set only the parser.
Use `*.configs["flat/base"]`.
https://eslint.vuejs.org/user-guide/#bundle-configurations-eslint-config-js
If you only want the parser, just import `vue-eslint-parser`.

PR #2508, opened 2024-07-11
feat: add await option for next-tick-style rule

### LINKED ISSUE #2485: New option for `next-tick-style`

`next-tick-style`

No.

If inside `async` function, provide autofix, otherwise no fix.

```vue
<script setup>
const foo = () => {
	nextTick(() => {
	  // ...
	})
}
</script>
```

Fixes my code to `nextTick().then(...)`

Enforce `await nextTick()`

**Additional context**

I want the promise style use of `nextTick()`, but I don't want it fix to `nextTick().then`, I want `await nextTick()`.

codiini:
This PR addresses #2485


lib/rules/next-tick-style.js
```diff
@@ -123,6 +159,61 @@ module.exports = {
...
+
+                // Handle callback to await conversion
+                if (callExpression.arguments.length > 0) {
+                  const [args] = callExpression.arguments
+                  let callbackBody = null
+
+                  callbackBody =
+                    args.type === 'ArrowFunctionExpression' ||
+                    args.type === 'FunctionExpression'
+                      ? extractCallbackBody(args, sourceCode)
```
fisker, line 181: We can't extract callback body directly, their might variable conflicts.

```js
export default {
  async created() {
     const foo = 1;

	  nextTick(() => {
       const foo = 2;
      doSomething(foo);
    })
   }
}
```

lib/rules/next-tick-style.js
```diff
@@ -79,17 +79,53 @@ function isAwaitedPromise(callExpression) {
...
+    callback.type !== 'FunctionExpression' &&
+    callback.type !== 'ArrowFunctionExpression'
+  ) {
+    return ''
+  }
+
+  if (callback.body.type === 'BlockStatement') {
+    return sourceCode
+      .getText(callback.body)
+      .slice(1, -1) // Remove curly braces
```
fisker, line 109: Maybe we can simply keep the block.

lib/rules/next-tick-style.js
```diff
@@ -123,6 +159,61 @@ module.exports = {
...
+                const sourceCode = context.getSourceCode()
+
+                // Handle callback to await conversion
+                if (callExpression.arguments.length > 0) {
+                  const [args] = callExpression.arguments
+                  let callbackBody = null
+
+                  callbackBody =
+                    args.type === 'ArrowFunctionExpression' ||
+                    args.type === 'FunctionExpression'
```
fisker, line 180:
```suggestion
                    (args.type === 'FunctionExpression' && !args.geneator)
```

lib/rules/next-tick-style.js
```diff
@@ -123,6 +159,61 @@ module.exports = {
...
+            !isAwaitedFunction(callExpression)
+          ) {
+            context.report({
+              node,
+              messageId: 'useAwait',
+              fix(fixer) {
+                const sourceCode = context.getSourceCode()
+
+                // Handle callback to await conversion
+                if (callExpression.arguments.length > 0) {
```
fisker, line 174:
```suggestion
                if (callExpression.arguments.length === 1) {
```

lib/rules/next-tick-style.js
```diff
@@ -123,6 +159,61 @@ module.exports = {
...
+                    args.type === 'FunctionExpression'
+                      ? extractCallbackBody(args, sourceCode)
+                      : `${sourceCode.getText(args)}()`
+
+                  const nextTickCaller = sourceCode.getText(
+                    callExpression.callee
+                  )
+                  return fixer.replaceText(
+                    callExpression.parent,
+                    `await ${nextTickCaller}();${callbackBody};`
```
fisker, line 189: We need make sure the `nextTick` call is an expression statement, otherwise this will cause sytax errors.


```
if (foo) nextTick(...)
```

```
(() => nextTick(...))();
```

lib/rules/next-tick-style.js
```diff
@@ -79,17 +79,53 @@ function isAwaitedPromise(callExpression) {
...
 }
 
+/**
+ * @param {CallExpression} callExpression
+ * @returns {boolean}
+ */
+function isAwaitedFunction(callExpression) {
+  return (
+    callExpression.parent.type === 'AwaitExpression' &&
+    callExpression.parent.parent.type !== 'MemberExpression'
```
fisker, line 89: Why this `.parent.parent` check needed.

lib/rules/next-tick-style.js
codiini: @fisker How would you suggest the fix method handle these cases?

fisker: Ignore or turn into a block.

fisker: "ignore" means no fix

codiini: Okay, I've added a check.

codiini: Please review the PR when you get the chance and let me know if there's any other changes you think I should make. Thanks

fisker: https://github.com/vuejs/eslint-plugin-vue/pull/2508#discussion_r1677865772

codiini: Isn't it related to this one: https://github.com/vuejs/eslint-plugin-vue/pull/2508#discussion_r1677868269 ?
I removed the `.slice()` method which keeps the block

fisker: Sorry didn't see that change.

codiini: No worries. Is everything good to go now?

fisker: I'm not a maintainer..

lib/rules/next-tick-style.js
```diff
@@ -79,17 +79,44 @@ function isAwaitedPromise(callExpression) {
...
+  if (
+    callback.type !== 'FunctionExpression' &&
+    callback.type !== 'ArrowFunctionExpression'
+  ) {
+    return ''
+  }
+
+  if (callback.body.type === 'BlockStatement') {
+    return sourceCode.getText(callback.body).trim()
+  }
```
fisker, line 102: We can remove this branch.

lib/rules/next-tick-style.js
fisker: This seems not the correct fix.. It changes behavior

```js
nextTick(() => console.log(2))
console.log(1)
```

Logs `1 2`

```js
await nextTick();  console.log(2)
console.log(1)
```

Logs `2 1`

Maybe use suggestions instead?

lib/rules/next-tick-style.js
```diff
@@ -123,6 +150,66 @@ module.exports = {
...
+                  }
+
+                  const [args] = callExpression.arguments
+                  let callbackBody = null
+
+                  callbackBody =
+                    args.type === 'ArrowFunctionExpression' ||
+                    (args.type === 'FunctionExpression' && !args.generator)
+                      ? extractCallbackBody(args, sourceCode)
+                      : `${sourceCode.getText(args)}()`
```
fisker, line 177: This is also not safe

```js
nextTick(foo || bar)
```

```diff
- await nextTick(); foo || bar()
+ await nextTick(); (foo || bar)()
```

If `foo` is truly, it will not be called

lib/rules/next-tick-style.js
```diff
@@ -79,17 +79,44 @@ function isAwaitedPromise(callExpression) {
...
+    callback.type !== 'ArrowFunctionExpression'
+  ) {
+    return ''
+  }
+
+  if (callback.body.type === 'BlockStatement') {
+    return sourceCode.getText(callback.body).trim()
+  }
+
+  return sourceCode.getText(callback.body)
```
fisker, line 104: The function body can't exact directly if it has `return`


```js
const foo = async () => {
	nextTick(() => {
		return;
	})

	return false
}
```

returns `false`


```js
const foo = async () => {
	await nextTick(); {
		return;
	}

	return false
}
```

returns `undefined`

fisker: It will also broken if the callback has `.id` or used `arguments` inside.

```js
const foo = async () => { 
	nextTick(function foo() {
    use(foo)
	  use(arguments)
	})
}
```

FloEdelmann (changes requested):
@codiini Could you please address @fisker's comments?

FloEdelmann:
It's been nearly two years now without any activity, so I'm going to close this. Feel free to open a new PR!

PR #2511, opened 2024-07-19
fix(attribute-hyphenation): apply to v-model directive

### LINKED ISSUE #2510: attribute-hyphenation should apply to v-model:propName
**What rule do you want to change?** attribute-hyphenation

**Does this change cause the rule to produce more or fewer warnings?** More

**How will the change be implemented? (New option, new default behavior, etc.)?** New default behavior

```vue
<my-component
    v-model:selectedItem="..."
   defaultValue="..."
/>
```

**What does the rule currently do for this code?** It flags that the `defaultValue` prop is not hyphenated (should be `default-value`), but it does not flag the `selectedItem` prop.

**What will the rule do after it's changed?** In addition to complaining that `defaultValue` should be hyphenated (`default-value`), it will also complain that `selectedItem` should be hyphenated (`v-model:selected-item`)

**Additional context**
If prop names should be hyphenated, then they should always be hyphenated, even when used with `v-model`.

waynzh:
fix #2510

### DIFF
--- a/lib/rules/attribute-hyphenation.js
+++ b/lib/rules/attribute-hyphenation.js
@@ -18,7 +18,7 @@ function getAttributeName(node) {
   }
 
   if (
-    node.key.name.name === 'bind' &&
+    (node.key.name.name === 'bind' || node.key.name.name === 'model') &&
     node.key.argument &&
     node.key.argument.type === 'VIdentifier'
   ) {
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):
Thanks, looks good to me :slightly_smiling_face:

ota-meshi (approved):
LGTM! Thank you!

PR #2513, opened 2024-07-21
Vuex and Pinia support for `no-undef-properties`

marcjfj-vmlyr:
This PR adds support for Vuex and Pinia methods to the `no-undef-properties` rule, allowing properties defined by `mapState`, `mapActions`, `mapGetters`, `mapMutations`, and `mapWritableState` to be recognized as defined properties.

Test cases have been included for this enhancement.


kazupon:
I think this rule should be supported by pinia or 3rd party plugin.
You can discuss in the issue
https://github.com/vuejs/pinia/issues/2612

marcjfj-vmlyr:
> I think this rule should be supported by pinia or 3rd party plugin. You can discuss in the issue [vuejs/pinia#2612](https://github.com/vuejs/pinia/issues/2612)

Sure but it looks like the maintainers are welcoming these kinds of enhancements, as similar fix was added for `vue/no-unused-properties`.  See discussion here https://github.com/vuejs/eslint-plugin-vue/issues/1675.

nirinsanity:
This is much required. Thanks to @marcjfj-vmlyr for working on this. Please merge as soon as possible.

lib/rules/no-undef-properties.js
```diff
@@ -111,6 +111,11 @@ module.exports = {
     ).map(toRegExp)
     const propertyReferenceExtractor = definePropertyReferenceExtractor(context)
     const programNode = context.getSourceCode().ast
+    /**
+     * Property names identified as defined via a Vuex or Pinia helpers
+     * @type {string[]}
+     */
+    const propertiesDefinedByVuexOrPiniaHelpers = []
```
FloEdelmann, line 118: This could be a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), which would make lookup more performant.

Also please rename to `propertiesDefinedByStoreHelpers` to make it a bit more concise.

lib/rules/no-undef-properties.js
```diff
@@ -331,6 +337,55 @@ module.exports = {
...
+              }
+            }
+          } else if (arg.type === 'ArrayExpression') {
+            // e.g. `mapMutations(['add'])`
+            for (const element of arg.elements) {
+              if (
+                !element ||
+                (element.type !== 'Literal' &&
+                  element.type !== 'TemplateLiteral')
+              ) {
```
FloEdelmann, line 379:
```suggestion
              if (!element || !utils.isStringLiteral(element)) {
```

FloEdelmann (changes requested):
Really nice, thank you!

I have some small suggestions.

marcjfj-vmlyr:
@FloEdelmann Thanks for reviewing! I've made the requested changes.

tests/lib/rules/no-undef-properties.js
FloEdelmann, line 1: Could you please add/modify some invalid test cases to make sure that undefined properties are still properly detected when Vuex/Pinia methods are used?

FloEdelmann (changes requested):
Thanks! I have one more change request, then it's good to go from my side :slightly_smiling_face:

marcjfj-vmlyr:
Invalid test cases have been added.

tests/lib/rules/no-undef-properties.js
```diff
@@ -561,7 +561,248 @@ tester.run('no-undef-properties', rule, {
         }
       }
     },
+    {
+      // Vuex
+      filename: 'test.vue',
+      code: `
+      <script>
+      import { mapState } from 'Vuex';
```
FloEdelmann, line 569: Please change `Vuex` to `vuex` in all test cases.

FloEdelmann (changes requested):
Thanks! One more minor thing I just noticed.

tests/lib/rules/no-undef-properties.js
marcjfj-vmlyr: Good catch. Will update!

FloEdelmann (approved):
Looks good to me now :slightly_smiling_face:

horosin:
this looks great, can't wait for it to ship!

ota-meshi (approved):
LGTM! Thank you!

tmcdos:
There is a problem with this fix - it works only if `mapGetters` or `mapState` or `mapMutations` appears in the source code **BEFORE** the actual reference to the property or method.
For example, in the following code no linting error will be emitted
```javascript
  created()
  {
    if (this.getCountries.length === 0) this.fetchCountries();
  },
methods:
{
        ...mapMutations(['setCountries']),
      fetchCountries()
      {
          // ... fetching from server
          this.setCountries(data);
      }
}
```
However, in the following code a linting error `setCountries is not defined vue/no-undef-properties` will be emitted
```javascript
  created()
  {
    if (this.getCountries.length === 0)
    {
          // ... fetching from server
          this.setCountries(data); // <==== linting error - but it should not be
    }
  },
methods:
{
        ...mapMutations(['setCountries']),
}
```

PR #2529, opened 2024-08-15
fix(no-deprecated-slot-attribute): handle v-for with dynamic slot

privatenumber:
Fixes an edge-case where the slot is dynamic derived from a `v-for`.

Previously, it would move the `slot` directive in a new `<template>` above the `v-for`, which broke it.

Now, it also moves the `v-for` to the new `<template>`.


FloEdelmann (approved):
Looks good to me, thanks for your contribution!

ota-meshi (approved):
LGTM! Thank you!

PR #2537, opened 2024-08-28
feat(require-toggle-inside-transition): add additionalDirectives option (#2535)

### LINKED ISSUE #2535: Allow custom toggles in vue/require-toggle-inside-transition

[vue/require-toggle-inside-transition](https://eslint.vuejs.org/rules/require-toggle-inside-transition.html) 

If option is configured, fewer warnings.

I would add a new option to the rule which allows adding additional custom "toggle" directives which will satisfy this check in addition to `v-show` or `v-if`.

```vue
<transition>
  <dialog v-dialog="visible">...</dialog>
</transition>
```

It flags it as an error, because there is no `v-show` or `v-if` on the `<dialog>`

If the rule is configured to include `v-dialog` as a toggle directive, it will no longer flag this case as an error.

**Additional context**

In the example above, `v-dialog` is a custom directive which shows/hides the element using `close()` and `showModal()`. Transitions are also applied in the directive.

cjpearson:
Resolves #2535

### DIFF
--- a/docs/rules/require-toggle-inside-transition.md
+++ b/docs/rules/require-toggle-inside-transition.md
@@ -33,7 +33,34 @@ This rule reports elements inside `<transition>` that do not control the display
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+  "vue/require-toggle-inside-transition": ["error", {
+    "additionalDirectives": []
+  }]
+}
+```
+
+- `additionalDirectives` (`string[]`) ... Custom directives which will satisfy this rule in addition to `v-show` and `v-if`. Should be added without the `v-` prefix.
+
+### `additionalDirectives: ["dialog"]`
+
+<eslint-code-block :rules="{'vue/require-toggle-inside-transition': ['error', {additionalDirectives: ['dialog']}]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <transition><div v-if="show" /></transition>
+  <transition><div v-show="show" /></transition>
+  <transition><dialog v-dialog="show" /></transition>
+
+  <!-- ✗ BAD -->
+  <transition><div /></transition>
+  <transition><div v-custom="show" /></transition>
+<template>
+```
+
+</eslint-code-block>
 
 ## :books: Further Reading
 
--- a/lib/rules/require-toggle-inside-transition.js
+++ b/lib/rules/require-toggle-inside-transition.js
@@ -20,6 +20,25 @@ function isValidBindAppear(vBindAppear) {
   return true
 }
 
+/**
+ * @param {string[]} directives
+ */
+function createDirectiveList(directives) {
+  let str = ''
+
+  for (const [index, directive] of directives.entries()) {
+    if (index === 0) {
+      str += `\`v-${directive}\``
+    } else if (index < directives.length - 1) {
+      str += `, \`v-${directive}\``
+    } else {
+      str += ` or \`v-${directive}\``
+    }
+  }
+
+  return str
+}
+
 module.exports = {
   meta: {
     type: 'problem',
@@ -30,14 +49,34 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/require-toggle-inside-transition.html'
     },
     fixable: null,
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          additionalDirectives: {
+            type: 'array',
+            items: {
+              type: 'string'
+            },
+            uniqueItems: true
+          }
+        },
+        additionalProperties: false
+      }
+    ],
     messages: {
       expected:
-        'The element inside `<transition>` is expected to have a `v-if` or `v-show` directive.'
+        'The element inside `<transition>` is expected to have a {{allowedDirectives}} directive.'
     }
   },
   /** @param {RuleContext} context */
   create(context) {
+    /** @type {Array<string>} */
+    const additionalDirectives =
+      (context.options[0] && context.options[0].additionalDirectives) || []
+    const allowedDirectives = ['if', 'show', ...additionalDirectives]
+    const allowedDirectivesString = createDirectiveList(allowedDirectives)
+
     /**
      * Check if the given element has display control.
      * @param {VElement} element The element node to check.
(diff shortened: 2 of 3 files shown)


lib/rules/require-toggle-inside-transition.js
```diff
@@ -30,14 +30,31 @@ module.exports = {
...
+              type: 'string'
+            }
+          }
+        },
+        additionalProperties: false
+      }
+    ],
     messages: {
       expected:
         'The element inside `<transition>` is expected to have a `v-if` or `v-show` directive.'
```
waynzh, line 49: Should we also update the message to display all allowed directives?

FloEdelmann: Yes, I think that would make sense.

lib/rules/require-toggle-inside-transition.js
```diff
@@ -30,14 +30,31 @@ module.exports = {
...
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          additionalDirectives: {
+            type: 'array',
+            items: {
+              type: 'string'
+            }
```
FloEdelmann, line 41:
```suggestion
            },
            uniqueItems: true
```

FloEdelmann (changes requested):
Looks good, apart from the two small open threads.

lib/rules/require-toggle-inside-transition.js
cjpearson: This sounds good to me as well. Added to the PR

lib/rules/require-toggle-inside-transition.js
```diff
@@ -30,14 +49,38 @@ module.exports = {
...
   create(context) {
+    /** @type {Array<string>} */
+    const additionalDirectives =
+      (context.options[0] && context.options[0].additionalDirectives) || []
+
+    const allowedDirectivesString = createDirectiveList([
+      'if',
+      'show',
+      ...additionalDirectives
+    ])
```
FloEdelmann, line 82:
```suggestion
    const additionalDirectives =
      (context.options[0] && context.options[0].additionalDirectives) || []
    const allowedDirectives = ['if', 'show', ...additionalDirectives]
    const allowedDirectivesString = createDirectiveList(allowedDirectives)
```

lib/rules/require-toggle-inside-transition.js
```diff
@@ -61,12 +104,18 @@ module.exports = {
         element.name !== 'slot' &&
         !utils.hasDirective(element, 'if') &&
         !utils.hasDirective(element, 'show') &&
-        !utils.hasDirective(element, 'bind', 'key')
+        !utils.hasDirective(element, 'bind', 'key') &&
+        !additionalDirectives.some((directive) =>
+          utils.hasDirective(element, directive)
+        )
```
FloEdelmann, line 110: With the above suggestion, this could be simplified to:

```js
      if (
        element.name !== 'slot' &&
        !allowedDirectives.some((directive) =>
          utils.hasDirective(element, directive)
        ) &&
        !utils.hasDirective(element, 'bind', 'key')
      )
```

lib/rules/require-toggle-inside-transition.js
```diff
@@ -20,6 +20,25 @@ function isValidBindAppear(vBindAppear) {
...
+
+  for (const [index, directive] of directives.entries()) {
+    if (index === directives.length - 1) {
+      str += ` or \`v-${directive}\``
+    } else if (index === 0) {
+      str += ` \`v-${directive}\``
+    } else {
+      str += `, \`v-${directive}\``
+    }
+  }
```
FloEdelmann, line 37: It feels odd to have the last string part in the first `if` condition. I'd reorder it to have it more readable:

```suggestion
  for (const [index, directive] of directives.entries()) {
    if (index === 0) {
      str += ` \`v-${directive}\``
    } else if (index < directives.length - 1) {
      str += `, \`v-${directive}\``
    } else {
      str += ` or \`v-${directive}\``
    }
  }
```

tests/lib/rules/require-toggle-inside-transition.js
```diff
@@ -132,6 +141,26 @@ tester.run('require-toggle-inside-transition', rule, {
...
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><transition><dialog v-dialog="show" /></transition></template>',
+      options: [
+        {
+          additionalDirectives: []
+        }
+      ],
+      errors: [{ messageId: 'expected' }]
```
FloEdelmann, line 153: Please also check for the `data` in the new test cases please.

FloEdelmann (changes requested):
Some more small suggestions.

docs/rules/require-toggle-inside-transition.md
```diff
@@ -33,7 +33,42 @@ This rule reports elements inside `<transition>` that do not control the display
...
+
+```vue
+<template>
+  <!-- ✗ BAD -->
+  <transition><div /></transition>
+  <transition><div v-custom="show" /></transition>
+<template>
+```
+
+</eslint-code-block>
```
FloEdelmann, line 71: These two code blocks could be merged into one, right?

````suggestion
<eslint-code-block :rules="{'vue/require-toggle-inside-transition': ['error', {additionalDirectives: ['dialog']}]}">

```vue
<template>
  <!-- ✓ GOOD -->
  <transition><div v-if="show" /></transition>
  <transition><div v-show="show" /></transition>
  <transition><dialog v-dialog="show" /></transition>

  <!-- ✗ BAD -->
  <transition><div /></transition>
  <transition><div v-custom="show" /></transition>
<template>
```

</eslint-code-block>
````

FloEdelmann (commented):
One tiny docs improvement, other than that it looks good to me :slightly_smiling_face: 

Thanks for your contribution!

FloEdelmann (approved):

docs/rules/require-toggle-inside-transition.md
cjpearson: Yep this should work

ota-meshi (approved):
LGTM! Thank you!

PR #2539, opened 2024-08-30
fix(v-on-handler-style): fixed the breaking change caused by the auto-fix on functions without parameters（#2538）

### LINKED ISSUE #2538: Rule vue/v-on-handler-style default auto fix will bring breaking change into code

`vue/v-on-handler-style`

Fewer

In the current default configuration, such as `vue/v-on-handler-style: 'warn'`,  the rule will apply the following formatting by default:

```vue
// before - variant 1
<div @click="()=>changeFunc()" />

// before - variant 2
<div @click="changeFunc()" />

// after
<div @click="changeFunc" />
```

**When the first argument of `changeFunc` is not an `event` and can be `undefined`**, this auto-fix may lead to a breaking change in the code logic. This is because using `@click="changeFunc"` will pass the event object to changeFunc by default, causing logic errors in the subsequent code.

By default, the rule will no longer automatically fix the following two types of `v-on` syntax:

```vue
// type 1
<div @click="()=>changeFunc()" />

// type 2
<div @click="changeFunc()" />
```

HEGGRIA:
Resolves #2538

### DIFF
--- a/lib/rules/v-on-handler-style.js
+++ b/lib/rules/v-on-handler-style.js
@@ -314,6 +314,10 @@ module.exports = {
             return null
           }
           const paramCount = methodParamCountMap.get(idCallExpr.callee.name)
+          // disable the auto-fixed when the node does't have params
+          if (paramCount == 0) {
+            return null
+          }
           if (paramCount != null && paramCount > 0) {
             // The behavior of target method can change given the arguments.
             return null
@@ -368,7 +372,8 @@ module.exports = {
           ) {
             return null
           }
-          if (!isSameParamsAndArgs(idCallExpr)) {
+          // disable the auto-fixed when the node does't have params
+          if (node.params.length === 0 || !isSameParamsAndArgs(idCallExpr)) {
             // It is not a call with the arguments given as is.
             return null
           }
--- a/lib/rules/v-on-handler-style.js
+++ b/lib/rules/v-on-handler-style.js
@@ -315,7 +315,7 @@ module.exports = {
           }
           const paramCount = methodParamCountMap.get(idCallExpr.callee.name)
           // disable the auto-fixed when the node does't have params
-          if (paramCount == 0) {
+          if (idCallExpr.arguments.length === 0 || paramCount === 0) {
             return null
           }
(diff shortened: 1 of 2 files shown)


tests/lib/rules/v-on-handler-style.js
```diff
@@ -81,11 +81,7 @@ tester.run('v-on-handler-style', rule, {
         <button @click="foo()" />
         <button @click="() => foo()" />
       </template>`,
-      output: `<template>
-        <button @click="foo" />
-        <button @click="foo" />
-        <button @click="foo" />
-      </template>`,
+      output: null,
```
FloEdelmann, line 84: The previous autofix should now be added as a [suggestion](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions), so that users can apply it manually.

FloEdelmann, line 84: The previous autofix should now be added as a [suggestion](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions), so that users can apply it manually.

FloEdelmann (changes requested):

tests/lib/rules/v-on-handler-style.js
Heggria: Thank you for your feedback. The code has been updated according to the suggestion.

lib/rules/v-on-handler-style.js
```diff
@@ -300,6 +305,36 @@ module.exports = {
         return false
       }
 
+      // disable the auto-fixed when the node does't have params
+      if (
+        !hasComment /* The statement contains comment and cannot be fixed. */ &&
+        idCallExpr /* The statement is not a simple identifier call and cannot be fixed. */ &&
+        idCallExpr.arguments.length === 0
+      ) {
+        const paramCount = methodParamCountMap.get(idCallExpr.callee.name)
```
ota-meshi, line 314: I think this rule also has the number of arguments defined in the method. If we check that, I think some code can be safely fixed automatically.
If it's unsafe or we can't identify the number of arguments for the method, I think it's good to make it a suggestion.

Issue #2121, opened 2023-04-03
Add support for props destructure


ota-meshi:
**Tell us about your environment**

- **ESLint version:**  v8.36.0
- **eslint-plugin-vue version:** v9.10.0
- **Node version:** v18.12.1

**The problem you want to solve.**

Add support for props destructure. It will be introduced in Vue 3.3.

```vue
<script setup lang="ts">
const { foo = 1, bar = 'ok' } = defineProps<{ foo?: number, bar?: string }>()
</script>
```

**Your take on the correct solution to problem.**

I think we should at least change the rules in the following list.

- [ ] Change `vue/no-setup-props-destructure` rule.
        Destructuring assignment immediately after defineProps() is allowed.
        However, Vue2 doesn't allow to use props destructure, so we may need to split rules.
- [ ] Change `vue/require-valid-default-prop` rule.
- [ ] Change `vue/require-default-prop` rule.
- [ ] Change `vue/no-boolean-default` rule.
        We need to support a new way to specify defaults.

**Additional context**

https://github.com/vuejs/core/pull/7986

ota-meshi:
Since it's still an experimental feature, eslint-plugin-vue won't support it yet until it's GA.

Jungzl:
It is now a [stable feature](https://github.com/vuejs/core/pull/7986#issue-1646843263) I think.

n0099:
It's now stabled in 3.5.0

henribru:
Perhaps a bit niche, but a rule that bans props destructure could also be useful, to keep consistency with existing code that doesn't use it

ota-meshi:
I believe we have merged the necessary changes into the master branch so I will close this issue.

PR #2550, opened 2024-09-13
Add support for props destructure to `vue/no-setup-props-reactivity-loss` rule

ota-meshi:
This PR modifies the `vue/no-setup-props-reactivity-loss` rule to add support for props destructure.

related to #2121

### DIFF
--- a/tests/lib/rules/no-setup-props-destructure.js
+++ b/tests/lib/rules/no-setup-props-destructure.js
@@ -508,21 +508,6 @@ tester.run('no-setup-props-destructure', rule, {
       filename: 'test.vue',
       code: `
       <script setup>
-      const {count} = defineProps({count:Number})
-      </script>
-      `,
-      errors: [
-        {
-          message:
-            'Destructuring the `props` will cause the value to lose reactivity.',
-          line: 3
-        }
-      ]
-    },
-    {
-      filename: 'test.vue',
-      code: `
-      <script setup>
       const props = defineProps({count:Number})
       const {count} = props
       ;({count} = props)
--- a/tests/lib/rules/no-setup-props-reactivity-loss.js
+++ b/tests/lib/rules/no-setup-props-reactivity-loss.js
@@ -251,6 +251,14 @@ tester.run('no-setup-props-reactivity-loss', rule, {
       const count = computed(() => props.count)
       </script>
       `
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+      const { foo = 1, bar = 'ok' } = defineProps({ foo: Number, bar: String })
+      </script>
+      `
     }
   ],
   invalid: [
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-setup-props-destructure.js
```diff
@@ -504,21 +504,6 @@ tester.run('no-setup-props-destructure', rule, {
...
-      </script>
-      `,
-      errors: [
-        {
-          message:
-            'Destructuring the `props` will cause the value to lose reactivity.',
-          line: 3
-        }
-      ]
-    },
```
FloEdelmann, line 521: Should we make this behavior dependent on the Vue version? Or make it configurable to still warn about this? (But not in the deprecated rule.)

Also, why is the same change not necessary in the `no-setup-props-reactivity-loss` tests?

tests/lib/rules/no-setup-props-reactivity-loss.js
FloEdelmann, line 1: Destructured props are only reactive inside `<script setup>`, right? So should we add a test (or is there one already?) that forbids destructuring in `setup()`?

tests/lib/rules/no-setup-props-destructure.js
ota-meshi: > Should we make this behavior dependent on the Vue version? Or make it configurable to still warn about this? (But not in the deprecated rule.)

If necessary, we should create a new rule for Vue2. However, since Vue2 has already reached its EOL, I'm wondering if a new rule are really necessary 🤔 

> Also, why is the same change not necessary in the `no-setup-props-reactivity-loss` tests?

Since `no-setup-props-reactivity-loss` was a deprecated rule and an alias, I wasn't planning on maintaining the test for it.
However, changes to the original rule caused a test to fail, so I removed it.

ota-meshi: The new test was added to `no-setup-props-reactivity-loss.js` 👍 

https://github.com/vuejs/eslint-plugin-vue/pull/2550/files#diff-0c8677d8ffefa333a1434ec8b0c44ca42be0f2553d6bd7f1ac315368007787e2R540-R544

FloEdelmann: Yes, I've seen that.

I was just wondering why the following test case was not part of `no-setup-props-reactivity-loss` before (because that should have been reported before Vue 3.5). Or if it is included, why it still is not failing (because it should not report this case anymore):

```vue
<script setup>
const {count} = defineProps({count:Number})
</script>
```

ota-meshi: I understand, thanks for the explanation 😄 
I added that test case as a `valid` test case.

FloEdelmann (approved):
Thanks for all your adjustments!

PR #2551, opened 2024-09-13
Add support for props destructure to `vue/require-valid-default-prop` rule

ota-meshi:
This PR modifies the `vue/require-valid-default-prop` rule to add support for props destructure.

related to #2121

### DIFF
--- a/tests/lib/rules/require-valid-default-prop.js
+++ b/tests/lib/rules/require-valid-default-prop.js
@@ -223,8 +223,7 @@ ruleTester.run('require-valid-default-prop', rule, {
         parser: require('@typescript-eslint/parser'),
         ecmaVersion: 6,
         sourceType: 'module'
-      },
-      errors: errorMessage('function')
+      }
     },
     {
       filename: 'test.vue',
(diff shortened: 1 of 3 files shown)


tests/lib/rules/require-valid-default-prop.js
```diff
@@ -259,8 +257,7 @@ ruleTester.run('require-valid-default-prop', rule, {
         parser: require('@typescript-eslint/parser'),
         ecmaVersion: 6,
         sourceType: 'module'
-      },
-      errors: errorMessage('function')
+      }
```
FloEdelmann, line 260: These changes are not really related to your changes, right? It was just missed before and you cleaned it up here?

FloEdelmann (approved):

tests/lib/rules/require-valid-default-prop.js
ota-meshi: That's right 😄

aethr:
I think default values for arrays and objects are handled slightly differently when using props destructuring. I don't believe it's necessary to use a function to construct them, as shown in the example in https://vuejs.org/api/sfc-script-setup.html#reactive-props-destructure:

```vue
interface Props {
  msg?: string
  labels?: string[]
}

const { msg = 'hello', labels = ['one', 'two'] } = defineProps<Props>()
```

I've opened an issue: https://github.com/vuejs/eslint-plugin-vue/issues/2578

PR #2565, opened 2024-09-23
fix(utils): also recognize .tsx files as Vue files

haoqunjiang:
I was testing `eslint-config-vue` integration with TypeScript projects.
I was surprised to find out that the rule `vue/multi-word-component-names` does not error on `.tsx` files, such as `Foo.tsx`, while it correctly showed errors for `Foo.jsx` and `Foo.vue`.

Digging into the rule source code I found that it's because `isVueFile('Foo.tsx')` always returns false:
https://github.com/vuejs/eslint-plugin-vue/blob/2dc606ca589ff5cdcd197c3c64ace7e575a6347d/lib/rules/multi-word-component-names.js#L120

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -2681,7 +2681,7 @@ function markVariableAsUsed(context, name, node) {
  * @param {string} path
  */
 function isVueFile(path) {
-  return path.endsWith('.vue') || path.endsWith('.jsx')
+  return path.endsWith('.vue') || path.endsWith('.jsx') || path.endsWith('.tsx')
 }
 
 /**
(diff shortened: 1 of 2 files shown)


FloEdelmann:
Could you please add a test case?

haoqunjiang:
Thanks for the feedback!
Initially, I didn't know there were corresponding tests for this simple utility function.
Now I've found them and added the TSX case.

FloEdelmann (approved):
Great, thank you!

ota-meshi (approved):
Thank you!

PR #2566, opened 2024-09-26
fix: type declarations for configs

thor-juhasz:
Adds proper types for the configs (in the yet unreleased index.d.ts).

If this PR gets rejected, note that the `configs` key is still incorrectly typed with the current `index.d.ts` file.

If only a fix for that is wanted, without the types for the configs themselves, consider this:

```diff
  configs: {
-    [name in keyof ConfigName]: any
+    [name in ConfigName]: any
  }
```

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,24 +1,21 @@
-type ConfigName =
-  | "base"
-  | "essential"
-  | "no-layout-rules"
-  | "recommended"
-  | "strongly-recommended"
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):
Thanks, this looks good! :slightly_smiling_face:

DavidVaness (approved):

tryforceful (approved):

ota-meshi (approved):
Thank you!

Issue #2574, opened 2024-10-14
The new config types in 9.29.0 are broken


thebanjomatic:
**Checklist**

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 9.11.0
- **eslint-plugin-vue version:** 9.29.0
- **Vue version:** 3.5.6
- **Node version:** 20.18.0
- **Operating System:** Windows 11

**Please show your full configuration:**

```js
import eslintPluginVue from 'eslint-plugin-vue';

const baseConfigs = eslintPluginVue.configs['flat/base'];
```

**What did you do?**	
When importing the configs, the types are essentially still any because the types `Linter.FlatConfig` and `Linter.LegacyConfig` are not actually imported or defined within your own declarations. This results in the type being treated as `any` if you have `skipLibChecks` set to `true`, or a build error if you don't.

![image](https://github.com/user-attachments/assets/5d3dd7b7-077c-414c-a420-09ef765e753e)

**What did you expect to happen?**
Config types should use the actual eslint types in their definition instead of just being typed as `any` for consumers..

**What actually happened?**
Types for `Linter` are not being imported in the .d.ts file.
https://unpkg.com/browse/eslint-plugin-vue@9.29.0/lib/index.d.ts

I believe its just missing an import at the top of the file. I'll file a PR.

**Repository to reproduce this issue**
None provided.

thebanjomatic:
@FloEdelmann are you able to push a release for this change, or does this project release on a schedule?

FloEdelmann:
I can't but @ota-meshi could :slightly_smiling_face:

PR #2575, opened 2024-10-14
fix(types): importing Linter namespace from eslint

thebanjomatic:
The types for `Linter.LegacyConfig` and `Linter.FlatConfig[]` weren't being imported or defined by the declaration file, so they were being interpretted as `any` and `any[]` respectively by consumers.

Fixes: #2574


FloEdelmann:
Thanks, but this is a duplicate of #2572, so I'll close this one (as the other one also fixes formatting and was earlier).

PR #2581, opened 2024-10-23
Add option to ignore comments in `vue/multiline-html-element-content-newline`

### LINKED ISSUE #2179: vue/multiline-html-element-content-newline option to ignore comments
vue/multiline-html-element-content-newline

Fewer if new option set

New option, or extension to the `ignores` option. Either a boolean option, default `false`, called something like `allowHTMLComments`, or a token such as `HTMLComments` added to the list of possible `ignores` values. Note I'm not sure about the latter, as my understanding is that ignores list is looking for the element names _before_ the subsequent content without newline (e.g. the `<div>`), where as we'd like to allow multiline for _any_ element, providing that the subsequent element is a comment.

To enable this to be Good:

```vue
<div class="something">  <!-- or use class="other" -->
  <Component />
</div>
```

Currently, it produces the error `vue/multiline-html-element-content-newline – Expected 1 line break after opening tab ('<div>'), but no line breaks found.`

Produce no errors / warnings with same-line HTML comments

**Additional context**
This is a common pattern for our JS code, and we'd like to be able to use the same kind of inline comments within the template:
```js
const meaning = 42  // obviously
```

dsl101:
Fixes #2179

I finally got around to looking at a PR. It seems deceptively simple, so not sure if it broke something elsewhere, but no other tests seemed affected.

### DIFF
--- a/docs/rules/multiline-html-element-content-newline.md
+++ b/docs/rules/multiline-html-element-content-newline.md
@@ -161,8 +161,6 @@ This rule enforces a line break before and after the contents of a multiline ele
 
 </eslint-code-block>
 
-
-
 ## :books: Further Reading
 
 - [no-multiple-empty-lines]
--- a/docs/rules/multiline-html-element-content-newline.md
+++ b/docs/rules/multiline-html-element-content-newline.md
@@ -82,7 +82,8 @@ This rule enforces a line break before and after the contents of a multiline ele
     "vue/multiline-html-element-content-newline": ["error", {
         "ignoreWhenEmpty": true,
         "ignores": ["pre", "textarea", ...INLINE_ELEMENTS],
-        "allowEmptyLines": false
+        "allowEmptyLines": false,
+        "ignoreComments": false
     }]
 }
 ```
@@ -92,6 +93,8 @@ This rule enforces a line break before and after the contents of a multiline ele
 - `ignores` ... the configuration for element names to ignore line breaks style.
     default `["pre", "textarea", ...INLINE_ELEMENTS]`.
 - `allowEmptyLines` ... if `true`, it allows empty lines around content. If you want to disallow multiple empty lines, use [no-multiple-empty-lines] in combination.  
+    default `false`
+- `ignoreComments` ... if `true`, it allows comments to be on the same line as the tag.
     default `false`
 
 ::: info
@@ -142,6 +145,23 @@ This rule enforces a line break before and after the contents of a multiline ele
 ```
 
 </eslint-code-block>
+
+### `"ignoreComments": true`
+
+<eslint-code-block fix :rules="{'vue/multiline-html-element-content-newline': ['error', { ignoreComments: true }]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <div class="red">  <!-- or "blue" -->
+    content
+  </div>
+</template>
+```
+
+</eslint-code-block>
+
+
 
 ## :books: Further Reading
 
--- a/docs/rules/singleline-html-element-content-newline.md
+++ b/docs/rules/singleline-html-element-content-newline.md
@@ -58,7 +58,9 @@ This rule enforces a line break before and after the contents of a singleline el
     "ignoreWhenNoAttributes": true,
     "ignoreWhenEmpty": true,
     "ignores": ["pre", "textarea", ...INLINE_ELEMENTS],
-    "externalIgnores": []
+    "externalIgnores": [],
+    "ignoreComments": false
+
   }]
 }
 ```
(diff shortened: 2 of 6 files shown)


dsl101:
If the approach is OK, I'll add to the docs

FloEdelmann:
Please go ahead and adjust the docs now :wink:

dsl101:
> And could you also add the same option to [`vue/singleline-html-element-content-newline`](https://eslint.vuejs.org/rules/singleline-html-element-content-newline.html) as well please?

I'm unsure how to do this, as a straight copy / paste of the changes into that rule doesn't quite work the same. For example, with `ignoreComments: true` this test failes:
```js
  valid: [{
    code: `
      <template>
        <div attr> <!-- comment --> </div>
      </template>`,
    options: [
      {
        ignoreComments: true
      }
    ]
  }]
```
with these errors:
```
  1) singleLineRule
       valid

      <template>
        <div attr> <!-- comment --> </div>
      </template>:

      AssertionError [ERR_ASSERTION]: Should have no errors but had 2: [
  {
    ruleId: 'singleLineRule',
    severity: 1,
    message: 'Expected 1 line break after opening tag (`<div>`), but no line breaks found.',
    line: 3,
    column: 19,
    nodeType: 'HTMLTagClose',
    messageId: 'unexpectedAfterClosingBracket',
    endLine: 3,
    endColumn: 37,
    fix: { range: [Array], text: '\n' }
  },
  {
    ruleId: 'singleLineRule',
    severity: 1,
    message: 'Expected 1 line break before closing tag (`</div>`), but no line breaks found.',
    line: 3,
    column: 19,
    nodeType: 'HTMLEndTagOpen',
    messageId: 'unexpectedBeforeOpeningBracket',
    endLine: 3,
    endColumn: 37,
    fix: { range: [Array], text: '\n' }
  }
]
```

But if I remove the whitespace around the comment:
```js
  valid: [{
    code: `
      <template>
        <div attr><!-- comment --></div>
      </template>`,
    options: [
      {
        ignoreComments: true
      }
    ]
  }]
```
the test passes fine. I didn't see this issue in the multiline rule.

In this test here:
```js
        if (
          ignoreWhenEmpty &&
          elem.children.length === 0 &&
          template.getFirstTokensBetween(
            elem.startTag,
            elem.endTag,
            getTokenOption
          ).length === 0
        ) {
          return
        }
```
I'm not sure why it's required to test `elem.children.length === 0` _and_ `getFirstTokensBetween().length === 0`—indeed, commenting that line out:
```js
        if (
          ignoreWhenEmpty &&
          // elem.children.length === 0 &&
          template.getFirstTokensBetween(
            elem.startTag,
            elem.endTag,
            getTokenOption
          ).length === 0
        ) {
          return
        }
```
then makes this valid:
```html
      <template>
        <div attr> <!-- comment --> </div>
      </template>`
```
but still reports errors for this:
```html
      <template>
        <div attr> content <!-- comment --> </div>
      </template>`
```

FloEdelmann:
> I'm not sure why it's required to test `elem.children.length === 0` _and_ `getFirstTokensBetween().length === 0`

Probably it is indeed not needed, seems like both were added at the same time in #684. I guess if the other tests are passing, then removing the `getFirstTokensBetween` check should be fine.

dsl101:
Ah, this is a bit more complicated than I thought. Without my proposed changes, the default settings for `singleline-html-element-content-newline` give these test results:

1. `<div></div>` → OK
2. `<div> </div>` → OK
3. `<div attr></div>` → OK
4. `<div attr> </div>` → Fail

I'm not sure if the behaviour here is really consistent—shouldn't 2 and 4 give the same result? Because the setting is `ignoreWhenNoAttributes` and defaults to `true`, I can't see a way to configure that case 4 passes the test. Is the only option here to just disable the `singleline-html-element-content-newline` rule altogether?

It complicates my use case, as it's very common to put white space around comments, but then that white space counts as 'content':

`<div attr><!-- comment --></div>` → OK
`<div attr> <!-- comment --> </div>` → Fail

But I think that's OK, and is consistent with the existing rules / tests. I'll push an update with documentation for you to see.

docs/rules/multiline-html-element-content-newline.md
FloEdelmann, line 1: By the way, while you're at it: Could you please add a "Related rules" section here and link to `vue/singleline-html-element-content-newline` from this rule docs page?

And also vice-versa a link from `vue/singleline-html-element-content-newline` to `vue/multiline-html-element-content-newline`.

**EDIT:** I've done that now in https://github.com/vuejs/eslint-plugin-vue/pull/2911.

---

*Note: the following comments are replies to https://github.com/vuejs/eslint-plugin-vue/pull/2581#issuecomment-2435604099 rather than this comment.*

dsl101: Yes, since comments are ignored, you would _almost_ always want `allowEmptyLines: true` since it's not clear how many empty lines you'll have. I couldn't see an easy way around that.

dsl101: I'd love to spend more time digging in, but I'm afraid I have to get back to the project I'm using this on! I guess I'll just have to disable those rules for now...

FloEdelmann: @dsl101 Do you think you'll get to this PR at some point? Or should we close it and let someone else start fresh?

dsl101: Yeah, I would love to, but I'm really not sure when. I guess there have been plenty of other changes since I started on this, and I'd need to refresh my memory or just start over with a better design given where this got to.

If someone else is keen, that's probably the best bet.

PR #2591, opened 2024-11-02
fix(require-explicit-slots): ignore attribute binding

### LINKED ISSUE #2589: `require-explicit-slots` reports on slots with dynamic names
**Checklist**

- **ESLint version:** 9.13.0
- **eslint-plugin-vue version:** 9.30.3
- **Vue version:** 3.5.12
- **Node version:** 
- **Operating System:** 

**What did you do?**

Hello! `vue/require-explicit-slots` reports on slots with dynamic names. Paste the following code [on the rule page](https://eslint.vuejs.org/rules/require-explicit-slots.html):

```vue
<template>
  <div>
    <slot :name="`${hello} ${world}`"></slot>

    <slot :name="'foo'"></slot>
  </div>
</template>

<script setup lang="ts">
const hello = 'hello';
const world = 'world';

defineSlots<{
  foo: () => unknown;
}>();
</script>
```

*(Obviously, I don't expect the second slot not be reported - I just decided to test both options)*

I think that either (A) such slots should just be ignored or (B) the rule should require the presence of something like `{[K in string]: () => unknown;}` in `defineSlots` (although the intersections `&` are not recognized at this moment).

**What actually happened?**

waynzh:
fixes #2589.

Previously: When `node.directive` was true, `slotName` was incorrectly set to 'default', and it would check if it was explicitly defined.
Now: If `node.directive` is true, the report is simply skipped.

### DIFF
--- a/lib/rules/require-explicit-slots.js
+++ b/lib/rules/require-explicit-slots.js
@@ -149,7 +149,7 @@ module.exports = {
           )
 
           /** @type {string | undefined} */
-          let slotName = ''
+          let slotName
           if (!nameNode) {
             // If no slot name is declared, default to 'default'
             slotName = 'default'
--- a/lib/rules/require-explicit-slots.js
+++ b/lib/rules/require-explicit-slots.js
@@ -137,13 +137,27 @@ module.exports = {
         }
       }),
       utils.defineTemplateBodyVisitor(context, {
+        /** @param {VElement} node */
         "VElement[name='slot']"(node) {
-          let slotName = 'default'
+          const nameNode = node.startTag.attributes.find(
+            (node) =>
+              (!node.directive && node.key.name === 'name') ||
+              (node.directive &&
+                node.key.name.name === 'bind' &&
+                node.key.argument?.type === 'VIdentifier' &&
+                node.key.argument?.name === 'name')
+          )
 
-          const slotNameAttr = utils.getAttribute(node, 'name')
-
-          if (slotNameAttr?.value) {
-            slotName = slotNameAttr.value.value
+          /** @type {string | undefined} */
+          let slotName = ''
+          if (!nameNode) {
+            // If no slot name is declared, default to 'default'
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-explicit-slots.js
```diff
@@ -160,6 +160,21 @@ tester.run('require-explicit-slots', rule, {
...
         }
       }
+    },
+    // ignore attribute binding
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+        <div>
+          <slot :name="'foo'"></slot>
```
FloEdelmann, line 170: Maybe we can add support for string literals, but ignore other attribute bindings? That shouldn't be too hard with the `isStringLiteral` and `getStringLiteralValue` util functions.

waynzh: Done ✅

lib/rules/require-explicit-slots.js
```diff
@@ -137,20 +165,28 @@ module.exports = {
...
+            const expression = nameNode.value?.expression
+            // ignore attribute binding except string literal
+            if (!expression || !utils.isStringLiteral(expression)) {
+              return
+            }
+
+            const name = utils.getStringLiteralValue(expression) || undefined
+            reportMissingSlot(node, name)
+          } else {
+            // check and report if slot name is declared or is undefined
```
FloEdelmann, line 188: I think this comment doesn't add much value, since two lines above the same thing happens and the function name is quite descriptive.

FloEdelmann (approved):
Looks good to me, I only have one more minor comment.

lib/rules/require-explicit-slots.js
waynzh: Yeah, makes sense to me. Thanks for the review!

ota-meshi (approved):
LGTM! Thank you very much for your contributions!

PR #2585, opened 2024-10-26
fix(custom-event-name-casing): check defineEmits variable name in template

waynzh:
fix #2577

PR #2592, opened 2024-11-02
refactor: store the defineEmits variable name

waynzh:
related to #2585

Refactored to store the `defineEmits` variable name for checking `defineEmits` variable name usage in the template.

### DIFF
--- a/lib/rules/no-unused-emit-declarations.js
+++ b/lib/rules/no-unused-emit-declarations.js
@@ -56,18 +56,6 @@ function hasReferenceId(value, setupContext) {
   )
 }
 
-/**
- * Check if the given name matches emitReferenceIds variable name
- * @param {string} name
- * @param {Set<Identifier>} emitReferenceIds
- * @returns {boolean}
- */
-function isEmitVariableName(name, emitReferenceIds) {
-  if (emitReferenceIds.size === 0) return false
-  const emitVariable = emitReferenceIds.values().next().value.name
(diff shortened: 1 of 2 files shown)


FloEdelmann (commented):
Nice! Does this add support for other cases previously not covered? If so, could you add tests for those please?

Otherwise, the refactoring looks fine to me.

waynzh:
yes, can check out here [no-unused-emit-declarations L105](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-unused-emit-declarations.js#L105) + [require-explicit-emits L474](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/require-explicit-emits.js#L474)

FloEdelmann:
I'm confused. It sounds like there are behavior changes the rules? But none of the existing tests fail, so the behavior changes should be covered by new tests.

waynzh:
> Does this add support for other cases previously not covered

Sry, I mean NO 🤖, this does not add support for any additional cases.
It is purely a refactor of the existing code, as I found the previous implementation to be somewhat redundant and wanted to keep it consistent with this https://github.com/vuejs/eslint-plugin-vue/pull/2585.

FloEdelmann (approved):
Ah, I see. Thanks for the clarification and for the refactoring! LGTM then.

ota-meshi (approved):
LGTM! Thanks!

PR #2595, opened 2024-11-03
feat(no-empty-component-block): support auto fix

### LINKED ISSUE #2564: autofix for vue/no-empty-component-block
- Rule to change is vue/no-empty-component-block

- This change would not change the amount of errors/warnings

- Preferably default behaviour

```vue
<script setup lang="ts">
  defineProps<{...}>()
</script>

<template>
  <h1>My awesome title</h1>
</template>

<style scoped lang="postcss"></style>
```

- Currently the rule marks the style tag as an error, but you would have to remove it manually

- After change eslint will be able to remove the style tag with `pnpm eslint --fix`

**Additional context**

waynzh:
resolve #2564

~~Chose this approach is because I couldn't find a better way to handle reporting & fixing each `emptyBlock` within a for loop,  as removing the first block will affect the range of the second block😵‍💫.~~

UPDATE: support auto delete empty blocks

### DIFF
--- a/docs/rules/no-empty-component-block.md
+++ b/docs/rules/no-empty-component-block.md
@@ -10,6 +10,8 @@ since: v7.0.0
 
 > disallow the `<template>` `<script>` `<style>` block to be empty
 
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
 ## :book: Rule Details
 
 This rule disallows the `<template>` `<script>` `<style>` block to be empty.
@@ -17,7 +19,7 @@ This rule disallows the `<template>` `<script>` `<style>` block to be empty.
 This rule also checks block what has attribute `src`.
 See [Vue Single-File Component (SFC) Spec](https://vue-loader.vuejs.org/spec.html#src-imports).
 
-<eslint-code-block :rules="{'vue/no-empty-component-block': ['error']}">
+<eslint-code-block fix :rules="{'vue/no-empty-component-block': ['error']}">
 
 ```vue
 <!-- ✓ GOOD -->
--- a/docs/rules/no-empty-component-block.md
+++ b/docs/rules/no-empty-component-block.md
@@ -19,7 +19,7 @@ This rule disallows the `<template>` `<script>` `<style>` block to be empty.
 This rule also checks block what has attribute `src`.
 See [Vue Single-File Component (SFC) Spec](https://vue-loader.vuejs.org/spec.html#src-imports).
 
-<eslint-code-block fix :rules="{'vue/no-empty-component-block': ['error']}">
+<eslint-code-block fix :rules="{'vue/no-empty-component-block': ['error', { autofix: true }]}">
 
 ```vue
 <!-- ✓ GOOD -->
@@ -64,7 +64,15 @@ p {
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+  "vue/no-duplicate-attributes": ["error", {
+    "autofix": false,
+  }]
+}
+```
+
+- `"autofix"` ... If `true`, enable autofix. (Default: `false`)
 
 ## :rocket: Version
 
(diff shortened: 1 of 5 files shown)


lib/rules/no-empty-component-block.js
```diff
@@ -85,6 +99,29 @@ module.exports = {
...
           ) {
+            emptyBlocks.push(componentBlock)
+          }
+        }
+
+        if (emptyBlocks.length === 0) return
+
+        if (autofix) {
+          const firstEmptyBlock = emptyBlocks[0]
+          context.report({
```
ota-meshi, line 110: I don't think the option changes the reported location.
They may be ignored by the `eslint-disable` directive, so I think it's necessary to report them one by one and auto fix them for each report.
Could you please change it?

ota-meshi (changes requested):
Thank you for this PR!

lib/rules/no-empty-component-block.js
ota-meshi: > Chose this approach is because I couldn't find a better way to handle reporting & fixing each emptyBlock within a for loop, as removing the first block will affect the range of the second block😵‍💫.

I think a good way to test removing them in one go is to have some whitespace between each block.
Adjacent autofixes will not be fixed at the same time to avoid conflicts.

waynzh: I'll look into it later, thanks.

lib/rules/no-empty-component-block.js
```diff
@@ -85,16 +99,29 @@ module.exports = {
...
+            messageId: 'unexpected',
+            data: {
+              blockName: componentBlock.name
+            },
+            fix: autofix
+              ? function* (fixer) {
+                  for (const componentBlock of emptyBlocks) {
+                    yield fixer.remove(componentBlock)
+                  }
+                }
```
ota-meshi, line 121: Could you please change it to auto-fix only reported block?

```suggestion
              ? (fixer) => fixer.remove(componentBlock)
```

docs/rules/no-empty-component-block.md
```diff
@@ -62,7 +64,15 @@ p {
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+  "vue/no-duplicate-attributes": ["error", {
```
FloEdelmann, line 69:
```suggestion
  "vue/no-empty-component-block": ["error", {
```

docs/rules/no-empty-component-block.md
```diff
@@ -62,7 +64,15 @@ p {
...
+    "autofix": false,
+  }]
+}
+```
+
+- `"autofix"` ... If `true`, enable autofix. (Default: `false`)
```
FloEdelmann, line 75: Why does this have to be an option? I think always adding the autofix would be fine and would reduce the complexity a bit.

FloEdelmann (changes requested):

docs/rules/no-empty-component-block.md
waynzh: I treated it as providing an auto-fix feature, and upgrading the package wouldn't change the current code. Of course, I think it's also reasonable to directly fix it in the next minor version.

ota-meshi: I think it's okay if there are no options.
My only concern is that if we have autofix on save enabled in our editor, it will be removed if we temporarily save while writing code.
But other than that, I think it's better to have autofix enabled all the time. If it gets removed, you can probably undo it with Ctrl(Option)+Z.

waynzh: I'll update the doc and tests later

lib/rules/no-empty-component-block.js
```diff
@@ -70,6 +70,9 @@ module.exports = {
 
     return {
       Program() {
+        /** @type {VElement[]} */
+        const emptyBlocks = []
```
FloEdelmann, line 74: I think this array can be removed again; better report the errors directly when iterating `componentBlocks`.

FloEdelmann: No worries :slightly_smiling_face:

PR #2608, opened 2024-11-15
fix(prefer-use-template-ref): add support for `shallowRef`

ntnyq:
This PR adds `shallowRef` support to rule `prefer-use-template-ref`.

### DIFF
--- a/lib/rules/prefer-use-template-ref.js
+++ b/lib/rules/prefer-use-template-ref.js
@@ -22,13 +22,13 @@ module.exports = {
     type: 'suggestion',
     docs: {
       description:
-        'require using `useTemplateRef` instead of `ref/shallowRef` for template refs',
+        'require using `useTemplateRef` instead of `ref`/`shallowRef` for template refs',
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/prefer-use-template-ref.html'
     },
     schema: [],
     messages: {
-      preferUseTemplateRef: "Replace 'ref/shallowRef' with 'useTemplateRef'."
+      preferUseTemplateRef: "Replace '{{name}}' with 'useTemplateRef'."
     }
   },
   /** @param {RuleContext} context */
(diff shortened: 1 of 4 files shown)


lib/rules/prefer-use-template-ref.js
```diff
@@ -18,13 +22,13 @@ module.exports = {
     type: 'suggestion',
     docs: {
       description:
-        'require using `useTemplateRef` instead of `ref` for template refs',
+        'require using `useTemplateRef` instead of `ref/shallowRef` for template refs',
```
FloEdelmann, line 25:
```suggestion
        'require using `useTemplateRef` instead of `ref`/`shallowRef` for template refs',
```

lib/rules/prefer-use-template-ref.js
```diff
@@ -18,13 +22,13 @@ module.exports = {
...
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/prefer-use-template-ref.html'
     },
     schema: [],
     messages: {
-      preferUseTemplateRef: "Replace 'ref' with 'useTemplateRef'."
+      preferUseTemplateRef: "Replace 'ref/shallowRef' with 'useTemplateRef'."
```
FloEdelmann, line 31: Can you please use a [message placeholder](https://eslint.org/docs/latest/extend/custom-rules#using-message-placeholders) to report either `ref` or `shallowRef` please? Please also adjust the tests then.

FloEdelmann (changes requested):
Thanks for the enhancement! I have a few minor comments but they should be easy to resolve.

lib/rules/prefer-use-template-ref.js
ntnyq: Done!

FloEdelmann (approved):
Thanks, looks good to me :slightly_smiling_face:

FloEdelmann:
I'll merge this without @ota-meshi's review now to make conflict resolution possible in #2612. But feel free to comment on it afterwards!

PR #2610, opened 2024-11-16
feat(no-v-text-v-html-on-component): add ignore namespace option

### LINKED ISSUE #2458: Namespace handling makes it impossible to lint recursive SVG / MathML components

Not exactly a specific rule, but I have been motivated by `vue/no-v-text-v-html-on-component` being falsely triggered on MathML tags. `isCustomComponent()` is being called from 17 rules. The problem lies with the fact that every composition component starts with a top-level namespace of HTML, which makes it impossible to create partial, especially recursive components within another namespace without confusing the linter.

Fewer.

There are multiple options:

1. Stop examining namespaces when determining whether something is an element or a component.
2. Treat the HTML namespace as valid namespace for SVG and MathML tags.
3. Looks for `xmlns` on the top-level `<template/>` and adjust namespace accordingly.
4. Implement (1) or (2), but gate it behind a configuration option.

```vue
<script setup type="ts">
// This lives inside MathExpression.vue

import { Expression } from './expression.ts'

defineProps<{
  expression: Expression
}>()
</script>

<template>
  <mfrac v-if="expression instanceof Fraction">
    <mrow>
      <MathExpression v-text="expression.numerator" />
    </mrow>
    <mrow>
      <MathExpression v-text="expression.denominator" />
    </mrow>
  </mfrac>
  <!-- more cases omitted for brevity -->
  <mn v-else v-text="expression.value" />
</template>
```

Triggers `vue/no-v-text-v-html-on-component` because it does not see `<mfrac/>` as being inside MathML namespace. Which is impossible to realize, because `<math/>` cannot be nested and is only specified at the top level. This component recurses into the data, rendering the full MathML structure.

In case of (1), (2), (4) it won't consider using `<mfrac/>` in this context as an error.
In case of (3):

```vue
<template xmlns="http://www.w3.org/1998/Math/MathML">
  <mi v-text="'x'" />
</template>
```

It will accept this otherwise invalid, but (probably) harmless syntax to indicate that we are inside MathML namespace and proceed from there on, possibly marking HTML and SVG elements as out of place.

**Additional context**

This obviously applies to SVG as well, although there is not that high probability somebody will be doing anything recursive in there. Math is a different story, though.

waynzh:
resolve #2458

### DIFF
--- a/docs/rules/no-v-text-v-html-on-component.md
+++ b/docs/rules/no-v-text-v-html-on-component.md
@@ -25,11 +25,15 @@ If you use v-text / v-html on a component, it will overwrite the component's con
   <!-- ✓ GOOD -->
   <div v-text="content"></div>
   <div v-html="html"></div>
+  <svg><g v-text="content" /></svg>
+  <math><mi v-text="content" /></math>
   <MyComponent>{{ content }}</MyComponent>
 
   <!-- ✗ BAD -->
   <MyComponent v-text="content"></MyComponent>
   <MyComponent v-html="html"></MyComponent>
+  <g v-text="content" />
+  <mi v-text="content" />
 </template>
 ```
 
@@ -39,14 +43,15 @@ If you use v-text / v-html on a component, it will overwrite the component's con
 
 ```json
 {
-  "vue/no-v-text-v-html-on-component": [
-    "error",
-    { "allow": ["router-link", "nuxt-link"] }
-  ]
+  "vue/no-v-text-v-html-on-component": ["error", {
+    "allow": ["router-link", "nuxt-link"],
+    "ignoreElementNamespaces": false
+  }]
 }
 ```
 
 - `allow` (`string[]`) ... Specify a list of custom components for which the rule should not apply.
+- `ignoreElementNamespaces` (`boolean`) ... Specify whether to ignore the namespace restrictions for SVG and MathML elements, so that the rule should not apply. Default is `false`.
 
 ### `{ "allow": ["router-link", "nuxt-link"] }`
 
@@ -60,6 +65,20 @@ If you use v-text / v-html on a component, it will overwrite the component's con
 
   <!-- ✗ BAD -->
   <MyComponent v-html="content" />
+</template>
+```
+
+</eslint-code-block>
+
+### `{ "ignoreElementNamespaces": true }`
+
+<eslint-code-block :rules="{'vue/no-v-text-v-html-on-component': ['error', { ignoreElementNamespaces: true }]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <g v-text="content" /> <!-- <svg> -->
+  <mi v-text="content" /> <!-- <math> -->
 </template>
 ```
 
--- a/lib/rules/no-v-text-v-html-on-component.js
+++ b/lib/rules/no-v-text-v-html-on-component.js
@@ -26,6 +26,9 @@ module.exports = {
               type: 'string'
             },
             uniqueItems: true
+          },
+          ignoreElementNamespaces: {
+            type: 'boolean'
           }
         },
         additionalProperties: false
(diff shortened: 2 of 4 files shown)


docs/rules/no-v-text-v-html-on-component.md
```diff
@@ -39,14 +43,15 @@ If you use v-text / v-html on a component, it will overwrite the component's con
...
-  ]
+  "vue/no-v-text-v-html-on-component": ["error", {
+    "allow": ["router-link", "nuxt-link"],
+    "ignoreElementNamespaces": false
+  }]
 }
 ```
 
 - `allow` (`string[]`) ... Specify a list of custom components for which the rule should not apply.
+- `ignoreElementNamespaces` (`boolean`) ... Specify whether to ignore the namespace restrictions for SVG and MathML elements, so that the rule should not apply. Default is `false`.
```
FloEdelmann, line 54:
```suggestion
- `ignoreElementNamespaces` (`boolean`) ... If `true`, always treat SVG and MathML tag names as HTML elements, even if they are not used inside a SVG/MathML root element. Default is `false`.
```

docs/rules/no-v-text-v-html-on-component.md
```diff
@@ -65,6 +70,20 @@ If you use v-text / v-html on a component, it will overwrite the component's con
...
 
+### `{ "ignoreElementNamespaces": true }`
+
+<eslint-code-block :rules="{'vue/no-v-text-v-html-on-component': ['error', { ignoreElementNamespaces: true }]}">
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <g v-text="content" /> <!-- <svg> -->
+  <mi v-text="content" /> <!-- <math> -->
```
FloEdelmann, line 81:
```suggestion
  <g v-text="content" /> <!-- SVG element not inside of <svg> -->
  <mi v-text="content" /> <!-- MathML element not inside of <math> -->
```

lib/rules/no-v-text-v-html-on-component.js
```diff
@@ -56,13 +61,29 @@ module.exports = {
       )
     }
 
+    /** @param {VElement} element */
+    function isCustomComponent(element) {
```
FloEdelmann, line 65: Maybe better add the new logic to the `isCustomComponent` util function with an optional second parameter `ignoreElementNamespaces`? E.g. something like this:

```js
  /**
   * Check whether the given node is a custom component or not.
   * @param {VElement} node The start tag node to check.
   * @param {boolean} [ignoreElementNamespaces=false] If `true`, ignore element namespaces.
   * @returns {boolean} `true` if the node is a custom component.
   */
  isCustomComponent(node, ignoreElementNamespaces = false) {
    if (
      hasAttribute(node, 'is') ||
      hasDirective(node, 'bind', 'is') ||
      hasDirective(node, 'is')
    ) {
      return true
    }

    const isHtmlName = this.isHtmlWellKnownElementName(node.rawName)
    const isSvgName = this.isSvgWellKnownElementName(node.rawName)
    const isMathName = this.isMathWellKnownElementName(node.rawName)

    if (ignoreElementNamespaces) {
      return !isHtmlName && !isSvgName && !isMathName
    }

    return (
      (this.isHtmlElementNode(node) && !isHtmlName) ||
      (this.isSvgElementNode(node) && !isSvgName) ||
      (this.isMathElementNode(node) && !isMathName)
    )
  }
```

tests/lib/rules/no-v-text-v-html-on-component.js
FloEdelmann, line 1: Good test cases!

FloEdelmann (changes requested):

FloEdelmann (approved):

FloEdelmann:
The test failure is probably caused by an ESLint update, so no need to worry about this.

ota-meshi (approved):
LGTM! Thank you!

PR #2611, opened 2024-11-16
feat: add `restricted-component-names` rule

### LINKED ISSUE #2568: Only allow specific compontent names

Enforce consistency component name prefix. This rule should accept options to enforce component matches names or name patterns.
 
[ ] Other (please specify:)

```vue
<!-- Good -->
<keep-alive></keep-alive> <!-- builtin -->
<button/> <!-- html tag -->
<foo-button/> <!-- allowed with option `/^foo\-/` -->

<!-- Bad -->
<bar-button/>
```

**Additional context**

N/A

waynzh:
resolve #2568

Referenced the [vue/no-reserved-component-names](https://eslint.vuejs.org/rules/no-reserved-component-names) rule.


lib/rules/valid-component-name.js
```diff
@@ -0,0 +1,101 @@
...
+const kebabCaseElements = [
+  'annotation-xml',
+  'color-profile',
+  'font-face',
+  'font-face-src',
+  'font-face-uri',
+  'font-face-format',
+  'font-face-name',
+  'missing-glyph'
+]
```
FloEdelmann, line 29: Are those actually tag names that should always be allowed? Where does this list come from?

lib/rules/valid-component-name.js
```diff
@@ -0,0 +1,101 @@
...
+            type: 'array',
+            items: { type: 'string' },
+            uniqueItems: true,
+            additionalItems: false
+          }
+        }
+      }
+    ],
+    messages: {
+      invalidName: 'Component name "{{name}}" is not valid.'
```
FloEdelmann, line 69:
```suggestion
      invalidName: 'Component name "{{name}}" is not allowed.'
```

lib/rules/valid-component-name.js
```diff
@@ -0,0 +1,101 @@
...
+  ...RESERVED_NAMES_IN_VUE,
+  ...RESERVED_NAMES_IN_VUE3,
+  ...RESERVED_NAMES_IN_OTHERS
+])
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce consistency in component names',
```
FloEdelmann, line 49:
```suggestion
      description: 'enforce using only specific component names',
```

FloEdelmann (changes requested):

lib/rules/valid-component-name.js
waynzh: I referred to a [previous PR comment](https://github.com/vuejs/eslint-plugin-vue/pull/757#discussion_r245944977). It's listed in the [Custom Elements Standard](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name).

I'll take another look to see if there's a more detailed explanation. If there are related inputs, feel free to add them.

lib/rules/restricted-component-names.js
```diff
@@ -0,0 +1,101 @@
...
+  ...RESERVED_NAMES_IN_VUE,
+  ...RESERVED_NAMES_IN_VUE3,
+  ...RESERVED_NAMES_IN_OTHERS
+])
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce using only specific in component names',
```
FloEdelmann, line 49:
```suggestion
      description: 'enforce using only specific component names',
```

lib/rules/valid-component-name.js
waynzh: Oh, I misunderstood this at first. These reserved names can be used in HTML, but they cannot be used as names for custom elements.
I referred to a [previous PR comment](https://github.com/vuejs/eslint-plugin-vue/pull/757#discussion_r245944977). It's listed in the [Custom Elements Standard](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name).

Deleted ✅

FloEdelmann: Ah, that makes sense :+1:

lib/rules/restricted-component-names.js
```diff
@@ -0,0 +1,90 @@
...
+  ...deprecatedHtmlElements,
+  ...svgElements
+])
+
+const reservedNames = new Set([
+  ...RESERVED_NAMES_IN_HTML,
+  ...RESERVED_NAMES_IN_VUE,
+  ...RESERVED_NAMES_IN_VUE3,
+  ...RESERVED_NAMES_IN_OTHERS
+])
```
FloEdelmann, line 32: There is no need to create multiple `Set`s from arrays, only to spread them into another array that is used to create the final `reservedNames` `Set`:

```suggestion
const htmlElements = require('../utils/html-elements.json')
const deprecatedHtmlElements = require('../utils/deprecated-html-elements.json')
const svgElements = require('../utils/svg-elements.json')
const vue2builtinComponents = require('../utils/vue2-builtin-components')
const vue3builtinComponents = require('../utils/vue3-builtin-components')

const reservedNames = new Set([
  ...htmlElements,
  ...deprecatedHtmlElements,
  ...svgElements,
  ...vue2builtinComponents,
  ...vue3builtinComponents
])
```

FloEdelmann (changes requested):

FloEdelmann (approved):
Looks good to me now, thanks! :slightly_smiling_face:

ota-meshi (approved):
LGTM! Thank you!!!

PR #2484, opened 2024-06-25
fix(require-explicit-slots): Add support for type reference

sanderdewinter:
Fixes #2398

PR #2617, opened 2024-11-19
fix(require-explicit-slots): add support for type references

### LINKED ISSUE #2398: `require-explicit-slots` doesn't recognize interface passed to the macro

- **ESLint version:** `8.56.0`
- **eslint-plugin-vue version:** `9.21.1`
- **Vue version:** `3.4.15`
- **Node version:** `20.10.0`
- **Operating System:** `Mac OS 14.2.1`

```json5
module.exports = {
  parser: 'vue-eslint-parser',

  parserOptions: {
  },

  extends: ['plugin:vue/base'],

  rules: {
    'vue/require-explicit-slots': 'error',
  },
};
```

**What did you do?**	

```vue
<template>
  <div>
    <slot />
  </div>
</template>

<script lang="ts" setup>
interface Slots {
  default: () => string;
}

defineSlots<Slots>();
</script>

```

No error reported.

**What actually happened?**
```
test.vue
  3:5  error  Slots must be explicitly defined  vue/require-explicit-slots

```

https://stackblitz.com/edit/stackblitz-starters-reg1dy?file=test.vue

```
~$ npm run lint
```

chadwickellis:
This pull request resolves #2398 using an approach requested within #2484.


typings/eslint-plugin-vue/util-types/utils.ts
```diff
@@ -191,6 +192,66 @@ export type ComponentEmit =
   | ComponentInferTypeEmit
   | ComponentUnknownEmit
 
+type ComponentArraySlotDetectName = {
```
ota-meshi, line 195: Hmm🤔. I think defineSlots only accepts Type literals or Interfaces or references that define function properties, so I don't think `type: 'array'` and `type: 'object'` are used at all. Am I wrong?
If there is logic or types that we don't use, can you remove them?

ota-meshi (changes requested):
Thank you for this PR!
It looks almost good, but I have one comment.

typings/eslint-plugin-vue/util-types/utils.ts
chadwickellis: Thank you for your feedback. I've removed the redundant type definitions and the associated utility method. Let me know if there's anything else that needs adjustment.

ota-meshi (approved):
LGTM! Thanks!

PR #2626, opened 2024-11-27
Remove legacy schema from `vue/custom-event-name-casing`

FloEdelmann:
See https://github.com/vuejs/eslint-plugin-vue/blame/207eb981df3bbab5d63846d7456d0925ab82394c/lib/rules/custom-event-name-casing.js#L89-L93

### DIFF
--- a/lib/rules/custom-event-name-casing.js
+++ b/lib/rules/custom-event-name-casing.js
@@ -54,18 +54,6 @@ function getCalleeMemberNode(node) {
   return null
 }
 
-const OBJECT_OPTION_SCHEMA = {
-  type: 'object',
-  properties: {
-    ignores: {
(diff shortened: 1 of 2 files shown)


lib/rules/custom-event-name-casing.js
```diff
@@ -85,11 +85,6 @@ module.exports = {
             },
             OBJECT_OPTION_SCHEMA
           ]
-        },
```
waynzh, line 88: should we also remove the `anyof` keyword

FloEdelmann: Oh yeah, I forgot that.

ota-meshi (approved):
LGTM!

PR #2631, opened 2024-11-30
docs: fix missing options `allow` value

ntnyq:
This PR adds the missing value for option `allow` to rule `vue/restricted-component-names` docs.


FloEdelmann (approved):
Thank you!

PR #2632, opened 2024-11-30
feat(prefer-use-template-ref): add support for fix option

Thomasan1999:


lib/utils/index.js
```diff
@@ -2208,6 +2208,16 @@ module.exports = {
...
     )
+  },
+
+  /**
+   * TODO: replace usages with 'array.at(-1)' once the minimum supported Node.js version is at least 16.6.0
+   * @template T
+   * @param array {T[]}
+   * @returns {T | undefined}
+   * */
+  getLastArrayElement(array) {
```
ota-meshi, line 2219: This is not necessary, just write it inline.
We can fix that later with a prefer-at rule.
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-at.md

Thomasan1999: done ✅

tests/lib/rules/prefer-use-template-ref.js
FloEdelmann, line 1: Can you please add an test case using TypeScript?

```vue
<template>
  <div ref="root" />
</template>
<script setup lang="ts">
  import { ref } from 'vue';
  const root = ref<HTMLDivElement>();
</script>
```

should get autofixed to

```vue
<template>
  <div ref="root" />
</template>
<script setup lang="ts">
  import { ref,useTemplateRef } from 'vue';
  const root = useTemplateRef<HTMLDivElement>('root');
</script>
```

Note that you'll need to specify the parser in the test:

```js
languageOptions: {
  parserOptions: {
    parser: require.resolve('@typescript-eslint/parser')
  }
},
```

lib/rules/prefer-use-template-ref.js
```diff
@@ -44,6 +44,60 @@ function getScriptRefsFromSetupFunction(body) {
...
+        (specifier) => specifier.type === 'ImportSpecifier'
+      )
+  )
+
+  if (vueDestructuredImport) {
+    const importSpecifiers = vueDestructuredImport.specifiers
+    const lastImportSpecifier = importSpecifiers[importSpecifiers.length - 1]
+
+    // @ts-ignore
+    return fixer.insertTextAfter(lastImportSpecifier, `,useTemplateRef`)
```
FloEdelmann, line 82:
```suggestion
    return fixer.insertTextAfter(lastImportSpecifier, `, useTemplateRef`)
```

lib/rules/prefer-use-template-ref.js
```diff
@@ -44,6 +44,60 @@ function getScriptRefsFromSetupFunction(body) {
...
+function addUseTemplateRefImport(context, fixer) {
+  const sourceCode = context.getSourceCode()
+
+  if (!sourceCode) {
+    return
+  }
+
+  const body = sourceCode.ast.body
+
+  const imports = body.filter((node) => node.type === 'ImportDeclaration')
```
FloEdelmann, line 67:
```suggestion
  const body = sourceCode.ast.body
  const imports = body.filter((node) => node.type === 'ImportDeclaration')
```

lib/rules/prefer-use-template-ref.js
```diff
@@ -44,6 +44,60 @@ function getScriptRefsFromSetupFunction(body) {
...
+    const importSpecifiers = vueDestructuredImport.specifiers
+    const lastImportSpecifier = importSpecifiers[importSpecifiers.length - 1]
+
+    // @ts-ignore
+    return fixer.insertTextAfter(lastImportSpecifier, `,useTemplateRef`)
+  }
+
+  const lastImport = imports[imports.length - 1]
+
+  const importStatement = "import {useTemplateRef} from 'vue';"
```
FloEdelmann, line 87: 1. Let's put the `importStatement` variable first because it's used in both following branches.
2. Remove the empty line to make it a little more concise.
3. Add spaces between the braces in the added code, to conform with the common Prettier style. (If users don't like it, they can and have to change it either way.)

```suggestion
  const importStatement = "import { useTemplateRef } from 'vue';"
  const lastImport = imports[imports.length - 1]
```

lib/rules/prefer-use-template-ref.js
```diff
@@ -108,6 +165,32 @@ module.exports = {
...
+                    context,
+                    fixer
+                  )
+
+                  if (missingImportFix) {
+                    fixes.push(missingImportFix)
+                  }
+                }
+
+                return fixes
```
FloEdelmann, line 193: I think it's a little easier to understand (and shorter) without the `fixes` array:

```suggestion
                const replaceFunctionFix = fixer.replaceText(
                  scriptRef.node,
                  `useTemplateRef('${scriptRef.ref}')`
                )

                if (!missingImportChecked) {
                  missingImportChecked = true

                  const missingImportFix = addUseTemplateRefImport(
                    context,
                    fixer
                  )

                  if (missingImportFix) {
                    return [replaceFunctionFix, missingImportFix]
                  }
                }

                return replaceFunctionFix
```

FloEdelmann (changes requested):

tests/lib/rules/prefer-use-template-ref.js
Thomasan1999: Sorry for the late response, I added the test but I am not sure about keeping the generic type parameter. I see as one of the main benefits of `useTemplateRef` function that you don't have to use the type parameters because the Vue Language Server would create it for you. With `ref` you did not have other option but to manually define it. Wouldn't it be better for the users to automatically get rid of it?

FloEdelmann: Often you can indeed get rid of the type parameter with `useTemplateRef`, but sometimes it is needed. I'd like to keep the type parameter intact in this rule. Maybe there can be another rule in the future that suggests removing unnecessary type parameters.

docs/rules/index.md
FloEdelmann, line 1: Please revert the formatting changes in this file (`git restore --source=origin/master docs/rules/index.md`) and rerun `npm run update`.

tests/lib/rules/prefer-use-template-ref.js
Thomasan1999: Makes sense, will update the implementation. 👍

lib/rules/prefer-use-template-ref.js
```diff
@@ -109,6 +172,30 @@ module.exports = {
...
+                const typeArgumentsString = scriptRef.typeArguments
+                  ? `<${scriptRef.typeArguments.join(', ')}>`
+                  : ''
+                const replaceFunctionFix = fixer.replaceText(
+                  scriptRef.node,
+                  `useTemplateRef${typeArgumentsString}('${scriptRef.ref}')`
+                )
+
+                if (!missingImportChecked) {
+                  missingImportChecked = true
```
ota-meshi, line 186: What is this flag for?
Perhaps to avoid autofix conflicts?
If so, ESLint's autofixes avoid conflicts as much as possible by core, so there's no need to handle that in the rules.

ota-meshi (changes requested):
It looks almost good, but I have one comment.

lib/rules/prefer-use-template-ref.js
FloEdelmann: @Thomasan1999 it would be great to get this merged. Could you please have a look into this?

FloEdelmann:
Closed in favor of #2983.

PR #2633, opened 2024-11-30
docs: align example variable name

waynzh:
image


FloEdelmann (approved):
Oh thanks for noticing!

PR #2637, opened 2024-12-03
feat: Add a test case to require-valid-default-prop

aryehb:
The documentation for Vue props, as well as the `require-valid-default-prop` rule, are not clear if setting a default value of `null` to a prop with type `Object` requires a factory function:

```js
props: {
  foo: { type: Object, default: () => null },
},
```

vs.

```js
props: {
  foo: { type: Object, default: null },
},
```

On one hand, `typeof null === 'object'`. On the other hand, the reason for using a factory function is so each instance of the prop has a different object reference, whereas `null` is always the same reference anyway.

This adds a valid test case for setting a default value of `null` to a prop with type `Object` without a factory function. The test passes, so it's evident that `null` does not require a factory function.


FloEdelmann (approved):
Looks good to me, thank you!

ota-meshi (approved):
Thank you!

PR #2639, opened 2024-12-05
Add `vue/no-implicit-coercion` rule

### LINKED ISSUE #2638: Rule Proposal: `vue/no-implicit-coercion`

It would be nice to make the core rule [no-implicit-coercion](https://eslint.org/docs/rules/no-implicit-coercion) work in `<template>` expressions.

[] Enforces code style (layout)
[ ] Other (please specify:)

```vue
<template>
  <div :data-foo="!!foo" />
  <div :data-foo="~foo.indexOf('.')" />
  <div :data-foo="+foo" />
  <div :data-foo="1 * foo" />
  <div :data-foo="'' + foo" />
  <div :data-foo="`` + foo" />
  <!-- if the `disallowTemplateShorthand` option is enabled -->
  <div :data-foo="`${foo}`" />
</template>
```

lozinsky:
Closes #2638


FloEdelmann (approved):
Nice, looks good to me :slightly_smiling_face:

tests/lib/rules/no-implicit-coercion.js
FloEdelmann, line 1: However, I see that some tests are failing. Could you please have a look at them? Possibly you have to skip specific test cases using a `semver` check, see e.g. https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/multiline-ternary.js.

FloEdelmann (changes requested):

tests/lib/rules/no-implicit-coercion.js
lozinsky: Yes, thanks for the reference. Done.

FloEdelmann (approved):
Perfect, thank you :slightly_smiling_face:

ota-meshi (approved):
LGTM! Thank you!

PR #2646, opened 2024-12-13
Fix sidebar categories in rule docs

FloEdelmann:
| Without this PR | With this PR |
|-----------------|--------------|
| grafik | grafik |


waynzh (approved):
Nice catch, LGTM!

UPDATE: The categories not shown in the screenshot (for example, 'Strongly Recommended for Vue.js 2.x') are omitted because their rules are already included in the previous rule sets.

PR #2653, opened 2024-12-17
Add vue/valid-define-options to essential config

waynzh:
Resolve the remaining todo in the code base.

Will added to #2166.

### DIFF
--- a/lib/rules/valid-define-options.js
+++ b/lib/rules/valid-define-options.js
@@ -12,9 +12,7 @@ module.exports = {
     type: 'problem',
     docs: {
       description: 'enforce valid `defineOptions` compiler macro',
-      // TODO Switch in the next major version
-      // categories: ['vue3-essential', 'vue2-essential'],
-      categories: undefined,
+      categories: ['vue3-essential', 'vue2-essential'],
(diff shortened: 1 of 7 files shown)


lib/rules/valid-define-options.js
```diff
@@ -12,9 +12,7 @@ module.exports = {
     type: 'problem',
     docs: {
       description: 'enforce valid `defineOptions` compiler macro',
-      // TODO Switch in the next major version
-      // categories: ['vue3-essential', 'vue2-essential'],
-      categories: undefined,
+      categories: ['vue3-essential', 'vue2-essential'],
```
FloEdelmann, line 15: Is `defineOptions` even supported in Vue 2? If not, then it makes no sense to add this rule to the `vue2-essentials` config.

waynzh: It can only be used in `<script setup>` syntax. Removed ✅

waynzh: Oh, I'm not very familiar with Vue 2.7, I'll check whether `defineOptions` was introduced in Vue 2.7 like `defineProps`.
`vue/valid-define-emits` and `vue/valid-define-props` are in the `essential` category for both Vue 2 and Vue 3.

PR #2663, opened 2025-01-03
Change `anyOf` to `oneOf` in rule schemas to make intent clearer

FloEdelmann:
See https://ajv.js.org/json-schema.html#oneof:

> The data is valid if it matches exactly one JSON Schema from this array.

See https://ajv.js.org/json-schema.html#anyof:

> The data is valid if it is valid according to one or more JSON Schemas in this array.


waynzh (approved):
LGTM! Thanks

PR #2665, opened 2025-01-07
Add `defaultOptions` to all rules

FloEdelmann:
ESLint [v9.17.0](https://github.com/eslint/eslint/releases/tag/v9.15.0) introduced `defaultOptions`:

* Documentation: https://eslint.org/docs/latest/extend/custom-rules#option-defaults
* RFC: https://github.com/eslint/rfcs/tree/main/designs/2023-rule-options-defaults

Even though earlier ESLint versions don't have official support for this property (i.e. `context.options` returns the raw user-provided options rather than merged user options and default options), adding it shouldn't hurt.

In the future, we could write our own helper function to merge user options and default options and thus standardize how rules normalize the user-provided options. Also, we could automate (parts of) the rule docs' "Options" sections.

### DIFF
--- a/eslint-internal-rules/no-invalid-meta-default-options.js
+++ b/eslint-internal-rules/no-invalid-meta-default-options.js
@@ -1,0 +1,156 @@
+/**
+ * @fileoverview Internal rule to enforce valid default options.
+ * @author Flo Edelmann
+ */
+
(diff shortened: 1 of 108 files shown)


FloEdelmann:
I'll try again in another PR.

PR #2668, opened 2025-01-28
Change legacy base config so that vue-eslint-parser is only used for `.vue`

### LINKED ISSUE #2036: Configs to set `.vue` extension via `overrides`
Currently, the docs recommend passing the `.vue` extension in via CLI: https://eslint.vuejs.org/user-guide/#usage

This is cumbersome to declare for each config because Vue SFCs are always in `.vue` files.

When an extension/pattern is set via `overrides`, the extension no longer needs to be manually passed into ESLint via the `--ext` flag.

I would like to recommend that configs use `overrides` so ESLint can automatically detect `.vue` files without the user needing to pass it in:

```js
{
	"overrides": [
		{
			"files": "*.vue",
			// ...
		}
	]
}
```

**Other info**
- Applying the plugin rules specifically to `.vue` extensions should make it easier to handle conflicts like this: https://github.com/vuejs/eslint-plugin-vue/blob/9673a61b08bdadf31b191e0dff30873678c7011b/docs/rules/script-indent.md?plain=1#L67

ota-meshi:
close #2036

### DIFF
--- a/lib/configs/base.js
+++ b/lib/configs/base.js
@@ -4,7 +4,6 @@
  * in order to update its content execute "npm run update"
  */
 module.exports = {
-  parser: require.resolve('vue-eslint-parser'),
   parserOptions: {
     ecmaVersion: 2020,
     sourceType: 'module'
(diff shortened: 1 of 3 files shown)


lib/configs/base.js
FloEdelmann, line 1: I think your changes will be overwritten when running `npm run update`, won't they?

ota-meshi: Thank you for letting me know. You were right 😅 
I fixed the problem.

PR #2669, opened 2025-01-28
Drop support for old versions of ESLint and Node.js

### LINKED ISSUE #1991: Update JS env and remove or update parserOptions.ecmaVersion

- **ESLint version:** 8.24.0
- **eslint-plugin-vue version:** 9.5.1
- **Node version:** 16.14.2

Currently, `eslint-plugin-vue` configures both `env` and `parserOptions.ecmaVersion` in its base config, with different ECMA versions – 2015 (ES6) and 2020: https://github.com/vuejs/eslint-plugin-vue/blob/467e85fdf29139050d92e61bd7ea217a53af87cb/lib/configs/base.js#L9-L14

The `es6` env means that globals that were added after 2015 can't be used without errors, like `BigInt`, `globalThis` or `WeakRef`.

`parser.ecmaVersion: 2020` means that parsing breaks on features like dynamic import or nullish assignments.

`env` automatically sets the version for the parser, but since this plugin explicitly configures the parser version in addition to `env`, that doesn't work anymore when users change the env in their config. So if users specify `es2022: true` as their env, the parser will *not* automatically switch to 2022 syntax [as described in the ESLint docs](https://eslint.org/docs/latest/user-guide/configuring/language-options#specifying-environments) but keep its 2020 config and continue to break on newer syntax.

1. Bump the version for the supported JS environment to `es2022` (latest). Or, if the outdated version config for the parser is intentional for compat, set the environment to `es2020` to align it with the parser version.

2. Either remove the `parserOptions.ecmaVersion` setting so it is automatically inferred from `env`, or set it to `latest` so the parser won't get in the way.

**Additional context**

Here's a repro repo: https://github.com/jonaskuske/vue-eslint-repro

ota-meshi:
Fixes #1991


.circleci/config.yml
```diff
@@ -2,126 +2,13 @@ workflows:
   version: 2
   node-multi-build:
     jobs:
-      - eslint-v6
-      - eslint-v7
-      - ts-eslint-v4
-      - node-v14
-      - node-v16
       - lint
```
FloEdelmann, line 5: If the lint job (which also runs on GitHub Actions) is the only remaining job, do we need the CircleCI config at all anymore?

lib/configs/base.js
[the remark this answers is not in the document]
ota-meshi: Thank you for letting me know. You were right 😅 
I fixed the problem.

.circleci/config.yml
ota-meshi: I stopped using CircleCI 😉

tools/update-lib-configs.js
```diff
@@ -58,7 +58,7 @@ function formatCategory(category) {
 module.exports = {
   parser: require.resolve('vue-eslint-parser'),
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: "latest",
     sourceType: 'module'
   },
   env: {
```
FloEdelmann, line 64: Should we also remove `env` to fix #1991 completely?

tools/update-lib-configs.js
```diff
@@ -58,7 +58,7 @@ function formatCategory(category) {
 module.exports = {
   parser: require.resolve('vue-eslint-parser'),
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: "latest",
```
FloEdelmann, line 61: To match Prettier formatting:

```suggestion
    ecmaVersion: 'latest',
```

tools/update-lib-configs.js
ota-meshi: Thank you for letting me know, I've changed it.

lib/configs/base.js
```diff
@@ -3,16 +3,14 @@
...
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: 'latest',
     sourceType: 'module'
   },
-  env: {
-    browser: true,
-    es6: true
-  },
+  globals: globals.browser,
```
FloEdelmann, line 13: Do we really need to specify globals here? Or should we remove that and rather instruct users what to do in case some global is missing?

FloEdelmann:
@ota-meshi I rebased `v10` onto master and then I rebased `drop-old` onto `v10` so everything is clean again. So please run `git pull --rebase` or `git reset --hard origin/drop-old` locally.

lib/configs/base.js
ota-meshi: I think it would make sense to remove `globals.browser` from the config provided by the plugin. In SSR those global objects don't exist.
We'd also need to change the Flat confign, so that might be better done in a separate PR. What do you think?

FloEdelmann: Yeah, let's do it separately :+1:

FloEdelmann (approved):

lib/configs/base.js
FloEdelmann: I've added a bullet point in #2166.

PR #2670, opened 2025-01-28
Change `vue-eslint-parser` to peer dependency

ota-meshi:
This PR changes vue-eslint-parser to peer dependency.

I make it easy to change the version used by making vue-eslint-parser a peer dependency.
Also, recent npm installs peer dependencies automatically by default, so I don't think it will be inconvenient for users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "node": "^14.17.0 || >=16.0.0"
   },
   "peerDependencies": {
-    "eslint": "^6.2.0 || ^7.0.0 || ^8.0.0 || ^9.0.0"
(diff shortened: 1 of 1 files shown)


FloEdelmann (approved):
Good idea!

mbrodala:
For completeness: this does not work with Yarn which does not install peer dependencies by default but warns that they are missing:

```
warning " > eslint-plugin-vue@10.0.0" has unmet peer dependency "vue-eslint-parser@^10.0.0".
```

See https://github.com/yarnpkg/yarn/issues/1503

PR #2673, opened 2025-01-30
Revert "Remove globals from config"

ota-meshi:
Reverts vuejs/eslint-plugin-vue#2672


ota-meshi:
I made a mistake in merging so I had to revert 😓

PR #2675, opened 2025-01-30
Deprecate Vue 2 only rules

FloEdelmann:
These rules are currently in configs for Vue 2 (in all of `essential`, `strongly-recommended` and `recommended`), but not in configs for Vue 3:

- [vue/no-custom-modifiers-on-v-model](https://eslint.vuejs.org/rules/no-custom-modifiers-on-v-model.html)
  → could be useful as an uncategorized rule for Vue 3 to disallow [custom modifiers](https://vuejs.org/guide/components/v-model.html#handling-v-model-modifiers), so let's not deprecate it
- [vue/no-multiple-template-root](https://eslint.vuejs.org/rules/no-multiple-template-root.html)
  → could be useful as an uncategorized rule for Vue 3 to prevent [attribute inheritance problems](https://vuejs.org/guide/components/attrs.html#attribute-inheritance-on-multiple-root-nodes), so let's not deprecate it
- [vue/no-v-for-template-key](https://eslint.vuejs.org/rules/no-v-for-template-key.html)
  → makes no sense for Vue 3, so deprecate it (see [Vue 3 Migration Guide: `key` Attribute](https://v3-migration.vuejs.org/breaking-changes/key-attribute.html))
- [vue/no-v-model-argument](https://eslint.vuejs.org/rules/no-v-model-argument.html)
  → makes no sense for Vue 3, so deprecate it (see [Vue 3 Migration Guide: `v-model`](https://v3-migration.vuejs.org/breaking-changes/v-model.html))
- [vue/valid-model-definition](https://eslint.vuejs.org/rules/valid-model-definition.html)
  → makes no sense for Vue 3, so deprecate it (see [Vue 3 Migration Guide: `v-model`](https://v3-migration.vuejs.org/breaking-changes/v-model.html))
- [vue/valid-v-bind-sync](https://eslint.vuejs.org/rules/valid-v-bind-sync.html)
  → makes no sense for Vue 3, so deprecate it (see [Vue 3 Migration Guide: `v-model`](https://v3-migration.vuejs.org/breaking-changes/v-model.html))

I didn't find any [uncategorized](https://eslint.vuejs.org/rules/#uncategorized) rules that would only be useful for Vue 2.

> [!NOTE]
> These rules will continue to work fine in the upcoming major version `v10`, but no longer receive much maintenance work, and we'll remove them in `v11`.

Thus, deprecating them is not technically a breaking change, but I think it will receive more attention if we do it in a major version.

### DIFF
--- a/docs/rules/index.md
+++ b/docs/rules/index.md
@@ -110,7 +110,7 @@ Rules in this category are enabled for all presets provided by eslint-plugin-vue
 | [vue/valid-define-emits] | enforce valid `defineEmits` compiler macro |  | :three::two::warning: |
 | [vue/valid-define-options] | enforce valid `defineOptions` compiler macro |  | :three::warning: |
 | [vue/valid-define-props] | enforce valid `defineProps` compiler macro |  | :three::two::warning: |
-| [vue/valid-model-definition] | require valid keys in model option |  | :two::warning: |
+| [vue/valid-model-definition] | require valid keys in model option | :no_entry_sign: | :two::warning: |
 | [vue/valid-next-tick] | enforce valid `nextTick` function calls | :wrench::bulb: | :three::two::warning: |
 | [vue/valid-template-root] | enforce valid template root |  | :three::two::warning: |
 | [vue/valid-v-bind-sync] | enforce valid `.sync` modifier on `v-bind` directives |  | :two::warning: |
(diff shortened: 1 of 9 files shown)


waynzh (approved):
LGTM!

PR #2679, opened 2025-02-08
extend prop-name-casing to ignore specific names

westberliner:
Hi,

is this something you would be interested in - at least I did 😄  ?
If yes I would extend the test for it as well. If not then close it.

Thx
Patrick

### DIFF
--- a/lib/rules/prop-name-casing.js
+++ b/lib/rules/prop-name-casing.js
@@ -15,6 +15,7 @@ const allowedCaseOptions = ['camelCase', 'snake_case']
 /** @param {RuleContext} context */
 function create(context) {
   const options = context.options[0]
+  const ignores = context.options[1]?.ignores || []
   const caseType = allowedCaseOptions.includes(options) ? options : 'camelCase'
   const checker = casing.getChecker(caseType)
 
@@ -27,7 +28,7 @@ function create(context) {
       if (propName == null) {
         continue
       }
-      if (!checker(propName)) {
+      if (!checker(propName) || ignores.includes(propName)) {
         context.report({
           node: item.node,
           messageId: 'invalidCase',
@@ -64,6 +65,12 @@ module.exports = {
     schema: [
       {
         enum: allowedCaseOptions
+      },
+      {
+        type: 'object',
+        properties: {
+          ignores: { type: 'array' }
+        }
       }
     ],
     messages: {


waynzh:
That makes sense to me. You can continue to add test cases and documentation when you have time. 
For reference, you might find similar ones in `vue/custom-event-name-casing` if you need it.

lib/rules/prop-name-casing.js
```diff
@@ -64,6 +65,12 @@ module.exports = {
     schema: [
       {
         enum: allowedCaseOptions
+      },
+      {
+        type: 'object',
+        properties: {
+          ignores: { type: 'array' }
+        }
       }
```
FloEdelmann, line 74: Please make the schema more strict:

```suggestion
      {
        type: 'object',
        properties: {
          ignores: {
            type: 'array',
            items: { type: 'string' },
            uniqueItems: true
          }
        },
        additionalProperties: false
      }
```

lib/rules/prop-name-casing.js
```diff
@@ -64,6 +65,12 @@ module.exports = {
     schema: [
       {
         enum: allowedCaseOptions
+      },
+      {
+        type: 'object',
+        properties: {
+          ignores: { type: 'array' }
```
FloEdelmann, line 72: Please rename the option from `ignores` to `ignoreProps`, so it's clear that you can't specify component names whose props will be ignored.

lib/rules/prop-name-casing.js
```diff
@@ -15,6 +15,7 @@ const allowedCaseOptions = ['camelCase', 'snake_case']
 /** @param {RuleContext} context */
 function create(context) {
   const options = context.options[0]
+  const ignores = context.options[1]?.ignores || []
```
FloEdelmann, line 18: Please allow ignore patterns as well, similar to the `ignores` option in other rules:

https://github.com/vuejs/eslint-plugin-vue/blob/b1fa3b9a444131a99b4fe4d7e025734f424c5dcd/lib/rules/custom-event-name-casing.js#L114

FloEdelmann (changes requested):
Thanks, this is a good addition to the rule! I have a few comments though. Like @waynzh already said, please also add tests and documentation.

westberliner:
Fixed everything as we talked about.

waynzh (approved):
LGTM!

docs/rules/prop-name-casing.md
```diff
@@ -39,12 +39,18 @@ export default {
...
+    {
+      "ignoreProps": []
+    }
+  ]
 }
 ```
 
 - `"camelCase"` (default) ... Enforce property names in `props` to camel case.
 - `"snake_case"` ... Enforce property names in `props` to snake case.
+- `ignoreProps` (`string[]`) ... An array of prop names (or patterns) that should be exempt from the case rule check. Use this option for prop names that are intentionally non-compliant, have special naming requirements for Vue components, or are defined by Vue libraries. You can define these entries as literal strings or as regular expressions (written as strings, e.g., `"/^name/"`).
```
waynzh, line 53: One tiny docs improvement, the explanation for the `ignoreProps` could perhaps be more concise. 
Something like: 

```suggestion
- `ignoreProps` (`string[]`) ... An array of prop names (or patterns) that don't need to follow the specified casing.
```

docs/rules/prop-name-casing.md
```diff
@@ -67,6 +73,31 @@ export default {
...
+<eslint-code-block :rules="{'vue/prop-name-casing': ['error', 'camelCase', {
+ignoreProps: ['foo-bar', '/^_[a-z]+/u'] }]}">
+
+```vue
+<script>
+export default {
+  props: {
+    /* ✓ GOOD */
+    greetingText: String,
+    foo-bar: String,
```
FloEdelmann, line 87:
```suggestion
    'foo-bar': String,
```

FloEdelmann (approved):
Apart from one syntax error in the docs' example code, this looks good to me :slightly_smiling_face:

docs/rules/prop-name-casing.md
westberliner: 🤦 of course, sry.

PR #2680, opened 2025-02-10
feat(no-ref-as-operand): use `ref.value` to replace `ref` when emit 

### LINKED ISSUE #2601: Rule proposal: `no-ref-emit`
Hello, first of all thanks for everything provided so far 😃 

Reports when trying to emit a ref itself, instead of emitting the value of the ref.

- [x] Warns about a potential error (problem)
- [ ] Other (please specify:)

***Example 1***
```vue
<template>
    <button @click="increment">Increment</button>
</template>

<script setup>
import { ref } from vue

const emit = defineEmits(['incremented'])

const counter = ref(0)

function increment() {
    counter.value++
    /* Probably an error, trying to emit a ref instead of it's value. */
    emit('incremented', counter) /* ✗ BAD */
    emit('incremented', counter.value) /* ✓ GOOD */
}
</script>
```

***Example 2***
```vue
<template>
    <button @click="counterA++">Increment A</button>
    <button @click="counterB++">Increment B</button>
</template>

<script setup>
import { ref, computed, watch } from vue

const emit = defineEmits(['updated'])

const counterA = ref(0)
const counterB = ref(0)

const result = computed(() => ({
    a: counterA.value,
    b: counterB.value,
}))

watch(result, () => {
    /*
    Probably an error, trying to emit a ref instead of it's value.
    Additionally, the parent might try to mutate the emitted value in this case where
    it should be a plain object but is instead a computed ref.
    Ideally we would use the argument provided by the watcher in this specific case though.
    */
    emit('updated', result) /* ✗ BAD */
    emit('updated', result.value) /* ✓ GOOD */
})
</script>
```

**Additional context**

I couldn't think of any usecase that's not against Vue recommendations where emitting the ref itself would be justified, but
if there is any feel free to discard this rule proposal.

Maybe this functionality can be added to [`no-ref-as-operand`](https://eslint.vuejs.org/rules/no-ref-as-operand) instead of creating a new rule as it is quite similar ?

chouchouji:
# Issue

close #2601

### DIFF
--- a/lib/rules/no-ref-as-operand.js
+++ b/lib/rules/no-ref-as-operand.js
@@ -125,11 +125,7 @@ module.exports = {
           ) {
             // verify setup(props,{emit}) {emit()}
             node.arguments
-              .filter(
-                (node) =>
-                  node.type === 'Identifier' &&
-                  isRefInit(refReferences.get(node))
-              )
+              .filter((node) => node.type === 'Identifier')
               .forEach((node) => {
                 reportIfRefWrapped(node)
               })
@@ -143,11 +139,7 @@ module.exports = {
             ) {
               // verify setup(props,context) {context.emit()}
               emit.member.parent.arguments
-                .filter(
-                  (node) =>
-                    node.type === 'Identifier' &&
-                    isRefInit(refReferences.get(node))
-                )
+                .filter((node) => node.type === 'Identifier')
                 .forEach((node) => {
                   reportIfRefWrapped(node)
                 })
--- a/lib/rules/no-ref-as-operand.js
+++ b/lib/rules/no-ref-as-operand.js
@@ -4,6 +4,7 @@
  */
 'use strict'
 
+const { findVariable } = require('@eslint-community/eslint-utils')
 const { extractRefObjectReferences } = require('../utils/ref-object-references')
 const utils = require('../utils')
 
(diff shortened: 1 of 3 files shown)


chouchouji:
@FloEdelmann Please review this pr when you have time. Thanks!

lib/rules/no-ref-as-operand.js
```diff
@@ -64,90 +100,236 @@ module.exports = {
...
+            name == null ||
+            // WritableComputedRef
+            name === 'effect'
+          ) {
+            return
+          }
+          reportIfRefWrapped(node)
+        }
+      },
+      utils.compositingVisitors(
```
FloEdelmann, line 238: Is this nesting of `utils.compositingVisitors` necessary? It can take multiple arguments: 

https://github.com/vuejs/eslint-plugin-vue/blob/827ab4b4ff3b86ce52da2dac837e9a0254aa2c6a/lib/utils/index.js#L2353-L2359

lib/rules/no-ref-as-operand.js
```diff
@@ -64,90 +100,236 @@ module.exports = {
...
-        if (
-          !data ||
-          !data.variableDeclaration ||
-          data.variableDeclaration.kind !== 'const'
-        ) {
-          return
+
+        // verify setup context
+        const setupContext = setupContexts.get(info ? info.node : programNode)
+        if (setupContext) {
```
FloEdelmann, line 120: Please return early here if `setupContext` is falsy.

lib/rules/no-ref-as-operand.js
```diff
@@ -64,90 +100,236 @@ module.exports = {
...
+            ) {
+              // verify setup(props,context) {context.emit()}
+              const nodes = node.arguments.filter(
+                (node) => node.type === 'Identifier'
+              )
+              for (const node of nodes) {
+                reportIfRefWrapped(node)
+              }
+            }
+          }
```
FloEdelmann, line 149: There is some duplicated code here. Can you reduce that by extracting some into a separate function please?

lib/rules/no-ref-as-operand.js
```diff
@@ -64,90 +100,236 @@ module.exports = {
...
+              utils.getScope(context, contextParam),
+              contextParam
+            )
+            if (!variable) {
+              return
+            }
+            for (const reference of variable.references) {
+              contextReferenceIds.add(reference.identifier)
+            }
+          }
```
FloEdelmann, line 322: Again, code duplication should be reduced if possible.

FloEdelmann (changes requested):
Thanks for your contribution! I now had a deeper look and apart from a few code maintenance/readbility improvements, it looks good to me :slightly_smiling_face:

chouchouji:
> Thanks for your contribution! I now had a deeper look and apart from a few code maintenance/readbility improvements, it looks good to me 🙂

Thanks for your reply!  I had updated code.

FloEdelmann (approved):
Thanks for implementing the feedback so quickly! LGTM now.

chouchouji:
> Thanks for implementing the feedback so quickly! LGTM now.

Thanks!

waynzh (approved):
LGTM! Thanks!

PR #2684, opened 2025-02-13
feat: add `vue/no-import-compiler-macros` rule

### LINKED ISSUE #2437: disallow importing macros: defineProps etc
It it worth tracking approved but closed #1906 as a new issue here?

Example playground that includes [vue/prefer-import-from-vue rule](https://eslint.vuejs.org/rules/prefer-import-from-vue.html) enabled: [link](https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJxNzsEKgzAMxvFXCTltsOG9dGWPsAfIRTRKQdPSRmGUvvsq7uAx/D/4peAQRkaDVnmNS6/sSADs6HdXCkwhQK22O04S211GNg/JR4XMukVYeplfhJoJW/NrDEmhwMiTF/6kEDNUmFJYgXDfmJDk0mw50IYZyJq8zCTV3e6HeCoOHxj7lDmhKU1pD7/1G/msT86LF+3+i1p/8rNMrg==)

This playground code triggers a compiler message about the macro import:

```
[@vue/compiler-sfc] `defineProps` is a compiler macro and no longer needs to be imported.
```

> **Please describe what the rule should do:** It should warn if defineProps, defineEmits and similar are imported from vue (autofixable)
> 
> **What category should the rule belong to?**
> 
>   [x] Enforces code style (layout)
>   [ ] Warns about a potential error (problem)
>   [ ] Suggests an alternate way of doing something (suggestion)
>   [ ] Other (please specify:)
> 
> **Provide 2-3 code examples that this rule should warn about:**
> 
> bad:
> 
> ```
> <script setup>
> import { defineProps } from '@vue/runtime-dom';
> 
> defineProps<{}>();
> </script>
> ```
> 
> good:
> 
> ```
> <script setup>
> defineProps<{}>();
> </script>
> ```
> 
> good:
> 
> ```
> <script setup>
> import { defineProps } from 'my-lib';
> 
> defineProps('whatever');
> ```
> 
> **Additional context** Useful when `define*` are added to eslint globals, there's no need to import them manually. If project was using imports and then dev switched to globals then autofix would just automatically remove unnecessary imports

waynzh:
resolve #2437

### DIFF
--- a/docs/rules/no-import-compiler-macros.md
+++ b/docs/rules/no-import-compiler-macros.md
@@ -42,6 +42,12 @@ import { ref } from 'vue'
 
 Nothing.
 
+## :books: Further Reading
+
+- [defineProps() & defineEmits()]
+
+[defineProps() & defineEmits()]: https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits
+
 ## :mag: Implementation
 
 - [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-import-compiler-macros.js)
--- a/docs/rules/no-import-compiler-macros.md
+++ b/docs/rules/no-import-compiler-macros.md
@@ -1,0 +1,48 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-import-compiler-macros
+description: disallow importing vue compiler macros
+---
+
+# vue/no-import-compiler-macros
+
+> disallow importing vue compiler macros
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> _**This rule has not been released yet.**_ </badge>
+- :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fix-problems) can automatically fix some of the problems reported by this rule.
+
+## :book: Rule Details
+
+This rule disallow importing vue compiler macros.
+
+<eslint-code-block fix :rules="{'vue/no-import-compiler-macros': ['error']}">
+
+```vue
+<script setup>
+/* ✗ BAD */
+import { defineProps, withDefaults } from 'vue'
+</script>
+```
+
+</eslint-code-block>
+
+<eslint-code-block fix :rules="{'vue/no-import-compiler-macros': ['error']}">
+
+```vue
+<script setup>
+/* ✓ GOOD */
+import { ref } from 'vue'
+</script>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :mag: Implementation
+
+- [Rule source](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-import-compiler-macros.js)
+- [Test source](https://github.com/vuejs/eslint-plugin-vue/blob/master/tests/lib/rules/no-import-compiler-macros.js)
--- a/docs/rules/no-import-compiler-macros.md
+++ b/docs/rules/no-import-compiler-macros.md
@@ -2,12 +2,12 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/no-import-compiler-macros
-description: disallow importing vue compiler macros
+description: disallow importing Vue compiler macros
 ---
 
 # vue/no-import-compiler-macros
 
-> disallow importing vue compiler macros
+> disallow importing Vue compiler macros
 
 - :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> _**This rule has not been released yet.**_ </badge>
 - :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fix-problems) can automatically fix some of the problems reported by this rule.
--- a/lib/rules/no-import-compiler-macros.js
+++ b/lib/rules/no-import-compiler-macros.js
@@ -23,13 +23,6 @@ function isComma(node) {
   return node.type === 'Punctuator' && node.value === ','
 }
 
-/**
- * @param {Token} node
- */
-function isLeftCurlyBrace(node) {
-  return node.type === 'Punctuator' && node.value === '{'
-}
-
 module.exports = {
   meta: {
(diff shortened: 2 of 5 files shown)


lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+function isLeftCurlyBrace(node) {
+  return node.type === 'Punctuator' && node.value === '{'
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow importing vue compiler macros',
```
FloEdelmann, line 34:
```suggestion
      description: 'disallow importing Vue compiler macros',
```

lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+ * @author Wayne Zhang
+ * See LICENSE file in root directory for full license.
+ */
+'use strict'
+
+const COMPILER_MACROS = new Set([
+  'defineProps',
+  'defineEmits',
+  'defineExpose',
+  'withDefaults'
```
FloEdelmann, line 11: Please also add `defineModel`, `defineOptions` and `defineSlots` (and tests for them):

- https://vuejs.org/api/sfc-script-setup.html#definemodel
- https://vuejs.org/api/sfc-script-setup.html#defineoptions
- https://vuejs.org/api/sfc-script-setup.html#defineslots

docs/rules/no-import-compiler-macros.md
```diff
@@ -0,0 +1,48 @@
...
+</script>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :mag: Implementation
```
FloEdelmann, line 45: Can you please add a link to https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits in the `📚 Further Reading` section?

lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+    docs: {
+      description: 'disallow importing vue compiler macros',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-import-compiler-macros.html'
+    },
+    fixable: 'code',
+    schema: [],
+    messages: {
+      noImportCompilerMacros:
+        "'{{name}}' is a compiler macro and no longer needs to be imported from '{{source}}'."
```
FloEdelmann, line 42:
```suggestion
        "'{{name}}' is a compiler macro and doesn't need to be imported."
```

I think "from {{ source }}" sounds like it should be imported from somewhere else. And "no longer" sounds like it had to be imported at some time, but I think it was never meant to be imported, was it?

lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+   * @returns {RuleListener}
+   */
+  create(context) {
+    const sourceCode = context.getSourceCode()
+
+    return {
+      ImportDeclaration(node) {
+        if (node.specifiers.length === 0) return
+
+        if (VUE_MODULES.has(node.source.value)) {
```
FloEdelmann, line 56: Better return early here, too, to remove one level of indentation and make the rest of the block easier to read:

```ts
        if (node.specifiers.length === 0 || !VUE_MODULES.has(node.source.value)) {
          return
        }
```

lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+          for (const specifier of node.specifiers) {
+            if (
+              specifier.type === 'ImportSpecifier' &&
+              COMPILER_MACROS.has(specifier.imported.name)
+            ) {
```
FloEdelmann, line 61: Again, please return early:

```ts
            if (
              specifier.type !== 'ImportSpecifier' ||
              !COMPILER_MACROS.has(specifier.imported.name)
            ) {
              continue
            }
```

lib/rules/no-import-compiler-macros.js
```diff
@@ -0,0 +1,94 @@
...
+                  const isFirstSpecifier = isLeftCurlyBrace(tokenBefore)
+
+                  const codeStart = hasCommaAfter
+                    ? tokenBefore.range[1]
+                    : isFirstSpecifier
+                      ? specifier.range[0]
+                      : tokenBefore.range[0]
+                  const codeEnd = hasCommaAfter
+                    ? tokenAfter.range[1]
+                    : specifier.range[1]
```
FloEdelmann, line 83: Nice that you implemented a quite elaborate auto-fix!

Can you please add test cases that these cases are handled correctly?

- `import { ref, defineProps } from '@vue/runtime-core'`
- `import { defineProps as definePropsFoo } from '@vue/runtime-core'` (or maybe we ignore this case)
- `import { ref as refFoo, defineProps, type computed } from '@vue/runtime-core'`

And maybe this logic would be easier to understand?

```ts
if (isOnlySpecifier) {
  removeWholeImportStatement()
} else if (isLastSpecifier) {
  removeSpecifierWithPrecedingComma()
} else {
  removeSpecifierWithSubsequentComma()
}

FloEdelmann (changes requested):
Thanks for adding this rule! :slightly_smiling_face:

lib/rules/no-import-compiler-macros.js
waynzh: It never needs to be imported. The "no longer" stuff was directly copied from Vue's console warning..

lib/rules/no-import-compiler-macros.js
waynzh: Oh right, got it

lib/rules/no-import-compiler-macros.js
waynzh: Yeah, that's more intuitive! Updated✅

FloEdelmann (approved):

ota-meshi (approved):
LGTM! Thank you!

PR #2685, opened 2025-02-14
feat(no-multiple-template-root): support disallowComments property

### LINKED ISSUE #2316: `vue/no-multiple-template-root` false negative with comment
**Checklist**

- **ESLint version:**  8.50.0
- **eslint-plugin-vue version:** 9.15.1
- **Node version:** 18.13.0
- **Operating System:** macOS

```json5
module.exports = {
  extends: ['../../../.eslintrc.js', 'plugin:storybook/recommended'],
  root: true,
  env: {
    node: true,
    browser: true
  },
  overrides: [
    {
      files: ['*.ts', '*.tsx', '*.vue'],
      parser: 'vue-eslint-parser',
      parserOptions: {
        project: './tsconfig.json',
        extraFileExtensions: ['.vue']
      },
      globals: {
        definePageMeta: 'readonly',
        $fetch: 'readonly',
        Stripe: 'readonly'
      },
      rules: {
        /**
         * This has unexpected behavior when used with other TS settings which produce
         * unexpected undefines, such as a map.get after testing for map.has
         */
        '@typescript-eslint/no-non-null-assertion': 'off'
      }
    },
    {
      files: ['**/pages/**/*.vue'],
      rules: {
        /**
         * Even though this is allowed in Vue3,
         * both Nuxt and Knockout need a single
         * root element.
         */
        'vue/no-multiple-template-root': 'error'
      }
    }
  ]
}

```

**What did you do?**	

```vue
<template>
  <!-- this is a comment -->
   <div></div>
</template>
```

**_IMPORTANT CONTEXT:  the Vue compiler has `comments: true` turned on (as a necessity). What this means is that the root template in this mode actually has multiple roots, including a blank text node, a comment node, and the div. I want to prevent developers from making this mistake, so I need a way to flag this as an error._**

I expected an ESLint error when there are multiple possible roots.

**What actually happened?**

There is no error, hence no output.

https://stackblitz.com/edit/vitejs-vite-mhdbqk?file=src%2FApp.vue

chouchouji:
# Issue

close #2316

### DIFF
--- a/lib/rules/no-multiple-template-root.js
+++ b/lib/rules/no-multiple-template-root.js
@@ -13,12 +13,12 @@ const utils = require('../utils')
  * @returns {(HTMLComment | HTMLBogusComment | Comment)[]}
  */
 function getReportComments(comments, elementRanges) {
-  return comments.filter(  
-    (comment) =>  
-      !elementRanges.some(  
-        (range) => range[0] <= comment.range[0] && comment.range[1] <= range[1]  
-      )  
-  )  
+  return comments.filter(
+    (comment) =>
+      !elementRanges.some(
+        (range) => range[0] <= comment.range[0] && comment.range[1] <= range[1]
+      )
+  )
 }
 
 module.exports = {
--- a/lib/rules/no-multiple-template-root.js
+++ b/lib/rules/no-multiple-template-root.js
@@ -50,30 +50,55 @@ module.exports = {
           return
         }
 
-        const commentRangesMap = new Map()
         const comments = element.comments
         if (disallowComments && comments.length > 0) {
-          for (const comment of comments) {
-            const [start, end] = comment.range
-            commentRangesMap.set(`${start}-${end}`, comment)
-          }
+          const commentRanges = comments.map((comment) => comment.range)
+          const elementRanges = element.children
+            .filter((child) => child.type === 'VElement')
+            .map((child) => child.range)
 
-          for (const child of element.children) {
-            if (child.type === 'VElement') {
-              for (const range of commentRangesMap.keys()) {
-                const ranges = range.split('-')
-                if (ranges[0] > child.range[0] && ranges[1] < child.range[1]) {
-                  commentRangesMap.delete(range)
-                }
-              }
+          let commentIndex = 0
+          let elementIndex = 0
+
+          const needReportComments = elementRanges.length === 0 ? comments : []
+          while (
+            commentIndex < commentRanges.length &&
+            elementRanges.length > 0
+          ) {
+            const [commentStart, commentEnd] = commentRanges[commentIndex]
+            const [elementStart, elementEnd] = elementRanges[elementIndex]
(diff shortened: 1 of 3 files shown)


chouchouji:
@FloEdelmann I am confused about how to determine if it is the comment of the root node. So I choose to filter invalid comments by node ranges. If you have better idea, please tell me. Thanks!

chouchouji:
@ota-meshi  I updated docs and test cases. Please review it if you have time. Thanks.

lib/rules/no-multiple-template-root.js
```diff
@@ -37,6 +102,20 @@ module.exports = {
...
+          const needReportComments = getReportComments(comments, element)
+          if (needReportComments.length > 0) {
+            for (const comment of needReportComments) {
+              context.report({
+                node: comment,
+                loc: comment.loc,
+                messageId: 'commentRoot'
+              })
+            }
+          }
```
ota-meshi, line 116: I don't think the `if` is necessary.

```suggestion
          for (const comment of getReportComments(comments, element)) {
            context.report({
              node: comment,
              loc: comment.loc,
              messageId: 'commentRoot'
            })
          }
```

tests/lib/rules/no-multiple-template-root.js
```diff
@@ -104,6 +164,112 @@ ruleTester.run('no-multiple-template-root', rule, {
...
+        <!-- When you have a comment in the root of your template in vue 3, 
+        using $el will point to the first text comment instead of the actual DOM element.   -->
+        <div>
+          12333
+          <!-- comments -->
+        </div>
+      </template>
+      `,
+      options: [{ disallowComments: true }],
+      errors: ['The template root disallows comments.']
```
ota-meshi, line 272: Could you please add the following test case to check if the report is correct?

```vue
      <template>
        <div />
        <!-- comments -->
        <div />
      </template>
```

lib/rules/no-multiple-template-root.js
```diff
@@ -6,6 +6,58 @@
...
+      needReportComments.push(comments[commentIndex])
+      commentIndex += 1
+    }
+
+    if (elementIndex < elementRanges.length - 1 && commentStart > elementEnd) {
+      elementIndex += 1
+    }
+  }
+
+  return needReportComments
```
ota-meshi, line 58: I think we can write it like this:

```suggestion
  return comments.filter(
    (comment) =>
      !elementRanges.some(
        (range) => range[0] <= comment.range[0] && comment.range[1] <= range[1]
      )
  )
```

ota-meshi (changes requested):
Thank you for this PR!

chouchouji:
@FloEdelmann Can you review this pr? I have encountered this problem and want to solve it. Thanks!

ota-meshi (approved):
Sorry for the late reply.
LGTM! Thanks!

PR #2694, opened 2025-03-03
feat(prefer-true-attribute-shorthand): add `except` option

### LINKED ISSUE #2687: `vue/prefer-true-attribute-shorthand`: add `except` option
[`vue/prefer-true-attribute-shorthand`](https://eslint.vuejs.org/rules/prefer-true-attribute-shorthand.html)

Depends on configuration. No changes by default.

New option `except` that allows specifying attribute names that should be specified in longhand form if the rule is set to `always` or in shorthand form if the rule is set to `never`.

```json
{
  "vue/prefer-true-attribute-shorthand": ["error", "always", {
    "except": ["value", "/^foo-/"]
  }]
}
```

```vue
<template>
  <MyComponent is-shorthand :value="true" foo-bar />
</template>
```

It doesn't warn for `is-shorthand` and `foo-bar`, but it warns for `value`.

It still shouldn't warn for `is-shorthand`, but it shouldn't warn for `value` and should warn for `foo-bar`.

**Additional context**
The motivating example uses PrimeVue's [`RadioButton`](https://primevue.org/radiobutton/) component:

```vue
<template>
  <RadioButton v-model="likesPizza" name="pizza" :value="true" />
  <RadioButton v-model="likesPizza" name="pizza" :value="false" />
</template>
```

Using the shorthand syntax for the first component's `value` prop would be less readable.

waynzh:
resolve #2687

### DIFF
--- a/docs/rules/prefer-true-attribute-shorthand.md
+++ b/docs/rules/prefer-true-attribute-shorthand.md
@@ -81,12 +81,18 @@ Default options is `"always"`.
 
 ```json
 {
-  "vue/prefer-true-attribute-shorthand": ["error", "always" | "never"]
+  "vue/prefer-true-attribute-shorthand": ["error",
+    "always" | "never",
+    {
+      except: []
+    }
+  ]
 }
 ```
 
 - `"always"` (default) ... requires shorthand form.
 - `"never"` ... requires long form.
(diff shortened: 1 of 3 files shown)


ota-meshi (approved):
LGTM! Thank you!

ota-meshi:
I want to work on the v10 release today so I will merge this PR.

PR #2698, opened 2025-03-05
Add vue-eslint-parser to Yarn install

mbrodala:
Since version 10.0.0 this has become a peer dependency which is not installed by Yarn.

See https://github.com/vuejs/eslint-plugin-vue/pull/2670


FloEdelmann (approved):
Thanks!

Issue #2690, opened 2025-02-21
[Enhancement] Add fixer for vue/no-export-in-script-setup to support autofix


chouchouji:
**Tell us about your environment**

- **ESLint version:**  **8.44.0**
- **eslint-plugin-vue version:**  **9.17.0**
- **Vue version:**  **3.3.4**
- **Node version:**  **18.17.0**

**The problem you want to solve.**

Can we add fixer to autofix this code?
![Image](https://github.com/user-attachments/assets/12b0161e-b714-49b4-9235-7b02163315bf)

**Your take on the correct solution to problem.**

**Additional context**

FloEdelmann:
In https://github.com/vuejs/eslint-plugin-vue/issues/2690, the rule was already improved to only report the `export` keyword for statements that might be relevant for local code, e.g.:

```ts
export const foo = 'bar'
console.log(foo)
```

As a potential follow-up improvement, a PR would be welcome that adds [suggestions](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) that remove the reported code (i.e. the whole statement for `export * from 'foo'` and only the `export` keyword for `export const foo = 'bar'`). For more context, see https://github.com/vuejs/eslint-plugin-vue/pull/2701#discussion_r1981066376.

Autofixes are considered too dangerous for this rule to break other code.

PR #2701, opened 2025-03-05
fix(no-export-in-script-setup): better report location

chouchouji:
support autofix for no-export-in-script-setup rule. 

relates to #2690

### DIFF
--- a/tests/lib/rules/no-export-in-script-setup.js
+++ b/tests/lib/rules/no-export-in-script-setup.js
@@ -94,6 +94,13 @@ ruleTester.run('no-export-in-script-setup', rule, {
       export class A {}
       </script>
       `,
+      output: `
+      <script setup>
+      
+      
+      
+      </script>
+      `,
       errors: [
         {
           message: '`<script setup>` cannot contain ES module exports.',
@@ -121,6 +128,16 @@ ruleTester.run('no-export-in-script-setup', rule, {
       export class A {}
       </script>
       `,
+      output: `
+      <script>
+      let foo;
+      </script>
+      <script setup>
+      
+      
+      
+      </script>
+      `,
       errors: [
         {
           message: '`<script setup>` cannot contain ES module exports.',
@@ -143,6 +160,13 @@ ruleTester.run('no-export-in-script-setup', rule, {
       export const Foo = {}
       export enum Bar {}
       export {}
+      </script>
+      `,
+      output: `
+      <script setup lang="ts">
+      
+      
+      
       </script>
       `,
       languageOptions: {
--- a/tests/lib/rules/no-export-in-script-setup.js
+++ b/tests/lib/rules/no-export-in-script-setup.js
@@ -92,27 +92,62 @@ ruleTester.run('no-export-in-script-setup', rule, {
       export * from 'foo'
       export default {}
       export class A {}
-      </script>
-      `,
-      output: `
-      <script setup>
-      
-      
-      
+      export const test = '123'
+      export function foo() {}
+      const a = 1
+      export { a }
+      export { fao } from 'bar'
       </script>
       `,
       errors: [
         {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-export-in-script-setup.js
```diff
@@ -92,6 +92,13 @@ ruleTester.run('no-export-in-script-setup', rule, {
...
       export default {}
       export class A {}
+      </script>
+      `,
+      output: `
+      <script setup>
+      
+      
+      
       </script>
```
FloEdelmann, line 102: That doesn't look right to me.

1. The first statement (`export * from 'foo'`) should be removed, like you already do. Maybe a suggestion would be better than an autofix though?
2. The second statement (`export default {}`) should have a suggestion to be removed (maybe the code is needed in this file, so an autofix should not remove it immediately).
3. For the third statement (`export class A {}`), only the `export` keyword should be removed, since maybe the class is used locally as well.

tests/lib/rules/no-export-in-script-setup.js
```diff
@@ -143,6 +160,13 @@ ruleTester.run('no-export-in-script-setup', rule, {
...
       export enum Bar {}
       export {}
+      </script>
+      `,
+      output: `
+      <script setup lang="ts">
+      
+      
+      
       </script>
```
FloEdelmann, line 170: Again, only the `export` keywords should be removed. Maybe the last statement (`export {}`) can be removed completely, but what about something more complex (e.g. `export { foo: 'bar' }`)? So I'd prefer a suggestion over an autofix at least in this case, too.

FloEdelmann (changes requested):

tests/lib/rules/no-export-in-script-setup.js
chouchouji: I think we can autofix `export * from 'foo'` only. It is not reasonable to support autofix for other cases. Like you had mentioned, code maybe work well.

chouchouji: Or we can close this pr.  The current behavior is not wrong, it's just a hint.

FloEdelmann: I'd be fine with changing this PR to

1. not autofix anything, but report suggestions instead
2. only remove the `export` keyword if the rest is still a valid statement (e.g. `export class A {}` → `class A` and `export const foo = 'bar'` → `const foo = 'bar'`)

chouchouji: I had update code. 

1. set `fixable` to `null`
2. remind user to remove `export` or whole statement by `node.loc`

chouchouji: The test file
image

FloEdelmann: Nice, that's already an improvement. Would you like to leave this PR as is, or do you want to add [suggestions](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) as well?

chouchouji: I think the current behavior is better. Like that image, the error code will be reminded by red line. Which one do you think is better?

FloEdelmann: Why not both? :smile: 

I'm fine with merging it as is. But additionally adding suggestions (that remove the code marked with the squiggly red line) would be an even better improvement IMO.

chouchouji: Can you update code in this pr? Both red line and suggestion.

FloEdelmann: If you don't want to add suggestions, that's fine. I don't care enough to do this myself, so I'll merge this PR as is.

FloEdelmann (approved):
Thanks @chouchouji!

PR #2703, opened 2025-03-05
fix: fix walking program AST when multiple script blocks

### LINKED ISSUE #2696: `vue/prefer-use-template-ref` does not work with multiple script blocks
**Checklist**

- **ESLint version:** 9.21.0
- **eslint-plugin-vue version:** 9.32.0
- **Vue version:** 3.5.13
- **Node version:** 23.9.0
- **Operating System:** Arch Linux

```js
import js from '@eslint/js'
import pluginVue from 'eslint-plugin-vue'

export default [
  {
    name: 'app/files-to-lint',
    files: ['**/*.{js,mjs,jsx,vue}'],
  },

  {
    name: 'app/files-to-ignore',
    ignores: ['**/dist/**', '**/dist-ssr/**', '**/coverage/**'],
  },

  js.configs.recommended,
  ...pluginVue.configs['flat/essential'],
  {
    rules: {
      'vue/prefer-use-template-ref': 'error'
    }
  }
]
```

**What did you do?**	

```vue
<template>
  <div ref="root" :data-a="A" />
</template>

<script>
const A = 'foo'
</script>

<script setup>
import { ref } from 'vue'
const root = ref()
</script>
```

Eslint should report an error for the rule `vue/prefer-use-template-ref`.

**What actually happened?**

No errors are reported, it does not detect the `ref` when there is multiple `<script>` in the same file. If I remove the first `<script>` block, it reports the error. The bug also occurs with two `<script setup>` and the order of the blocks does not matter, as soon as there is more than one `<script>`, the rule stops working.

https://github.com/doominator42/vue-eslint-use-template-ref-bug

ntnyq:
This PR fixes #2696 and adds test cases for it.

When multiple script blocks exist and walk AST of node `Program`, `inScriptSetup(node)` will return false.

The range of node Program contains the range of script setup block.

### DIFF
--- a/tests/lib/rules/prefer-use-template-ref.js
+++ b/tests/lib/rules/prefer-use-template-ref.js
@@ -263,6 +263,23 @@ tester.run('prefer-use-template-ref', rule, {
           })
         </script>
       `
+    },
+    {
+      filename: 'multiple-scripts.vue',
+      code: `
+      <template>
+        <div ref="root" :data-a="A" />
+      </template>
+
+      <script>
+      const A = 'foo'
+      </script>
+
+      <script setup>
+      import { useTemplateRef } from 'vue'
+      const root = useTemplateRef('root')
+      </script>
+      `
     }
   ],
   invalid: [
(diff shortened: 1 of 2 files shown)


tests/lib/rules/prefer-use-template-ref.js
```diff
@@ -263,6 +263,23 @@ tester.run('prefer-use-template-ref', rule, {
...
+
+      <script>
+      const A = 'foo'
+      </script>
+
+      <script setup>
+      import { useTemplateRef } from 'vue'
+      const root = useTemplateRef('root')
+      </script>
+      `
```
FloEdelmann, line 282: Can you please add 2 test cases (one valid, one invalid) with `<script setup>` first and `<script>` second?

FloEdelmann (changes requested):
Thank you! Looks good to me, I have just one small suggestion.

FloEdelmann (approved):

PR #2705, opened 2025-03-06
refactor: migrate @types/eslint-visitor-keys → eslint-visitor-keys

ntnyq:
This PR removes package `@types/eslint-visistor-keys` as it's a stub package, deprecated and no longer provides type definitions.

`eslint-visitor-keys` is added to provide type definitions instead.


FloEdelmann (approved):
Thanks!

PR #2717, opened 2025-03-16
ci: prevent action running on forks

ntnyq:
This PR prevents the schedule workflow running on forks.


FloEdelmann (approved):
Makes sense to me, thank you!

PR #2724, opened 2025-03-26
Better report location for `vue/html-self-closing`

### LINKED ISSUE #2723: Better report location for `html-self-closing`
The html-self-closing rule currently highlights the entire element when there's an error. I suggest changing it to highlight only the closing part, like `/>` or `</Element>`.

Adjust `report({ loc })`

```
<template>
  <div>
    <div />
    <div></div>
  </div>
</template>
```

截屏2025-03-26 14 13 22
截屏2025-03-26 14 13 03

Image
Image

It will be easier to focus on the specific error.

**Additional context**

meowtec:
Fixes #2723

### DIFF
--- a/lib/rules/html-self-closing.js
+++ b/lib/rules/html-self-closing.js
@@ -150,8 +150,7 @@ module.exports = {
             isEmpty(node, sourceCode)
           ) {
             context.report({
-              node,
-              loc: node.loc,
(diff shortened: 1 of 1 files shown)


FloEdelmann (approved):
Looks good to me, thanks 🙂

Issue #2730, opened 2025-04-01
Omit HTML entities from `vue/no-bare-strings-in-template`


mrleblanc101:
**What rule do you want to change?**
`vue/no-bare-strings-in-template`

**Does this change cause the rule to produce more or fewer warnings?**
Fewer

**How will the change be implemented? (New option, new default behavior, etc.)?**
New value for default option

**Please provide some example code that this change will affect:**

```vue
<span v-if="$slots.default" class="text-sm">
    <slot /><span class="ml-2 whitespace-nowrap"
        >&zwj;<UHelpTooltip v-if="help" :content="help" class="align-middle"
    /></span>
</span>
```

**What does the rule currently do for this code?**
> Unexpected non-translated string used.

**What will the rule do after it's changed?**
No error reported

**Additional context**

waynzh:
There are some challenges for supporting a complete list of HTML entities. There are many HTML entities, and including all entities will make the `allowlist` very long.

IMO, the original intention of the rule design is to allow common punctuation marks and special characters, not for all. Maybe a more feasible approach could be to add only the HTML entities you actually use in your project to the `allowlist` option. If you frequently use various HTML entities, you may need to consider disabling this rule for specific code segments, or bypassing the check in other ways, like `{{ '&zwj;' }}`.

waynzh:
I'll close this issue for now, since [v10.1.0](https://github.com/vuejs/eslint-plugin-vue/releases/tag/v10.1.0) supports regexp.

Issue #2695, opened 2025-03-03
`vue/no-bare-strings-in-template` allow numbers


mrleblanc101:
**What rule do you want to change?**
`vue/no-bare-strings-in-template`

**Does this change cause the rule to produce more or fewer warnings?**
Fewer

**How will the change be implemented? (New option, new default behavior, etc.)?**
Either add an option to allow numbers:
```diff
{
  "vue/no-bare-strings-in-template": ["error", {
+   "allowNumbers": true,
    "allowlist": [
      "(", ")", ",", ".", "&", "+", "-", "=", "*", "/", "#", "%", "!", "?", ":", "[", "]", "{", "}", "<", ">", "\u00b7", "\u2022", "\u2010", "\u2013", "\u2014", "\u2212", "|"
    ],
    "attributes": {
      "/.+/": ["title", "aria-label", "aria-placeholder", "aria-roledescription", "aria-valuetext"],
      "input": ["placeholder"],
      "img": ["alt"]
    },
    "directives": ["v-text"]
  }]
}
```
or add all number to the default allowList:
```js
{
  "vue/no-bare-strings-in-template": ["error", {
    "allowlist": [
      "(", ")", ",", ".", "&", "+", "-", "=", "*", "/", "#", "%", "!", "?", ":", "[", "]", "{", "}", "<", ">", "\u00b7", "\u2022", "\u2010", "\u2013", "\u2014", "\u2212", "|", 0, 1, 2, 3, 4, 5, 6, 7, 8, 9
    ],
    "attributes": {
      "/.+/": ["title", "aria-label", "aria-placeholder", "aria-roledescription", "aria-valuetext"],
      "input": ["placeholder"],
      "img": ["alt"]
    },
    "directives": ["v-text"]
  }]
}
```

**Please provide some example code that this change will affect:**

```vue
<div>1</div>
```

**What does the rule currently do for this code?**
> Unexpected non-translated string used.

**What will the rule do after it's changed?**
No error, number don't need to be "translated" (unless they are formatted, which is not alwats needed)

**Additional context**

mrleblanc101:
If there was a way to append to the `allowlist` without overwritting it, maybe we wouldn't need this.
Maybe it possible, but I'm not sure how.

```
'vue/no-bare-strings-in-template': [
    'error',
    {
        allowlist: [ ...allowlist, 1, 2, 3, 4, 5, 6, 7, 8, 9 ],
        attributes: {
            '/.+/': [
[...]

PR #2734, opened 2025-04-06
feat(no-bare-strings-in-template): `allowlist` support regex

### LINKED ISSUE #2613: vue/no-bare-strings-in-template: per-component ignoring of slot content
The `v-icon` component in Vuetify for Vue 2 uses the default slot to define the icon ID. This is not user-readable text, so I'd like to disable `vue/no-bare-strings-in-template` for all contents of any `<v-icon>` components. 

For Vue3, this has been moved from a slot to an `icon` prop, so it's a non-issue there. But ignoring slot contents for some components seems like a good addition to this rule. 

`vue/no-bare-strings-in-template`

Fewer

New option to allow some components or tags to be exempted (by name).  For example: 
 
```js
  'vue/no-bare-strings-in-template': ['error',
    {
      ignoreSlots: {
        'v-icon': [ true /* true means default slot */, 'some-other-slot-name' ]
      }
    }
  ]
```

```vue
          <v-icon color="error">
            $power
          </v-icon>
```

Error, because the rule thinks that `$power` should be localized.

No error. 

**Additional context**

waynzh:
resolve #2613
related #2730, #2695

### DIFF
--- a/docs/rules/no-bare-strings-in-template.md
+++ b/docs/rules/no-bare-strings-in-template.md
@@ -12,7 +12,7 @@ since: v7.0.0
 
 ## :book: Rule Details
 
-This rule disallows the use of bare strings in `<template>`.  
+This rule disallows the use of bare strings in `<template>`.
 In order to be able to internationalize your application, you will need to avoid using plain strings in your templates. Instead, you would need to use a template helper specializing in translation.
 
 This rule was inspired by [no-bare-strings rule in ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-bare-strings.md).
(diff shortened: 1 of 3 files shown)


docs/rules/no-bare-strings-in-template.md
```diff
@@ -72,7 +72,7 @@ If you want to report these string literals, enable the [vue/no-useless-v-bind]
 }
 ```
 
-- `allowlist` ... An array of allowed strings.
+- `allowlist` ... An array of allowed strings or regular expression patterns (e.g. `/\d+/` to allow number).
```
FloEdelmann, line 75:
```suggestion
- `allowlist` ... An array of allowed strings or regular expression patterns (e.g. `/\d+/` to allow numbers).
```

lib/rules/no-bare-strings-in-template.js
```diff
@@ -168,7 +184,21 @@ module.exports = {
      * @param {string} str
      */
     function getBareString(str) {
-      return str.trim().replace(allowlistRe, '').trim()
+      let result = str.trim()
```
FloEdelmann, line 187: I think we don't need the first `trim()`, do we? (Since the result is trimmed at the end anyway.)

waynzh: There is a distinction when it comes to matching with regex such as `/^foo$/`.
It wouldn't match with `<h1> foo</h1>` with whitespaces,  and as a result, it would report an error.

FloEdelmann: Ah, I see.

FloEdelmann (approved):
Thanks!

ota-meshi (approved):
LGTM! Thank you!

PR #2736, opened 2025-04-14
feat: add `define-props-destructuring` rule

### LINKED ISSUE #2688: Enforcing props destructuring
The role of this rule would be to suggest [props destructuring](https://vuejs.org/guide/components/props.html#reactive-props-destructure). It could also warns on its usage if necessary.

[ ] Other (please specify:)

```vue
// vue/prefer-props-destructuring

const props = defineProps<>({...}) // 🚨

const {...} = defineProps<>({...}) // 👍
```

**Additional context**

waynzh:
resolve #2688

### DIFF
--- a/lib/rules/define-props-destructuring.js
+++ b/lib/rules/define-props-destructuring.js
@@ -14,7 +14,7 @@ module.exports = {
       categories: undefined,
       url: 'https://eslint.vuejs.org/rules/define-props-destructuring.html'
     },
-    fixable: undefined,
+    fixable: null,
     schema: [
       {
         type: 'object',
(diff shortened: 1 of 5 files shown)


lib/rules/define-props-destructuring.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+'use strict'
+
+const utils = require('../utils')
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'enforce consistent style for prop destructuring',
```
FloEdelmann, line 13:
```suggestion
      description: 'enforce consistent style for props destructuring',
```

lib/rules/define-props-destructuring.js
```diff
@@ -0,0 +1,78 @@
...
+              })
+            }
+          } else {
+            if (hasDestructure) {
+              context.report({
+                node,
+                messageId: 'avoidDestructuring'
+              })
+            }
+          }
```
FloEdelmann, line 73:
```suggestion
          if (destructurePreference === 'never') {
            if (hasDestructure) {
              context.report({
                node,
                messageId: 'avoidDestructuring'
              })
            }
            return
          }

          if (!hasDestructure) {
            context.report({
              node,
              messageId: 'preferDestructuring'
            })
            return
          }

          if (hasWithDefaults) {
            context.report({
              node: node.parent.callee,
              messageId: 'avoidWithDefaults'
            })
          }
```

waynzh: Thanks for the review!

PR #2737, opened 2025-04-15
feat(prefer-template): ignore style block

### LINKED ISSUE #2712: `vue/prefer-template` tries to apply to `style scoped`
`vue/prefer-template`

Fewer

The rule doesn't check for `style scoped` or there's a way to disable it on a per-line basis.

```vue
<script setup lang="ts">
  const someWidth = computed(() => 
    // Some computation process
  );
</script>

<style scoped>
.some-element {
  width: v-bind(someWidth + 'px');
}

.some-other-element {
  /* same issue if we try to disable the rule, it doesn't seem to work */
  /* eslint-disable-next-line vue/prefer-template */
  width: v-bind(someWidth + 'px');
}
</style>
```

It want us to replace `someWidth + 'px'` with : 
```
`{someWidth}px`
```
and this is not valid CSS handled by the IDE (VSCode in this case) so it shows another error.

Either :
- Not react to `style scoped`
- Allow to disable the rule on the style somehow or react to the `eslint-disable-next-line` properly.

**Additional context**
The only way we are able to skip this issue entirely for now is to create a new computed which does : 
```vue
const someWidthStr = computed(() => `${someWidth.value}px`);
```
which makes us need to maintain yet another `computed` which we would prefer to avoid.

waynzh:
fix #2712

### DIFF
--- a/lib/rules/prefer-template.js
+++ b/lib/rules/prefer-template.js
@@ -6,6 +6,4 @@
 const { wrapCoreRule } = require('../utils')
 
 // eslint-disable-next-line internal/no-invalid-meta
-module.exports = wrapCoreRule('prefer-template', {
-  applyDocument: true
-})
+module.exports = wrapCoreRule('prefer-template')
--- a/tests/lib/rules/prefer-template.js
+++ b/tests/lib/rules/prefer-template.js
@@ -28,7 +28,15 @@ tester.run('prefer-template', rule, {
     .text {
       color: v-bind('\`#\${hex}\`')
     }
-    </style>`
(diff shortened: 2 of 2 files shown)


waynzh:
Ah, I found that the original test case behavior is correct though. 
The `v-bind` in CSS must be wrapped in quotes. So the user's issue should be resolved by changing `width: v-bind(someWidth + 'px');` to `width: v-bind("someWidth + 'px'");`.

waynzh:
I'm going to close this PR for now and will further elaborate in the issue discussion.

Issue #2661, opened 2024-12-27
Enforce defineComponent for options API components with Typescript


GeoffreyParrier:
**Please describe what the rule should do:**

When we use Option API with typescript it is necessary to use defineComponent to enable props type inference (1).
A rule should be available to enforce it.

**What category should the rule belong to?**

[ ] Enforces code style (layout)
[x] Warns about a potential error (problem)
[ ] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

Should error on:
```vue
<script lang="ts">
export default {
 // ...
}
</script>
```

Should not error on:
```vue
<script lang="ts">
export default defineComponent({
  // ...
})
</script>
```

```vue
<script>
export default {
  // ...
}
</script>
```

```vue
<script setup>
// ...
</script>
```

```vue
<script lang="ts" setup>
// ...
</script>
```

**Additional context**

(1) source : https://vuejs.org/guide/typescript/composition-api#without-script-setup

FloEdelmann:
Thanks for the suggestion! I think it would be a good rule, but not only for `lang="ts"`, but all Vue components (that are defined using the Options API, not for `<script setup>`). This is because even for JavaScript components, using `defineComponent` helps autocompletion in the editor, prepares for a possible TypeScript migration, and increases consistency, with no downsides.

So it should also report an error for:

```vue
<script>
export default {
  // ...
}
</script>
```

And for Vue 2's way of getting TypeScript support:

```vue
<script>
export default Vue.extend({
  // ...
})
</script>
```

As a name, I'd suggest `vue/prefer-define-component`. In [v10](https://github.com/vuejs/eslint-plugin-vue/issues/2166), I think this could be added to the recommended configs. PRs to implement this are welcome :slightly_smiling_face:

The implementation might be somewhat similar to [`vue/require-direct-export`](https://eslint.vuejs.org/rules/require-direct-export.html), [`vue/require-default-export`](https://eslint.vuejs.org/rules/require-default-export.html) and [`vue/one-component-per-file`](https://eslint.vuejs.org/rules/one-component-per-file.html).

absidue:
[...]

PR #2738, opened 2025-04-25
Add new `vue/prefer-define-component` rule

thesheppard:
This is implementing the new proposed rule in #2661


absidue (commented):
I would suggest reverting all of the formatting changes to the markdown files, as it makes the diff a lot larger than it needs to be, which makes it harder to review.

docs/rules/index.md
```diff
@@ -546,6 +546,7 @@ The following rules extend the rules provided by ESLint itself and apply them to
 [vue/padding-line-between-blocks]: ./padding-line-between-blocks.md
 [vue/padding-line-between-tags]: ./padding-line-between-tags.md
 [vue/padding-lines-in-component-definition]: ./padding-lines-in-component-definition.md
+[vue/prefer-define-component]: ./prefer-define-component.md
```
FloEdelmann, line 549: This link definition is now unused. Please run `npm run update` to regenerate the docs, then [both Markdownlint errors](https://github.com/vuejs/eslint-plugin-vue/actions/runs/15364424799/job/43235511714?pr=2738) should be fixed.

2nofa11:
I noticed it's been about a month since the last update. Are you still planning to address the formatting feedback and add the changeset? If you're busy or would like help moving this forward, I'd be happy to assist 👍
(No pressure at all - just wanted to check in since this would be a valuable addition to the plugin.)

Issue #2744, opened 2025-05-14
Issue in documentation for vue/block-order


sbrauch:
[In the documentation](https://eslint.vuejs.org/rules/block-order.html  ) I found this example which didn't work. I made it work by removing the additional parentheses.
**Did not work**
{
  "vue/block-order": ["error", {
    "order": [ [ "script", "template" ], "style" ]
  }]
}

**Worked for me**
"vue/block-order": ["error", { order: ["script", "template", "style"] }],

If I'm not the only one, I think it's just a small documentation error.

waynzh:
The docs seems to be fine. Please follow the bug report template and provide more detailed information.
The outer array defines the overall order, while the inner array indicates that the order among these elements is interchangeable.

sbrauch:
Hey,
Thanks for your quick reply.
I've tried again with a college and another computer but it doesn't seem to work. I have a simple configuration where some files in my project have the order `template, script, style` and some `script, template, style`.
I want all of them to change to and always be `script, template, style`.

When I implement it without the inner array it works. This works meaning the import order becomes the one that I have defined: 
<img width="1309" alt="Image" src="https://github.com/user-attachments/assets/e4666183-6647-4d0a-b6d6-f54941e0e8bb" />

This doesn't work, meaning when I save the import order does not change.

<img width="1292" alt="Image" src="https://github.com/user-attachments/assets/5c2678d9-1bb5-4310-b0e1-f70e927206a4" />

FloEdelmann:
The inner array means "the order of these is irrelevant", i.e. `[["script", "template"], "style"]` means both `["script", "template", "style"]` and `["template", "script", "style"]` are valid. This is exactly what you noticed. If you want to force a specific order, please specify it without an inner array.

I'll close this issue because the rule is working as intended and the docs are correct. If you think the docs can be improved, feel free to comment or open a PR 🙂

PR #2745, opened 2025-05-14
docs: fix block-order default value

mrleblanc101:
fix: #2744

### DIFF
--- a/docs/rules/block-order.md
+++ b/docs/rules/block-order.md
@@ -22,14 +22,14 @@ This rule warns about the order of the top-level tags, such as `<script>`, `<tem
 ```json
 {
   "vue/block-order": ["error", {
-    "order": [ [ "script", "template" ], "style" ]
+    "order": [ "script", "template", "style" ]
   }]
 }
 ```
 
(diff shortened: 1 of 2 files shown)


ntnyq:
The docs is aligned with the default options:

https://github.com/vuejs/eslint-plugin-vue/blob/87793a579d80c8127e1e272a7ab9e33126550a4f/lib/rules/block-order.js#L14

The issue owner need to provide more information about his code, config, expect behavior and actually behavior.

mrleblanc101:
@ntnyq Why is there a nested array ? It's not explained anywhere

ntnyq:
> @ntnyq Why is there a nested array ? It's not explained anywhere... All the example use flat array

The default option means leave scripts and templates as it's original order, but moves all styles to the bellow.

mrleblanc101:
@ntnyq There is one, out of all the others, and it's not explained.
Maybe there should be an explanation on the syntax ?

waynzh:
As @ ntnyq and https://github.com/vuejs/eslint-plugin-vue/issues/2744#issuecomment-2888910794 mentioned, there's no issue with the documentation.
Feel free to continue improving the documentation if you're interested.

Issue #2741, opened 2025-05-09
Regression with props destructuring and optional props between 10.0.0 and 10.1.0


mrleblanc101:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 9.26.0
- **eslint-plugin-vue version:** 10.1.0
- **Vue version:** 3.5.13
- **Node version:** v20.18.0
- **Operating System:** macOS 15.4.1

**Please show your full configuration:**

```json5
import eslintConfigPrettier from 'eslint-config-prettier';
import prettierPlugin from 'eslint-plugin-prettier';
import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended';

import withNuxt from './.nuxt/eslint.config.mjs';

export default withNuxt({
    plugins: {
        prettier: prettierPlugin,
    },
    rules: {
        ...prettierPlugin.configs.recommended.rules,
        ...eslintConfigPrettier.rules,
        ...eslintPluginPrettierRecommended.rules,
        'vue/no-bare-strings-in-template': [
            'error',
            {
                attributes: {
                    '/.+/': [
                        'title',
                        'description',
                        'label',
                        'help',
                        'tooltip',
                        'placeholder',
                        'errors',
                        'content',
                        'confirmButtonLabel',
                        'confirmButtonVariant',
                    ],
                    'UTab': ['name'],
                },
                directives: ['v-text'],
            },
        ],
        'vue/v-slot-style': ['error', 'shorthand'],
        'vue/define-macros-order': 'error',
    },
});

```

**What did you do?**	

```vue
<template>
    <Component
        :is="to ? NuxtLink : 'button'"
        :type="!to ? type : null"
        :to="to"
        :class="
            twMerge(
                'relative inline-flex h-14 items-center justify-center gap-2.5 rounded border-2 border-transparent px-8 text-center text-sm font-bold leading-6 text-white transition duration-300 ease-in-out disabled:pointer-events-none disabled:opacity-25 [&[disabled]]:pointer-events-none [&[disabled]]:opacity-25',
                variant === 'info' && 'bg-blue hover:bg-blue-80 focus-visible:bg-blue-80',
                variant === 'primary' && 'bg-teal hover:bg-teal-80 focus-visible:bg-teal-80',
[...]

Issue #2725, opened 2025-03-26
`vue/require-default-prop` doesn't trigger with props destructure


hartbit:
**Checklist**

- [X] I have tried restarting my IDE and the issue persists.
- [X] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 9.21.0
- **eslint-plugin-vue version:** 10.0.0
- **Vue version:** 3.5.13
- **Node version:** 22.14.0
- **Operating System:** macOS

**Please show your full configuration:**
```ts
import pluginVue from "eslint-plugin-vue";
import { defineConfigWithVueTs, vueTsConfigs } from "@vue/eslint-config-typescript";

export default defineConfigWithVueTs(
  {
    name: "app/files-to-lint",
    files: ["**/*.{ts,mts,tsx,vue}"],
  },

  {
    name: "app/files-to-ignore",
    ignores: ["**/dist/**", "**/dist-ssr/**", "**/coverage/**"],
  },

  pluginVue.configs["flat/recommended"],
  vueTsConfigs.recommended
);

```

**What did you do?**	
```vue
const { msg } = defineProps<{
  msg?: string;
}>();
```

**What did you expect to happen?**
I switched from using the `withDefaults` to the new [props destructure syntax](https://vuejs.org/api/sfc-script-setup.html#reactive-props-destructure) and I expected the `vue/require-default-prop` to continue to check when defaults were not defined in the destructured object. I expected the above example to complain that I did not define defaults for `msg`.

**What actually happened?**
I got no errors in the above example.

**Repository to reproduce this issue**
https://github.com/hartbit/eslint-require-default-prop-destructured-issue
* The `src/components/HelloWorld.vue` component generates a warning, as expected, as it uses `withDefaults`,
* The `src/components/HelloWordIssue.vue` component does not generate a warning

ZexTech:
The potentially unwanted change with this rule is that it now requires props to be defined (with undefined value).
This results in unnecessarily verbose code, despite the types being already correctly defined

```ts
// Before ✅
const { optional } = defineProps<{
  optional?: boolean;
  optionalLabel?: string;
}>();

// After ❌
const { optional, optionalLabel = undefined } = defineProps<{
  optional?: boolean;
  optionalLabel?: string;
}>();
```

[...]

PR #2768, opened 2025-06-24
fix(require-default-prop): avoid requiring defaults for optional props not included in destructuring

### LINKED ISSUE #2767: `vue/require-default-prop`: Avoid requiring defaults for optional props not included in destructuring

`vue/require-default-prop`

This change will produce fewer warnings. Specifically, it will stop producing warnings for optional props that are not included in the destructuring pattern when using Vue 3.5's props destructuring syntax.

New default behavior - no new options needed. The rule will be modified to check if an optional prop is actually destructured before requiring a default value. If an optional prop is not included in the destructuring pattern, it won't require a default value.

```vue
<script setup lang="ts">
// Case 1: Optional prop NOT destructured - should NOT require default (new behavior)
const { requiredProp } = defineProps<{
  requiredProp: string;
  optionalNotUsed?: string; // ← This should NOT cause an error
}>()

// Case 2: Optional prop IS destructured - should require default (existing behavior)
const { requiredProp, optionalUsed } = defineProps<{
  requiredProp: string;
  optionalUsed?: string; // ← This should still cause an error without default
}>()

// Case 3: Correct usage with defaults
const { requiredProp, optionalUsed = 'default' } = defineProps<{
  requiredProp: string;
  optionalUsed?: string;
}>()
</script>
```

Currently, the rule produces warnings for ALL optional props when using props destructuring, even if they are not included in the destructuring pattern. This is unnecessarily verbose and forces developers to destructure props they don't need just to assign `= undefined`.

Example of current behavior:
```typescript
// Current: Forces unnecessary destructuring
const { requiredProp, optionalNotUsed = undefined } = defineProps<{
  requiredProp: string;
  optionalNotUsed?: string; // Must be destructured even if not used
}>()
```

After the change, the rule will only require default values for optional props that are actually included in the destructuring pattern. Optional props that are not destructured will be ignored by this rule, allowing for cleaner and more intuitive code.

**Additional context**

This change addresses the regression introduced in version 10.1.0 as reported in issues #2725 and #2741.

**Background from #2725:**
The original issue was that `vue/require-default-prop` didn't trigger with props destructure at all. This was fixed in PR #2728, but the fix was too strict and started requiring default values for ALL optional props, even those not used in destructuring.

**Background from #2741:**
Users reported a regression where they had to destructure all optional props and assign `= undefined` even when those props were only used in templates, leading to unnecessarily verbose code.

**The Problem:**
```typescript
// Before the fix in #2728: No warnings (incorrect)
const { msg } = defineProps<{ msg?: string }>()

// After the fix in #2728: Requires unnecessary defaults (too strict)
const { used, unused = undefined } = defineProps<{
  used?: string;
  unused?: string; // unused in script, only in template
}>()
```

**Proposed Solution:**
Only require default values for optional props that are explicitly included in the destructuring pattern:

```typescript
// Should NOT warn (unused is not destructured)
const { used } = defineProps<{
  used?: string;    // ← Should warn (destructured without default)
  unused?: string;  // ← Should NOT warn (not destructured)
}>()

// Should warn only for 'used'
// Should NOT warn for 'unused'
```

This approach maintains the intent of the original fix while avoiding the verbose regression that affects users who don't destructure all their optional props.

yuasa-engineer:
Fixes https://github.com/vuejs/eslint-plugin-vue/issues/2767

If this approach doesn't align with the project's direction, please don't hesitate to reject this proposal. In that case, you may also close the corresponding issue.

## What type of PR is this?

bug fix / enhancement

## What does this PR do? / Why do we need it?

This PR fixes a regression in the `vue/require-default-prop` rule introduced in v10.1.0. The rule was requiring default values for ALL optional props when using props destructuring, even if those props were not included in the destructuring pattern.

### Before this fix
```typescript
// Was forced to destructure all optional props, even unused ones
const { used, unused = undefined } = defineProps<{
  used?: string;
  unused?: string; // Only used in template, forced to assign undefined
}>()
```

### After this fix
```typescript
// Only need to provide defaults for actually destructured optional props
const { used = 'default' } = defineProps<{
  used?: string;    // Destructured, so needs default
  unused?: string;  // Not destructured, no default needed
}>()
```

## What changes did you make? (Give an overview)

1. **Modified `lib/rules/require-default-prop.js`**:
   - Added logic to check if an optional prop is actually included in the destructuring pattern
   - Only requires default values for optional props that are explicitly destructured
   - Uses `utils.getPropsDestructure(node)` to get destructured props and check if optional prop exists

2. **Updated test cases in `tests/lib/rules/require-default-prop.js`**:
   - Added valid test cases for optional props not included in destructuring
   - Updated invalid test cases to properly test mixed scenarios
   - Added comprehensive test coverage for the new behavior

## Which issue(s) does this PR fix?

Fixes #2767
Related to #2725 and #2741

## Additional context

This change maintains backward compatibility and only reduces false positive warnings. The rule continues to work as expected for:
- Required props (still checked)
- Optional props with explicit destructuring (still requires defaults)  
- Boolean props (still exempt)
- Props using `withDefaults()` (still works)

The fix specifically addresses the user feedback that the previous fix was too aggressive and made Vue 3.5's props destructuring unnecessarily verbose.

### DIFF
--- a/.changeset/wise-bugs-peel.md
+++ b/.changeset/wise-bugs-peel.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+[vue/require-default-prop](https://eslint.vuejs.org/rules/require-default-prop.html) now ignores optional props that are not included in Vue 3.5 props destructuring pattern.
--- a/lib/rules/require-default-prop.js
+++ b/lib/rules/require-default-prop.js
@@ -207,6 +207,19 @@ module.exports = {
               if (defaultsByWithDefaults[prop.propName]) {
                 return true
               }
+              // If using props destructure and this is an optional prop that is NOT included
+              // in the destructure pattern, exclude it from the report
+              if (
+                isUsingPropsDestructure &&
+                !prop.required &&
+                prop.propName != null
+              ) {
+                const destructuredProps = utils.getPropsDestructure(node)
+                if (!destructuredProps[prop.propName]) {
+                  // Optional prop is not destructured, so no default value is needed
+                  return true
+                }
+              }
             }
             if (!isUsingPropsDestructure) {
               return false
--- a/tests/lib/rules/require-default-prop.js
+++ b/tests/lib/rules/require-default-prop.js
@@ -413,6 +413,40 @@ ruleTester.run('require-default-prop', rule, {
         ...languageOptions,
         parserOptions: { parser: require.resolve('@typescript-eslint/parser') }
       }
+    },
+    // Optional props that are not destructured should not require default values
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup lang="ts">
+      const { bar } = defineProps<{
+        foo?: string;
+        bar: string;
+      }>()
+      </script>
+      `,
+      languageOptions: {
+        parser: require('vue-eslint-parser'),
+        ...languageOptions,
+        parserOptions: { parser: require.resolve('@typescript-eslint/parser') }
+      }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup lang="ts">
+      const { bar } = defineProps<{
+        foo?: string;
+        bar: string;
+        baz?: number;
+      }>()
+      </script>
+      `,
+      languageOptions: {
+        parser: require('vue-eslint-parser'),
+        ...languageOptions,
+        parserOptions: { parser: require.resolve('@typescript-eslint/parser') }
+      }
     }
   ],
 
@@ -728,6 +762,7 @@ ruleTester.run('require-default-prop', rule, {
     },
     {
       // https://github.com/vuejs/eslint-plugin-vue/issues/2725
+      // Optional props that ARE destructured should still require default values
       filename: 'type-with-props-destructure.vue',
       code: `
       <script setup lang="ts">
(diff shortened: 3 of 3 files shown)


waynzh:
Sorry for the late reply.

I'm not sure about others' use cases. For example, in the previous `withDefaults` case, any optional fields without a defined default value would result in errors. So why are we choosing to ignore this error specifically when destructuring props?

```
withDefaults(defineProps<{
  used?: string;
  unused?: string;  // error! Prop 'unused' requires default value to be set.
}>(), {
  used: 'default'
})
```

It looks like there's some feedback about this being too verbose. Maybe we could add an option to control whether optional props not included in destructuring should be ignored? This way, users who know what they're doing can explicitly configure it.

I'll hold off for now and see what others think.

yuasa-engineer:
@waynzh 
  Thanks for the feedback! You're absolutely right about the need for consistency.

  This approach actually reflects the **withDefaults era philosophy**:
  - Props in `withDefaults()` → intentionally want meaningful defaults
  - Props not in `withDefaults()` → undefined is acceptable

  Destructuring follows the same logic:
  - Props in destructuring → intentionally want meaningful defaults
  - Props not destructured → undefined is acceptable (accessed via `props.propName`)

  **You're right that this should be an option.** There are valid use cases for both approaches:

  1. **Strict mode**: All optional props need defaults (current behavior)
  2. **Intent-based mode**: Only actively used props need defaults

  I should implement this as a configurable option that works consistently across **both withDefaults and
   destructuring patterns**.

  Would you like me to:
  1. Add an option like `"ignoreNonUsed": true/false`
  2. Apply the same logic to both withDefaults and destructuring cases
  3. Default to `false` (current behavior) for backward compatibility

  What do you think about this approach?

waynzh:
@yuasa-engineer
Thank you for the detailed explanation! I completely agree that this approach would be much better.

I have a question though - what would "non-used" props mean in the `withDefaults` cases? It seems like apply this logic to `withDefaults` cases would have the same effect as just turning off the rule entirely.

So I think the option could simply be called `ignoreNonDestructured` and only handle destructuring cases. This keeps the solution focused and avoids unnecessary complexity.

yuasa-engineer:
Thank you for the detailed feedback! After further consideration, I realize my earlier analysis was incorrect.

This option would only be useful in edge cases where props are used solely in templates with undefined being acceptable. However, even in these cases, simply disabling the rule with  provides a simpler and more appropriate solution.

Given the limited utility and added complexity, I believe withdrawing this proposal is the right approach.

Thank you for your time reviewing this, and apologies for the confusion.

yuasa-engineer:
@waynzh 
Thank you for the detailed feedback! Having given it further thought, I realize that my earlier analysis was incorrect.

This option would only be useful in edge cases where props with 'undefined' as an acceptable default value are used in the template section. However, even in these cases, simply disabling the rule with 'eslint-disable-line' provides a simpler and more appropriate solution.

Given its limited utility and added complexity, I believe that withdrawing this proposal is the right approach.

Thank you for your review, and I apologize for any confusion.

Issue #2756, opened 2025-06-10
Upgrade ESLint from v8 to v9


2nofa11:
## Summary
Upgrade ESLint from v8 to v9 to support the latest features and enable [eslint-typegen](https://github.com/antfu/eslint-typegen) integration.

## Motivation
- Required for eslint-typegen compatibility (depends on ESLint v9)
- Better compatibility with modern tooling ecosystem

## Changes needed
- [ ] Upgrade ESLint to v9.x
- [ ] Update @typescript-eslint/parser to v8.x for compatibility
- [ ] Update @types/eslint to v9.x
- [ ] Fix TypeScript compilation errors due to @typescript-eslint API

## Related work
This upgrade is a prerequisite for [eslint-typegen](https://github.com/antfu/eslint-typegen) integration which will provide TypeScript definitions and IntelliSense for ESLint rule configurations.

ota-meshi:
If we were to make a breaking change to this package, there would be other things I would like to do.
For example, I would like to change this plugin to the ESLint language plugin.
https://eslint.org/docs/latest/extend/languages
Changing to the language plugin would allow to use directive comments such as `<!-- eslint vue/attribute-hyphenation: off -->`.

When I have time, I will list the breaking changes I would like to make to this plugin.

PR #2770, opened 2025-06-26
feat: add TypeScript IntelliSense support with eslint-typegen

### LINKED ISSUE #2757: Add TypeScript IntelliSense support with eslint-typegen
## Summary
Add [eslint-typegen](https://github.com/antfu/eslint-typegen) integration to provide TypeScript definitions and IntelliSense for ESLint rule configurations.

## Motivation
- Improve developer experience with TypeScript IntelliSense
- Provide type safety for ESLint configuration
- Reduce configuration errors through better tooling support

## Dependencies
Requires ESLint v9 upgrade (https://github.com/vuejs/eslint-plugin-vue/issues/2756)

## Benefits
- Full TypeScript IntelliSense for all ESLint rules
- Type safety prevents configuration errors

2nofa11:
## Summary

Add TypeScript IntelliSense support via `eslint-typegen` with ESM migration for ESLint v9.0+ compatibility.

## Changes

- Integrate `eslint-typegen` for TypeScript definitions
- Convert `eslint.config.js` to `eslint.config.mjs` (required for eslint-typegen)
- Install with `--legacy-peer-deps` due to ESLint version requirements
- Update all imports to ESM format

## Benefits

- TypeScript IntelliSense for ESLint configurations
- Modern ESLint v9.0+ compatibility
- Improved developer experience with type safety

## Screenshot

| Before(No TypeScript Support) | After(TypeScript IntelliSense Active) |
|--------|--------|
| No TypeScript Support | TypeScript IntelliSense Active|

close  #2757

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "docs:build": "vitepress build docs",
     "generate:version": "env-cmd -e version npm run update && npm run lint -- --fix",
     "changeset:version": "changeset version && npm run generate:version && git add --all",
-    "changeset:publish": "changeset publish"
+    "changeset:publish": "npm run typegen && changeset publish"
   },
   "files": [
     "lib"
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "eslint-plugin-prettier": "^5.2.1",
     "eslint-plugin-unicorn": "^56.0.0",
     "eslint-plugin-vue": "file:.",
+    "eslint-typegen": "^2.2.0",
     "eslint-visitor-keys": "^4.2.0",
     "espree": "^10.4.0",
     "events": "^3.3.0",
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "version": "npm run generate:version && git add .",
     "update": "node ./tools/update.js",
     "update-resources": "node ./tools/update-resources.js",
+    "typegen": "node ./tools/generate-typegen.mjs",
     "docs:watch": "vitepress dev docs",
     "predocs:build": "npm run update",
     "docs:build": "vitepress build docs",
--- a/tools/generate-typegen.mjs
+++ b/tools/generate-typegen.mjs
@@ -1,0 +1,9 @@
+import fs from 'node:fs/promises'
+import { pluginsToRulesDTS } from 'eslint-typegen/core'
+import plugin from '../lib/index.js'
+
+const dts = await pluginsToRulesDTS({
+  vue: plugin
+})
+
+await fs.writeFile('lib/eslint-typegen.d.ts', dts)
--- a/.changeset/neat-swans-argue.md
+++ b/.changeset/neat-swans-argue.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Add TypeScript IntelliSense support with eslint-typegen
--- a/.changeset/neat-swans-argue.md
+++ b/.changeset/neat-swans-argue.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": minor
 ---
 
-Add TypeScript IntelliSense support with eslint-typegen
+Add TypeScript IntelliSense support via [eslint-typegen](https://github.com/antfu/eslint-typegen)
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ yarn-error.log
 /docs/.vitepress/.temp
 /docs/.vitepress/cache
 typings/eslint/lib/rules
+eslint-typegen.d.ts
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,13 +1,19 @@
-'use strict'
+import globals from 'globals'
+import eslintPluginEslintPlugin from 'eslint-plugin-eslint-plugin/configs/all'
+import eslintPluginJsonc from 'eslint-plugin-jsonc'
+import eslintPluginNodeDependencies from 'eslint-plugin-node-dependencies'
+import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended'
+import eslintPluginUnicorn from 'eslint-plugin-unicorn'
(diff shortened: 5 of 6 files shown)


package.json
FloEdelmann, line 1: Please update to ESLint v9 in the dev dependencies, so we don't need to use `--legacy-peer-deps`. This would also fix the Netlify deployment that is currently failing due to that peer dependency mismatch.

Updating that version is irrelevant for users of this package, as ESLint v8 is still supported as a peer dependency. We also run the tests in the CI with ESLint v8, so everything is well tested.

ota-meshi:
Thank you for this PR!
Furthermore, can you make it so that we can provide rule types from this package using the low-level API?
https://github.com/antfu/eslint-typegen/blob/main/README.md#low-level-api

package.json
2nofa11: Thank you for the feedback :)

I initially anticipated that ESLint v9 would be a blocker for this issue, so I created a separate [issue for the ESLint v9 upgrade](https://github.com/vuejs/eslint-plugin-vue/issues/2756). 
However, I received [advice that eslint-typegen doesn't strictly require ESLint v9](https://github.com/vuejs/eslint-plugin-vue/issues/2757#issuecomment-2961738922), and I was able to confirm it works locally with the current setup using `--legacy-peer-deps`.

From a project perspective, which approach would you prefer ?
* ESLint v9 as blocker
  Update ESLint v9 in (https://github.com/vuejs/eslint-plugin-vue/issues/2756), then integrate eslint-typegen without `--legacy-peer-deps`
* Proceed with ESLint v8  
  Keep the current ESLint v8 setup and use `--legacy-peer-deps` for eslint-typegen integration


I'm happy to proceed with whichever approach aligns 😄

2nofa11:
@ota-meshi 
I've implemented rule type generation using eslint-typegen's low-level API. ([856784e](https://github.com/vuejs/eslint-plugin-vue/pull/2770/commits/856784e053c9853a529a1734d5a51be7d0543f6a))

  **Implementation:**
  - Added `npm run typegen` command (like that [@antfu/eslint-config](https://github.com/antfu/eslint-config/blob/95963ac345d27cd06a4eeb5ebc16e1848cb2fd81/package.json#L33))
  - Uses `pluginsToRulesDTS` from `eslint-typegen/core`
  - Generated file `lib/eslint-typegen.d.ts` is currently gitignored

Should I also add lib/eslint-typegen.d.ts to this PR, or would you prefer CI automation for the distribution strategy?

package.json
```diff
@@ -21,6 +21,7 @@
     "version": "npm run generate:version && git add .",
     "update": "node ./tools/update.js",
     "update-resources": "node ./tools/update-resources.js",
+    "typegen": "node ./tools/generate-typegen.mjs",
```
ota-meshi, line 24: Awesome! Could you please modify the `changeset:publish` script so that it automatically updates with releases?

```json
    "changeset:publish": "npm run typegen && changeset publish"
```

tools/generate-typegen.mjs
```diff
@@ -0,0 +1,9 @@
+import fs from 'node:fs/promises'
+import { pluginsToRulesDTS } from 'eslint-typegen/core'
+import plugin from '../lib/index.js'
+
+const dts = await pluginsToRulesDTS({
+  vue: plugin
+})
+
+await fs.writeFile('lib/eslint-typegen.d.ts', dts)
```
ota-meshi, line 9: I'm guessing maybe we need to reference the generated file from `lib/index.d.ts`, but is that not necessary?

package.json
FloEdelmann: > received [advice that eslint-typegen doesn't strictly require ESLint v9](https://github.com/vuejs/eslint-plugin-vue/issues/2757#issuecomment-2961738922), and I was able to confirm it works locally with the current setup using `--legacy-peer-deps`.

The fact that you have to use `--legacy-peer-deps` means that eslint-typegen officially only supports ESLint v9. It may work with ESLint v8, but it might break without warning. So I'd definitely prefer updating the development dependency ESLint to v9 (which, as I wrote, would _not_ be a breaking change).

I think https://github.com/vuejs/eslint-plugin-vue/issues/2756#issuecomment-2961759090 refers to updating our publicly supported peer dependency version of ESLint from `^8.57.0 || ^9.0.0` to something like `^9.0.0`. That would indeed be a breaking change that I'd also like to avoid for some time.

2nofa11: I misunderstood—sorry about that! 🙇‍♂️ 
Taking your advice, I’ve updated the ESLint version in our dev-dependencies to v9 (4c5914a4473f8cca1a79cb065da294dc6e363541 and 750151792c73e8337548ce077c52640343eed9c5).
The upgrade exposed a few additional lint errors, which I fixed in 01b2ccc12b05af17ab1dd805e80fb28fe44417e2 and d41249b4dc0e71512659d1012ec88218370af009.

Thanks again for catching this—looking forward to your review!

package.json
2nofa11: Thank you for the suggestion!
I added npm run typegen to changeset:publish 👍
(df481e355c415f8a8cdf2710fb3c82c4f37b5985)

tools/generate-typegen.mjs
2nofa11: As you pointed out, I needed to add the types to `lib/index.d.ts`. (254c69708a4704a99afb1c56ee293e610fb13f34)

After doing that, I now get TypeScript IntelliSense like the ones shown in the screenshot☺️
I could not receive TypeScript IntelliSense without the above commitments.

screenshot

package.json
FloEdelmann: https://github.com/vuejs/eslint-plugin-vue/pull/2775 is now merged, so this prerequisite is fulfilled now. Can you please merge `master` into your branch to resolve the conflicts?

2nofa11: I’ve confirmed it’s been merged and pulled the latest master branch into this PR🙂
Please review whenever you have a moment!

PR #2775, opened 2025-07-01
chore: Use ESLint v9 in development env

ota-meshi:
This PR will change the development environment to use ESLint v9.

After looking into the documentation build errors in #2770, I found that a lot of changes are needed.
I want to address those issues first with this PR.

Also, some tests were failing for the script-indent rule when using typescript-eslint v8, so I fixed those as well.


ota-meshi:
@2nofa11 Once this PR is merged, please rebase your PR 🙏

.changeset/gentle-glasses-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed false negatives when using typescript-eslint v8 in `vue/script-indent` rule
```
FloEdelmann: Can we add a CI job for typescript-eslint v7, so we don't break it accidentally?

FloEdelmann: Also, should we maybe add `"@typescript-eslint/parser": "^7.0.0 || ^8.0.0"` as an optional peer dependency?

FloEdelmann: Let's link rules in changesets, so changelogs are more helpful to users:

```suggestion
Fixed false negatives when using typescript-eslint v8 in [`vue/script-indent`](https://eslint.vuejs.org/rules/script-indent.html) rule
```

ota-meshi: Thank you for your review! I added a test job for it.

.github/workflows/CI.yml
```diff
@@ -66,3 +66,18 @@ jobs:
...
+  test-with-typescript-eslint-v7:
+    name: Test with typescript-eslint v7
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install Node.js
+        uses: actions/setup-node@v4
+      - name: Install Packages
+        run: npm install -f
```
FloEdelmann: Do we still need `-f` in all the `npm install` commands in all CI jobs?

.changeset/gentle-glasses-tie.md
ota-meshi: I don't think there is any need to specify optional dependency because `typescript-eslint` is a dependency resolved in `eslint.config.js` and is not required for users who don't use `typescript`.

.github/workflows/CI.yml
ota-meshi: The force flag seems unnecessary so I removed it 👍

waynzh (approved):
Thanks!

.changeset/gentle-glasses-tie.md
FloEdelmann: I agree, for our own codebase it's irrelevant.

But some rules (e.g. `vue/script-indent`) change their behavior for `lang="ts"` blocks and if `@typescript-eslint/parser` is available. But we only support specific versions of `@typescript-eslint/parser` in these rules (namely v7 and v8; while actually v8 has been untested untl now). The issue you just fixed arose from different major versions of typescript-eslint behaving differently.

Specifying exactly which versions we support might help users by e.g. delaying typescript-eslint updates until we support them, so they don't break some rules implicitly.

ota-meshi: I agree that it's hard for users and us to know which version of `typescript-eslint` we support.
For now, the only way to find out is by checking the Actions definition.

Do you think this could be solved by mentioning somewhere in the documentation what version of `typescript-eslint` we are testing against?

FloEdelmann: I'm not opposed to it, but I think it will likely be missed once we update our supported versions. Updating a peer dependency version will likely not be missed because the update happens in the `package.json` anyway; and if we miss it, then users will complain because they can't update.

ota-meshi: Hmm. I think adding `@typescript-eslint/parser` to peer dependencies will install `typescript` locally for users who don't use `typescript`.
I've seen a lot of users who don't want `typescript` installed.
https://github.com/eslint-community/eslint-plugin-n/pull/449
https://github.com/eslint-stylistic/eslint-stylistic/issues/571

FloEdelmann: But only for required peer dependencies, not for optional ones, right?

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependenciesmeta

> Npm will not automatically install optional peer dependencies. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.

docs/.vitepress/build-system/build.mts
```diff
@@ -47,11 +75,11 @@ function transform(code: string, injects: string[]) {
 ${injects
   .map(
     (inject) =>
-      `import $inject_${inject.replace(/-/gu, '_')}$ from '${inject}';`
+      `import $inject_${inject.replace(/[\-:]/gu, '_')}$ from '${inject}';`
```
2nofa11, line 77: Ah, I see—you’re replacing the colon as well because we can get values like `node:stream`. Understood!

.changeset/gentle-glasses-tie.md
ota-meshi: Oh, I didn't know that we could avoid installing it with `peerDependenciesMeta.*.optional=true`.
If so, I think it would be useful for us 😄

ota-meshi: I have added `@typescript-eslint/parser` to peer dependencies 👍

docs/.vitepress/config.mts
```diff
@@ -142,24 +143,32 @@ export default async () => {
...
           tslib: path.join(dirname, '../../node_modules/tslib/tslib.es6.js'),
-          esquery: path.join(dirname, './build-system/shim/esquery.mjs'),
-          globby: path.join(dirname, './build-system/shim/globby.mjs')
+          globby: path.join(dirname, './build-system/shim/empty.mjs'),
+          'fast-glob': path.join(dirname, './build-system/shim/empty.mjs'),
+          'node:fs': 'fs',
+          'node:path': 'path',
+          'node:util': 'util'
         }
       },
```
2nofa11, line 168: When I run `npm run docs:build`, Vite automatically detects Node.js built-in modules such as `module`, `fs`, and `stream`, and shows warnings like:

```
[plugin:vite:resolve] Module "module" has been externalized for browser compatibility
  imported by "/Users/~/eslint-plugin-vue/docs/.vitepress/build-system/shim/vue-eslint-parser.mjs".
  See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
```

These warnings appear because those modules aren’t available in browser environments.
To resolve the issue, explicitly list those modules in `build.rollupOptions.external` .

```
build: {
  rollupOptions: {
    external: [
      'module',
      'stream', 
      'node:stream',
      'os',
      'node:os', 
      'util',
      'node:util',
      'path', 
      'node:path',
      'fs',
      'node:fs'
    ]
  }
```

docs/.vitepress/config.mts
```diff
@@ -142,24 +143,32 @@ export default async () => {
...
-          path: path.join(dirname, './build-system/shim/path.mjs'),
 
           tslib: path.join(dirname, '../../node_modules/tslib/tslib.es6.js'),
-          esquery: path.join(dirname, './build-system/shim/esquery.mjs'),
-          globby: path.join(dirname, './build-system/shim/globby.mjs')
+          globby: path.join(dirname, './build-system/shim/empty.mjs'),
+          'fast-glob': path.join(dirname, './build-system/shim/empty.mjs'),
+          'node:fs': 'fs',
+          'node:path': 'path',
+          'node:util': 'util'
```
2nofa11: Setting `build.rollupOptions.external` might make it unnecessary!

docs/.vitepress/config.mts
ota-meshi: In fact, some of these modules can be replaced by shims and work, so they shouldn't be externalized.
However, I've adjusted the configuration so that the warning doesn't appear 👍

package.json
```diff
@@ -58,7 +58,13 @@
...
   "peerDependencies": {
     "eslint": "^8.57.0 || ^9.0.0",
-    "vue-eslint-parser": "^10.0.0"
+    "vue-eslint-parser": "^10.0.0",
+    "@typescript-eslint/parser": "^7.0.0 || ^8.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@typescript-eslint/parser": {
+      "optional": true
+    }
```
FloEdelmann, line 67: Can you please add another changeset (with a `minor` version update) explaining this change?

ota-meshi: I have added the changeset 👍

Issue #2710, opened 2025-03-07
`vue/no-deprecated-slot-attribute` conflicts with web component named slots


raoufswe:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** v9
- **eslint-plugin-vue version:** v10.0.0
- **Vue version:** v3
- **Node version:** v20
- **Operating System:** Mac OS

**Please show your full configuration:**

```json5

import pluginVue from 'eslint-plugin-vue'
import globals from 'globals'

export default [
  // add more generic rulesets here, such as:
  // js.configs.recommended,
  ...pluginVue.configs['flat/recommended'],
  // ...pluginVue.configs['flat/vue2-recommended'], // Use this if you are using Vue.js 2.x.
  {
    rules: {
      // override/add rules settings here, such as:
      // 'vue/no-unused-vars': 'error'
    },
    languageOptions: {
      sourceType: 'module',
      globals: {
        ...globals.browser
      }
    }
  }
]

```

**What did you do?**	

```vue
 // my-element is a custom element / web component. 
<my-element>
  <div slot="named">hello</div>
</my-element>

```

**What did you expect to happen?**

Vue has a documented limitation when using named slots with web components. If a web component that accepts a named slot is used in a Vue app, the `vue/no-deprecated-slot-attribute` linting rule conflicts with this behavior.

> When passing named slots, use the slot attribute instead of the v-slot directive.

https://vuejs.org/guide/extras/web-components#slots

**What actually happened?**

Since we must use the native slot attribute when working with named slots in web components and vue, consumers will encounter the following linting error unless it is explicitly disabled:

```
  11:22  error    `slot` attributes are deprecated vue/no-deprecated-slot-attribute
```

**Repository to reproduce this issue**

The issue can be reproduced in [this nuxt@3 example](https://github.com/justeattakeaway/pie-aperture/blob/dsw-000-add-eslint-vue/nuxt-app/pages/components/button.vue#L11)

To see the lint error:

Navigate to the nuxt-app directory:

```
cd nuxt-app
```

Run the following command and you should be able to see the error:
```
yarn dlx eslint ./pages/components/button.vue
```

raoufswe:
Hi @FloEdelmann
[...]

PR #2784, opened 2025-07-09
feat(no-deprecated-slot-attribute): add ignoreParents option

ST-DDT:
Adds #2710

- #2710

---

Adds an `ignoreParents: string[]` option to `no-deprecated-slot-attribute` to allow ignoring web-component usecases:

````vue
<template>
	<web-component>
 		<Entry slot="entry" />
	</web-component>
</template>
````

Preview:

- [`rules/no-deprecated-slot-attribute` (new)](https://deploy-preview-2784--eslint-plugin-vue.netlify.app/rules/no-deprecated-slot-attribute.html)
- [`rules/no-deprecated-slot-attribute` (current)](https://eslint.vuejs.org/rules/no-deprecated-slot-attribute.html)

---

Note to self: Next PR(s):

- Increase the specificity of the rule test case assertions https://github.com/vuejs/eslint-plugin-vue/pull/2773#discussion_r2176359714

### DIFF
--- a/.changeset/beige-teams-camp.md
+++ b/.changeset/beige-teams-camp.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Added `ignoreParents` option to `no-deprecated-slot-attribute`
--- a/.changeset/beige-teams-camp.md
+++ b/.changeset/beige-teams-camp.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': minor
 ---
 
-Added `ignoreParents` option to `no-deprecated-slot-attribute`
+Added `ignoreParents` option to [`vue/no-deprecated-slot-attribute`](https://eslint.vuejs.org/rules/no-deprecated-slot-attribute.html)
--- a/docs/rules/no-deprecated-slot-attribute.md
+++ b/docs/rules/no-deprecated-slot-attribute.md
@@ -50,7 +50,7 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
 ```
 
 - `"ignore"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) that ignore these rules. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
-- `"ignoreParents"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) for parents that ignore these rules. This option is especially useful for Web-Components. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
+- `"ignoreParents"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) for parents that ignore these rules. This option is especially useful for [Web-Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components). This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
 
 ### `"ignore": ["my-component"]`
 
@@ -117,6 +117,7 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
 ## :books: Further Reading
 
 - [API - slot](https://v2.vuejs.org/v2/api/#slot-deprecated)
+- [Web - slot](https://developer.mozilla.org/en-US/docs/Web/API/Element/slot)
 
 ## :rocket: Version
 
--- a/docs/rules/no-deprecated-slot-attribute.md
+++ b/docs/rules/no-deprecated-slot-attribute.md
@@ -89,19 +89,19 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
 
 ```vue
 <template>
-  <my-web-component'>
+  <my-web-component>
     <!-- ✓ GOOD -->
     <template v-slot:name>
       {{ props.title }}
     </template>
-  </my-web-component'>
+  </my-web-component>
 
-  <my-web-component'>
+  <my-web-component>
     <!-- ✓ GOOD -->
     <my-component slot="name">
       {{ props.title }}
     </my-component>
-  </my-web-component'>
+  </my-web-component>
 
   <other-component>
     <!-- ✗ BAD -->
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: docs/rules/no-deprecated-slot-attribute.md [hunk 1]
```diff
@@ -81,6 +83,37 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
...
+    </my-component>
+  </other-component>
+</template>
+```
+
+</eslint-code-block>
+
 ## :books: Further Reading
 
 - [API - slot](https://v2.vuejs.org/v2/api/#slot-deprecated)
```
ST-DDT, line 119: Should we add a link to native web-components here as well?

e.g.

- https://developer.mozilla.org/en-US/docs/Web/API/Web_components
- https://developer.mozilla.org/en-US/docs/Web/API/Element/slot

docs/rules/no-deprecated-slot-attribute.md
```diff
@@ -43,16 +43,18 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
...
 }
 ```
 
 - `"ignore"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) that ignore these rules. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
+- `"ignoreParents"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) for parents that ignore these rules. This option is especially useful for Web-Components. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
 
 ### `"ignore": ["my-component"]`
 
-<eslint-code-block fix :rules="{'vue/no-dupe-keys': ['error', {ignore: ['my-component']}]}">
+<eslint-code-block fix :rules="{'vue/no-deprecated-slot-attribute': ['error', {ignore: ['my-component']}]}">
```
ST-DDT, line 57: I also noticed that the rule config was wrong here, if I should move it to a separate PR please let me know.

.changeset/beige-teams-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Added `ignoreParents` option to `no-deprecated-slot-attribute`
```
FloEdelmann:
```suggestion
Added `ignoreParents` option to [`vue/no-deprecated-slot-attribute`](https://eslint.vuejs.org/rules/no-deprecated-slot-attribute.html)
```

docs/rules/no-deprecated-slot-attribute.md
FloEdelmann: Oh, good catch! It's fine in this PR.

docs/rules/no-deprecated-slot-attribute.md
FloEdelmann: Yeah, sounds like a good idea to me. Maybe link to the `slot` documentation here, and link from the `ignoreParents` description to the general Web components documentation.

docs/rules/no-deprecated-slot-attribute.md
```diff
@@ -43,16 +43,18 @@ This rule reports deprecated `slot` attribute in Vue.js v2.6.0+.
...
   "vue/no-deprecated-slot-attribute": ["error", {
-    "ignore": ["my-component"]
+    "ignore": ["my-component"],
+    "ignoreParents": ["my-web-component"],
   }]
 }
 ```
 
 - `"ignore"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) that ignore these rules. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
+- `"ignoreParents"` (`string[]`) An array of tags or regular expression patterns (e.g. `/^custom-/`) for parents that ignore these rules. This option is especially useful for Web-Components. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
```
FloEdelmann: I think web components can only be kebab-case, so should we really automatically check PascalCase-ified versions, too?

ST-DDT: This might be relevant for other things other than web components. Web components is just my usecase.

To be honest I wouldnt have added either kebab or pascal case (for ignore and ignoreParents) but I wanted to keep it consistent with the existing ignore option.

FloEdelmann: It might, but I think this is pretty theoretical. This rule is reporting a _deprecated_ attribute after all. The already existing `ignore` option is there for people using `slot` as a prop or attribute name in certain components, so there catching both kebab-case and PascalCase notations with a single option makes sense. But I can't think of a reason (apart from web components) where you might ignore the `slot` attribute if a _parent_ (Vue) component has a certain name.

ST-DDT: So should we use only the given patterns and ignore/remove the explicit case conversions?
If web-components are the only expected value, then the user should only define patterns matching the web-component naming convention.

FloEdelmann: Yes, I think that would be better.

ST-DDT: Done

FloEdelmann (approved):
Looks good to me now. Thank you! :slightly_smiling_face:

tests/lib/rules/no-deprecated-slot-attribute.js
```diff
@@ -732,6 +756,90 @@ tester.run('no-deprecated-slot-attribute', rule, {
...
+      <template>
+        <my-component>
+          <one slot="one">
+            A
+          </one>
+        </my-component>
+        <my-component2>
+          <template v-slot:two>\n<two >
+            B
+          </two>\n</template>
```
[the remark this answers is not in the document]
ST-DDT: This is reused from above examples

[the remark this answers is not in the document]
FloEdelmann: Nah, let's keep it as is.

waynzh (approved):
LGTM! Thank you!

Issue #2785, opened 2025-07-10
vue/require-explicit-slots does not work when type reference is imported


Waxoussou:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 8.57 
- **eslint-plugin-vue version:** 10.0.0 
- **Vue version:** ^3.5 
- **Node version:** 18.20 
- **Operating System:** Ubuntu 24.10 

**Please show your full configuration:**

```json5
// baseConfig
{
    root: true,
    env: {
        browser: true,
        es6: true,
        jest: true,
        node: true
    },
    extends: ["eslint:recommended", "plugin:vue/recommended", "@vue/prettier"],
    parser: "vue-eslint-parser",
    parserOptions: {
        parser: "@typescript-eslint/parser",
        ecmaVersion: 2020,
        sourceType: "module",
        ecmaFeatures: {
            jsx: true
        }
    },
    ignorePatterns: ["**/target/", "**/generated/"],
    overrides: [
        {
            files: ["**/*.ts", "**/*.tsx"],
            extends: [
                "eslint:recommended",
                "plugin:@typescript-eslint/eslint-recommended",
                "plugin:@typescript-eslint/recommended",
                "plugin:vue/recommended",
                "@vue/prettier"
            ]
        }
    ],
    rules: {
        "@typescript-eslint/no-unused-vars": ["warn"],
        "vue/multi-word-component-names": [
            "error",
            {
                ignores: []
            }
        ],
        "vue/require-explicit-slots": ["warn"]
    }
}

```

**What did you do?**	

```vue
<script setup lang="ts">
import type {  BmButtonSlots } from "./button.type";
// ...
const slots = defineSlots<BmButtonSlots>();
</script>

<template>
    <BmButton>
        <slot v-else name="icon">
            <BmIcon v-if="icon" :icon="icon" />
        </slot>
        <span class="label-wrapper" :class="{ 'sr-only': isIconVariant }">
            <slot>
                {{ label }}
            </slot>
        </span>
    </BmButton>
</template>
```

**What did you expect to happen?**
should not warn about require-explicit-slot

**What actually happened?**

```raw 
Slots must be explicitly defined.eslint[vue/require-explicit-slots](https://eslint.vuejs.org/rules/require-explicit-slots.html)
```

[...]

PR #2786, opened 2025-07-11
test(require-explicit-slots): import slot type

waynzh:
Related: #2785


FloEdelmann (approved):
Good idea to add a test case! Thank you! :slightly_smiling_face:

Issue #2769, opened 2025-06-25
Rule suggestion: `vue/no-negated-condition` for `v-if`/`v-else` in the template


FloEdelmann:
**Please describe what the rule should do:**
Similar to the core ESLint rule [no-negated-condition](https://eslint.org/docs/latest/rules/no-negated-condition) (and its autofixable variant [unicorn/no-negated-condition](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-negated-condition.md)), this rule would report negated conditions in the template (only if both `if` and `else` cases are handled). Avoiding unnecessary negated conditions makes the code a bit easier to read.

**What category should the rule belong to?**
[ ] Enforces code style (layout)
[ ] Warns about a potential error (problem)
[x] Suggests an alternate way of doing something (suggestion)
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule should warn about:**

```vue
<template>
  <div>
    <!-- BAD -->
    <div v-if="!foo" />
    <span v-else />

    <!-- GOOD -->
    <span v-if="foo" />
    <div v-else />
  </div>
</template>
```

**Additional context**
Should it also report ternaries with negated conditions used in the template?

```vue
<template>
  <div>
    <!-- BAD -->
    <div :class="!foo ? 'bar' : 'baz'" />

    <!-- GOOD -->
    <div :class="foo ? 'baz' : 'bar'" />
  </div>
</template>
```

Or should that be a different rule? Then the names could be `vue/no-negated-condition` for the [no-negated-condition](https://eslint.org/docs/latest/rules/no-negated-condition) extension rule that applies to ternaries, and `vue/no-negated-v-if-condition` for the rule proposed above.

PR #2794, opened 2025-07-14
Add `vue/no-negated-v-if-condition` rule

waynzh:
related #2769

Found that the check and fix logic for ternaries is quite different. So, I extract `v-if` condition into a separate rule.

I’ll submit a follow-up PR later to implement `vue/no-negated-condition` as an extension of the `no-negated-condition`.

Update: https://github.com/vuejs/eslint-plugin-vue/pull/2795


docs/rules/no-negated-v-if-condition.md
```diff
@@ -0,0 +1,63 @@
...
+</template>
+```
+
+</eslint-code-block>
+
+## :wrench: Options
+
+Nothing.
+
+## :couple: Related Rules
```
FloEdelmann, line 55: I merged https://github.com/vuejs/eslint-plugin-vue/pull/2795 first. Could you link to it here (and vice-versa) please?

.changeset/early-worlds-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Add new `vue/no-negated-v-if-condition` rule
```
FloEdelmann:
```suggestion
Added new [`vue/no-negated-v-if-condition`](https://eslint.vuejs.org/rules/no-negated-v-if-condition.html) rule
```

docs/rules/no-negated-v-if-condition.md
waynzh: Thank you!  Link updated ✅

waynzh:
Before merging, should we consider adding support for `v-show` in this rule as well?

docs/rules/no-negated-v-if-condition.md
```diff
@@ -0,0 +1,64 @@
+---
+pageClass: rule-details
+sidebarDepth: 0
+title: vue/no-negated-v-if-condition
+description: disallow negated conditions in v-if/v-else
+---
+
+# vue/no-negated-v-if-condition
```
waynzh, line 8: ~~Should we consider adding support for `v-show` in this rule as well?~~

FloEdelmann: https://vuejs.org/guide/essentials/conditional.html#v-show:

> `v-show` doesn't support the `<template>` element, nor does it work with `v-else`.

If it doesn't work with `v-else`, how would you rewrite `<foo v-show="!bar" />`?

waynzh: haha, I had a brain fade🥶

FloEdelmann: No worries, that's what reviews are for :relaxed: :sweat_smile:

lib/rules/no-negated-v-if-condition.js
```diff
@@ -0,0 +1,203 @@
...
+      }
+    }
+
+    /**
+     * @param {VElement} element
+     * @returns {string}
+     */
+    function getElementContent(element) {
+      if (element.children.length === 0) return ''
+      if (!element.endTag) return ''
```
FloEdelmann:
```suggestion
      if (element.children.length === 0 || !element.endTag) {
        return ''
      }
```

FloEdelmann (approved):
Really cool, thank you! I have just one minor code style nitpick, the rest looks good to me 🙂

lib/rules/no-negated-v-if-condition.js
waynzh: Thanks for the review!

ota-meshi (approved):
(Sorry for the late reply. I've been very busy lately 😓)
LGTM! Thank you!

PR #2795, opened 2025-07-14
Add new `vue/no-negated-condition` rule

### LINKED ISSUE #2769: Rule suggestion: `vue/no-negated-condition` for `v-if`/`v-else` in the template

waynzh:
Resolve #2769
Related PR https://github.com/vuejs/eslint-plugin-vue/pull/2794

### DIFF
--- a/.changeset/true-pumas-open.md
+++ b/.changeset/true-pumas-open.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': minor
 ---
 
-Add new `vue/no-negated-condition` rule
+Added new [`vue/no-negated-condition`](https://eslint.vuejs.org/rules/no-negated-condition.html) rule
--- a/.changeset/true-pumas-open.md
+++ b/.changeset/true-pumas-open.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Add new `vue/no-negated-condition` rule
--- a/docs/rules/no-negated-condition.md
+++ b/docs/rules/no-negated-condition.md
@@ -15,6 +15,10 @@ description: Disallow negated conditions in `<template>`
 
 This rule is the same rule as core [no-negated-condition] rule but it applies to the expressions in `<template>`.
 
+## :couple: Related Rules
+
+- [unicorn/no-negated-condition](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-negated-condition.md)
+
 ## :books: Further Reading
 
 - [no-negated-condition]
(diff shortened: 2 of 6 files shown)


.changeset/true-pumas-open.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Add new `vue/no-negated-condition` rule
```
FloEdelmann: The link doesn't work yet, but will do as soon as the changelog is released:

```suggestion
Added new [`vue/no-negated-condition`](https://eslint.vuejs.org/rules/no-negated-condition.html) rule
```

FloEdelmann (approved):
Thank you! 🙂

docs/rules/no-negated-condition.md
```diff
@@ -0,0 +1,29 @@
...
+
+## :book: Rule Details
+
+This rule is the same rule as core [no-negated-condition] rule but it applies to the expressions in `<template>`.
+
+## :books: Further Reading
+
+- [no-negated-condition]
+
+[no-negated-condition]: https://eslint.org/docs/rules/no-negated-condition
```
FloEdelmann:
```suggestion
## :book: Rule Details

This rule is the same rule as core [no-negated-condition] rule but it applies to the expressions in `<template>`.

## :couple: Related Rules

- [unicorn/no-negated-condition](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-negated-condition.md)

## :books: Further Reading

- [no-negated-condition]

[no-negated-condition]: https://eslint.org/docs/rules/no-negated-condition
```

Issue #1951, opened 2022-08-25
Deprecate shareable configs


ota-meshi:
**The problem you want to solve.**

Most of the shareable configs provided by this plugin are based on style guide.
However, the link to Vue's style guide seems to have been removed.

https://github.com/vuejs/docs/blob/74449a840c88ee9488e2337081302a993e14e1f6/.vitepress/config.ts#L18
https://github.com/vuejs/docs/commit/4acd127dc4975df302b52ee3313e6743d50a5c66

So, I suspect that these configs are based on deprecated documentation.

**Your take on the correct solution to problem.**

Deprecate shareable configurations other than `plugin:vue/base`.
Provide shareable configurations based on the new standards, if needed.

**Additional context**

FloEdelmann:
I'm not sure if that's a good idea. I think most people use the `recommended` config and they would need to enable dozens of rules manually, which is quite a bit of work and bloats up the ESLint config.

Maybe we can instead create new shareable configs based on other criteria? E.g. "error prevention" and "layout" rules?

Or just leave it as it is, and curate the configs ourselves based on good judgement?

sodatea:
https://vuejs.org/style-guide/
They are not complete and not up-to-date, but I think they are still quite useful, especially the `essential` ones.

Maybe we can start a project to discuss the new guidance?
And before it's finished, put a small warning on the ESLint plugin documentation page, like what the official style guide does now?

ota-meshi:
Thank you for your comment.

> I think most people use the recommended config and they would need to enable dozens of rules manually, which is quite a bit of work and bloats up the ESLint config.
> Maybe we can instead create new shareable configs based on other criteria? E.g. "error prevention" and "layout" rules?

If we deprecate shareable configs provided by this plugin, I would publish a new shareable config as a personal project. It's probably two configurations like "error prevention", "layout".
However these are configured according to my preferences. Not official. Because I think the official recommended style is now deprecated.

> Maybe we can start a project to discuss the new guidance?

I think it's good to discuss the new(?) style guide. But I'm not sure of the right place to discuss style guides. Additionally, I personally prefer the Options API. Therefore, I am a little worried about whether I can properly participate in the discussion :sweat_smile: 

If we create a new style guide, the following PR might be a good starting point.
https://github.com/vuejs/v2.vuejs.org/pull/2839

[...]

PR #2796, opened 2025-07-15
feat: support having all rules in config report at `error` instead of `warn` using env variable

### LINKED ISSUE #2774: Allow having all rules error instead of warn

- **ESLint version:** 9.16.0
- **eslint-plugin-vue version:** 9.32.0
- **Vue version:** 3.x
- **Node version:** 22.14.0

Currently some of the configs set rules to `warn` rather than `error`, which we've found easily leads to changes we'd want not being applied as warnings don't get surfaced that well - IDEs tend to use a more subtle yellow highlighting rather than angry red, and by default ESLint won't return a non-zero exit code meaning that CIs will be green.

While we can use `--max-warnings 0` to have CI fail if there are any warnings, that cannot be configured as part of config to ensure all linting runs apply that and it also doesn't improve the IDE DX.

Allow having the configs use `error` instead of `warn` - since flat config is the future and means plugins and configs are always being used in a js runtime situation, that could be done using a function or an env variable e.g.

```
const pluginVue = require('eslint-plugin-vue')

// we could do this...
process.env.ESLINT_VUE_ALWAYS_ERROR = 'true';

module.exports = [
  // ...or this
  ...pluginVue.configs['flat/recommended'](true),
]
```

Personally right now my vote would be for the env variable as I think that is simpler, but wanted to highlight flat config does give more flexibility in future

**Additional context**

I'm happy to do a PR for this 🙂

G-Rath:
I'm happy to take alternative names for the environment and local config variables

Resolves #2774

### DIFF
--- a/docs/user-guide/index.md
+++ b/docs/user-guide/index.md
@@ -73,6 +73,8 @@ You can use the following configs by adding them to `eslint.config.js`.
 
 :::warning Reporting rules
 By default, all rules from **base** and **essential** categories report ESLint errors. Other rules - because they're not covering potential bugs in the application - report warnings. What does it mean? By default - nothing, but if you want - you can set up a threshold and break the build after a certain amount of warnings, instead of any. More information [here](https://eslint.org/docs/user-guide/command-line-interface#handling-warnings).
+
+Alternatively, you can set `process.env.VUE_ESLINT_ALWAYS_ERROR` to `true` in your configuration file to have `error` be used by all rules.
 :::
 
 #### Specifying Globals (`eslint.config.js`)
--- a/tools/update-lib-configs.js
+++ b/tools/update-lib-configs.js
@@ -44,7 +44,9 @@ function formatRules(rules, categoryId) {
       return [rule.ruleId, options]
     })
   )
-  return JSON.stringify(obj, null, 2)
+  // use the ruleLevel variable for rules set to warn so that they can
+  // be made to error with an env variable if desired
+  return JSON.stringify(obj, null, 2).replaceAll('"warn"', 'ruleLevel')
 }
 
 function formatCategory(category) {
@@ -55,6 +57,8 @@ function formatCategory(category) {
  * This file has been automatically generated,
  * in order to update its content execute "npm run update"
  */
+const ruleLevel = process.env.VUE_ESLINT_ALWAYS_ERROR === 'true' ? 'error' : 'warn'
+
 module.exports = {
   parserOptions: {
     ecmaVersion: 'latest',
@@ -78,6 +82,8 @@ module.exports = {
  * This file has been automatically generated,
  * in order to update its content execute "npm run update"
  */
+const ruleLevel = process.env.VUE_ESLINT_ALWAYS_ERROR === 'true' ? 'error' : 'warn'
+
 module.exports = {
   extends: require.resolve('./${extendsCategoryId}'),
   rules: ${formatRules(category.rules, category.categoryId)}
--- a/tools/update-lib-configs.js
+++ b/tools/update-lib-configs.js
@@ -51,13 +51,18 @@ function formatRules(rules, categoryId) {
 
 function formatCategory(category) {
   const extendsCategoryId = extendsCategories[category.categoryId]
+  const formattedRules = formatRules(category.rules, category.categoryId)
+  const ruleLevelVariable = formattedRules.includes('ruleLevel')
+    ? "const ruleLevel = process.env.VUE_ESLINT_ALWAYS_ERROR === 'true' ? 'error' : 'warn'"
+    : ''
+
   if (extendsCategoryId == null) {
     return `/*
  * IMPORTANT!
  * This file has been automatically generated,
  * in order to update its content execute "npm run update"
  */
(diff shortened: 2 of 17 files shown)


G-Rath:
@waynzh would it be possible to get a review or even just a yay/nay on if you're open to this?

FloEdelmann:
I'm unsure whether an environment variable is the correct approach to this. Maybe it would be better to duplicate the configs with an `-error` variant for each one?

G-Rath:
I'm personally not fussed so if you're happy having double the amount of configs I'm fine to do it that way too - I just figured the env variable was a nice way of leveraging the fact that going forward you have to use a JS file for the ESLint config (meaning its always possible to ensure that the property on `process.env` is always set when ESLint is running) without doing a breaking change

FloEdelmann:
@ota-meshi @waynzh What do you think?

ota-meshi:
I've been thinking about this for a few days, and my personal opinion is that I'm against adding new shareable configs or making major changes to them.
It's impossible to satisfy everyone's preferences, so I think it's better for users to create their own configs.
I think this is also related to https://github.com/vuejs/eslint-plugin-vue/issues/1951#issuecomment-1660237710.
However, if @FloEdelmann and @waynzh both think this config change is a good idea, there's no problem with accepting it.

> @ota-meshi @waynzh What do you think?

I'm not a fan of using environment variables.

waynzh:
For the IDE DX issue, you can address it by configuring the IDE-ESLint integration like this:

```json
"eslint.rules.customizations": [
  { "rule": "*", "severity": "error" }
  // { "rule": "vue/*", "severity": "error" }
]
```

I'm not a big fan of controlling this via environment variables either. 
One thing worth noting is that there's already a (though not very active) discussion in the ESLint core (https://github.com/eslint/eslint/issues/14679) about adding `severity` settings into ESLint config. If implemented, this would allow severity to be configured through a single field, removing the need for each plugin to provide separate sharable configs for different severity levels.

G-Rath:
> For the IDE DX issue, you can address it by configuring the IDE-ESLint integration like this:

That only works for VSCode

> I'm not a big fan of controlling this via environment variables either.

Sounds like folks are against that, so I'll switch to generating new configs

G-Rath:
@FloEdelmann @waynzh @ota-meshi I've updated the implementation to generate dedicated configs instead of using an env variable

docs/user-guide/index.md
FloEdelmann, line 187: Please make the same adjustments to the "Bundle Configurations (eslint.config.js)" section.

tools/update-lib-configs.js
```diff
@@ -47,8 +48,16 @@ function formatRules(rules, categoryId) {
...
 
-function formatCategory(category) {
-  const extendsCategoryId = extendsCategories[category.categoryId]
+function hasWarningRules(categoryId) {
+  return (
+    categoryId !== 'base' &&
+    categoryId !== 'vue3-essential' &&
+    categoryId !== 'vue2-essential'
+  )
+}
```
FloEdelmann: This seems brittle; can we somehow compute this instead of relying on a hardcoded list? (same in `tools/update-lib-flat-configs.js`)

.changeset/brown-eels-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Use `error` instead of `warn` in configs when `VUE_ESLINT_ALWAYS_ERROR` env variable is `"true"`
```
FloEdelmann:
```suggestion
Add [new configs](https://eslint.vuejs.org/user-guide/#bundle-configurations-eslint-config-js) `strongly-recommended-error` and `recommended-error` with all rules' severity set to `error`
```

tools/update-lib-configs.js
G-Rath: I've refactored this to use `errorCategories` since somehow I missed that already exists 🤦

I think it might technically be possible to avoid needing a hardcoded list for this new code, but I'm not sure if that would be any less brittle - either way though, at least now we only have one list 😅

.changeset/brown-eels-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Add [new configs](https://eslint.vuejs.org/user-guide/#bundle-configurations-eslint-config-js) `strongly-recommended-error` and `recommended-error` with all rules' severity set to `error`
```
FloEdelmann: Another small suggestion to keep the changelog entries' wording consistent:

```suggestion
Added [new configs](https://eslint.vuejs.org/user-guide/#bundle-configurations-eslint-config-js) `strongly-recommended-error` and `recommended-error` with all rules' severity set to `error`
```

FloEdelmann (approved):
Thanks, looks good to me now 🙂

FloEdelmann:
@G-Rath can you please have a look at Copilot's suggestions? Some of them seem to make sense. Feel free to just comment on and resolve the invalid ones.

G-Rath:
@FloEdelmann should be all good to go now

PR #2797, opened 2025-07-17
feat(component-api-style): add scriptSetupVapor option

### LINKED ISSUE #2791: Rule proposal: vue/prefer-vapor-mode
This rule should suggest using the new `vapor` attribute on `<script setup>` blocks in Vue 3.6+.

Vapor Mode is a new feature introduced in Vue 3.6 (currently in alpha), and this rule helps developers start adopting it early by recommending `vapor` where it's appropriate. The behavior of components stays the same, but performance can be improved.

The rule should detect components using `<script setup>` and suggest adding the `vapor` attribute when possible.

[ ] Other (please specify:)

```vue
<!-- ❌ This component could use vapor mode, but doesn't -->
<script setup>
defineProps(['msg'])
</script>

<template>
  <div>{{ msg }}</div>
</template>
```
```vue
<!-- ✅ This component uses vapor mode correctly -->
<script setup vapor>
defineProps(['msg'])
</script>

<template>
  <div>{{ msg }}</div>
</template>
```

**Additional context**

[Vapor Mode](https://github.com/vuejs/core/releases/tag/v3.6.0-alpha.1) was introduced in Vue 3.6, which is currently in alpha.

This rule will become more important once Vue 3.6 is officially released, helping developers adopt the new mode and benefit from improved performance in eligible components.

waynzh:
resolve #2791 

Waiting for the official release of Vue 3.6...

### DIFF
--- a/.changeset/strong-masks-fetch.md
+++ b/.changeset/strong-masks-fetch.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+Fixed inconsistent quotes in [`vue/block-lang`](https://eslint.vuejs.org/rules/block-lang.html) error messages
--- a/lib/rules/block-lang.js
+++ b/lib/rules/block-lang.js
@@ -40,7 +40,7 @@ const DEFAULT_LANGUAGES = {
  * @param {NonNullable<BlockOptions['lang']>} lang
  */
 function getAllowsLangPhrase(lang) {
-  const langs = [...lang].map((s) => `"${s}"`)
+  const langs = [...lang].map((s) => `'${s}'`)
   switch (langs.length) {
     case 1: {
       return langs[0]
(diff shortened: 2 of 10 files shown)


ota-meshi (approved):
LGTM! Thanks!

FloEdelmann (commented):
The code looks good to me 🙂 

Unfortunately, this new option will not solve the request in #2791 ("detect components using `<script setup>` and suggest adding the `vapor` attribute when possible"). Should this be a new rule instead? Or a new additional rule, while first only allowing `script-setup-vapor` here?

waynzh:
> this new option will not solve the request in #2791

Users setting `['script-setup-vapor']` like [this](https://github.com/vuejs/eslint-plugin-vue/pull/2797/files#diff-ec5f01dbaecce463aaba312674adfd96ec927c721c77eb0db6ba00af7d9492b0R891) should resolve the issue, right?
The error message will suggest: "Use `<script setup vapor>` instead."

FloEdelmann:
True, that might work. But you have to migrate everything at once to `<script setup vapor>`, or decrease the rule severity to `warn` (but then, also other unwanted styles might creep back in).

ota-meshi:
To implement that, I think we need to check whether it's safe to assign the `vapor` attribute. I don't know how to do that 🤔 
Do you know how to check that?

PR #2800, opened 2025-07-18
test(array-bracket-spacing): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `array-bracket-spacing` to include both error message and full location checks.


ST-DDT:
Do you prefer these small PRs or would you prefer larger PRs?
How many test files should I update in one PR?

- 1:1
- 5:1
- One PR for each starting letter
- One PR for each starting word
- All in one PR
- Other?

I myself prefer small PRs as they have a lower chance of merge conflicts.
But I can understand, if this would cause issues on your side.

FloEdelmann (approved):
Thanks! These small PRs are fine for me, but larger ones (like all rules starting with the same letter) would be fine, too.

PR #2802, opened 2025-07-18
test(attribute-hyphenation): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `attribute-hyphenation` to include both error message and full location checks.


FloEdelmann (approved):
Thanks!

PR #2805, opened 2025-07-18
fix(block-lang): use consistent quotation marks in error messages

ST-DDT:
Follow-Up on #2804

- #2804

---

While updating the tests, I noticed that the error messages use `"`, `'` and <code>`</code> for quoting/highlighting things.
This PR changes this to only use <code>'</code>.

Do I need to create an issue for this one?

---

Blocked by #2804

### DIFF
--- a/.changeset/strong-masks-fetch.md
+++ b/.changeset/strong-masks-fetch.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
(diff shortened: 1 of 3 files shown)


FloEdelmann (approved):
Thank you! I'd say this does qualify as a `fix`, so thanks for adding a changeset! An issue is not required 🙂

waynzh (approved):
LGTM! Thank you!

PR #2806, opened 2025-07-18
test(block-order): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `block-order` to include both error message and full location checks.


waynzh (approved):
Thanks!

PR #2807, opened 2025-07-18
test(block-spacing): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `block-spacing` to include both error message and full location checks.


waynzh (approved):
Thank you!

PR #2814, opened 2025-07-20
test(component-api-style): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `component-api-style` to include both error message and full location checks.


FloEdelmann (approved):
Thanks!

PR #2815, opened 2025-07-20
test(component-definition-name-casing): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `component-definition-name-casing` to include both error message and full location checks.


FloEdelmann (approved):
Thanks!

PR #2819, opened 2025-07-20
test(define-emits-declaration): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `define-emits-declaration` to include both error message and full location checks.


waynzh (approved):
Thanks!

PR #2824, opened 2025-07-20
test(dot-location): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `dot-location` to include both error message and full location checks.


waynzh (approved):
LGTM! Thank you!

PR #2830, opened 2025-07-25
test(func-call-spacing): remove obsolete compatibility code

ST-DDT:
Continuation of #2793
- #2793
---
This PR converts all error assertions for `func-call-spacing` to include both error message and full location checks.

### DIFF
--- a/tests/lib/rules/func-call-spacing.js
+++ b/tests/lib/rules/func-call-spacing.js
@@ -3,8 +3,7 @@
  */
 'use strict'
 
-const { RuleTester, ESLint } = require('../../eslint-compat')
-const semver = require('semver')
+const { RuleTester } = require('../../eslint-compat')
 const rule = require('../../../lib/rules/func-call-spacing')
 
 const tester = new RuleTester({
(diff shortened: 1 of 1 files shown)


tests/lib/rules/func-call-spacing.js
```diff
@@ -61,10 +60,9 @@ tester.run('func-call-spacing', rule, {
       `,
       errors: [
         {
-          message: semver.lt(ESLint.version, '7.0.0')
```
FloEdelmann, line 64: Always good to see legacy code removed 🙂

FloEdelmann (dismissed):
Thank you!

tests/lib/rules/func-call-spacing.js
FloEdelmann, line 1: The tests are failing, could you have a look at those please?

ST-DDT: It looks like it only fails when eslint stylistic is missing.
Any idea why this might be the case?

If not, I can revert adding the `column` property.

waynzh: It might be a difference in how `@stylistic/func-call-spacing` and ESLint's `func-call-spacing` rule report locations.  

Additionally, the version of `@stylistic/eslint-plugin` in the project isn’t the latest, which could be causing this issue.

ST-DDT:
I removed the column property for now, so this is no longer blocked.
In the future, we should probably look into why this reports different locations depending on whether stylistic is present or not.

FloEdelmann (approved):
Thanks! I'll look into the differences. For reference: https://github.com/vuejs/eslint-plugin-vue/pull/2884

PR #2848, opened 2025-07-25
test(max-lines-per-block): make tests more strict

ST-DDT:
Continuation of #2793

- #2793

---

This PR converts all error assertions for `max-lines-per-block` to include both error message and full location checks.


waynzh (approved):
LGTM! Thanks!

PR #2857, opened 2025-07-29
feat: Add `ignorePattern` to `no-v-html`

lucaswerkmeister:
This allows configuring the rule such that certain variables are allowed in `v-html=`. For instance, we would like to allowlist variables named “*Html” (as in the test), where the name already marks them as containing safe HTML that’s expected to be used with `v-html=`.

----

I tried to follow existing code style in here, but still, this is my first time contributing to any eslint plugin (I think), so let me know if anything needs to be changed or improved :) the option name mimics e.g. `no-unused-vars` but could of course be changed. (Also, I’m not completely sure if it’s correct that the option value seems to require slashes around it in order to not be interpreted as a literal string.)


lucaswerkmeister:
> (Also, I’m not completely sure if it’s correct that the option value seems to require slashes around it in order to not be interpreted as a literal string.)

Huh, apparently `vue/no-unused-vars` just uses `new RegExp(ignorePattern, 'u')` rather than the `toRegExp` helper. (I found the `toRegExp` helper in some other rule, don’t remember which one.)

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,32 @@ module.exports = {
...
     },
     fixable: null,
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties: {
+        ignorePattern: {
+          type: 'string'
+        }
+      }
```
FloEdelmann, line 26:
```suggestion
      },
      additionalProperties: false
```

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,32 @@ module.exports = {
...
       unexpected: "'v-html' directive can lead to XSS attack."
     }
   },
   /** @param {RuleContext} context */
   create(context) {
+    const options = context.options[0]
+    let ignoredVarMatcher = null
+    if (options?.ignorePattern) {
+      ignoredVarMatcher = toRegExp(options.ignorePattern, { remove: 'g' })
+    }
```
FloEdelmann:
```suggestion
    const ignoredVarMatcher = options?.ignorePattern
      ? toRegExp(options.ignorePattern, { remove: 'g' })
      : undefined
```

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,32 @@ module.exports = {
...
+
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VDirective} node */
       "VAttribute[directive=true][key.name.name='html']"(node) {
+        if (ignoredVarMatcher !== null && node.value.expression.type === 'Identifier' && ignoredVarMatcher.test(node.value.expression.name)) {
```
FloEdelmann:
```suggestion
        if (ignoredVarMatcher && node.value.expression.type === 'Identifier' && ignoredVarMatcher.test(node.value.expression.name)) {
```

tests/lib/rules/no-v-html.js
```diff
@@ -28,6 +28,11 @@ ruleTester.run('no-v-html', rule, {
     {
       filename: 'test.vue',
       code: '<template><div v-if="foo" v-bind="bar"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-html="knownHtml"></div></template>',
+      options: [{ignorePattern: '/^(?:html|.+Html)$/'}]
```
FloEdelmann: Please run `npm run update` to fix formatting inconsistencies (like missing spaces between the curly brackets here.

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,32 @@ module.exports = {
       url: 'https://eslint.vuejs.org/rules/no-v-html.html'
     },
     fixable: null,
-    schema: [],
+    schema: [{
+      type: 'object',
+      properties: {
+        ignorePattern: {
```
FloEdelmann: Please document this option in the rule docs.

FloEdelmann (changes requested):
Thanks for your contribution! I have some suggestions / requested changes.

tests/lib/rules/no-v-html.js
lucaswerkmeister: Thanks, I thought I was in the clear when `npm t` didn’t complain…

lib/rules/no-v-html.js
lucaswerkmeister: Yup, I totally forgot about this.

lucaswerkmeister: Added, though it could perhaps be expanded more. (Should we include an example?)

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,32 @@ module.exports = {
...
+          type: 'string'
+        }
+      },
+      additionalProperties: false
+    }],
     messages: {
       unexpected: "'v-html' directive can lead to XSS attack."
```
lucaswerkmeister, line 29: I wonder if a separate message would make sense if `ignorePattern` is defined… what do you think?

docs/rules/no-v-html.md
```diff
@@ -32,7 +32,15 @@ This rule reports all uses of `v-html` directive in order to reduce the risk of
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+    "vue/no-v-html": ["error", {
+        "ignorePattern": "/Html$"
```
FloEdelmann:
```suggestion
        "ignorePattern": "/Html$/"
```

docs/rules/no-v-html.md
```diff
@@ -32,7 +32,15 @@ This rule reports all uses of `v-html` directive in order to reduce the risk of
...
+    }]
+}
+```
+
+- `ignorePattern` ... disables reporting when the v-html directive references a variable matching this pattern. By default, all v-html uses are forbidden.
```
FloEdelmann:
```suggestion
- `ignorePattern` ... disables reporting when the `v-html` directive references a variable matching this pattern. By default, all `v-html` uses are forbidden.
```

lib/rules/no-v-html.js
FloEdelmann: I don't think that is necessary.

lib/rules/no-v-html.js
FloEdelmann: I think an example would indeed be nice, because that would make clear what sort of expressions the pattern matches against.

lib/rules/no-v-html.js
```diff
@@ -14,16 +15,38 @@ module.exports = {
...
+    ],
     messages: {
       unexpected: "'v-html' directive can lead to XSS attack."
     }
   },
   /** @param {RuleContext} context */
   create(context) {
+    const options = context.options[0]
+    const ignoredVarMatcher = options?.ignorePattern
+      ? toRegExp(options.ignorePattern, { remove: 'g' })
```
FloEdelmann: > Huh, apparently `vue/no-unused-vars` just uses `new RegExp(ignorePattern, 'u')` rather than the `toRegExp` helper. (I found the `toRegExp` helper in some other rule, don’t remember which one.)

I think `new RegExp(ignorePattern, 'u')` is indeed more suitable here. `toRegExp` is used in multiple rules that have an `ignore` option that allows both strings (that are compared as is) and strings containing patterns (e.g. `"/foo/"`).

Here, it is intentionally only one string option called "ignore***Pattern***", so the leading and trailing slashes are not necessary I think.

docs/rules/no-v-html.md
FloEdelmann, line 1: While you're at it, would you fancy adding a `## :couple: Related Rules` section linking to https://eslint.vuejs.org/rules/no-v-text.html and https://eslint.vuejs.org/rules/no-v-text-v-html-on-component.html, and linking from those pages to this one and each other, too?

I just noticed they are missing and this could be done in this PR, but it's fine to do it separately, too 🙂

lib/rules/no-v-html.js
lucaswerkmeister: Ok :)

docs/rules/no-v-html.md
lucaswerkmeister: Oops (but also, becomes obsolete if we stop using `toRegExp()` as you pointed out below).

lib/rules/no-v-html.js
lucaswerkmeister: Makes sense, will change.

docs/rules/no-v-html.md
lucaswerkmeister: Done :)

lib/rules/no-v-html.js
lucaswerkmeister: Added.

lucaswerkmeister:
(I’ve been adding commits under the assumption that you’ll squash them on merge and throw away the intermediate commit messages – let me know if you’d like the Git history cleaned up into a version that would be suitable for a merge commit ^^)

tests/lib/rules/no-v-html.js
```diff
@@ -45,6 +50,12 @@ ruleTester.run('no-v-html', rule, {
       filename: 'test.vue',
       code: '<template><section v-html/></template>',
       errors: ["'v-html' directive can lead to XSS attack."]
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-html="unsafeString"></div></template>',
+      options: [{ ignorePattern: '.Html$' }],
+      errors: ["'v-html' directive can lead to XSS attack."]
```
FloEdelmann: While you're at it, can you please change these tests to check the exact error location, similar to <https://github.com/vuejs/eslint-plugin-vue/pull/2878>?

docs/rules/no-v-html.md
```diff
@@ -32,12 +32,42 @@ This rule reports all uses of `v-html` directive in order to reduce the risk of
 
 ## :wrench: Options
 
-Nothing.
+```json
+{
+    "vue/no-v-html": ["error", {
+        "ignorePattern": ".Html$"
```
FloEdelmann: I think the `.` is not necessary, since the pattern is not matched against the whole string. Please remove it everywhere in the docs and tests.

lucaswerkmeister: The idea was that there should be something before the `Html`, so that the full variable name is in lowerCamelCase. (Earlier I had a more complicated regex that also allowed `^html$`, but decided that was a pointless overcomplication.)

How about we make the examples match a *prefix* `html` instead, i.e. `^html`? It doesn’t really matter one way or the other, but the regex is simpler that way IMHO :)

tests/lib/rules/no-v-html.js
lucaswerkmeister: Sure, done.

docs/rules/no-v-html.md
FloEdelmann: > How about we make the examples match a _prefix_ `html` instead, i.e. `^html`?

Sounds good to me :slightly_smiling_face:

lucaswerkmeister: Ok, done :)

lib/rules/no-v-html.js
```diff
@@ -14,16 +14,38 @@ module.exports = {
...
+    const options = context.options[0]
+    const ignoreRegEx = options?.ignorePattern
+      ? new RegExp(options.ignorePattern, 'u')
+      : undefined
+
     return utils.defineTemplateBodyVisitor(context, {
       /** @param {VDirective} node */
       "VAttribute[directive=true][key.name.name='html']"(node) {
+        if (
+          ignoreRegEx &&
```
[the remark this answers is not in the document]
FloEdelmann: I think this is a reasonable assumption.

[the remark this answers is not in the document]
waynzh: This check is necessary. I’ll commit the suggestion.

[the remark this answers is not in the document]
FloEdelmann: Oh, you're right! Thanks for double-checking!

waynzh (approved):
LGTM! Thanks you!

PR #2870, opened 2025-07-30
test(no-deprecated-dollar-listeners-api): make tests more strict

ST-DDT:
Continuation of #2793
- #2793
---
This PR converts all error assertions for `no-deprecated-dollar-listeners-api` to include both error message and full location checks.


FloEdelmann (approved):
Thanks! While at it, I sorted the objects to put `messageId` before `line` everywhere.

PR #2876, opened 2025-07-30
test(no-deprecated-html-element-is): make tests more strict

ST-DDT:
Continuation of #2793
- #2793
---
This PR converts all error assertions for `no-deprecated-html-element-is` to include both error message and full location checks.


FloEdelmann (approved):
Thanks! While at it, I sorted the objects to put `messageId` before `line` everywhere.

PR #2901, opened 2025-07-31
test(no-multiple-objects-in-class): make tests more strict

ST-DDT:
Continuation of #2793
- #2793
---
This PR converts all error assertions for `no-multiple-objects-in-class` to include both error message and full location checks.


waynzh (approved):
Thanks!

PR #2909, opened 2025-07-31
test(no-side-effects-in-computed-properties): make tests more strict

ST-DDT:
Continuation of #2793
- #2793
---
This PR converts all error assertions for `no-side-effects-in-computed-properties` to include both error message and full location checks.


waynzh (approved):
Thanks!

PR #2913, opened 2025-08-03
chore: introduce `@eslint/markdown` and `eslint-plugin-markdown-preferences`

ota-meshi:
This PR introduces `@eslint/markdown` and `eslint-plugin-markdown-preferences` to the repository.

We prefer to have links in rule names in release notes and changelog, and this PR will allow the `markdown-preferences/prefer-linked-words` rule to automatically detect and fix that.

https://ota-meshi.github.io/eslint-plugin-markdown-preferences/rules/prefer-linked-words.html

After merging this PR, I believe the `changeset:version` script should automatically include the link without having to include it in the changeset file.


eslint.config.mjs
```diff
@@ -1,24 +1,58 @@
+import { defineConfig } from 'eslint/config'
 import globals from 'globals'
 import eslintPluginEslintPlugin from 'eslint-plugin-eslint-plugin/configs/all'
 import eslintPluginJsonc from 'eslint-plugin-jsonc'
 import eslintPluginNodeDependencies from 'eslint-plugin-node-dependencies'
 import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended'
 import eslintPluginUnicorn from 'eslint-plugin-unicorn'
+import eslintMarkdown from '@eslint/markdown'
+import markdownPreferences from 'eslint-plugin-markdown-preferences'
```
FloEdelmann: Please rename this variable to `eslintPluginMarkdownPreferences` for consistency, even if it's long.

eslint.config.mjs
```diff
@@ -1,24 +1,58 @@
...
+  ...MD_BASE_LINKS
+}
+
 export default typegen([
   {
     ignores: [
       '.nyc_output',
       'coverage',
       'node_modules',
+      '.changeset/**/*.md',
```
FloEdelmann, line 55: Why are those files ignored completely?

lib/rules/no-setup-props-reactivity-loss.js
```diff
@@ -139,15 +139,26 @@ module.exports = {
...
+            expression.consequent,
+            expression.alternate
+          )
+          break
+        }
+        case 'Identifier': {
+          expressions.push(expression)
+          break
+        }
+      }
```
FloEdelmann, line 161: This change is unrelated, right? It looks like a good change, so no need to revert it or anything, I just stumbled a bit over it.

docs/rules/multiline-html-element-content-newline.md
```diff
@@ -146,7 +146,9 @@ This rule enforces a line break before and after the contents of a multiline ele
 ## :couple: Related Rules
 
 - [vue/singleline-html-element-content-newline](./singleline-html-element-content-newline.md)
-- [no-multiple-empty-lines](https://eslint.org/docs/rules/no-multiple-empty-lines)
+- [no-multiple-empty-lines]
+
+[no-multiple-empty-lines]: https://eslint.org/docs/rules/no-multiple-empty-lines
```
FloEdelmann, line 151: Why extract one of them but not both? I think extracting only really makes sense when the link is reused multiple times, so here, I'd keep it inlined.

docs/rules/no-async-in-computed-properties.md
```diff
@@ -112,7 +112,9 @@ Nothing.
 
 ## :books: Further Reading
 
-- [vue-async-computed](https://github.com/foxbenjaminfox/vue-async-computed)
+- [vue-async-computed]
+
+[vue-async-computed]: https://github.com/foxbenjaminfox/vue-async-computed
```
FloEdelmann, line 117: Again, why extract the link target here?

FloEdelmann (commented):
Very nice improvement, thank you! 🙂

eslint.config.mjs
ota-meshi: Those markdown files are often created on GitHub, and I think it would be annoying if the created Markdown files would result in errors due to CI.
Even if there are problems with those markdown files, they will eventually be merged into the CHANGELOG and the CHANGELOG will be checked by CI, so I don't think the problems will remain.

lib/rules/no-setup-props-reactivity-loss.js
ota-meshi: The original code was flagged by `unicorn/no-nested-ternary`, so I'm not sure why it wasn't flagged by CI until now 😅

docs/rules/multiline-html-element-content-newline.md
ota-meshi: Since I introduced `@eslint/mardkown`, the following code has been flagged by `markdown/no-missing-label-refs`.

https://github.com/vuejs/eslint-plugin-vue/blob/a198b52b0ac26c8b8069fef1e295758ef7e5d932/docs/rules/multiline-html-element-content-newline.md?plain=1#L94

`[no-multiple-empty-lines]` was specified, but the link definition was not specified.
I changed it to use the link definition to resolve the warning.

docs/rules/no-async-in-computed-properties.md
ota-meshi: Since I introduced `@eslint/mardkown`, the following code has been flagged by `markdown/no-missing-label-refs`.

https://github.com/vuejs/eslint-plugin-vue/blob/a198b52b0ac26c8b8069fef1e295758ef7e5d932/docs/rules/no-async-in-computed-properties.md?plain=1#L16

I changed it to use the link definition to resolve the warning.

eslint.config.mjs
FloEdelmann: But the release notes on GitHub are created from exactly those Markdown files, right? I think having the links is most helpful there, so people can quickly jump into the relevant documentation pages when they see a new release.

ota-meshi: Links from rule names will be created automatically.
Take a look at the following PR.

https://github.com/ota-meshi/eslint-plugin-markdown-preferences/pull/35

The changeset file only lists the rule name and not the link, but ESLint's autofix will automatically change the rule name to a link during the release workflow.
This will be used for changelogs and release notes.

FloEdelmann: Ah, cool!

waynzh (approved):
Thank you!

eslint.config.mjs
FloEdelmann: @ota-meshi Unfortunately, that seems to not work:

After merging https://github.com/vuejs/eslint-plugin-vue/pull/2918, the [`Release` job](https://github.com/vuejs/eslint-plugin-vue/actions/runs/17284277081/job/49058352469) failed with an ESLint error:

```
/home/runner/work/eslint-plugin-vue/eslint-plugin-vue/CHANGELOG.md
Error:   13:11  error  The word "vue/define-macros-order" should be a link  markdown-preferences/prefer-linked-words

  1 error and 0 warnings potentially fixable with the `--fix` option.

Error: Error: The process '/usr/local/bin/npm' failed with exit code 1
Error: The process '/usr/local/bin/npm' failed with exit code 1
```

And thus, the commit (and following ones) never made it to the [`Version Packages` PR](https://github.com/vuejs/eslint-plugin-vue/pull/2910).

ota-meshi: Oh, `npm run lint -- --fix` won't work. Instead, we need to use `npm run lint:fix`.
I'll submit a PR for this.

FloEdelmann: * https://github.com/vuejs/eslint-plugin-vue/pull/2922

PR #2914, opened 2025-08-03
test: add valid example for empty `defineEmits`

Sysix:
Covering this part of the code: 

https://github.com/vuejs/eslint-plugin-vue/blob/a198b52b0ac26c8b8069fef1e295758ef7e5d932/lib/rules/valid-define-emits.js#L121-L128

Found in https://github.com/oxc-project/oxc/pull/12545

### DIFF
--- a/tests/lib/rules/valid-define-emits.js
+++ b/tests/lib/rules/valid-define-emits.js
@@ -62,6 +62,17 @@ tester.run('valid-define-emits', rule, {
     {
       filename: 'test.vue',
       code: `
+      <script>
+        export default { emits: [] }
+      </script>
+      <script setup>
+        defineEmits();
+      </script>
+      `
+    },
+    {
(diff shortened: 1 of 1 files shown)


ota-meshi:
I don't understand, could you please explain in more detail?

Sysix:
> I don't understand, could you please explain in more detail?

It just extends the test suite with this special use case. (See linked code part).
Do be fair, I do not know if this is valid in the real world. But the rule allows it.

I am currently working on the implementation of the rule in `oxlint`. 
This test should help to be consistent across different tools.
`oxlint` will generate their test code from test source code of the plugin.

If you think, this is not helpful as a test case, feel free to close. 
I wanted to be sure we do not include inconsistency between both tools.

ota-meshi:
> If you think, this is not helpful as a test case, feel free to close.

I'll close it.

PR #2920, opened 2025-08-29
fix(valid-define-options): Allow the use of simple constants

### LINKED ISSUE #2919: Support literal constant references in `vue/valid-define-options` rule
**What rule do you want to change?**valid-define-options

no

Add option to ignoreVariables in valid-define-options

before
```ts
const COMPONENT_NAME = 'ElForm'
defineOptions({
  name: COMPONENT_NAME, // ❌ ESLint: referencing locally declared variables
})
```
config:
```json
{
  "vue/valid-define-options": [
    "error",
    { "allowVariableNamePattern": "^[A-Z_]+$" }
  ]
}
```
after
```ts
const COMPONENT_NAME = 'ElForm'

defineOptions({
  name: COMPONENT_NAME, // ✅ allowed
})
```

“Because the COMPONENT_NAME variable will be used multiple times in the <script> block, but it will never change.”

**Additional context**

FrontEndDog:
Fixes #2919

### DIFF
--- a/.changeset/fresh-taxis-glow.md
+++ b/.changeset/fresh-taxis-glow.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': minor
 ---
 
-vue/valid-define-options now supports local literal constant references
+Changed `vue/valid-define-options` to allow local literal constant references


lib/rules/valid-define-options.js
```diff
@@ -89,12 +89,19 @@ module.exports = {
...
+                  if (utils.inRange(defineOptions.range, def.name)) return false
+                  if (
+                    def.type === 'Variable' &&
+                    def.parent.kind === 'const' &&
+                    def.node.init &&
+                    def.node.init.type === 'Literal'
+                  )
+                    return false
+                  return true
+                })
```
FloEdelmann:
```suggestion
                variable.defs.every(
                  (def) =>
                    def.type !== 'ImportBinding' &&
                    utils.inRange(scriptSetup.range, def.name) &&
                    !utils.inRange(defineOptions.range, def.name) &&
                    !(
                      def.type === 'Variable' &&
                      def.parent.kind === 'const' &&
                      def.node.init &&
                      def.node.init.type === 'Literal'
                    )
                )
```

FrontEndDog: @FloEdelmann Thanks for your review！I have already made the modifications.

FrontEndDog:
I haven't changed to any other rules, I don't know why automated testing failed.

FloEdelmann:
> I don't know why automated testing failed

That failure is indeed unrelated; I opened an issue about it:

* https://github.com/vuejs/vue-eslint-parser/issues/274

FloEdelmann:
https://github.com/vuejs/vue-eslint-parser/issues/274 is now fixed, I updated your branch.

FloEdelmann (approved):
Thanks! 🙂

PR #2921, opened 2025-09-02
chore: add vitest to deps and use vitest.config.mts

wChenonly:
fix 
image


FloEdelmann (approved):
Thanks, makes sense to me 🙂

waynzh (approved):
Thanks!

PR #2928, opened 2025-09-05
feat(component-name-in-template-casing): `global` option support regex

### LINKED ISSUE #2912: component-name-in-template-casing options.globals should support regex

I want the `vue/component-name-in-template-casing` to support regexes for its `options.globals` option the way it does for its `ignores` options.

In my use case, we have prepended every component in our design system with `c-` and globally registered these components. Rather than producing and keeping a list of all those components in sync as they are developed, I'd like to use the regex `/^c-/` and capture them.

**Does this change cause the rule to produce more or fewer warnings?** Neither. Should make it simpler to create the same warnings.

I imagine by also mapping with the `toRegex...` helper function used on the `ignores` option

```vue

```

It requires you enumerate every single global component name you want to include in the linting.

A regex could potentially cover many globally registered component names if they follow a namespacing pattern.

**Additional context**

waynzh:
resolve #2912

### DIFF
--- a/.changeset/rude-heads-serve.md
+++ b/.changeset/rude-heads-serve.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Changed vue/component-name-in-template-casing `globals` option to support regex patterns
--- a/.changeset/rude-heads-serve.md
+++ b/.changeset/rude-heads-serve.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': minor
 ---
 
-Changed vue/component-name-in-template-casing `globals` option to support regex patterns
+Changed `vue/component-name-in-template-casing` `globals` option to support regex patterns
--- a/tests/lib/rules/component-name-in-template-casing.js
+++ b/tests/lib/rules/component-name-in-template-casing.js
@@ -1243,7 +1243,7 @@ tester.run('component-name-in-template-casing', rule, {
         <template>
           <c-button />
           <c-card />
-          <c-input />
+          <CInput />
         </template>
       `,
       output: `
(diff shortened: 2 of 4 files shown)


.changeset/rude-heads-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Changed vue/component-name-in-template-casing `globals` option to support regex patterns
```
FloEdelmann:
```suggestion
Changed `vue/component-name-in-template-casing` `globals` option to support regex patterns
```

tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -1223,6 +1238,86 @@ tester.run('component-name-in-template-casing', rule, {
         }
       ]
     },
+    {
+      code: `
+        <template>
+          <c-button />
+          <c-card />
+          <c-input />
```
FloEdelmann: Please change this to


```
          <c-button />
          <c-card />
          <CInput />
```

… and check that the last line is not reported.

tests/lib/rules/component-name-in-template-casing.js
```diff
@@ -1223,6 +1238,86 @@ tester.run('component-name-in-template-casing', rule, {
...
+          endColumn: 19
+        }
+      ]
+    },
+    {
+      code: `
+        <template>
+          <CButton />
+          <CCard />
+          <CInput />
```
FloEdelmann: Please change this to


```
          <CButton />
          <CCard />
          <c-input />
```

… and check that the last line is not reported.

FloEdelmann (approved):
Thank you! 🙂

PR #2934, opened 2025-09-09
feat: add `vue/no-duplicate-class-names` rule

### LINKED ISSUE #2932: Rule proposal: `vue/no-duplicate-class-names`
Hello I was looking for a rule to make sure I don't add duplicated class names in my code but I couldn't find any so I created a custom rule for my project but would be nice to have support for it in this plugin.

This rule can be used to void duplication of class names inside class attributes.

- [ ] Warns about a potential error (problem)
- [x] Suggests an alternate way of doing something (suggestion)
- [ ] Other (please specify:)

```vue
<template>
  <!-- ✓ GOOD -->
  <div class="foo"></div>

  <!-- ✗ BAD -->
  <div class="foo foo"></div>
</template>
```

```vue
<template>
  <!-- ✓ GOOD -->
  <div :class="{ 'foo': true }"></div>

  <!-- ✗ BAD -->
  <div :class="{ 'foo foo': true }"></div>
</template>
```

```vue
<template>
  <!-- ✓ GOOD -->
  <div :class="['foo']"></div>

  <!-- ✗ BAD -->
  <div :class="['foo foo']"></div>
</template>
```

~I'd say this rule should check duplicates within each class attribute separately so it should not report cases where a class appears once in `class=""` and once in `:class=""`~

~And also should not report duplicates withing each object key/array items or mixed inside `:class=""`~

I can help with the PR, tests and documentation as needed :)

Yizack:
resolves #2932

### DIFF
--- a/tests/lib/rules/no-duplicate-class-names.js
+++ b/tests/lib/rules/no-duplicate-class-names.js
@@ -535,16 +535,16 @@ tester.run('no-duplicate-class-names', rule, {
     },
     {
       filename: 'duplicate-class-logical-expression-in-array.vue',
-      code: `<template><div :class="['foo', isActive ?? 'foo']"></div></template>`,
-      output: null,
+      code: `<template><div :class="['foo', isActive ?? 'bar bar']"></div></template>`,
+      output: `<template><div :class="['foo', isActive ?? 'bar']"></div></template>`,
       errors: [
         {
-          message: "Duplicate class name 'foo'.",
-          type: 'ArrayExpression',
+          message: "Duplicate class name 'bar'.",
+          type: 'Literal',
           line: 1,
-          column: 24,
+          column: 44,
           endLine: 1,
-          endColumn: 50
+          endColumn: 53
         }
       ]
     },


Yizack:
I believe this rule is now ready!

tests/lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,233 @@
...
+      filename: 'no-duplicate-class-in-directive-literal.vue',
+      code: '<template><div :class="`foo bar baz`"></div></template>'
+    },
+    {
+      filename: 'no-duplicate-class-in-directive-object.vue',
+      code: `<template><div :class="{ 'foo bar baz': true }"></div></template>`
+    },
+    {
+      filename: 'duplicate-class-and-directive.vue',
+      code: `<template><div class="foo" :class="'foo'"></div></template>`
```
FloEdelmann: See https://github.com/vuejs/eslint-plugin-vue/issues/2932#issuecomment-3306121579 → let's discuss this there first.

FloEdelmann (changes requested):

docs/rules/no-duplicate-class-names.md
```diff
@@ -0,0 +1,57 @@
...
+
+```vue
+<template>
+  <!-- ✓ GOOD -->
+  <div class="foo bar"></div>
+  <div :class="'foo bar'"></div>
+  <div :class="{ 'foo bar': isActive }"></div>
+  <div :class="['foo', 'bar']"></div>
+  <div :class="isActive ? 'foo' : 'bar'"></div>
+  <div :class="'foo ' + 'bar'"></div>
```
FloEdelmann: Let's pick a more common example for the last "good" example:


```suggestion
  <div class="foo" :class="{ bar: isActive }"></div>
```

tests/lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,361 @@
...
+      filename: 'duplicate-class-in-different-directive-object-keys.vue',
+      code: `<template><div :class="{ 'foo': true, 'foo bar': true }"></div></template>`
+    },
+    {
+      filename: 'class-conditional-expression.vue',
+      code: `<template><div :class="isActive ? 'foo' : 'bar'"></div></template>`
+    },
+    {
+      filename: 'class-conditional-duplicate-expression-value.vue',
+      code: `<template><div :class="isActive ? 'foo' : 'foo'"></div></template>`
```
FloEdelmann, line 54: This looks like an actual mistake that should in the best case be reported; but I'm not sure this rule is the correct way to do it. So it's fine to list it explicitly here and handle it this way.

tests/lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,361 @@
...
+      filename: 'class-object-duplicate-value.vue',
+      code: `<div :class="{ 'foo bar': isActive, 'foo': isAnotherActive }"></div>`
+    }
+  ],
+  invalid: [
+    {
+      filename: 'duplicate-class.vue',
+      code: '<template><div class="foo foo"></div></template>',
+      output: '<template><div class="foo"></div></template>',
+      errors: [
```
FloEdelmann, line 66: Can you please add `line`, `column`, `endLine` and `endColumn` to all expected errors?

tests/lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,361 @@
...
+        }
+      ]
+    },
+    {
+      filename: 'multiple-duplicate-classes.vue',
+      code: '<template><div class="foo bar foo baz bar"></div></template>',
+      output: '<template><div class="foo bar baz"></div></template>',
+      errors: [
+        {
+          message: "Duplicate class name 'foo, bar'.",
```
FloEdelmann: This message would be a bit better. Can you change the code and tests to this format?

```suggestion
          message: "Duplicate class names 'foo', 'bar'.",
```

tests/lib/rules/no-duplicate-class-names.js
FloEdelmann, line 1: In general: nice test cases!

lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,314 @@
...
+
+        /** @type {Map<string, {node: ASTNode, unconditional: boolean}>} */
+        const collected = new Map()
+
+        const classNodes = extractClassNodes(node)
+        for (const { node: reportNode, unconditional } of classNodes) {
+          // report fixable duplicates and collect reported class names
+          const reportedClasses = reportDuplicateClasses(reportNode)
+          if (reportedClasses) {
+            for (const classes of reportedClasses) reported.add(classes)
```
FloEdelmann:
```suggestion
            for (const reportedClass of reportedClasses) reported.add(reportedClass)
```

Yizack:
> Thanks for your contribution! Sorry that reviewing took so long. I only have a few small suggestions/requests; in general the code looks really good! 🙂

Hi! No problem, thanks for the review 🙂, I just applied your suggestions and requests

lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,319 @@
...
+    type: 'suggestion',
+    docs: {
+      url: 'https://eslint.vuejs.org/rules/no-duplicate-class-names.html',
+      description: 'disallow duplication of class names in class attributes',
+      categories: undefined
+    },
+    fixable: 'code',
+    schema: [],
+    messages: {
+      duplicateClassNames: 'Duplicate class names {{names}}.'
```
FloEdelmann: Sorry I wasn't clear enough: The error message should be `Duplicate class names 'foo', 'bar'.` but `Duplicate class name 'foo'.`.

FloEdelmann (changes requested):

lib/rules/no-duplicate-class-names.js
Yizack: Oh yes, let me adjust it, I'm not sure if this is the correct way

FloEdelmann (approved):
Thank you! :partying_face:

tests/lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,477 @@
...
+      filename: 'class-conditional-expression.vue',
+      code: `<template><div :class="isActive ? 'foo' : 'bar'"></div></template>`
+    },
+    {
+      filename: 'class-conditional-duplicate-expression-value.vue',
+      code: `<template><div :class="isActive ? 'foo' : 'foo'"></div></template>`
+    },
+    {
+      filename: 'class-object-duplicate-value.vue',
+      code: `<div :class="{ 'foo bar': isActive, 'foo': isAnotherActive }"></div>`
```
waynzh, line 58: Nice case!

lib/rules/no-duplicate-class-names.js
```diff
@@ -0,0 +1,324 @@
...
+ * @param {VDirective} node
+ * @param {Expression} [expression]
+ * @param {boolean} [unconditional=true] whether the expression is unconditional
+ * @return {IterableIterator<{ node: Literal | TemplateElement, unconditional: boolean }>}
+ */
+function* extractClassNodes(node, expression, unconditional = true) {
+  const nodeExpression = expression ?? node.value?.expression
+  if (!nodeExpression) return
+
+  switch (nodeExpression.type) {
```
waynzh, line 20: `LogicalExpression` is not handled. For example, here's a less common case:  
`<div :class="isActive ?? 'foo foo'"></div>`"

Yizack: Omg you are right! I forgot about that one, adding right now

Yizack: I have a question to discuss:

for the following case, should `'bar bar'` be reported? Or should we assume `'foo'` will always be rendered so it should not be reported since this is an actual mistake.

cc @FloEdelmann 

```vue
<div :class="'foo' || 'bar bar'"></div>
```

waynzh: IMO, always reporting seems like a correct and low-cost solution to implement.

Yizack: Perfect! added the `LogicalExpression` case + tests

FloEdelmann (approved):
Thanks again!

PR #2938, opened 2025-10-04
refactor(no-import-compiler-macros): check only in `<script setup>`

Sysix:
After reading https://github.com/vuejs/eslint-plugin-vue/issues/2437, I found out that the compiler warnings are only reported at `<script setup>`:
https://github.com/vuejs/core/blob/c16f8a94c7eda79f51f44b7b3c64c72343df0d38/packages/compiler-sfc/src/compileScript.ts#L321-L355

Changed the rule to only look inside `<script setup>` and not other files like `someMacro.ts`.

Copied the logic from https://github.com/vuejs/eslint-plugin-vue/blob/553abe61c4d7a8964fb154069ea6a82d14b2b3b6/lib/rules/valid-define-emits.js#L31-L46

### DIFF
--- a/lib/rules/no-import-compiler-macros.js
+++ b/lib/rules/no-import-compiler-macros.js
@@ -4,6 +4,8 @@
  */
 'use strict'
 
+const utils = require('../utils')
+
 const COMPILER_MACROS = new Set([
   'defineProps',
   'defineEmits',
@@ -43,9 +45,13 @@ module.exports = {
    * @returns {RuleListener}
    */
   create(context) {
+    const scriptSetup = utils.getScriptSetupElement(context)
+    if (!scriptSetup) {
+      return {}
+    }
     const sourceCode = context.getSourceCode()
 
-    return {
+    return utils.defineScriptSetupVisitor(context, {
       ImportDeclaration(node) {
         if (node.specifiers.length === 0 || !VUE_MODULES.has(node.source.value))
           return
@@ -96,6 +102,6 @@ module.exports = {
           })
         }
       }
-    }
+    })
   }
 }


tests/lib/rules/no-import-compiler-macros.js
```diff
@@ -30,7 +30,8 @@ tester.run('no-import-compiler-macros', rule, {
       filename: 'test.vue',
       code: `
       <script>
-      import { defineProps } from 'some-other-package'
```
FloEdelmann, line 33: This was a good test case as well; please duplicate it (and use `<script setup>` there).

Sysix:
> it would not really make sense to import them outside <script setup>, too, would it?

That was my guess too. But they can be wrongly imported by the user. 
Do not know what the bundler/vue-compiler will do. Maybe a new ESLint rule like `no-compilier-macros-outside-script-setup`?

Updated the tests 👍

FloEdelmann:
> But they can be wrongly imported by the user.

But the `vue/no-import-compiler-macros` already warns about that case, while this PR removes exactly this warning. So I think the current behavior is better.

Do you agree? If so, feel free to update the PR to only improve test coverage for the non-`<script setup>` case. Otherwise, I'm open for discussion :slightly_smiling_face:

Sysix:
I would prefer an own error/diagnostic message for using compiler macros outside Vue files.
Do you think it is okay to include it in this rule? 
Something like 
```
'{{name}} is a compiler macro and is only valid inside `<script>`with "setup".
```

FloEdelmann:
Changing the message based on whether the import occurs inside `<script setup>` or not seems like a good improvement to me :slightly_smiling_face: :+1:

tests/lib/rules/no-import-compiler-macros.js
```diff
@@ -214,6 +214,27 @@ tester.run('no-import-compiler-macros', rule, {
           endColumn: 60
         }
       ]
+    },
+    {
```
FloEdelmann, line 245:
```suggestion
    },
    {
      filename: 'test.vue',
      code: `
      <script>
      // not in <script setup>
      import { defineProps } from 'vue'
      </script>
      `,
      output: `
      <script>
      // not in <script setup>
      
      </script>
      `,
      errors: [
        {
          messageId: 'onlyValidInScriptSetup',
          data: {
            name: 'defineProps'
          },
          line: 4,
          column: 16,
          endLine: 4,
          endColumn: 27
        }
      ]
    },
    {
```

PR #2947, opened 2025-10-30
chore: Update postcss-selector-parser@7.1.0

pioug:
The breaking changes affect only the `insertBefore`, `insertAfter`, and `prepend` methods https://github.com/postcss/postcss-selector-parser/commit/4fa6e860a176985e03fda969b2b435fe168f31c2 which don’t appear to be used here.

Updating this dependency can prevent having multiple versions of `postcss-selector-parser` in a project

### DIFF
--- a/.changeset/neat-coats-allow.md
+++ b/.changeset/neat-coats-allow.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": patch
 ---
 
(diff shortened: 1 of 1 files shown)


FloEdelmann (approved):
Looks good to me, thank you! 🙂

PR #2953, opened 2025-11-11
fix(define-macros-order): improve report messages

### LINKED ISSUE #2940: vue/define-macros-order confusing error message
**Checklist**

- **ESLint version:** 9.37.0
- **eslint-plugin-vue version:** 10.5.0
- **Vue version:** 3.5.22
- **Node version:** 24.7.0
- **Operating System:** Gentoo Linux

**What did you do?**

Given the following configuration (taken from the [rule documentat](https://eslint.vuejs.org/rules/define-macros-order#order-defineoptions-definemodel-defineprops-defineemits-defineslots)):	

```ts
{
  rules: {
    "vue/define-macros-order": ["error", { "order": ["defineOptions", "defineModel", "defineProps", "defineEmits",defineSlots"] },
  },
}
```

And the following code (also from the documentation, but with `defineOptions` moved to the top):

```vue
<script setup>
defineOptions({/* ... */})
defineEmits(/* ... */)
const slots = defineSlots()
defineProps(/* ... */)
const model = defineModel()
</script>
```

> defineModel should be above defineEmits

Since we know the order of the macros (from configuration) it would make sense if the error contained the expected ordering.

If we compare with the `vue/order-in-components` rule where we get an error similar to:

> The "..." property should be above the "..." property on line [..]

**What actually happened?**

> defineModel should be the first statement in `<script setup>` (after any potential import statements or type definitions).

Autofixing the error works, `defineModel()` is moved to the proper location. But the error is confusing and would make one try to move it to the top above `defineOptions` but doing that puts `defineOptions` in the wrong order and you get a new error instead. If multiple of the macros are out-of-order (as might happen if you just turn this on) you get an endless supply of these confusing errors. Almost like a puzzle, you keep trying all kinds of combinations until you at some point happen to stumble upon the correct ordering.

kzhrk:
Closes #2940 

## Summary

Improves error messages for the `vue/define-macros-order` rule to distinguish between two distinct cases:

1. **macrosNotAtTop**: When a macro comes after non-macro statements (e.g., `console.log()`)
2. **macrosUnordered**: When macros are not in the correct order relative to each other

## Motivation

Previously, the rule always reported: "X should be the first statement in `<script setup>`", which was confusing when multiple macros were out of order.
Users couldn't easily understand whether they needed to move a macro to the top or reorder it relative to other macros.

With this improvement, users get more specific feedback:
- "defineModel should be placed at the top of `<script setup>`" - when it's after a non-macro
- "defineModel should be above defineEmits" - when it's in the wrong order with other macros

This makes it much clearer what action is needed to fix the issue.

## Changes

- Updated messages object with two new message types: `macrosNotAtTop` and `macrosUnordered`
- Modified `reportNotOnTop()` function to detect whether the target statement is a macro from the order list
- Uses `macrosUnordered` when the target is a macro, `macrosNotAtTop` otherwise
- Updated all test expectations to match the new behavior (42 tests passing)

### DIFF
--- a/lib/rules/define-macros-order.js
+++ b/lib/rules/define-macros-order.js
@@ -221,14 +221,9 @@ function create(context) {
    */
   function reportNotOnTop(macro, nodes, before) {
     // Determine if 'before' is a macro from the order list
-    let beforeMacro = null
-    for (const macroName of order) {
-      const macroNodes = macrosNodes.get(macroName) ?? []
-      if (macroNodes.includes(before)) {
-        beforeMacro = macroName
-        break
-      }
-    }
+    const beforeMacro = order.find((macroName) =>
+      (macrosNodes.get(macroName) ?? []).includes(before)
+    )
 
     // Decide which message to use
     const messageId = beforeMacro ? 'macrosUnordered' : 'macrosNotAtTop'


lib/rules/define-macros-order.js
```diff
@@ -220,13 +220,25 @@ function create(context) {
...
   function reportNotOnTop(macro, nodes, before) {
+    // Determine if 'before' is a macro from the order list
+    let beforeMacro = null
+    for (const macroName of order) {
+      const macroNodes = macrosNodes.get(macroName) ?? []
+      if (macroNodes.includes(before)) {
+        beforeMacro = macroName
+        break
+      }
+    }
```
FloEdelmann:
```suggestion
    const beforeMacro = order.find((macroName) =>
      (macrosNodes.get(macroName) ?? []).includes(before)
    )
```

kzhrk:
@FloEdelmann 
> Is this PR deliberately still in draft status?

No, I was just waiting for the CI 😴
I'll check your suggestion.

PR #2954, opened 2025-11-13
feat(no-ref-as-operand): support ref detection from composable functions

### LINKED ISSUE #2519: `vue/no-ref-as-operand` to support `ref` from composable

Current `vue/no-ref-as-operand` rule only support the local `ref`. If a `ref` is being returned from a composable and being use with operand (eg: returnedRef++) , it didnt throw any warning as expected.

[ ] Other (please specify:)

```ts
const ok = ref(true)
const msg = ok ? 'yes' : 'no' // this throws warning as expected

// ---- //

const ok = useOk() // returned as Ref<boolean>
const msg = ok ? 'yes' : 'no' // No warning was shown
```

**Additional context**

kzhrk:
Closes #2519

## Summary

Extends the `vue/no-ref-as-operand` rule to detect and report when ref objects
returned from composable functions are used without accessing the `.value` property.

Previously, the rule only detected direct ref declarations like `ref(0)`.
Now it also detects ref objects returned from composable functions:

```typescript
// ❌ Before: No warning (bug)
const ok = useOk() // returns Ref<boolean>
const msg = ok ? 'yes' : 'no' // No warning was shown

// ✅ After: Warning shown
const ok = useOk() // returns Ref<boolean>
const msg = ok ? 'yes' : 'no' // Warning: Must use `.value` to read or write the value wrapped by `useOk()`.
```

## Motivation

Composable functions are a common pattern in Vue 3 for code reuse. When they return
Ref objects, developers should use the .value property to access the wrapped value.
However, the rule previously had no way to detect when these composable-returned refs
were used incorrectly.

This enhancement provides early feedback to developers, helping them follow Vue 3
best practices and preventing subtle bugs.

## Implementation Details

- Function body analysis: Uses AST analysis to detect when functions return ref objects
- Pattern support: Handles multiple return patterns:
  - Direct ref() calls: return ref(0)
  - Object properties: return { data: ref(0) }
  - Array elements: return [ref(0)]
- All scope contexts: Properly searches all scopes to find function definitions
- Error messages: Shows composable function names in error messages for clarity

## Scope and Limitations

The detection works for composable functions defined in the same file. Due to
ESLint's single-file analysis model, composable functions imported from external
modules cannot be analyzed. This is a fundamental ESLint framework constraint.

## Testing

- Added 6 comprehensive test cases covering:
  - Simple composable functions returning ref
  - Arrow functions returning ref
  - Array destructuring from composable returns
  - Proper .value access detection
  - Error detection when .value is missing
- All 55 tests pass (49 existing + 6 new)
- No breaking changes to existing functionality

### DIFF
--- a/.changeset/support-composable-ref-detection.md
+++ b/.changeset/support-composable-ref-detection.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Enhanced `vue/no-ref-as-operand` rule to detect ref objects returned from composable functions
--- a/lib/utils/ref-object-references.js
+++ b/lib/utils/ref-object-references.js
@@ -276,6 +276,72 @@ function* iterateIdentifierReferences(id, globalScope) {
 }
 
 /**
+ * Check if a function returns a ref() call
+ * @param {FunctionDeclaration | FunctionExpression | ArrowFunctionExpression} node
+ * @returns {boolean}
+ */
+function checkFunctionReturnsRef(node) {
+  const body = node.body
+  if (!body) {
+    return false
+  }
+
+  // For arrow functions with expression body
+  if (
+    node.type === 'ArrowFunctionExpression' &&
+    body.type !== 'BlockStatement'
+  ) {
+    return isRefCall(body)
+  }
+
+  // For function declarations and arrow functions with block body
+  if (body.type === 'BlockStatement') {
+    for (const stmt of body.body) {
+      if (stmt.type === 'ReturnStatement' && stmt.argument) {
+        return isRefCall(stmt.argument)
+      }
+    }
+  }
+
+  return false
+}
+
+/**
+ * Check if an expression is a ref() call or returns a ref object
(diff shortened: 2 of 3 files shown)


FloEdelmann:
This only works for composables defined inside the component itself, right? So not for imported ones? If so, it's probably not _that_ useful and will likely result in exactly that issue raised right after release :sweat_smile: 

As I wrote in https://github.com/vuejs/eslint-plugin-vue/issues/2519#issuecomment-2304409514, we could use variable types (`Ref<…>`) when using `typescript-eslint`. Would you be willing to look into that?

kzhrk:
Thank you for the feedback! You're right – this implementation only works for composables defined within the same file, which significantly limits its usefulness in real-world projects 🫠 

I agree that leveraging TypeScript's type information (`Ref<…>`) through typescript-eslint would be a much more comprehensive solution. I'll look into that approach.

I'm going to close this PR for now. Thanks again for pointing me in the right direction!

PR #2960, opened 2025-11-20
chore(lint): setup typescript-eslint

9romise:


eslint.config.mjs
```diff
@@ -78,7 +83,7 @@ export default typegen([
     }
   },
   ...defineConfig({
-    files: ['**/*.js'],
+    files: ['**/*.?([cm])[jt]s'],
```
FloEdelmann: This should work, too, right? I think it's a little easier to read and reason about:

```suggestion
    files: ['**/*.{js,cjs,mjs,ts,cts,mts}'],
```

eslint.config.mjs
```diff
@@ -105,7 +110,21 @@ export default typegen([
...
+        sourceType: 'module'
+      }
+    },
+    plugins: {
+      '@typescript-eslint': eslintPluginTs
+    }
+  },
+
+  {
+    files: ['**/*.?([cm])[jt]s'],
```
FloEdelmann:
```suggestion
    files: ['**/*.{js,cjs,mjs,ts,cts,mts}'],
```

eslint.config.mjs
```diff
@@ -46,10 +48,13 @@ const MD_LINKS_FOR_DOCS = {
   ...MD_BASE_LINKS
 }
 
+const GLOB_TS = '**/*.?([cm])ts'
+
```
FloEdelmann: Again, this could be `'**/*.{ts,cts,mts}'`, right?

Also, please use this string literal instead of this constant.

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
```diff
@@ -351,7 +351,7 @@ export interface PrivateIdentifier extends HasParentNode {
 }
 export interface Literal extends HasParentNode {
   type: 'Literal'
-  value: string | boolean | null | number | RegExp | BigInt
+  value: string | boolean | null | number | RegExp | bigint
```
FloEdelmann, line 354: I guess `BigInt` is a wrapper type (like `String` instead of `string`) and thus should be avoided?

docs/.vitepress/vite-plugin.mts
```diff
@@ -57,27 +58,23 @@ function transformRequire(code: string) {
         return match
       }
 
-      let id =
-        '__' +
-        moduleString.replace(/[^a-zA-Z0-9_$]+/gu, '_') +
-        Math.random().toString(32).substring(2)
+      let id = `__${moduleString.replace(
```
FloEdelmann: These changes make this file harder to read. Can you please disable the `prefer-template` rule in this file only and only correct the other lint violations?

docs/.vitepress/theme/index.ts
```diff
@@ -1,23 +1,18 @@
...
 // @ts-expect-error -- Cannot change `module` option
 import type { Theme } from 'vitepress'
 // @ts-expect-error -- Cannot change `module` option
 import DefaultTheme from 'vitepress/theme'
-// @ts-expect-error -- ignore
 import Layout from './Layout.vue'
-// @ts-expect-error -- ignore
 import ESLintCodeBlock from './components/eslint-code-block.vue'
-// @ts-expect-error -- ignore
 import RulesTable from './components/rules-table.vue'
```
FloEdelmann, line 19: Don't these throw TypeScript errors anymore? Why doesn't `@ts-expect-error` then fail? I thought that was the advantage over `@ts-ignore`?

9romise: Sorry about that, I misread the ESLint error.

9romise:
I think we will never use `.cjs` or `.cts` extensions, so I haven't added support for them.

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
FloEdelmann: In [`3fc1431`](https://github.com/vuejs/eslint-plugin-vue/pull/2960/commits/3fc14310e526f187b66b1862205e1d496096a0e5#diff-641223a90b3deb4eb0aed81b84b454fab8ebad94295bd051c1348b33ee51fa34R355), you restored the status quo and disabled the respective ESLint rule.

I was not saying that ESLint complaining here was wrong, I only wanted to understand the issue. It might very well be a genuine mistake in our code that ESLint now found.

docs/.vitepress/vite-plugin.mts
```diff
@@ -58,19 +59,22 @@ function transformRequire(code: string) {
...
       modules.set(id, moduleString)
-      return id + '()'
+      return `${id}()`
     }
   )
 
   return (
+    // eslint-disable-next-line prefer-template
     [...modules]
+      // eslint-disable-next-line arrow-body-style
```
FloEdelmann: Please remove this inline disable comment regarding `arrow-body-style` and instead fix the issue. [My comment above](https://github.com/vuejs/eslint-plugin-vue/pull/2960#discussion_r2546288430) was referring to the `prefer-template` rule only.

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
9romise: I think ESLint is correct. Here are some related links:
https://typescript-eslint.io/rules/no-wrapper-object-types/
https://github.com/typescript-eslint/typescript-eslint/blob/28cf8032c2492bb3c55dd7dd145249f2246034ad/packages/ast-spec/src/expression/literal/BigIntLiteral/spec.ts#L3

eslint.config.mjs
```diff
@@ -249,6 +266,35 @@ export default typegen([
       'internal/require-eslint-community': ['error']
     }
   },
+
+  {
+    files: ['**/*.mjs'],
+    languageOptions: {
+      sourceType: 'module'
+    }
+  },
```
FloEdelmann, line 275: Shouldn't `languageOptions.sourceType` be set to `module` for `.mts` files as well?


```suggestion
  {
    files: ['**/*.{mjs,mts}'],
    languageOptions: {
      sourceType: 'module'
    }
  },
```

typings/eslint-plugin-vue/util-types/ast/es-ast.ts
FloEdelmann: Ok, thanks for the references. Please keep ESLint's suggestion then.

eslint.config.mjs
9romise: Yes, you're right.

9romise:
The type definitions in our repository appear to be outdated. Just curious — why aren't we directly referencing some of the node types from `typescript-eslint`?

9romise:
Additionally, is the team interested in migrating to a pnpm workspace? I believe this would allow `.mts` files in the `docs` directory to use the `module` type directly instead of following the root's `commonjs` configuration.

FloEdelmann:
> why aren't we directly referencing some of the node types from `typescript-eslint`?

I don't know. @ota-meshi can you elaborate?

ota-meshi:
> > why aren't we directly referencing some of the node types from `typescript-eslint`?
>
> I don't know. @ota-meshi can you elaborate?

Since the nodes provided by `typescript-eslint` don't know that the parent has a `VExpressionContainer`, I thought it might be difficult to check them properly.
https://github.com/vuejs/vue-eslint-parser/blob/master/docs/ast.md#vexpressioncontainer
For this reason, I avoid using the nodes provided by typescript-eslint directly. I'd welcome any better solutions.

PR #2916, opened 2025-08-22
feat: introduce `tsdown`, support mixed js & ts in codebase

9romise:
Hi 👋, Vue team!

I've been trying to contribute to `eslint-plugin-vue` and recently noticed there might be plans to [rewrite it in TypeScript](#2777), though things seem to have stalled. I’d like to help move this effort forward.

This PR introduces `tsdown`, a tool that can gradually transform mixed JavaScript/TypeScript codebases into pure JavaScript output. This allows us to incrementally migrate the entire project to TypeScript. I’ve already successfully rewritten several files — it works well.

Going forward, I can split the changes into multiple PRs to make review easier.

If there’s anything you’d like me to adjust in this PR, or if you have any feedback, I’d really appreciate your thoughts! 💚

Related: #2777

PR #2961, opened 2025-11-20
refactor: migrate to pnpm workspace

9romise:
> Additionally, is the team interested in migrating to a pnpm workspace? I believe this would allow .mts files in the docs directory to use the module type directly instead of following the root's commonjs configuration.

https://github.com/vuejs/eslint-plugin-vue/pull/2960#issuecomment-3558823510

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,9 @@ yarn.lock
 yarn-error.log
 typings/eslint/lib/rules
 eslint-typegen.d.ts
+
+docs/.vitepress/dist
+docs/.vitepress/build-system/shim/@typescript-eslint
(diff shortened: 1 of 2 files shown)


FloEdelmann:
I'm a bit on the fence about this. This is an extra step for most contributors (which we can only expect to have plain `npm` installed), only to simplify the CJS/ESM-specific code in the docs repo.

We could probably use proper ESM syntax there, too, by switching to the `.mts`/`.js` file extension, can we?

And in the long term, we'll probably switch over to ESM for the whole repo anyway (and generate CJS in the build step, which you introduced in #2916).

PR #2963, opened 2025-11-22
chore(lint): setup `eslint-config-flat-gitignore`

9romise:
[`eslint-config-flat-gitignore`](https://github.com/antfu/eslint-config-flat-gitignore) automatically excludes files listed in `.gitignore` during linting, so we don’t need to specify them manually.

### DIFF
--- a/.changeset/rich-bags-turn.md
+++ b/.changeset/rich-bags-turn.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed false positives for `TSImportType` in `vue/script-indent` rule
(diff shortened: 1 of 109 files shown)


eslint.config.mjs
```diff
@@ -1,5 +1,6 @@
 import { defineConfig } from 'eslint/config'
 import globals from 'globals'
+import ignore from 'eslint-config-flat-gitignore'
```
FloEdelmann: Please rename to `eslintConfigFlatGitIgnore` for consistency.

FloEdelmann (changes requested):
Good idea, this makes it more readable and less likely to diverge 🙂

PR #2345, opened 2023-12-04
feat(no-unused-properties): mark $props.<prop> as used

levchak0910:
fixes #2344 

@FloEdelmann could you please mark my issue as open again?

PR #2968, opened 2025-11-22
test: rewrite some files in `.ts`

9romise:
In earlier Node.js versions, importing `ts` files directly wasn't supported, so the failing test files from #2916 have been changed to use the `.ts` extension in this PR.

### DIFF
--- a/typings/eslint/index.d.ts
+++ b/typings/eslint/index.d.ts
@@ -271,7 +271,7 @@ export namespace SourceCode {
 }
 
 export namespace Rule {
-  interface RuleModule extends ESLintRule.RuleModule {
+  interface RuleModule /*extends ESLintRule.RuleModule*/ {
     meta: RuleMetaData
     create(context: RuleContext): Rule.RuleListener
   }


tests/lib/rules/block-order.ts
```diff
@@ -1,38 +1,39 @@
...
 })
 
 const tester = new RuleTester({
   languageOptions: {
-    parser: require('vue-eslint-parser')
+    parser: parserVue
   }
 })
 
+// @ts-expect-error rule ts(#2345)
```
FloEdelmann: Is this [TypeScript error 2345](https://stackoverflow.com/questions/42421501/error-ts2345-argument-of-type-t-is-not-assignable-to-parameter-of-type-objec) or [GitHub issue #2345](https://togithub.com/vuejs/eslint-plugin-vue/pull/2345)?

In any case, please provide a better comment than just the error code.

FloEdelmann (changes requested):
Cool!

tests/lib/rules/block-order.js
```diff
@@ -71,44 +72,36 @@ tester.run('block-order', rule, {
     // order
     {
       code: '<script></script><template></template><style></style>',
-      output: null,
```
9romise, line 74: `output` does not exist in type `ValidTestCase`

PR #2982, opened 2025-11-27
refactor(prefer-use-template-ref): optimize performance and type safety

9romise:
- Optimize array processing with `flatMap`.
- Add stricter type checks for `ref()`/`shallowRef()` calls.
- Use a `Map` to match the `ref` in `template` and `script`.


FloEdelmann (approved):
Thanks, very cool!

PR #2983, opened 2025-11-27
feat(prefer-use-template-ref): support auto-fix

9romise:
Re-implement of #2632 (closes #2632). (I really love this feature!)
But this PR doesn't insert `import { useTemplateRef } from 'vue'` automatically.
Because I think that falls outside the scope of this rule. Additionally, some users rely on `unplugin-auto-import` and would have to remove the unnecessary import to maintain clean imports.

Please merge #2982 first.

### DIFF
--- a/.changeset/thin-scissors-heal.md
+++ b/.changeset/thin-scissors-heal.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Added autofix to `vue/prefer-template-ref`


tests/lib/rules/prefer-use-template-ref.js
```diff
@@ -311,14 +311,25 @@ tester.run('prefer-use-template-ref', rule, {
         const root = ref();
       </script>
       `,
+      output: `
+      <template>
+          <div ref="root"/>
+      </template>
+      <script setup>
+        import { ref } from 'vue';
+        const root = useTemplateRef('root');
```
ota-meshi, line 320: I think autofixes shouldn't break source code.
I think adding `import { useTemplateRef } from 'vue'` will inconvenience users who use `unplugin-auto-import`, but I think not adding `import { useTemplateRef } from 'vue'` will break it for users who don't use `unplugin-auto-import`.
Could you change it to add `import { useTemplateRef } from 'vue'`?
Or could you change it to a suggestion instead of autofix?

9romise: That makes sense. I'll have a try.

9romise:
Would it make sense to add an option to control whether imports are included during autofix?

FloEdelmann:
> option to control whether imports are included during autofix

I'd keep it simple for now and either always or never add them. But if you think that an option would make sense and you'd like to spend the extra effort, go ahead :slightly_smiling_face:

PR #2984, opened 2025-11-27
feat(vue/no-negated-v-if-condition): upgrade rule suggestion to autofix

### LINKED ISSUE #2979: `no-negated-v-if-condition`: promote suggestions to autofixes

`no-negated-v-if-condition`

N/A

`no-negated-v-if-condition`'s suggestion to swap elements can be safely replaced with autofix, which would enhance DX and bring the rule on par with [the similar rules](https://eslint.vuejs.org/rules/no-negated-v-if-condition.html#related-rules).

N/A

**Additional context**

rzzf:
close #2979

This is my first contribution to this amazing project. Please feel free to point out anything that’s not quite right. Thank you!

### DIFF
--- a/.changeset/rich-zebras-type.md
+++ b/.changeset/rich-zebras-type.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Changed `vue/no-negated-v-if-condition` suggestion to autofix


docs/rules/no-negated-v-if-condition.md
```diff
@@ -50,7 +51,15 @@ Negated conditions make the code less readable. When there's an `else` clause, i
...
-Nothing.
+```json
+{
+  "vue/no-negated-v-if-condition": ["error", {
+    "autofix": false,
+  }]
+}
+```
+
+- `"autofix"` ... If `true`, enable autofix. (Default: `false`)
```
FloEdelmann: I think the autofix is safe, so it can always be enabled and there doesn't need to be an option.

rzzf: Thank you for reviewing my PR so quickly! 😀
Question: do we still need to keep suggest?

<hr />

I checked most of the rules, and it seems that suggest and fix usually don’t exist simultaneously. If I’m mistaken, please feel free to correct me. Thank you! 🙇

FloEdelmann: I think they *could* coexist, but it usually makes no sense; if the issue can be autofixed, why offer suggestions that have to be applied manually?

In this specific case, the suggestion definitely has to be replaced with the autofix, since they would do the same thing. I see that you have already implemented it this way, so I'll resolve this conversation.

ota-meshi (approved):
LGTM! Thank you!

PR #2985, opened 2025-11-28
test: remove unnecessary `semver` calls

9romise:


tests/lib/rules/space-in-parens.js
```diff
@@ -3,20 +3,10 @@
...
-  ? (obj) => {
-      const messageId = obj.messageId
-      delete obj.messageId
-      obj.message = messageId.startsWith('missing')
-        ? 'There must be a space inside this paren.'
-        : 'There should be no spaces inside this paren.'
-      return obj
-    }
-  : (obj) => obj
+const errorMessage = (obj) => obj
```
FloEdelmann: That function could be removed now, too.

9romise: Oh! I missed it, thanks!

PR #2990, opened 2025-12-10
feat: add vue/no-undef-directives rule

### LINKED ISSUE #2946: Rule proposal: `vue/no-undef-directives`

The rule should warn when a custom directive is used, but not imported. The rule should also accept some kind of `ignore` option to avoid false positives with globally-registered custom directives.

- [x] Warns about a potential error (problem)
- [ ] Other (please specify:)

Should warn on:

```vue
<template>
  <input v-focus>
</template>

<script setup>
// vFocus is not imported or defined anywhere
</script>
```

```vue
<template>
  <a v-sref="someValue">
    <!-- anchor contents -->
  </a>
</template>

<script setup>
// vSref is not imported or defined anywhere
</script>
```

Should not warn on:

```vue
<template>
  <input v-focus>
</template>

<script setup>
import vFocus from '../path/to/vFocus';
</script>
```

```vue
<template>
  <a v-sref="someValue">
    <!-- anchor contents -->
  </a>
</template>

<script setup>
import vSref from '../path/to/vSref';
</script>
```

**Additional context**

There is a runtime warning in the browser console, but it would be cool if we can catch these issues when linting too.

```
[Vue warn]: Failed to resolve directive: sref
```

rzzf:
resolves #2946

### DIFF
--- a/lib/rules/no-undef-directives.js
+++ b/lib/rules/no-undef-directives.js
@@ -28,11 +28,10 @@ function getRegisteredDirectives(componentObject) {
     return []
   }
 
-  // @ts-ignore
   return directivesNode.value.properties.flatMap((node) => {
     const name =
       node.type === 'Property' ? utils.getStaticPropertyName(node) : null
-    return name ? [{ node, name }] : []
+    return name ? [{ node: /** @type {Property} */ (node), name }] : []
   })
 }
 


docs/rules/no-undef-directives.md
```diff
@@ -0,0 +1,89 @@
...
+
+> disallow use of undefined custom directives
+
+- :exclamation: <badge text="This rule has not been released yet." vertical="middle" type="error"> _**This rule has not been released yet.**_ </badge>
+
+## :book: Rule Details
+
+This rule reports directives that are used in the `<template>`, but that are not registered in the `<script setup>` or the Options API's `directives` section.
+
+Undefined directives will be resolved from globally registered directives. However, if you are not using global directives, you can use this rule to prevent run-time errors.
```
FloEdelmann:
```suggestion
Undefined directives will be resolved from globally registered directives. However, if you are not using global directives, you can use this rule to prevent runtime errors.
```

docs/rules/no-undef-directives.md
```diff
@@ -0,0 +1,89 @@
...
+{
+  "vue/no-undef-directives": ["error", {
+    "ignorePatterns": ["foo"]
+  }]
+}
+```
+
+- `ignorePatterns` (`string[]`) ... An array of regex pattern strings to ignore.
+
+### `ignorePatterns`
```
FloEdelmann:
```suggestion
### "ignorePatterns": ["foo"]
```

tests/lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,471 @@
...
+      `
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <template>
+        <div v-foo></div>
+      </template>
+      `,
+      options: [{ ignorePatterns: ['foo'] }]
```
FloEdelmann:
```suggestion
    {
      filename: 'test.vue',
      code: `
      <template>
        <div v-foo></div>
        <div v-barfoobaz></div>
      </template>
      `,
      options: [{ ignorePatterns: ['foo'] }]
```

tests/lib/rules/no-undef-directives.js
FloEdelmann, line 1: Nice test cases!

tests/lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,471 @@
...
+    }
```
FloEdelmann, line 242: Idea for another test case that should be supported: `foo-bar` should ignore both `v-fooBar` and `v-foo-bar`:

```suggestion
    },
    {
      filename: 'test.vue',
      code: `
      <template>
        <div v-fooBar></div>
        <div v-foo-bar></div>
      </template>
      `,
      options: [{ ignorePatterns: ['foo-bar'] }]
    }
```

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
+    ],
+    messages: {
+      undef: "The 'v-{{name}}' directive has been used, but not defined."
+    }
+  },
+  /** @param {RuleContext} context */
+  create(context) {
+    const options = context.options[0] || {}
+    /** @type {string[]} */
+    const ignorePatterns = options.ignorePatterns || []
```
FloEdelmann: For consistency, please rename the option to `ignore` and use the `regexp.toRegExpGroupMatcher` util. Then the option supports both literals (`"foo"` would only ignore `v-foo` but not `v-foo-bar`), and regexp patterns (`"/foo/"` would also ignore `v-foo-bar`).

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  /**
+   * @param {string[]} names
+   */
+  addName(...names) {
+    for (const name of names) {
+      this.names.add(name)
+    }
+  }
```
FloEdelmann: This is never called with multiple names, so can be simplified:


```suggestion
  /**
   * @param {string} name
   */
  addName(name) {
    this.names.add(name)
  }
```

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
```
FloEdelmann: This is never called with multiple names, so can be simplified:


```suggestion
  /**
   * @param {string} name
   */
  addName(name) {
    this.names.add(name)
  }
```

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
+  return directivesNode.value.properties
+    .filter((node) => node.type === 'Property')
+    .map((node) => {
+      const name = utils.getStaticPropertyName(node)
+      return name ? { node, name } : null
+    })
+    .filter((res) => !!res)
+}
+
+class DefinedInSetupDirectives {
```
FloEdelmann: With `addName` only accepting a single name, this class is basically only a wrapper around `Set` with one accessor function. Maybe remove the class and use a `Set` directly with a separate function `isDefinedInSetupDirectives` instead?

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
+   * @param {string} rawName
+   */
+  isDefinedDirective(rawName) {
+    const camelName = camelize(rawName)
+    const variableName = `v${casing.capitalize(camelName)}`
+    return this.names.has(variableName)
+  }
+}
+
+class DefinedInOptionDirectives {
```
FloEdelmann: With `addName` only accepting a single name, this class is basically only a wrapper around `Set` with one accessor function. Maybe remove the class and use a `Set` directly with a separate function `isDefinedInOptionDirectives` instead?

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,260 @@
...
+      const ignored = ignorePatterns.some((pattern) =>
+        new RegExp(pattern).test(rawName)
+      )
+      return !ignored
+    }
+
+    /** @type { (rawName:string, reportNode: ASTNode) => void } */
+    let verifyName
+    /** @type {RuleListener} */
+    let scriptVisitor = {}
```
FloEdelmann: Instead of using `let` here, better `return` two distinct `utils.defineTemplateBodyVisitor` calls from inside `if (utils.isScriptSetup(context))` and `else`.

docs/rules/no-undef-directives.md
```diff
@@ -0,0 +1,89 @@
...
+{
+  "vue/no-undef-directives": ["error", {
+    "ignorePatterns": ["foo"]
+  }]
+}
+```
+
+- `ignorePatterns` (`string[]`) ... An array of regex pattern strings to ignore.
+
+### "ignorePatterns": ["foo"]
```
FloEdelmann: Sorry, I forgot the code quotes in my earlier suggestion

```suggestion
### `"ignorePatterns": ["foo"]`
```

tests/lib/rules/no-undef-directives.js
rzzf: Thank you for the review! ❤️
 With `toRegExpGroupMatcher`, `barfoobaz` shouldn’t pass the regex test, right? 🤔

FloEdelmann: * `"/foo/"` should match both `v-foo` `v-barfoobaz`
* `"foo"` should only match `v-foo` and not `v-barfoobaz`

lib/rules/no-undef-directives.js
rzzf: Thank you for your suggestion 👍, I’ve made the changes.

docs/rules/no-undef-directives.md
```diff
@@ -0,0 +1,90 @@
...
+
+```json
+{
+  "vue/no-undef-directives": ["error", {
+    "ignore": ["foo"]
+  }]
+}
+```
+
+- `"ignore"` (`string[]`) An array of tags or regular expression patterns (e.g. `"/^custom-/"`) that ignore these rules. This option will check both kebab-case and PascalCase versions of the given tag names. Default is empty.
```
FloEdelmann:
```suggestion
- `"ignore"` (`string[]`) An array of directive names or regular expression patterns (e.g. `"/^custom-/"`) that ignore these rules. This option will check both kebab-case and PascalCase versions of the given directive names. Default is empty.
```

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,217 @@
...
+
+      return utils.defineTemplateBodyVisitor(
+        context,
+        templateBodyVisitor,
+        scriptVisitor
+      )
+    } else {
+      // For Options API
+      /** @type {Set<string>} */
+      const definedInOptionDirectives = new Set()
```
FloEdelmann: This variable initialization is done in both `if` and `else`, so you can deduplicate it by moving it to the function scope.

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,217 @@
...
+          isDefinedInSetup(rawName, definedInSetupDirectives) ||
+          isDefinedInOptions(rawName, definedInOptionDirectives)
+      )
+
+      return utils.defineTemplateBodyVisitor(
+        context,
+        templateBodyVisitor,
+        scriptVisitor
+      )
+    } else {
```
FloEdelmann: `else` is not needed anymore when you `return` in the `if` case.

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,217 @@
...
+
+      const globalScope = context.sourceCode.scopeManager.globalScope
+      if (globalScope) {
+        for (const variable of globalScope.variables) {
+          definedInSetupDirectives.add(variable.name)
+        }
+        const moduleScope = globalScope.childScopes.find(
+          (scope) => scope.type === 'module'
+        )
+        for (const variable of (moduleScope && moduleScope.variables) || []) {
```
FloEdelmann: I think this is a little more concise:

```suggestion
        for (const variable of moduleScope?.variables ?? []) {
```

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,217 @@
...
+
+const utils = require('../utils')
+const casing = require('../utils/casing')
+const regexp = require('../utils/regexp')
+
+/**
+ * @param {string} str
+ * @returns {string}
+ */
+function camelize(str) {
```
FloEdelmann: Can you please use `casing.camelCase` instead of defining this function?

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,217 @@
...
+    return []
+  }
+
+  return directivesNode.value.properties
+    .filter((node) => node.type === 'Property')
+    .map((node) => {
+      const name = utils.getStaticPropertyName(node)
+      return name ? { node, name } : null
+    })
+    .filter((res) => !!res)
```
FloEdelmann: A little faster, because it only loops over the properties once:


```suggestion
  return directivesNode.value.properties.flatMap((node) => {
    const name = node.type === 'Property'
      ? utils.getStaticPropertyName(node)
      : null
    return name ? [{ node, name }] : []
  })
```

lib/rules/no-undef-directives.js
rzzf: Thank you again 🙇.  Yeah, this does make things simpler.

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,207 @@
...
+      return true
+    }
+  }
+
+  return false
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
```
waynzh: Align with other similar rules.

```suggestion
    type: 'suggestion',
```

And please run `npm run update` again.

lib/rules/no-undef-directives.js
```diff
@@ -0,0 +1,207 @@
...
+      {
+        type: 'object',
+        properties: {
+          ignore: {
+            type: 'array',
+            items: { type: 'string' },
+            uniqueItems: true
+          }
+        },
+        additionalProperties: true
```
waynzh:
```suggestion
        additionalProperties: false
```

lib/rules/no-undef-directives.js
rzzf: Thanks for your review, I’ve made the changes.

waynzh (approved):
Thanks!

lib/rules/no-undef-directives.js
```diff
@@ -36,21 +28,20 @@ function getRegisteredDirectives(componentObject) {
...
   }
 
-  return directivesNode.value.properties
-    .filter((node) => node.type === 'Property')
-    .map((node) => {
-      const name = utils.getStaticPropertyName(node)
-      return name ? { node, name } : null
-    })
-    .filter((res) => !!res)
+  // @ts-ignore
```
FloEdelmann: Why is a `@ts-ignore` comment needed?

rzzf: I wasn’t sure how to assert the `node` type before, but now I understand it. 
please review again, Thank you!

PR #2991, opened 2025-12-11
fix(vue/valid-v-for): allow using an empty alias in value

### LINKED ISSUE #2733: `vue/valid-v-for` should not trigger on looping through only values of an object
[vue/valid-v-for](https://eslint.vuejs.org/rules/valid-v-for.html)

fewer

new default behavior

```vue
  v-for="(, foo) in bar"
```

`error  Invalid alias ''  vue/valid-v-for`

not error out on looping through only keys of an Object

**Additional context**

rzzf:
resolves #2733

This is indeed valid syntax.
See [Vue Playground](https://play.vuejs.org/#eNp9kc1OwzAQhF9l5RNIoRE/p5AiAeoBDoCAoy9psglOHduynRAU+d1ZJ2rpAfVmz8yOv5Undm/MauiRZSx3pRXGg0PfG5CFatacecfZHVelVs6D3rawhglqrTO4TGBb2AyuINxylafLNGXp4rEzsvBIN4C8EgMMF7W21HeWwA5/zkGo2Ebd0xQFCCFPKUcDeXo0zRIioMdr0axapxVhTrGTs1J3Rki0r8YLguMsg9mJXiGl/n6eNW97TPZ6+YXl7h+9dWPUOHuz6NAOyNnB84Vt0C/25uMFRzofzE5XvaT0CfMdnZZ9ZFxiD72qCPsoN9M+dUZbL1Tz6TajR+X2S0XQmAxznjP6q8cTq//hXq9u5jmuAgu/BU6fZQ==)

### DIFF
--- a/.changeset/silly-falcons-explain.md
+++ b/.changeset/silly-falcons-explain.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fix `vue/valid-v-for` rule to allow empty value alias


tests/lib/rules/valid-v-for.js
```diff
@@ -183,11 +191,6 @@ tester.run('valid-v-for', rule, {
...
       errors: ["'v-for' directives require that attribute value."]
     },
-    {
-      filename: 'test.vue',
-      code: '<template><div><div v-for="(,a,b) in list"></div></div></template>',
-      errors: ["Invalid alias ''."]
-    },
     {
       filename: 'test.vue',
       code: '<template><div><div v-for="(a,,b) in list"></div></div></template>',
```
waynzh, line 196: This is also a valid syntax, we consider it as invalid since no one is actually using the object's index, right?

waynzh: If using the `index` makes sense, I wonder if `invalidEmptyAlias` was intentionally added to avoid potential errors. Maybe this rule recommends explicitly enumerating all variables before use (and pairing it with other rules to handle unused variables).

If that goes beyond this rule, I think cases like `(value, , index)`, `(value,,)`or `(, , index)`—while valid syntax—should also be allowed. (~or `invalidEmptyAlias` could just focus on cases like `(value,,)` where the alias is empty at the end~)

Let me know what you guys think.

rzzf: > If using the index makes sense, I wonder if invalidEmptyAlias was intentionally added to avoid potential errors.

To be honest, I’m not sure what the original background or motivation for `invalidEmptyAlias`.

> `invalidEmptyAlias` could just focus on cases like `(value,,)` where the alias is empty at the end.

I actually like this idea, it feels more reasonable to me. What do you think?
Under this definition, the behavior would be:

```js
// GOOD
(value, , index) in list
(, key) in list
(, key, index) in list
(,, index) in list

// BAD
(,) in list
(,,) in list
(,,,) in list
(value,,) in list
(value, key,,) in list
```

waynzh: I’d still lean towards keeping the check for `invalidEmptyAlias`.  

As an easy adjustment, we could provide an option to disable this check, such as adding an `allowEmptyAlias` option. By default, it could be set to `false`, but for those who prefer a different style, they could explicitly set it to `true` if needed.

If you're still interested, feel free to implement it during your free time! 😊 @rzzf

rzzf: Thanks for the suggestion. 🙇 
I’ll give it a try when I have some free time.

PR #2997, opened 2026-01-03
docs: improve editor interaction experience

rzzf:
Sometimes we want to type code directly in the documentation to quickly understand the details of the rules. However, typing “/” triggers the search input. It would be better if this behavior could be avoided.

### DIFF
--- a/docs/.vitepress/theme/components/eslint-code-block.vue
+++ b/docs/.vitepress/theme/components/eslint-code-block.vue
@@ -11,6 +11,7 @@
       dark
       :format="format"
       :fix="fix"
+      @keydown.stop
     />
   </div>
 </template>


FloEdelmann:
That does make sense to me, but I can't reproduce the bug you're experiencing. On which page and in which text field does typing `/` trigger the search bar instead of entering that character?

rzzf:
> That does make sense to me, but I can't reproduce the bug you're experiencing. On which page and in which text field does typing `/` trigger the search bar instead of entering that character?

Typing `/` in any code block triggers the search. My device is macOS 15.6.1.

This seems to only happen in Chrome and Edge.

<details>
<summary>video</summary>

https://github.com/user-attachments/assets/19e5a322-7cb5-4c16-a28c-1026b8bbcd34

</details>

waynzh (approved):
I can confirm the issue. Thanks for fixing it!

FloEdelmann (approved):
Ah indeed, I can confirm the bug and your fix in Chromium. Thank you! 🙂

PR #3000, opened 2026-01-14
feat: add vue/no-literals-in-template rule

### LINKED ISSUE #2516: Proposal for a New ESLint Rule: No Literals in Vue Templates

Object literals, array literals, and function literals can cause unnecessary re-renders in Vue components because these literals always generate new objects.

We hope to have a new ESLint rule that can alert us when we unconsciously use literals in Vue templates. It is very convenient and quick to casually write a literal, which can significantly degrade the performance of the component.

For example, in many sample codes of ant-design-vue, object literals are directly used. This is especially true for the table component. When rendering a large amount of data, even if data outside the table is modified, it will still cause the table to re-render.

[ ] Other (please specify:)

```vue
<div :arr="[]"></div>
<div :arr="[1,2,3]"></div>
<div :obj="{}"></div>
<div :obj="{name:'Tom', age: 123}"></div>
<div :callback="() => someFunction(someArgs)"></div>
```

**Additional context**

[https://github.com/vuejs/vue/issues/4060](https://github.com/vuejs/vue/issues/4060)

rzzf:
resolves #2516

### DIFF
--- a/lib/rules/no-literals-in-template.js
+++ b/lib/rules/no-literals-in-template.js
@@ -41,8 +41,8 @@ module.exports = {
      */
     function hasReferenceUpperElementVariable(node) {
       for (const element of upperElements) {
-        for (const vv of element.variables) {
-          for (const reference of vv.references) {
+        for (const variable of element.variables) {
+          for (const reference of variable.references) {
             const { range } = reference.id
             if (node.range[0] <= range[0] && range[1] <= node.range[1]) {
               return true


lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,57 @@
...
+    type: 'suggestion',
+    docs: {
+      description: 'disallow object literals in template',
+      categories: undefined,
+      url: 'https://eslint.vuejs.org/rules/no-literals-in-template.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      unexpected: 'Unexpected object literals in template.'
```
FloEdelmann: Please adapt the message depending on the type for each specific error,
i.e. `'Unexpected {{type}} literal in template.'`
with `data: { type: 'object' | 'array' | 'function' | 'arrow function' }`

rzzf: Done, thank you!

lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,62 @@
...
+  ArrayExpression: 'array',
+  FunctionExpression: 'function',
+  ArrowFunctionExpression: 'arrow function'
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow object literals in template',
```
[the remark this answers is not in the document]
FloEdelmann: Good catch. @rzzf could you please adjust the description and the rule docs and run `npm run update` again?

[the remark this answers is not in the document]
rzzf: sure 🚀

tests/lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,365 @@
...
+      code: '<template><div :class="{ active: isActive }"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-bind:class="[]"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-bind:class.camel="[]"></div></template>'
+    },
```
FloEdelmann: `class` can't have modifiers, can it? I'd just remove this test case (the rule doesn't have to have special logic for invalid usage).


```suggestion
```

tests/lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,365 @@
...
+      code: '<template><div :style="[baseStyles, overridingStyles]"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-bind:style="[]"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-bind:style.camel="[]"></div></template>'
+    }
```
FloEdelmann, line 160: Same for `style`:


```suggestion
    }
```

tests/lib/rules/no-literals-in-template.js
rzzf: You’re right. 👍

tests/lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,357 @@
...
+      code: `<template><div :style="{ color: 'red' }"></div></template>`
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div :style="[baseStyles, overridingStyles]"></div></template>'
+    },
+    {
+      filename: 'test.vue',
+      code: '<template><div v-bind:style="[]"></div></template>'
+    }
```
ota-meshi, line 160: Could you please add the tests for a template that uses a function with a variable defined in `v-for`?

```vue
<template>
  <template v-for="i in arr">
    <MyComponent :callback="() => someFunction(someArgs, i)" />
  </template>
</template>
```

I don't think this should be reported, but what do you think?

ota-meshi (changes requested):

tests/lib/rules/no-literals-in-template.js
rzzf: > I don't think this should be reported, but what do you think?

Thanks for your review. I have a few questions:

1. Is this rule limited to functions only?
2. If a function’s parameter list does not reference the v-for variables, should it still be reported?
3. Should scoped slots follow the same usage pattern as well?
… and some other cases that I haven’t thought of yet.
```html
<Child>
  <template #default="{ foo }">
    <ChildComponent :callback ="() => someFunction(someArgs, foo)" />
  </template>
</Child>
```

Handling these cases seems quite complex, and any guidance would be greatly appreciated. 🫡

ota-meshi: > Is this rule limited to functions only?

After thinking about it further, I think it might be necessary to exclude objects as well.


```vue
<template>
  <template v-for="i in arr">
    <MyComponent :data="{ index: i }" />
  </template>
</template>
```

> If a function’s parameter list does not reference the v-for variables, should it still be reported?

Yes. That's what I think.

> Should scoped slots follow the same usage pattern as well?

Yeah, v-slot and v-for are directives that allow users to define variables in templates.

https://github.com/vuejs/vue-eslint-parser/blob/master/docs/ast.md#:~:text=VEndTag%20%7C%20null-,variables,-%3A%20%5B%20Variable

---

I think the `v-on-handler-style`'s `hasReferenceUpperElementVariable` function might be helpful.

https://github.com/vuejs/eslint-plugin-vue/blob/e9d3f4366ae6bce8cdb7c162aa08a2f7d0e5ce33/lib/rules/v-on-handler-style.js#L268

rzzf: Thank you very much for your guidance. 👍 
If anything looks incorrect, please feel free to point it out or make changes directly.

lib/rules/no-literals-in-template.js
```diff
@@ -0,0 +1,91 @@
...
+    const upperElements = new Set()
+
+    /**
+     * Checks whether the given node refers to a variable of the element.
+     * @param {Expression | VForExpression | VOnExpression | VSlotScopeExpression | VFilterSequenceExpression} node
+     */
+    function hasReferenceUpperElementVariable(node) {
+      for (const element of upperElements) {
+        for (const vv of element.variables) {
+          for (const reference of vv.references) {
```
FloEdelmann:
```suggestion
        for (const variable of element.variables) {
          for (const reference of variable.references) {
```

PR #3001, opened 2026-01-16
feat: add `-error` configs to types

FloEdelmann:
Follow-up to #2796, where we forgot to update the type definitions.


FloEdelmann:
No changeset necessary, as the main MR (#2796) has one.

ota-meshi (approved):
👍

PR #3003, opened 2026-01-16
Update development dependencies

FloEdelmann:
See individual commit messages


netlify.toml
```diff
@@ -0,0 +1,2 @@
+[build.environment]
+  NODE_VERSION = "24"
```
ota-meshi, line 2: Even if I changed the Node version in the Netlify settings, Node 18 was still used, so I added this file to change the Node version.

docs/.vitepress/build-system/shim/empty.mjs
```diff
@@ -1,2 +1 @@
-export {}
```
FloEdelmann, line 1: The ESLint editors in the preview deployment are broken now, maybe because of this change.

FloEdelmann: Fixed with [`c6194cf`](https://github.com/vuejs/eslint-plugin-vue/pull/3003/commits/c6194cfb6e6efff7ac05edf212f3f7dea55b2a5d) and [`f10ff77`](https://github.com/vuejs/eslint-plugin-vue/pull/3003/commits/f10ff771be5fce0b292986c47da7f4ab41d3d31a). @ota-meshi could you please have a quick look at those commits?

ota-meshi: Thank you! It looks good to me!

PR #3002, opened 2026-01-16
refactor: move `lib/configs` to TypeScript

9romise:
Resolve #3004

This PR also upgrades `tsdown` to `0.20.3`, and fixes some incorrect test config.

PR #3007, opened 2026-01-20
refactor: migrate `lib/utils` to TypeScript

### LINKED ISSUE #3005: Internal utils no longer available for custom rules
This breaks custom Vue rules - not sure if officially supported - but it was working in past.

In 10.6 and before you could create your own rules by using `defineTemplateBodyVisitor` from `'eslint-plugin-vue/lib/utils/index.js'`.

But this is no longer available in the package.
Meaning you can no longer create your own Vue Template related rules.

_Originally posted by @susnux in https://github.com/vuejs/eslint-plugin-vue/issues/2916#issuecomment-3768816322_

9romise:
Resolve #3005


lib/utils/style-variables/index.ts
```diff
@@ -27,19 +26,13 @@ class StyleVariablesContext {
...
 }
 
-module.exports = {
-  getStyleVariablesContext,
-  StyleVariablesContext
-}
-
-/** @type {WeakMap<VElement, StyleVariablesContext>} */
-const cache = new WeakMap()
+const cache: WeakMap<VElement, StyleVariablesContext> = new WeakMap()
```
FloEdelmann:
```suggestion
const cache = new WeakMap<VElement, StyleVariablesContext>()
```

lib/utils/casing.ts
FloEdelmann, line 1: Since the type annotations are in TypeScript syntax rather than JSDoc now, I think the `@param` and `@returns` descriptions don't add much value over the JSDoc function descriptions anymore. Can you please remove them please? (But keep the JSDoc function descriptions!)

lib/utils/casing.ts
```diff
@@ -0,0 +1,173 @@
...
+export function getExactConverter(
+  name: CaseType
+): ConverterReturnType[CaseType] {
+  const converter = getConverter(name)
+  const checker = getChecker(name)
+  return (str) => {
+    const result = converter(str)
+    return checker(result) ? result : str /* cannot convert */
+  }
+}
```
FloEdelmann: My suggestion below…

1. …removes `@param` and `@returns`, as suggested above
2. …moves `type` declarations above "map" objects and adds type annotations for them
3. …replaces `type CheckerReturnType = Record<CaseType, (str: string) => boolean>` with `type CheckerType = (str: string) => boolean` and uses that directly instead of `CheckerReturnType[CaseType]` (likewise for the converter)

```suggestion
type CaseType = 'camelCase' | 'kebab-case' | 'PascalCase' | 'snake_case'
type ConverterType = (str: string) => string
type CheckerType = (str: string) => boolean

const convertersMap: Record<CaseType, ConverterType> = {
  'kebab-case': kebabCase,
  snake_case: snakeCase,
  camelCase,
  PascalCase: pascalCase
}
const checkersMap: Record<CaseType, CheckerType> = {
  'kebab-case': isKebabCase,
  snake_case: isSnakeCase,
  camelCase: isCamelCase,
  PascalCase: isPascalCase
}

/**
 * Return case checker function
 */
export function getChecker(name: CaseType): CheckerType {
  return checkersMap[name] || isPascalCase
}


/**
 * Return case converter function
 */
export function getConverter(name: CaseType): ConverterType {
  return convertersMap[name] || pascalCase
}

/**
 * Return case exact converter.
 * If the converted result is not in correct case, the original value is returned.
 */
export function getExactConverter(name: CaseType): ConverterType {
  const converter = getConverter(name)
  const checker = getChecker(name)
  return (str) => {
    const result = converter(str)
    return checker(result) ? result : str /* cannot convert */
  }
}
```

lib/utils/comments.ts
```diff
@@ -0,0 +1,8 @@
+export const isJSDocComment = (node: Comment): boolean =>
```
FloEdelmann: Instead of `boolean`, can you make this a type predicate (`node is BlockComment` or similar)? But I don't know if there is a type for block comment AST nodes. Even if we don't need it right now, it seems to make sense and should work as a boolean everywhere else.

(same for the function below)

lib/utils/html-comments.ts
```diff
@@ -1,45 +1,58 @@
...
-const TYPE_HTML_COMMENT_VALUE = 'HTMLCommentValue'
-/** @type { 'HTMLCommentClose' } */
-const TYPE_HTML_COMMENT_CLOSE = 'HTMLCommentClose'
-/** @type { 'HTMLCommentCloseDecoration' } */
-const TYPE_HTML_COMMENT_CLOSE_DECORATION = 'HTMLCommentCloseDecoration'
+const TYPE_HTML_COMMENT_OPEN = 'HTMLCommentOpen' as const
+const TYPE_HTML_COMMENT_OPEN_DECORATION = 'HTMLCommentOpenDecoration' as const
+const TYPE_HTML_COMMENT_VALUE = 'HTMLCommentValue' as const
+const TYPE_HTML_COMMENT_CLOSE = 'HTMLCommentClose' as const
+const TYPE_HTML_COMMENT_CLOSE_DECORATION = 'HTMLCommentCloseDecoration' as const
```
FloEdelmann: Is `as const` needed in a const declaration?

lib/utils/indent-utils.ts
FloEdelmann, line 1: Again

1. please remove `@param` and `@returns` where it is just boilerplate
2. return type predicates where possible

lib/utils/property-references.ts
FloEdelmann, line 1: Please keep the JSDoc descriptions for the classes and `extractFrom*` functions in this file.

lib/utils/ts-utils/ts-types.ts
FloEdelmann, line 1: Please keep the JSDoc descriptions for the `getComponent*FromTypeDefineTypes` functions.

lib/utils/ts-utils/index.ts
FloEdelmann, line 1: Please keep the JSDoc descriptions in this file.

FloEdelmann:
Similar to https://github.com/vuejs/eslint-plugin-vue/pull/3002#issuecomment-3898884467, this is now also unblocked, right?

PR #3008, opened 2026-01-20
fix(no-unused-vars): detect slot props used as component tags

genu:
## Problem
The rule incorrectly reports slot props as unused when they are used as component tags:

```vue
<template #wrapper="{ items, Iteration }">
  <Iteration v-for="(key, index) in items" :key="key" />
</template>
```
`Iteration` in this case is flagged as unused even though it's being used as a component.

## Solution
Fix false positive in `no-unused-vars` when slot props are used as component tags by adding a `isUsedAsComponentTag()` helper to detect when a destructured slot prop is used as a component tag in descendants

We then search the element's descendants for matching tag names -- we specifically use exact case matching since Vue requires slot props to match the tag name exactly (unlike registered components which support kebab-case).

### DIFF
--- a/.changeset/icy-pens-greet.md
+++ b/.changeset/icy-pens-greet.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': patch
 ---
 
-detect slot props used as component tags
+Fixed `vue/no-unused-vars` to detect components passed as slot props


.changeset/icy-pens-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+detect slot props used as component tags
```
FloEdelmann:
```suggestion
Fixed `vue/no-unused-vars` to detect components passed as slot props
```

lib/rules/no-unused-vars.js
```diff
@@ -49,6 +49,28 @@ function isDestructuringVar(variable) {
...
+/**
+ * Checks if the given variable name is used as a component tag in the element's descendants.
+ * @param {string} variableName the variable name to check
+ * @param {VElement} element the element node to search within
+ * @returns {boolean} `true` if the variable name is used as a component tag.
+ */
+function isUsedAsComponentTag(variableName, element) {
+  for (const child of element.children) {
+    if (child.type === 'VElement') {
+      // Check if the tag name matches the variable name (PascalCase component)
```
waynzh, line 61: Should components written in `kebab-case` be supported?

`<template><my-component v-slot="{ CompItem }"><comp-item>content</comp-item></my-component></template>`

FloEdelmann: @genu actually wrote this in the PR description:

> We then search the element's descendants for matching tag names -- we specifically use exact case matching since Vue requires slot props to match the tag name exactly (unlike registered components which support kebab-case).

waynzh: oh, thanks for the explanation.

PR #2962, opened 2025-11-21
feat: allow ESLint v10 as peer dependency

FloEdelmann:
Fixes https://github.com/vuejs/eslint-plugin-vue/issues/3017.

See https://eslint.org/blog/2026/02/eslint-v10.0.0-released/ and https://github.com/vuejs/vue-eslint-parser/pull/277.

PR #3012, opened 2026-01-22
feat(attributes-order): add `ignoreVBindObject` option

### LINKED ISSUE #2999: vue/attributes-order: add option to ignore v-bind="object" (attribute spread) from ordering/fixing
Hi! We use `vue/attributes-order` with a custom `order`. We want to keep attribute spread (`v-bind="attrs"`) after event listeners for readability, but the rule always reports and auto-fixes it to be placed before events.

This is especially painful because `v-bind="object"` is often used as a “rest props” pass-through and developers may want it at the end (or at least not enforced by the ordering rule).

**Repro (Vue 3):**
```vue
<template>
  <MyButton @click="onClick" v-bind="attrs" />
</template>
```

**Config:**
```js
{
  rules: {
    'vue/attributes-order': ['warn', {
      order: [
        'DEFINITION',
        ['CONDITIONALS', 'LIST_RENDERING'],
        'SLOT',
        'ATTR_STATIC',
        'ATTR_SHORTHAND_BOOL',
        'ATTR_DYNAMIC',
        'TWO_WAY_BINDING',
        'OTHER_DIRECTIVES',
        'RENDER_MODIFIERS',
        'CONTENT',
        'EVENTS',
      ],
    }],
  },
}
```

**Actual behavior:**
`vue/attributes-order` reports and auto-fixes by moving `v-bind="attrs"` before `@click`.

**Expected behavior:**
Provide a way to exclude `v-bind="object"` from ordering enforcement (and fixing), e.g.:

```js
'vue/attributes-order': ['warn', {
  order: [/* ... */],
  ignore: ['v-bind'], // ignore only v-bind="object"
}]
```

**Proposed semantics (minimal):**
- `ignore: ['v-bind']` ignores only `v-bind="object"` (directive `bind` with no argument).
- `v-bind:foo="bar"` and other directives are unaffected.

**Why not `eslint-disable-next-line`:**
We can workaround with `// eslint-disable-next-line vue/attributes-order` but it’s noisy and not scalable.

waynzh:
Resolve #2999

Currently, `v-bind="object"` is only excluded from sorting checks when adjacent to `ATTR_DYNAMIC`.  

Propose adding a new boolean option, `ignoreVBindObject`, to allow developers to globally exclude `v-bind="object"` from sorting checks, regardless of its position(skipped both near `ATTR_DYNAMIC` and `EVENT`).

### DIFF
--- a/docs/.vitepress/build-system/build.mts
+++ b/docs/.vitepress/build-system/build.mts
@@ -26,7 +26,7 @@ build(
 )
 
 build(
-  path.join(dirname, '../../../node_modules/vue-eslint-parser/index.js'),
+  path.join(dirname, '../../../node_modules/vue-eslint-parser/dist/index.cjs'),
   path.join(dirname, './shim/vue-eslint-parser.mjs'),
   [
     'node:path',
--- a/docs/rules/no-unused-properties.md
+++ b/docs/rules/no-unused-properties.md
@@ -15,7 +15,7 @@ since: v7.0.0
 This rule is aimed at eliminating unused properties.
 
 ::: warning Note
-This rule cannot check for use of properties by other components (e.g. `mixins`, property access via `$refs`) and use in places where the scope cannot be determined. Some access to properties might be implied, for example accessing data or computed via a variable such as `this[varName]`. In this case, the default is to assume all properties, methods, etc. are 'used'. See the `unreferencedOptions` for a more strict interpretation of 'use' in these cases.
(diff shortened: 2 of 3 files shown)


FloEdelmann (approved):
Looks good to me, thank you!

.changeset/dark-cars-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
```
FloEdelmann, line 2: Please don't merge this yet, I'd like to release a patch version first.

FloEdelmann: Since the TypeScript bugfixes took longer than expected and we'll want to have https://github.com/vuejs/eslint-plugin-vue/pull/2962 in the next release, I'll merge this now.

Issue #3010, opened 2026-01-21
Expose the generated types in a stable way


thebanjomatic:
**Tell us about your environment**

- **ESLint version:** 9.39.2
- **eslint-plugin-vue version:** 10.7.0
- **Vue version:** 3.4.x
- **Node version:** 24.18.0

**The problem you want to solve.**

In 10.4.0 typescript type augmentation of the eslint types was disabled. See https://github.com/vuejs/eslint-plugin-vue/issues/2788

Generally, I'm fine with this, but the types aren't able to be accessed in a consistent and reliable way since previously we could `import {RuleOptions as VueRuleOptions} from 'eslint-plugin-vue/lib/eslint-typegen.d.ts'` but in 10.7.0, the import path changed to: `eslint-plugin-vue/dist/eslint-typegen.d.ts`.

**Your take on the correct solution to problem.**

I'm not a fan of reaching into undeclared exports of a package like this, so a few ideas I guess are:

1. Add an `exports` field to package.json to declare a `./types` export to point to this package.
2. Re-export these types from the index.js

**Additional context**

FloEdelmann:
This is similar to #3005, but this time it actually is a user-facing API. So I think introducing `exports` would be a good idea, e.g. something like this:

```jsonc
"exports": {
  ".": {
    "import": "./dist/index.js", // does this break things, or should we rather leave it out?
    "require": "./dist/index.js",
    "types": "./dist/index.d.ts"
  },
  "./types": "./dist/eslint-typegen.d.ts",
  "./dist": "./dist", // to allow importing utils etc., but not officially supported
  "./lib": "./dist" // same, but maybe this is not even necessary or even confusing?
}
```

@9romise what do you think?

thebanjomatic:
A few thoughts:

For correctness, the "types" condition is supposed to go first in order to make sure typescript matches it instead of finding the .d.ts file that sits along-side the .js import. In this case the end result would be the same since it is the same file, but pointing it out because the order of the conditions matters so sorting them is an anti-pattern.

The "./dist" export needs to use a wildcard to be able to match against any files contained in dist and not just the directory itself.

You might also consider adding an export for package.json in case someone is importing/requiring that in order to get the version number or something like that. I found a few instances of people doing that sort of thing with require: https://github.com/search?q=eslint-plugin-vue%2Fpackage.json&type=code

```json
"exports": {
  ".": {
    "types": "./dist/index.d.ts",
[...]

PR #3013, opened 2026-01-23
feat: expose typegen types as a package export

thebanjomatic:
Adding an exports field to the package.json.

For compatibility with the existing 10.7.0 artifacts (where someone might be importing from `eslint-plugin-vue/dist/*`, we are also exposing those files through a wildcard export.

In a future major release, the wildcard will likely be dropped.

Closes: #3010


.changeset/twelve-weeks-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
```
thebanjomatic, line 2: Let me know if you feel that a patch is more appropriate for this, but I figured it is new functionality rather than a fix.

thebanjomatic:
```
[ERR_MODULE_NOT_FOUND]: Cannot find module '/home/runner/work/eslint-plugin-vue/eslint-plugin-vue/tests/integrations/eslint-plugin-import/node_modules/eslint-plugin-vue/dist/index.js' imported from /home/runner/work/eslint-plugin-vue/eslint-plugin-vue/tests/integrations/eslint-plugin-import/eslint.config.mjs
```
Hmm that's definitely not expected. I'll take a look.

PR #3025, opened 2026-02-14
feat(padding-line-between-tags): support :single-line/:multi-line pseudo-classes

### LINKED ISSUE #1974: [padding-line-between-tags] Support singleline/multiline options

`vue/padding-line-between-tags`

More warnings.

It could be an additional option like singleline/multiline.

```json
{ "blankLine": "always", "prev": "div-singleline", "next": "span-multiline" }
```

But it might be a bad example as `<div-singleline />` might be the correct tag/component. So it could be also:

```json
{ "blankLine": "always", "prevMulti": "div", "nextSingle": "span" }
```

```html
<div>
  This is a multiline tag
<div>
<span>This is a singleline tag</span>
```

Will trigger an error with configuration like that:

```json
{ "blankLine": "always", "prevMulti": "div", "nextSingle": "span" }
```

Moreover, it will be possible to force the required padding line between multiline and singleline tags in general.

For the ruleset like this:

```json
[
  { "blankLine": "always", "prevMulti":  "*", "nextSingle": "*" },
  { "blankLine": "always", "prevSingle": "*", "nextMulti":  "*" },
  { "blankLine": "always", "prevMulti":  "*", "nextMulti":  "*" },
  { "blankLine": "never",  "prevSIngle": "*", "nextSingle": "*" }
]
```

It will be possible to have code like this:

```html
<div>
  This is a multiline tag
<div>

<div>
  This is a multiline tag
<div>

<div>
  This is a multiline tag
<div>

<span>This is a singleline tag</span>
<span>This is a singleline tag</span>

<div>
  This is a multiline tag
<div>

<span>This is a singleline tag</span>

<div>
  This is a multiline tag
<div>

<span>This is a singleline tag</span>
<span>This is a singleline tag</span>
<span>This is a singleline tag</span>
```

Perdolique:
## Changes

Add `:single-line` and `:multi-line` pseudo-class support for `prev`/`next` selectors in `padding-line-between-tags` rule.

Fixes #1974

## Usage

```javascript
{
  "vue/padding-line-between-tags": ["error", [
    { "blankLine": "never",  "prev": "*:single-line", "next": "*:single-line" },
    { "blankLine": "always", "prev": "*:single-line", "next": "*:multi-line" },
    { "blankLine": "always", "prev": "*:multi-line",  "next": "*:single-line" },
    { "blankLine": "always", "prev": "*:multi-line",  "next": "*:multi-line" }
  ]]
}
```

Enforces:
```vue
<template>
  <div>
    <!-- No blank lines between single-line tags -->
    <span>inline</span>
    <span>inline</span>

    <!-- Blank line before/after multi-line tags -->
    <div>
      block
    </div>

    <span>inline</span>

    <div>
      block
    </div>
  </div>
</template>
```

Tag names work too, e.g. `"div:single-line"` or `"span:multi-line"`.

### DIFF
--- a/docs/rules/padding-line-between-tags.md
+++ b/docs/rules/padding-line-between-tags.md
@@ -169,10 +169,10 @@ You can use the `:single-line` and `:multi-line` pseudo-classes in `prev`/`next`
 ```json
 {
   "vue/padding-line-between-tags": ["error", [
-    { blankLine: 'always', prev: '*:single-line', next: '*:multi-line' },
-    { blankLine: 'always', prev: '*:multi-line', next: '*:single-line' },
-    { blankLine: 'always', prev: '*:multi-line', next: '*:multi-line' },
-    { blankLine: 'never', prev: '*:single-line', next: '*:single-line' }
+    { "blankLine": "always", "prev": "*:single-line", "next": "*:multi-line" },
+    { "blankLine": "always", "prev": "*:multi-line", "next": "*:single-line" },
+    { "blankLine": "always", "prev": "*:multi-line", "next": "*:multi-line" },
+    { "blankLine": "never", "prev": "*:single-line", "next": "*:single-line" }
   ]]
 }
 ```


ota-meshi:
Thank you for working on this PR!

However, I apologize for contradicting what was discussed in the issue (#1974), but I think
CSS selector-like options like `prev: "*:single-line"` and `prev: "*:multi-line"` are better than `prevSingle: "*"` and `prevMulti: "*"`.
I think there's potential for supporting CSS selector-like options in the future if users of the `padding-line-between-tags` rule want to set padding for specific classes. e.g. `prev: ".my-class"`
Pseudo-classes like `:single-line` don't exist in actual CSS, but I think these options would be easier to expand in the future.


For reference, the `block-order` rule currently supports CSS selector-like options.
However, the `padding-line-between-tags` rule doesn't yet need to fully support CSS selector-like options. This is only in case there's a request. For now, we can simply branch based on whether the option contains the string `:single-line` or `:multi-line`.

What do you think? @Perdolique 

cc @FloEdelmann

Perdolique:
@ota-meshi I'm ok with anything when it comes to naming. But in this specific case, it might add some confusion because, in the syntax `*:{something}`, `something` will have two meanings: a CSS pseudo-element and a linting-controlling option at the same time.

Perdolique:
For me personally, both cases will work. Even in the initially discussed proposal, it will be easy to depreciate `prevMulti`/`prevSingle`/`...` in the future, in case a wider range of CSS-selectors will be supported, and this syntax is no longer needed, as the current comparison is trivial and isolated in one function.

ota-meshi:
Thank you for your opinion!
I prefer the selector format, so could you please change it to the selector format if that's okay with you?

Perdolique:
Like this? 🤔

lib/rules/padding-line-between-tags.js
```diff
@@ -91,6 +91,54 @@ function removeExcessLines(context, endTag, sibling, lineDifference) {
...
+
+/**
+ * Match a selector string against an element.
+ * Supports pseudo-classes `:single-line` and `:multi-line`.
+ * @param {string} selector
+ * @param {VElement} element
+ * @returns {boolean}
+ */
+function matchesTag(selector, element) {
+  const colonIdx = selector.lastIndexOf(':')
```
FloEdelmann: Can you please rename the variable to `colonIndex`?

docs/rules/padding-line-between-tags.md
```diff
@@ -160,6 +162,42 @@ A configuration is an object which has 3 properties; `blankLine`, `prev` and `ne
...
+### Distinguish between single-line and multi-line tags
+
+You can use the `:single-line` and `:multi-line` pseudo-classes in `prev`/`next` to enforce different spacing rules based on whether tags are single-line or multi-line.
+
+<eslint-code-block fix :rules="{'vue/padding-line-between-tags': ['error', [
+  { blankLine: 'always', prev: '*:single-line', next: '*:multi-line' },
+  { blankLine: 'always', prev: '*:multi-line', next: '*:single-line' },
+  { blankLine: 'always', prev: '*:multi-line', next: '*:multi-line' },
+  { blankLine: 'never', prev: '*:single-line', next: '*:single-line' }
+]]}">
```
FloEdelmann, line 185: The configuration is not shown in the rendered docs, see https://deploy-preview-3025--eslint-plugin-vue.netlify.app/rules/padding-line-between-tags.html#distinguish-between-single-line-and-multi-line-tags

This makes it hard to understand the options when reading the docs. Can you please add a code block with the settings used in the code example?

lib/rules/padding-line-between-tags.js
```diff
@@ -91,6 +91,54 @@ function removeExcessLines(context, endTag, sibling, lineDifference) {
...
+
+/**
+ * Check if the configuration matches the given elements
+ * @param {{blankLine: string, prev: string, next: string}} configure
+ * @param {VElement} prevElement
+ * @param {VElement} nextElement
+ * @returns {boolean}
+ */
+function matchesConfiguration(configure, prevElement, nextElement) {
+  return matchesTag(configure.prev, prevElement) && matchesTag(configure.next, nextElement)
```
FloEdelmann: Please fix the lint issue here

FloEdelmann (changes requested):
Thanks for the implementation! I have to agree with @ota-meshi, using CSS selector syntax makes sense here 🙂

tests/lib/rules/padding-line-between-tags.js
```diff
@@ -280,6 +280,104 @@ tester.run('padding-line-between-tags', rule, {
...
+          </div>
+        </div>
+      </template>
+      `,
+      options: [
+        [
+          { blankLine: 'never', prev: '*:single-line', next: '*:single-line' },
+          { blankLine: 'always', prev: '*:multi-line', next: '*:multi-line' }
+        ]
+      ]
```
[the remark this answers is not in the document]
Perdolique: Added some tests, just in case.

docs/rules/padding-line-between-tags.md
```diff
@@ -59,8 +59,10 @@ A configuration is an object which has 3 properties; `blankLine`, `prev` and `ne
   - `always` requires one or more blank lines.
   - `never` disallows blank lines.
   - `consistent` requires or disallows a blank line based on the first sibling element.
-- `prev` any tag name without brackets.
-- `next` any tag name without brackets.
+- `prev` and `next` accept a tag name (e.g., `"div"`, `"br"`) or `"*"` to match any tag. Pseudo-classes can be appended to filter by layout:
+  - `"*:single-line"` - matches only single-line tags (opening and closing tags on the same line).
```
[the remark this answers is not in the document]
Perdolique: Changed

lib/rules/padding-line-between-tags.js
Perdolique: Done

docs/rules/padding-line-between-tags.md
Perdolique: Added

lib/rules/padding-line-between-tags.js
Perdolique: Done as well

docs/rules/padding-line-between-tags.md
```diff
@@ -160,6 +162,53 @@ A configuration is an object which has 3 properties; `blankLine`, `prev` and `ne
 
 </eslint-code-block>
 
+### Distinguish between single-line and multi-line tags
+
+You can use the `:single-line` and `:multi-line` pseudo-classes in `prev`/`next` to enforce different spacing rules based on whether tags are single-line or multi-line.
+
+```json
```
FloEdelmann, line 169: Please change the code in the code block to be valid JSON (double quotes instead of single quotes, quoted keys)

FloEdelmann (changes requested):
Please don't force-push, that makes it hard to review (I can't see the changes you made after I reviewed earlier). We'll squash all commits when merging this PR to `master` anyway.

docs/rules/padding-line-between-tags.md
Perdolique: Done. I thought you were reviewing changes between the latest reviewed changes and the new one. Tried to keep history clean. 😅

FloEdelmann: > I thought you were reviewing changes between the latest reviewed changes and the new one.

That's exactly what I was trying to do, but when you force-push multiple times in a row, GitHub only allows to view the last force-push's changes. (Try clicking on the "Compare" button [above](https://github.com/vuejs/eslint-plugin-vue/pull/3025#event-22984727193)).

But no worries, it wasn't too difficult for this PR :slightly_smiling_face:

Perdolique: I suppose this button is simply a toggle between the previous and selected changes. I'm about this feature that I'm using in the screenshot. Maybe it will be useful as well.

image

PR #3032, opened 2026-02-24
feat: extend `require-default-prop` and `require-valid-default-prop` to check `defineModel`

### LINKED ISSUE #2369: Extend `vue/require-valid-default-prop` to check inside defineModel

Extend `vue/require-valid-default-prop` to check defineModel

Potentially the same

Add support for defineModel to check that it has a default prop

```vue
const modelValue = defineModel({
  type: Boolean,
  default: false,
});
```

It does not check for a default value

A default value must be defined

**Additional context**

seanogdev:
Closes #2369.

### Changes

Extends require-default-prop and require-valid-default-prop to check defineModel calls. 

Since defineModel is effectively a prop+emit pair, both rules now enforce the same constraints they apply to defineProps: non-required, non-Boolean props must have a default value, and any provided default must be valid for the declared type.

Tests should validate the change.


lib/rules/require-default-prop.js
```diff
@@ -218,6 +218,32 @@ module.exports = {
...
+          if (
+            model.options.properties.some(
+              (p) =>
+                p.type === 'Property' &&
+                utils.getStaticPropertyName(p) === 'type' &&
+                isValueNodeOfBooleanType(p.value)
+            )
+          ) {
+            return
+          }
```
FloEdelmann:
```suggestion
          if (
            !model.options ||
            model.options.type !== 'ObjectExpression' ||
            propIsRequired(model.options) ||
            propHasDefault(model.options) ||
            model.options.properties.some(
              (p) =>
                p.type === 'Property' &&
                utils.getStaticPropertyName(p) === 'type' &&
                isValueNodeOfBooleanType(p.value)
            )
          ) {
            return
          }
```

tests/lib/rules/require-default-prop.test.ts
```diff
@@ -413,6 +413,103 @@ ruleTester.run('require-default-prop', rule, {
...
+      languageOptions: {
+        parser: require('vue-eslint-parser'),
+        ...languageOptions
+      }
+    },
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+      defineModel(String)
```
FloEdelmann: Why is this case allowed? Shouldn't this be reported as there is no default value?

tests/lib/rules/require-default-prop.test.ts
```diff
@@ -413,6 +413,103 @@ ruleTester.run('require-default-prop', rule, {
         ...languageOptions,
         parserOptions: { parser: require.resolve('@typescript-eslint/parser') }
       }
+    },
+    // defineModel — valid cases
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+      defineModel()
```
FloEdelmann: Why is this case allowed? Shouldn't this be reported as there is no default value?

tests/lib/rules/require-default-prop.test.ts
```diff
@@ -738,6 +835,58 @@ ruleTester.run('require-default-prop', rule, {
...
+      defineModel({ type: String })
+      </script>
+      `,
+      languageOptions: {
+        parser: require('vue-eslint-parser'),
+        ...languageOptions
+      },
+      errors: [
+        {
+          message: "Prop 'modelValue' requires default value to be set."
```
FloEdelmann: Please add at least `line` numbers to expected errors (better also `column`, `endLine` and `endColumn`)

tests/lib/rules/require-default-prop.test.ts
```diff
@@ -413,6 +413,103 @@ ruleTester.run('require-default-prop', rule, {
...
+    // defineModel — valid cases
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+      defineModel()
+      </script>
+      `,
+      languageOptions: {
+        parser: require('vue-eslint-parser'),
```
FloEdelmann: Please use the imported `vueEslintParser` instead of `require('vue-eslint-parser')` (also in the other test file)

tests/lib/rules/require-valid-default-prop.test.ts
```diff
@@ -340,6 +340,62 @@ ruleTester.run('require-valid-default-prop', rule, {
...
+    // defineModel — no default (nothing to check)
+    {
+      filename: 'test.vue',
+      code: `
+      <script setup>
+        defineModel({ type: Boolean })
+      </script>
+      `,
+      languageOptions: { parser: require('vue-eslint-parser') }
+    },
```
FloEdelmann, line 344:
```suggestion
    {
      filename: 'test.vue',
      code: `
      <script setup>
        defineModel({ type: Boolean })
      </script>
      `,
      languageOptions: { parser: vueEslintParser }
    },
    {
      filename: 'test.vue',
      code: `
      <script setup>
        defineModel({ type: String, required: true })
      </script>
      `,
      languageOptions: { parser: vueEslintParser }
    },
```

tests/lib/rules/require-valid-default-prop.test.ts
FloEdelmann, line 1: Please also add test cases to both rules that use the TypeScript syntax (e.g. `defineModel<string>({ required: true })`) and make sure they are handled correctly.

FloEdelmann (changes requested):
Sorry for the late review!

In general the code looks good, but I have a few remarks, see below.

tests/lib/rules/require-default-prop.test.ts
seanogdev: Good shout. I think I was too insistent on matching the prop behavior, but there's a distinct difference here alright

FloEdelmann (changes requested):
@seanogdev can you please have a look at the Copilot suggestions above?

lib/rules/require-default-prop.js
```diff
@@ -132,6 +134,32 @@ module.exports = {
...
+     * Boolean prop) does not require a default value.
+     * @param {ComponentModel} model
+     * @param {ObjectExpression | null} options The model's options object, if any.
+     * @return {boolean}
+     */
+    function isBooleanModel(model, options) {
+      if (
+        options &&
+        options.properties.some(
+          (p) =>
```
FloEdelmann: No single-letter variable names please.

lib/rules/require-default-prop.js
```diff
@@ -212,6 +240,27 @@ module.exports = {
...
+              : null
+
+          // A required model, or one that already declares a default, is fine.
+          if (options && (propIsRequired(options) || propHasDefault(options))) {
+            return
+          }
+          // Boolean models don't require a default value, just like Boolean props.
+          if (isBooleanModel(model, options)) {
+            return
+          }
```
FloEdelmann: I think the comments are not needed.

```suggestion
          if (options && (propIsRequired(options) || propHasDefault(options))) {
            return
          }

          if (isBooleanModel(model, options)) {
            return
          }
```

FloEdelmann (changes requested):
Thanks for the quick update! I have two small code-style suggestions.

seanogdev:
Totally forgot about this, sorry!

lib/rules/require-default-prop.js
```diff
@@ -132,6 +134,38 @@ module.exports = {
...
+                utils.getStaticPropertyName(property) === 'type' &&
+                isValueNodeOfBooleanType(property.value)
+            )
+          ) {
+            return true
+          }
+        } else if (isValueNodeOfBooleanType(options)) {
+          // Shorthand constructor form, e.g. `defineModel(Boolean)`.
+          return true
+        }
```
FloEdelmann: These can be turned around to save some indentation:

```suggestion
        if (isValueNodeOfBooleanType(options)) {
          // Shorthand constructor form, e.g. `defineModel(Boolean)`.
          return true
        }
        if (
          options.type === 'ObjectExpression' &&
          options.properties.some(
            (property) =>
              property.type === 'Property' &&
              utils.getStaticPropertyName(property) === 'type' &&
              isValueNodeOfBooleanType(property.value)
          )
        ) {
          return true
        }
```

lib/rules/require-default-prop.js
```diff
@@ -212,6 +246,26 @@ module.exports = {
...
+              ? model.options
+              : null
+
+          if (options && (propIsRequired(options) || propHasDefault(options))) {
+            return
+          }
+
+          if (isBooleanModel(model)) {
+            return
+          }
```
FloEdelmann: This could be the first check before `const options` so that variable is only initialized if needed.

FloEdelmann (changes requested):
No worries 🙂

FloEdelmann (approved):

.changeset/lucky-waves-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+feat: extend `require-default-prop` and `require-valid-default-prop` to check `defineModel`
```
waynzh:
```suggestion
Added `defineModel` support to `vue/require-default-prop` and `vue/require-valid-default-prop` rule
```

.changeset/lucky-waves-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
```
waynzh:
```suggestion
"eslint-plugin-vue": minor
```

waynzh (approved):
Thanks!

troyere:
It is not at all the same thing; providing default values ​​for every `defineModel` is not necessarily advisable and certainly should not be done systematically : 

image

Source : https://vuejs.org/guide/components/v-model.html#under-the-hood

PR #3034, opened 2026-02-26
refactor(test): improve `ts-utils` testing

9romise:
Tip: per‑commit review might be easier than a full‑diff review.

- Migrate to snapshot testing.
- Migrate to TypeScript.
- Fix cross-file race condition: The original tests all shared a single `src/test.ts` scratch file. When Vitest runs test files in parallel, concurrent writes to this file caused flaky failures. Split into `test-props.ts` and `test-emits.ts` so each test file has its own fixture.
- Extract shared `verifyWithTsParser` helper.

### DIFF
--- a/tests/lib/utils/ts-utils/index/get-component-slots.ts
+++ b/tests/lib/utils/ts-utils/index/get-component-slots.ts
@@ -1,60 +1,20 @@
-import type { Linter as ESLintLinter } from 'eslint'
-import assert from 'node:assert'
 import path from 'node:path'
-import tsParser from '@typescript-eslint/parser'
-import vueEslintParser from 'vue-eslint-parser'
-import { Linter } from '../../../../eslint-compat'
-import { defineScriptSetupVisitor } from '../../../../../lib/utils/index'
+import { FIXTURES_ROOT, verifyWithTsParser } from './utils'
 
-const FIXTURES_ROOT = path.resolve(
-  __dirname,
-  '../../../../fixtures/utils/ts-utils'
-)
-const TSCONFIG_PATH = path.resolve(FIXTURES_ROOT, './tsconfig.json')
 const SNAPSHOT_ROOT = path.resolve(FIXTURES_ROOT, './get-component-slots')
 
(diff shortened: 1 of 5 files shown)


tests/lib/utils/ts-utils/index/get-component-slots.ts
```diff
@@ -0,0 +1,61 @@
...
+          type: slot.type,
+          name: slot.slotName
+        }))
+      )
+    }
+  })
+  return result
+}
+
+describe('getComponentSlotsFromTypeDefineTypes', () => {
```
FloEdelmann, line 21: No `.sequential` needed here? Or is it not needed anywhere anymore after your refactoring?

vitest.config.mts
```diff
@@ -8,7 +8,8 @@ export default defineConfig({
       '**/dist/**',
       'tests/fixtures/**',
       'tests/integrations/flat-config/eslint.config.js',
-      'tests/lib/rules/no-unsupported-features/utils.js'
+      'tests/lib/rules/no-unsupported-features/utils.js',
+      'tests/lib/utils/ts-utils/index/utils.ts'
```
FloEdelmann, line 12: Should we maybe exclude all util files with something like `**/utils.{js,ts}`?

tests/lib/utils/ts-utils/index/get-component-slots.ts
9romise: The other two tests use the same `test.ts` (now `test-props.ts`/`test-emits.ts`), so they have to run sequentially to avoid conflicts.
`get-component-slots` doesn't touch `test.ts`, so no race condition and no `.sequential` needed.
Maybe we could give each test its own temp file, so they can run in parallel and we don’t have to commit empty test files.

vitest.config.mts
9romise: This seems like a good approach for now, but I think we could perhaps switch to using `.test.ts` / `.spec.ts` files for all tests in the future.

PR #3036, opened 2026-02-27
refactor(test): migrate `rules` to TypeScript

9romise:
- Move all test files to TypeScript

Test files count:
[Before](https://github.com/vuejs/eslint-plugin-vue/actions/runs/22436023448/job/64966030962):
image

[After](https://github.com/vuejs/eslint-plugin-vue/actions/runs/22475429291/job/65101318114?pr=3036):
image

PR #3038, opened 2026-02-28
refactor(test): migrate `utils` to TypeScript

9romise:
part of https://github.com/vuejs/eslint-plugin-vue/pull/3036

### DIFF
--- a/tests/lib/utils/index.test.ts
+++ b/tests/lib/utils/index.test.ts
@@ -467,7 +467,7 @@ describe('editdistance', () => {
 })
 function nodeToJson(nodes: ESNode[], ignores: string[] = []) {
   return JSON.stringify(nodes, replacer, 2)
-  function replacer(key: string, value: string) {
+  function replacer(key: string, value: unknown) {
     if (key === 'parent' || key === 'start' || key === 'end') {
       return undefined
     }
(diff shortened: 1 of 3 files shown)


tests/lib/utils/regexp.test.ts
```diff
@@ -21,7 +19,9 @@ describe('escape()', () => {
   })
 
   it('should return an empty string for empty values', () => {
+    // @ts-expect-error escape only accept `string`
     assert.strictEqual(escape(null), null)
+    // @ts-expect-error escape only accept `string`
```
FloEdelmann:
```suggestion
    // @ts-expect-error escape only accepts `string`
    assert.strictEqual(escape(null), null)
    // @ts-expect-error escape only accepts `string`
```

tests/lib/utils/vue-component.test.ts
```diff
@@ -1,32 +1,38 @@
 /**
  * @author Armano
  */
-'use strict'
+import { Linter } from 'eslint'
```
FloEdelmann:
```suggestion
import type { Linter } from 'eslint'
```

tests/lib/utils/index.test.ts
```diff
@@ -430,9 +465,9 @@ describe('editdistance', () => {
     assert.equal(editDistance('computed', 'computd'), 1)
   })
 })
-function nodeToJson(nodes, ignores = []) {
+function nodeToJson(nodes: ESNode[], ignores: string[] = []) {
   return JSON.stringify(nodes, replacer, 2)
-  function replacer(key, value) {
+  function replacer(key: string, value: string) {
```
FloEdelmann: `value` should be type `unknown`

PR #3054, opened 2026-03-09
refactor: migrate indent related utils and rules to TypeScript

9romise:


lib/utils/indent-common.ts
```diff
@@ -347,14 +326,14 @@ module.exports.defineVisitor = function create(
    * Process the given node list.
    * The first node is offsetted from the given left token.
    * Rest nodes are adjusted to the first node.
-   * @param {(MaybeNode|null)[]} nodeList The node to process.
-   * @param {MaybeNode|Token|null} left The left parenthesis token.
-   * @param {MaybeNode|Token|null} right The right parenthesis token.
-   * @param {number} offset The offset to set.
-   * @param {boolean} [alignVertically=true] The flag to align vertically. If `false`, this doesn't align vertically even if the first node is not at beginning of line.
```
FloEdelmann, line 354: Please keep this description

lib/utils/indent-common.ts
```diff
@@ -317,11 +295,12 @@ module.exports.defineVisitor = function create(
   /**
    * Get the first and last tokens of the given node.
    * If the node is parenthesized, this gets the outermost parentheses.
-   * @param {MaybeNode} node The node to get.
-   * @param {number} [borderOffset] The least offset of the first token. Defailt is 0. This value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
-   * @returns {{firstToken:Token,lastToken:Token}} The gotten tokens.
+   * This `borderOffset` value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
```
FloEdelmann:
```suggestion
   * The `borderOffset` value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
```

lib/utils/indent-common.js
```diff
@@ -824,12 +777,12 @@ module.exports.defineVisitor = function create(
 
   /**
    * Validate indentation of the line that the given tokens are on.
-   * @param {Token[]} tokens The tokens on the same line to validate.
-   * @param {Token[]} comments The comments which are on the immediately previous lines of the tokens.
-   * @param {Token|null} lastToken The last validated token. Comments can adjust to the token.
```
FloEdelmann, line 829: Please keep these descriptions.

lib/utils/indent-ts.ts
```diff
@@ -2,154 +2,110 @@
...
- * @typedef {import('@typescript-eslint/types').TSESTree.TSOptionalType} TSOptionalType
- * @typedef {import('@typescript-eslint/types').TSESTree.TSNonNullExpression} TSNonNullExpression
- * @typedef {import('@typescript-eslint/types').TSESTree.TSAsExpression} TSAsExpression
- * @typedef {import('@typescript-eslint/types').TSESTree.TSSatisfiesExpression} TSSatisfiesExpression
- * @typedef {import('@typescript-eslint/types').TSESTree.TSTypeReference} TSTypeReference
- * @typedef {import('@typescript-eslint/types').TSESTree.TSInstantiationExpression} TSInstantiationExpression
- * @typedef {import('@typescript-eslint/types').TSESTree.JSXChild} JSXChild
- * @typedef {import('@typescript-eslint/types').TSESTree.TypeNode} TypeNode
- *
+ * Deprecated in @typescript-eslint/parser v5
```
FloEdelmann, line 21: Is this addition correct and intended?

9romise: This is not an addition. I just inherited the comment.

FloEdelmann: Ah, now I understand: Previously the comment referred to both `ClassProperty` and `TSAbstractClassProperty` typedefs inside the comment. Now you added it to both type definitions. So that's fine.

lib/utils/indent-utils.ts
```diff
@@ -42,10 +40,8 @@ function isNotWhitespace(token) {
 
 /**
  * Check whether the given token is a comment.
- * @param {Token|undefined|null} token The token to check.
- * @returns {boolean} `true` if the token is a comment.
  */
-function isComment(token) {
+export function isComment(token: Token | undefined | null): token is Comment {
```
FloEdelmann: `isComment` is declared as a type guard to `Comment`, but the implementation also returns true for tokens like `Shebang` and any token whose type ends with `Comment` (e.g. `HTMLComment`). Those token shapes are not compatible with the Comment type alias, so this type predicate is unsound and can lead to incorrect narrowing. Consider changing the predicate return type to a broader comment-token union (e.g. `Token & { type: ... }`) or drop the type predicate and return `boolean`.

lib/utils/indent-utils.ts
```diff
@@ -78,40 +74,23 @@ function isNotComment(token) {
 
 /**
  * Check whether the given node is not an empty text node.
- * @param {ASTNode} node The node to check.
- * @returns {boolean} `false` if the token is empty text node.
  */
-function isNotEmptyTextNode(node) {
+export function isNotEmptyTextNode(node: ASTNode): node is VText {
```
FloEdelmann: `isNotEmptyTextNode` declares `node is VText`, but it returns true for any non-`VText` node as well. This type guard is incorrect (it can narrow element/container nodes to `VText`). Either remove the type predicate (return `boolean`) or adjust the predicate to reflect that it keeps non-`VText` nodes and only filters out empty `VText` nodes.

lib/utils/indent-common.ts
```diff
@@ -1890,15 +1818,13 @@ module.exports.defineVisitor = function create(
     // ----------------------------------------------------------------------
     // COMMONS
     // ----------------------------------------------------------------------
-    /** @param {Statement} node */
     // Process semicolons.
-    ':statement, PropertyDefinition'(node) {
+    ':statement, PropertyDefinition'(node: Statement) {
```
FloEdelmann: This visitor key matches both `:statement` and `PropertyDefinition`, but the parameter is typed as `Statement` only. That annotation is incorrect for `PropertyDefinition` callbacks and weakens type-safety. Update the parameter type to include `PropertyDefinition` (or a shared base like `ASTNode`/`MaybeNode`).

lib/utils/indent-ts.ts
```diff
@@ -2,154 +2,110 @@
...
- * @property {ParserServices.TokenStore | SourceCode} tokenStore
- * @property {SetOffset} setOffset
- * @property {CopyOffset} copyOffset
- * @property {ProcessSemicolons} processSemicolons
- * @property {GetFirstAndLastTokens} getFirstAndLastTokens
  */
+type GetFirstAndLastTokens = (
+  node: MaybeNode,
+  /**
+   * The least offset of the first token. Defailt is 0. This value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
```
FloEdelmann:
```suggestion
   * The least offset of the first token. Default is 0. This value is used to prevent false positive in the following case: `(a) => {}` The parentheses are enclosing the whole parameter part rather than the first parameter, but this offset parameter is needed to distinguish.
```

lib/utils/indent-utils.ts
```diff
@@ -75,7 +75,9 @@ export function isNotComment(
 /**
  * Check whether the given node is not an empty text node.
  */
-export function isNotEmptyTextNode(node: ASTNode): node is VText {
+export function isNotEmptyTextNode(
+  node: ASTNode
+): node is Exclude<ASTNode, VText> {
   return !(node.type === 'VText' && node.value.trim() === '')
```
FloEdelmann: That's not correct either, is it?

The function returns true for non-`VText` nodes, but also for non-empty `VText` nodes. So we can not deduce anything for the return type, i.e. it should be `boolean` only.

9romise: my fault 🤦‍♂️

PR #3057, opened 2026-03-11
chore: enable internal rules for `.ts` rules

9romise:


eslint.config.mjs
FloEdelmann, line 317: Found by Copilot: With this glob now including `.ts` rule sources, `internal/no-invalid-meta-docs-categories` will run on TS rules that use `export default { ... }`, but that internal rule currently only detects `module.exports = ...` assignments (no `ExportDefaultDeclaration` handler) and will silently skip validation for TS rules. To actually enforce categories on TS rules, update `eslint-internal-rules/no-invalid-meta-docs-categories.js` to capture `default` exports similarly to `no-invalid-meta.js` (and/or share a common export-detection helper).

PR #3067, opened 2026-04-12
fix: false positive for returns in exhaustive switch

### LINKED ISSUE #2100: `vue/require-render-return` false positive for switch-case on enum without default
**Checklist**

- **ESLint version:** 8.4.10
- **eslint-plugin-vue version:** 8.7.1
- **Node version:** 18.13.0

```json5
/* eslint-env node */
require("@rushstack/eslint-patch/modern-module-resolution");

module.exports = {
  root: true,
  plugins: ["vitest"],
  extends: [
    "plugin:vue/vue3-essential",
    "eslint:recommended",
    "plugin:jest-formatting/recommended",
    "@vue/eslint-config-typescript/recommended",
    "@vue/eslint-config-prettier",
  ],
  env: {
    "vue/setup-compiler-macros": false,
  },
  rules: {
    eqeqeq: ["warn", "always"],

    // https://github.com/veritem/eslint-plugin-vitest
    // "vitest/expect-expect": "warn", // TODO: enable after https://github.com/veritem/eslint-plugin-vitest/issues/20 is actually fixed
    // "vitest/max-nested-describe": "warn", // TODO: enable after https://github.com/veritem/eslint-plugin-vitest/issues/23 is fixed
    // "vitest/lower-case-title": "warn", // I don't like that it converts all my describe blocks with camelCase method names...

    "vitest/no-conditional-tests": "warn",
    "vitest/no-focused-tests": "warn",
    "vitest/no-identical-title": "warn",
    "vitest/no-skipped-tests": "warn",
    "require-await": "warn",
  },
  overrides: [
    {
      files: ["cypress/integration/**.spec.{js,ts,jsx,tsx}"],
      extends: ["plugin:cypress/recommended"],
    },
  ],
};

```

**What did you do?**	

```vue
import { defineComponent, type PropType } from "vue";

enum DialogType {
  First,
  Second,
}

  name: "Dialog",
  props: {
    dialogType: {
      type: Number as PropType<DialogType>,
      required: true,
    },
  },
  render() {
    switch (this.dialogType) {
      case DialogType.First:
        return <h1>First</h1>;
      case DialogType.Second:
        return <h1>Second</h1>;
    }
  },
});

```

No error because in my switch-case I've exhausted all the enum options, so no `default` should be needed.

**What actually happened?**

```
  16:3  error  Expected to return a value in render function        vue/require-render-return
```

### LINKED ISSUE #2142: vue/return-in-computed-property requires return even if Typescript union type is exhausted in switch statement
**Checklist**

- **ESLint version:** 8.39.0
- **eslint-plugin-vue version:** 9.11.0
- **Node version:** Problem in both v20.0.0 and v16.14.2
- **Operating System:** macOS 13.3.1

```js
/* eslint-env node */

module.exports = {
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  }
}
```
(Default Vue 3 setup)

**What did you do?**	

```vue
<script setup lang="ts">

interface Foo {
  category: 'foo'
}

interface Baz {
  category: 'baz'
}

type unionType = Foo | Baz

//Use Math random so typescript can't infer the type
const foo: unionType = Math.random() > 0.5 ? { category: 'foo' } : { category: 'baz' }

const computedStuff = computed(() => {  //This is the line that throws the error
  switch (foo.category) {
    case 'foo':
      return 'foo'
    case 'baz':
      return 'baz'
  }
})

</script>
```

Expected ESLint to recognize that the switch exhausts the union type and therefore a value would always be returned

**What actually happened?**
```
  17:32  error  Expected to return a value in computed function  vue/return-in-computed-property
```

ESLint complains and says the computed property is missing a return value.

https://github.com/Larsmyrup/return-in-computed-property-bug-repro

Let me know if i need to elaborate on anything

EdRW:
- resolves #2142
- resolves #2100

## Description

Currently `return-in-computed-property` and `require-render-return` report an error for switch statements without a `default` case, when the switch is the final statement. For Typescript users the error is shown even when the final statement is a switch statements that exhaustively calls `return` or `throw` for every case. 

This PR updates [`executeOnFunctionsWithoutReturn`](https://github.com/EdRW/eslint-plugin-vue/blob/303e74496186c4a8273e04c1975015f225d37377/lib/utils/index.js#L1822) to use `typescript-eslint/parser` services (when available) to verify exhaustive switch statements.

**No behavior changes for Javascript users.** Switch statements without `default` case will still show an error when typing info is not available.

 ### Supported patterns

This implementation borrows much from typescript-eslint's implementation of [switch-exhaustiveness-check.ts](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts), and supports:

  - String/number/bigint literal unions
  - Numeric and string enums
  - Boolean (`true | false`)
  - Nullable types (`T | null`, `T | undefined`)
  - Branded/intersection types
  - Discriminated unions (e.g., `switch (event.type)`)
  - Generic components
  - Arbitrary nesting in `if`/`else` and `try`/`catch`/`finally`

Most of the lines added in this PR are just test cases to check all of these patterns.

## Screenshots

<details>
<summary>

### return-in-computed-property

</summary>
Examples come from #2142

#### Exhaustive

Screenshot 2026-04-12 at 20 38 57

#### Non-exhaustive

Screenshot 2026-04-12 at 20 38 49

</details>

<details>
<summary>

### require-render-return

</summary>

Examples come from #2100

#### Exhaustive

Screenshot 2026-04-12 at 21 05 10

#### Non-exhaustive

Screenshot 2026-04-12 at 21 09 00

</details>


lib/utils/ts-utils/ts-types.js
```diff
@@ -334,3 +335,130 @@ function* iterateTypes(type) {
...
+function isTypeLiteralLike(type, ts) {
+  return (
+    (type.flags &
+      (ts.TypeFlags.Literal |
+        ts.TypeFlags.Undefined |
+        ts.TypeFlags.Null |
+        ts.TypeFlags.UniqueESSymbol)) !==
+    0
+  )
+}
```
EdRW, line 369: Basically copied from [switch-exhaustiveness-check.ts#L415](https://github.com/typescript-eslint/typescript-eslint/blob/7c9e06f669065bfb24b66c1f2bfb05ae2b512c09/packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts#L415)

lib/utils/ts-utils/ts-types.js
```diff
@@ -334,3 +335,130 @@ function* iterateTypes(type) {
...
+  return type.isUnion() ? type.types : [type]
+}
+
+/**
+ * @param {Type} type
+ * @returns {Type[]}
+ */
+function intersectionConstituents(type) {
+  return type.isIntersection() ? type.types : [type]
+}
```
EdRW, line 353: Inlined TS utility functions from [`ts-api-utils`](https://github.com/JoshuaKGoldberg/ts-api-utils/blob/417acd9459c6ba059ae1060e86ce77917b4b0776/src/types/utilities.ts#L41)

lib/utils/ts-utils/ts-types.js
```diff
@@ -334,3 +335,130 @@ function* iterateTypes(type) {
...
+  }
+  return false
+}
+
+/**
+ * @param {RuleContext} context
+ * @param {SwitchStatement} switchNode
+ * @returns {boolean}
+ */
+function isExhaustiveSwitch(context, switchNode) {
```
EdRW, line 392: This function is derived from [getSwitchMetadata in switch-exhaustiveness-check.ts](https://github.com/typescript-eslint/typescript-eslint/blob/7c9e06f669065bfb24b66c1f2bfb05ae2b512c09/packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts#L155)

lib/utils/ts-utils/ts-types.js
EdRW, line 1: **question**
Most of the work was done here because it's TS related, but I wasn't sure if this is actually the most appropriate location.
Is there a more suitable place or is it ok keep the exhaustive switch-related logic here?

tests/lib/rules/return-in-computed-property.test.ts
EdRW, line 1: Since `return-in-computed-property` and `require-render-return` share the same logic for exhaustive switch statements, the thorough tests of the full range of supported types probably only need to happen in one place. So they were added here.

tests/lib/rules/return-in-computed-property.test.ts
```diff
@@ -144,6 +145,277 @@ ruleTester.run('return-in-computed-property', rule, {
...
+        import { computed, ref } from 'vue'
+        const flag = ref<boolean>(true)
+        const result = computed(() => {
+          switch (flag.value) {
+            case true: return 'yes'
+            case false: return 'no'
+          }
+        })
+        </script>`,
+      ...getTypeScriptFixtureTestOptions()
```
ota-meshi, line 214: (edited)
Could you please add the following test case?

```vue
<script setup lang="ts">
import { computed, ref } from 'vue'
const flag = ref<boolean>(true)
const result = computed(() => {
  if (x !== 0) {
    switch (flag.value) {
      case true: return 'yes'
      case false: return 'no'
    }
  } else {
    return 'x === 0'
  }
})
</script>
```


Is it supported?

ota-meshi (changes requested):
Thank you for this PR!

tests/lib/rules/return-in-computed-property.test.ts
EdRW: Thanks for the review and suggestion. 

> Is it supported?

Unfortunately, it's not supported yet. I mainly focused on the case where switch is the last statement in the function. In this example, the switch is nested in an if-statement.

> Could you please add the following test case?

Sure I can try. It think it'll involve a lot of recursive traversal though😨 . I might also take a look at try/catch as well.

ota-meshi: In my opinion, rather than using recursion, it would probably be better to use ESLint's Code Path Analysis to process the relevant switch statements.

https://eslint.org/docs/latest/extend/code-path-analysis

However, you don't need to add support in this PR :smile: 
I think even just the changes in this PR are a valuable improvement.
Could you please add test cases and leave comments about anything that isn't yet supported?

EdRW: @ota-meshi Sorry for the delay in updating the PR. Would you mind taking another look when you have a chance? 

> In my opinion, rather than using recursion, it would probably be better to use ESLint's Code Path Analysis to process the relevant switch statements.
> 
> [eslint.org/docs/latest/extend/code-path-analysis](https://eslint.org/docs/latest/extend/code-path-analysis)

Thanks so much for pointing me in the right direction. First I explored a recursive solution but felt like I was "reinventing the wheel". 
After switching to the eslint code path analysis approach, I was able to handle even more complex test cases. I think most common use-cases should be supported now 🎉 

> Could you please add test cases and leave comments about anything that isn't yet supported?

I added test cases similar to the case you suggested, as well as many many more cases in the below commits:

- 7d9c017
- ff29ea9
- bc2e0f0

We can support nested switch statements, try/catch/finally, loops, and more.

**Limitations**

In 4c6299e I added test cases to document potential limitations. 

With the current implementation, the below cases will report an errors. But I think it's probably acceptable 🤔 

```vue
  <script setup lang="ts">
  import { computed, ref } from 'vue'
  import { Status } from './test01'
  const status = ref<Status>('active')
  const result = computed(() => {
    switch (status.value) {
      case 'active': throw new Error('a')
      case 'inactive': throw new Error('i')
      case 'pending': throw new Error('p')
    }
  })
  </script>
```
```vue
  <script setup lang="ts">
  import { computed, ref } from 'vue'
  import { Status } from './test01'
  function panic(): never { throw new Error() }
  const status = ref<Status>('active')
  const result = computed(() => {
    switch (status.value) {
      case 'active': return 'a'
      case 'inactive': return 'i'
      case 'pending': panic()
    }
  })
  </script>
```

lib/utils/ts-utils/ts-types.js
```diff
@@ -334,3 +335,90 @@ function* iterateTypes(type) {
...
+    if (!caseTSNode) return false
+    let caseType = checker.getTypeAtLocation(caseTSNode)
+    if (!caseType) return false
+    caseType = checker.getBaseConstraintOfType(caseType) ?? caseType
+    caseTypes.add(caseType)
+  }
+
+  const constituents = unionConstituents(discriminantType)
+    .flatMap(intersectionConstituents)
+    .filter((part) => isTypeLiteralLike(part, ts))
```
ota-meshi: I think this means that if anything other than a TypeLiteral is included here, it won't be handled properly.
For example, if you have `'a' | 'b' | {x:string}`, the last object cannot be handled by the switch.
Therefore, I think it needs to be processed with a for loop rather than filtered. What do you think?

ota-meshi (changes requested):
Thank you for improving this PR!

lib/utils/ts-utils/ts-types.js
EdRW: Thanks so much for spotting this!
I agree, we should properly handle that situation and show an error since there's no way to specifically cover `{x:string}` with a switch case.

In a16d928 I've added a test to cover the below case:

```vue
<script setup lang="ts">
import { computed, ref } from 'vue'

const val = ref<'a' | 'b' | { x: string }>('a')
// Error: Expected to return a value in computed function
const result = computed(() => {
  switch (val.value) {
    case 'a':
      return 1
    case 'b':
      return 2
  }
})
</script>
```

In ade7e1a I've switched to use a loop for checking each constituent of a union, to see if it is covered by a switch case.
Now we get the expected error for the situation in the example above 🚀

ota-meshi (approved):
LGTM! Thank you!

waynzh (approved):
Thanks!

PR #3068, opened 2026-04-13
perf(no-child-content): bail out sooner for unrelated directives

absidue:
Move the `directives.has(directiveName)` check higher up to avoid unnecessarily gathering the child nodes for `v-if`, `v-for` etc.

### DIFF
--- a/.changeset/honest-tigers-crash.md
+++ b/.changeset/honest-tigers-crash.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Improve performance in `vue/no-child-content` rule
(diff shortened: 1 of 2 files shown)


FloEdelmann (approved):
Thank you! 🙂 

How did you find this issue?
And do you have a benchmark that shows the speed improvement?

absidue:
I ran ESLint with `TIMING=1` and saw that this rule was one of the longer running ones and noticed that the check could be moved earlier on to skip unnecessary work. After spending way longer that I would like to admit investigating why this rather simple rule was taking so long and that none of my changes made any difference that couldn't be explained away by run-to-run variance, I realised that the timing included the overhead of parsing all the Vue files as it was the first Vue rule to run. Even if this doesn't make a major difference I still thought that it was worth opening a pull request for this change.

PR #3073, opened 2026-04-21
feat(no-literals-in-template): add support for ignored attributes in …

ZacharyBear:
Allow pass in an array of ignored attribute names. They could be `string` or `RegExp`.


FloEdelmann:
@rzzf was a bit faster than you, so I'm closing this as duplicate of #3072. But thank you anyway! :slightly_smiling_face:

PR #3075, opened 2026-04-30
fix: require correct peerDep version for `vue-eslint-parser`

### LINKED ISSUE #3076: v10.9.0 regression: Vue files fail with 'Parsing error: Unexpected token <' due to broken vue-eslint-parser reference in flat config
### Description

After upgrading from `eslint-plugin-vue` v10.8.0 to v10.9.0, ESLint fails to parse **all** `.vue` files with:

```
  1:1  error  Parsing error: Unexpected token <
```

This is a regression introduced by the TypeScript migration of configs in v10.9.0 ([#3002](https://github.com/vuejs/eslint-plugin-vue/pull/3002)).

### Root Cause

In `dist/configs/flat/base.js` (v10.9.0), the compiled output does:

```js
let vue_eslint_parser = require("vue-eslint-parser");
vue_eslint_parser = require_runtime.__toESM(vue_eslint_parser);

// ...
languageOptions: {
  parser: vue_eslint_parser.default,  // <-- undefined!
  sourceType: "module"
},
```

The problem is that `vue-eslint-parser` is a CommonJS module that sets `__esModule: true` on its exports (a TypeScript compilation artifact) but does **not** export a `default` property:

```js
const vueParser = require('vue-eslint-parser');
// vueParser.__esModule === true
// vueParser.default === undefined  (!)
// vueParser.parse === function  (the actual parser)
```

The rolldown `__toESM` helper sees `__esModule: true` and skips the `{ default: mod }` wrapping, so `vue_eslint_parser.default` is `undefined`. The Vue parser is never applied to `.vue` files.

**In v10.8.0**, the config used a direct `require`:
```js
// dist/configs/flat/base.js (v10.8.0)
parser: require("vue-eslint-parser"),  // worked fine
```

### Reproduction

```js
const p = require('eslint-plugin-vue');
const cfg = p.configs['flat/essential'];
const parserConfig = cfg[1]; // the config with files: ['*.vue', '**/*.vue']
console.log(parserConfig.languageOptions.parser); // undefined in 10.9.0, parser object in 10.8.0
```

### Versions

- `eslint-plugin-vue`: 10.9.0 (broken), 10.8.0 (works)
- `vue-eslint-parser`: 10.2.0
- `eslint`: 9.39.4
- Node.js: 24.15.0

### Fix Suggestion

In the TypeScript source (`lib/configs/flat/base.ts`), change:

```ts
import vueParser from 'vue-eslint-parser'
```

to:

```ts
import * as vueParser from 'vue-eslint-parser'
```

Or handle the CJS/ESM interop explicitly so the compiled output uses the module directly rather than `.default`.

haug1:
currently wrong specified may require manual intervention from consumer repo maintainers

### DIFF
--- a/.changeset/unlucky-bugs-mix.md
+++ b/.changeset/unlucky-bugs-mix.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Updated peer dependency version for `vue-eslint-parser` to fix parsing errors in Vue SFCs
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@stylistic/eslint-plugin": "^2.0.0 || ^3.0.0 || ^4.0.0 || ^5.0.0",
     "@typescript-eslint/parser": "^7.0.0 || ^8.0.0",
     "eslint": "^8.57.0 || ^9.0.0 || ^10.0.0",
-    "vue-eslint-parser": "^10.0.0"
+    "vue-eslint-parser": "^10.3.0"
   },
   "peerDependenciesMeta": {
     "@stylistic/eslint-plugin": {


haug1:
Symptoms I experienced in my consumer project:

```
error  Parsing error: '>' expected
```

All SFC vue files give this error.

zoispag (approved):

FloEdelmann (approved):
Thank you @haug1! I didn't know it was such a severe issue for anyone not manually specifying a `vue-eslint-parser` dependency themselves. A more elaborate PR/issue description would have been helpful.

Fortunately, @zoispag opened that issue today and led my attention to it, so it can now be fixed properly. Thank you, too!

haug1:
@FloEdelmann @zoispag Great that someone described it better and it's getting fixed, thanks! Sorry that I wasn't more descriptive, I really didn't give it much effort, did manual `npm-update` in my repos and went on with my actual work

PR #3079, opened 2026-05-08
fix(custom-event-name-casing): check segments of colon-separated names

### LINKED ISSUE #2439: `vue/custom-event-name-casing`: should check and report `update:kebab-case`/`update:camelCase` event names too
**Checklist**

- **ESLint version:** 8.57.0 
- **eslint-plugin-vue version:** 9.24.0
- **Vue version:** 3.4.21
- **Node version:** 18.18.0

```json5
import pluginVue from 'eslint-plugin-vue';
import configPrettier from 'eslint-config-prettier';

export default [
  ...pluginVue.configs['flat/recommended'],
  configPrettier,

  {
    rules: {
      'vue/require-default-prop': 'off',
      'vue/attribute-hyphenation': ['error', 'never'],
      'vue/custom-event-name-casing': ['error', 'camelCase'],
      // 'vue/custom-event-name-casing': ['error', 'kebab-case'],
      'vue/v-on-event-hyphenation': ['error', 'never', { autofix: true }],
    },
  },
];
```

**What did you do?**	

```vue
<script setup>
const emit = defineEmits(['update:model-value',]);

const handleEmitClick = () => {
  emit('update:model-value', 'new foo');
};
</script>

<template>
  <button type="button" @click="$emit('update:model-value', 'new foo')">
    Emit
  </button>
</template>

```

`emit` and `$emit` calls with `update:kebab-case` like event names should be reported because of wrong event name casing but no error is reported. 
side note: `vue/v-on-event-hyphenation` are able to report such event names as error as expected.

**What actually happened?**

Nothing, no error is reported.
(It seems the rule only works with plain `kebab-case`/`camelCase` event names.)

[Repo with minimal reproduction](https://stackblitz.com/edit/vitejs-vite-w2xrze?file=eslint.config.js,src%2FApp.vue,src%2Fcomponents%2FFooBar.vue&view=editor)
Just open a new terminal or terminate dev server in current terminal then run `npm run lint` to see the problem. Feel free to change the eslint rule setting to `kebab-case` it won't report errors either.

rodindev:
fix #2439

Each segment of colon-separated event names is now checked against the configured casing, instead of skipping any name starting with `update:`.
This also applies to non-`update:` namespaced names like `fooBar:barBaz`.

Updated two existing valid tests that pinned the old shortcut behaviour.

### DIFF
--- a/.changeset/quiet-events-split.md
+++ b/.changeset/quiet-events-split.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/custom-event-name-casing` to check segments of colon-separated event names like `update:*`
--- a/.changeset/quiet-events-split.md
+++ b/.changeset/quiet-events-split.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": patch
 ---
 
-Fixed `vue/custom-event-name-casing` to check segments of colon-separated event names like `update:*`
+Fixed `vue/custom-event-name-casing` to check segments of colon-separated event names like `update:foo-bar`
--- a/docs/rules/custom-event-name-casing.md
+++ b/docs/rules/custom-event-name-casing.md
@@ -30,6 +30,8 @@ See [Guide - Custom Events] for more details.
 
 This rule enforces camelCase by default.
 
+For namespaced event names like `update:myProp`, each segment between `:` is checked against the configured casing.
+
 <eslint-code-block :rules="{'vue/custom-event-name-casing': ['error']}">
 
 ```vue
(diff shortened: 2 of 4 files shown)


waynzh (approved):
Look good to me! Thanks.

.changeset/quiet-events-split.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/custom-event-name-casing` to check segments of colon-separated event names like `update:*`
```
FloEdelmann:
```suggestion
Fixed `vue/custom-event-name-casing` to check segments of colon-separated event names like `update:foo-bar`
```

FloEdelmann (changes requested):
Thank you! As noted by Copilot, please also adjust the rule docs accordingly.

FloEdelmann (approved):
Thanks!

PR #3081, opened 2026-05-13
fix(prefer-import-from-vue): don't report names not exported by vue

### LINKED ISSUE #2354: Incorrect `prefer-import-from-vue` for values that are not exported by `vue`
**Checklist**

- **ESLint version:** 8.56.0
- **eslint-plugin-vue version:** 9.19.2
- **Node version:** 19.3.0
- **Operating System:** Win 10

```json5
{
  "extends": [
    "plugin:@typescript-eslint/recommended",
    "plugin:vue/vue3-recommended"
  ],
  "parserOptions": {
    "parser": "@typescript-eslint/parser",
    "ecmaVersion": 2020,
    "ecmaFeatures": {
      "jsx": false
    },
    "sourceType": "module"
  },
  "rules": {
    "@typescript-eslint/ban-ts-comment": "off",
    "@typescript-eslint/ban-types": "off",
    "@typescript-eslint/explicit-module-boundary-types": "off",
    "@typescript-eslint/no-empty-function": "off",
    "@typescript-eslint/no-explicit-any": "off",
    "@typescript-eslint/no-non-null-assertion": "off",
    "@typescript-eslint/no-inferrable-types": "off",
    "@typescript-eslint/no-unused-vars": [
      "error",
      {
        "varsIgnorePattern": "^_",
        "argsIgnorePattern": "^_"
      }
    ],
    "prefer-const": [
      "error",
      {
        "destructuring": "all"
      }
    ],
    "vue/html-self-closing": [
      "error",
      {
        "html": {
          "void": "any",
          "normal": "any"
        }
      }
    ],
    "vue/max-attributes-per-line": "off",
    "vue/no-dupe-keys": "off",
    "vue/no-mutating-props": "off",
    "vue/require-default-prop": "off",
    "vue/require-v-for-key": "off",
    "vue/singleline-html-element-content-newline": "off",
    "vue/valid-v-slot": "off",
    "vue/attributes-order": "off",
    "vue/attribute-hyphenation": "off",
    "vue/multi-word-component-names": "off",
    "vue/html-indent": "off"
  }
}
```

**What did you do?**	
```vue
import { trigger, track } from "@vue/reactivity"
```

Nothing, that's fine code because `vue` itself does not export `trigger` nor `track`.

**What actually happened?**
Error `vue/prefer-import-from-vue` suggesting to import those members from `vue`, which is impossible.

**Suggested fix**
It's generally good practice to import from `vue` directly, so I'd like to keep this rule enabled.
The rule should have a list of all `vue` exports and not trigger when the imported value is not one of them (false positive).

rodindev:
fix #2354

For `@vue/reactivity` and `@vue/shared`, named imports/exports of bindings not re-exported by `vue` (e.g. `pauseTracking` from `@vue/reactivity`) are no longer reported. ~`@vue/runtime-dom` and `@vue/runtime-core` are unchanged — `vue` re-exports everything from them.~

Two existing tests that pinned the false-positive behaviour moved from `invalid` to `valid`. Added one `invalid` case for mixed-named imports (at least one name in `vue`, no autofix possible).

### DIFF
--- a/.changeset/prefer-import-from-vue-non-exports.md
+++ b/.changeset/prefer-import-from-vue-non-exports.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": patch
 ---
 
-Fixed `vue/prefer-import-from-vue` to not report imports/exports of names from `@vue/reactivity` and `@vue/shared` that are not re-exported by `vue`.
+Fixed `vue/prefer-import-from-vue` to not report imports/exports of names that are not re-exported by `vue`.
--- a/.changeset/prefer-import-from-vue-non-exports.md
+++ b/.changeset/prefer-import-from-vue-non-exports.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/prefer-import-from-vue` to not report imports/exports of names from `@vue/reactivity` and `@vue/shared` that are not re-exported by `vue`.
--- a/docs/rules/prefer-import-from-vue.md
+++ b/docs/rules/prefer-import-from-vue.md
@@ -30,6 +30,8 @@ Imports from the following modules are almost always wrong. You should import fr
 - `@vue/reactivity`
 - `@vue/shared`
 
+For `@vue/reactivity` and `@vue/shared`, only named imports/exports that `vue` actually re-exports are reported.
+
 <eslint-code-block fix :rules="{'vue/prefer-import-from-vue': ['error']}" filename="example.js" language="javascript">
 
 ```js
--- a/docs/rules/prefer-import-from-vue.md
+++ b/docs/rules/prefer-import-from-vue.md
@@ -30,7 +30,7 @@ Imports from the following modules are almost always wrong. You should import fr
 - `@vue/reactivity`
 - `@vue/shared`
 
-For `@vue/reactivity` and `@vue/shared`, only named imports/exports that `vue` actually re-exports are reported.
+Only named imports/exports of bindings that `vue` actually re-exports are reported.
 
 <eslint-code-block fix :rules="{'vue/prefer-import-from-vue': ['error']}" filename="example.js" language="javascript">
 
--- a/lib/rules/prefer-import-from-vue.js
+++ b/lib/rules/prefer-import-from-vue.js
@@ -85,6 +85,17 @@ module.exports = {
         // Skip imports without specifiers in `.d.ts` files
         if (node.specifiers.length === 0 && context.filename.endsWith('.d.ts'))
           return
+        // Don't report imports where no name is re-exported by 'vue'.
+        if (
+          !SUBSET_AT_VUE_MODULES.has(node.source.value) &&
+          node.specifiers.length > 0 &&
+          node.specifiers.every(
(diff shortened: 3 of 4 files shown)


docs/rules/prefer-import-from-vue.md
```diff
@@ -30,6 +30,8 @@ Imports from the following modules are almost always wrong. You should import fr
 - `@vue/reactivity`
 - `@vue/shared`
 
+For `@vue/reactivity` and `@vue/shared`, only named imports/exports that `vue` actually re-exports are reported.
```
FloEdelmann: Why only for those two modules and not the other two?

rodindev: I kept `@vue/runtime-dom` and `@vue/runtime-core` reporting unconditionally because `vue` re-exports everything from them, but happy to apply the check uniformly to all four if you'd prefer 🙌🏻

FloEdelmann: Yes, let's apply the check to all exports. Then the check can also be moved to `verifySource`. BTW, also be sure to remove the no longer needed condition in the auto-fixers.

rodindev: Done, uniform check moved into verifySource, subset shortcut in the auto-fixer narrowed to namespace/default imports. Let me know if you'd prefer to drop the shortcut entirely.

.changeset/prefer-import-from-vue-non-exports.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/prefer-import-from-vue` to not report imports/exports of names from `@vue/reactivity` and `@vue/shared` that are not re-exported by `vue`.
```
FloEdelmann:
```suggestion
Fixed `vue/prefer-import-from-vue` to not report imports/exports of names that are not re-exported by `vue`.
```

lib/rules/prefer-import-from-vue.js
```diff
@@ -86,41 +94,62 @@ module.exports = {
...
-          }
-          for (const name of extractImportNames(node)) {
-            if (name == null) {
-              return false // import all
-            }
-            if (!vue3ExportNames.has(name)) {
-              // If there is a name that is not exported from 'vue', it will not be auto-fixed.
-              return false
+        const hasOnlyNamedSpecifiers = node.specifiers.every(
+          (s) => s.type === 'ImportSpecifier'
```
FloEdelmann: Please don't use single-character parameter names (e.g. use `specifier` in this case)

FloEdelmann (changes requested):

docs/rules/prefer-import-from-vue.md
FloEdelmann: I think the "subset" shortcut only makes the code less readable without much benefit, so I'd argue to drop it completely.

rodindev: All applied, thanks!

FloEdelmann (approved):
Thank you! 🙂

waynzh (approved):
thanks!

PR #3082, opened 2026-05-18
Fix use-v-on-exact false positive for keydown stop listener

### LINKED ISSUE #2455: vue/use-v-on-exact - false positive on multiple @keydown.* attributes
**Checklist**

- **ESLint version:** 8.57.0
- **eslint-plugin-vue version:** 9.25.0
- **Vue version:** 3.4.23
- **Node version:** v20.11.1
- **Operating System:** macOS 12.7.4 (21H1123)

```json5
/* eslint-env node */
module.exports = {
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  }
}
```

**What did you do?**	

```vue
<script lang="ts">
export default {
  methods: {
    save(): void {}
  }
}
</script>

<template>
  <!-- OK -->
  <input @keydown.ctrl.enter.exact="save()" />
  <!-- OK -->
  <input @keydown.stop />
  <!-- ERROR - Fails with vue/use-v-on-exact -> Consider to use '.exact' modifier on the @keydown.stop line -->
  <input
    @keydown.ctrl.enter.exact="save()"
    @keydown.stop
  />
</template>
```

I expected that `@keydown.stop` is OK to use and doesn't need `.exact` modifier.

**What actually happened?**

`vue/use-v-on-exact` returned an error that .exact should be used - on the `@keydown.stop` line.
```
$ npm run lint

> vue-project@0.0.0 lint
> eslint . --ext .vue,.js,.jsx,.cjs,.mjs --fix --ignore-path .gitignore


/<redacted>/bugreports/vue-project/src/components/HelloWorld.vue
  17:5  error  Consider to use '.exact' modifier  vue/use-v-on-exact

```

https://github.com/petrkotek/bugreports (See https://github.com/petrkotek/bugreports/blob/main/vue-project/src/components/HelloWorld.vue)

sjh9714:
## Summary

Fixes a `vue/use-v-on-exact` false positive for a global `@keydown.stop` listener paired with an exact key-specific listener.

## What changed

- Added regression coverage for `@keydown.ctrl.enter.exact` used with `@keydown.stop`.
- Added a guard case to keep reporting `@keydown.prevent` with `@keydown.ctrl` when `.exact` is still needed.
- Updated conflict detection so a keyboard `stop` listener is not treated as needing `.exact` when the other handler already uses `.exact`.

## Why

Issue #2455 reports that `@keydown.stop` is flagged when it appears next to `@keydown.ctrl.enter.exact`, even though the handler intentionally stops propagation for all keydown events.

The implementation is scoped to keyboard `stop` listeners and only skips the report when the matching key-specific handler already has `.exact`, preserving existing conflict reports for non-exact modifier combinations.

## Testing

- `npx vitest run tests/lib/rules/use-v-on-exact.test.ts`
- `npm run lint`
- `npm run build`
- `npm test`
- `git diff --check`

Fixes #2455


tests/lib/rules/use-v-on-exact.test.ts
```diff
@@ -158,6 +166,23 @@ ruleTester.run('use-v-on-exact', rule, {
         }
       ]
     },
+    {
+      code: `<template>
+        <input
+          @keydown.prevent="foo"
+          @keydown.ctrl="bar"
```
FloEdelmann, line 173: I think `@keydown.stop="foo"` should still report an error, right? So only `@keydown.stop` without a handler should be silenced.

sjh9714: Adjusted this so only handler-less `@keydown.stop` is skipped. `@keydown.stop="foo"` now remains reported, with regression coverage added.

.changeset/key-listeners-stop.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/use-v-on-exact` to avoid reporting `@keydown.stop` when paired with an exact key-specific listener.
```
FloEdelmann:
```suggestion
Fixed `vue/use-v-on-exact` to avoid reporting `@keydown.stop` when paired with an exact key-specific listener
```

FloEdelmann (approved):
Looks good to me, apart from a very minor changelog wording fix. Thank you!

waynzh (approved):
LGTM, thanks!

PR #3084, opened 2026-05-31
feat(sort-keys): add allowLineSeparatedGroups option

### LINKED ISSUE #2212: FR: introduce allowLineSeperatedGroups option on sort-keys
**What rule do you want to change?** `sort-keys`

**Does this change cause the rule to produce more or fewer warnings?** fewer

**How will the change be implemented? (New option, new default behavior, etc.)?** new option

```vue
  setup() {
    return {
      // feature 1
      a: 'wow',
      b: 'wow',

      // feature 2
      a2: 'wow',
      b2: 'wow',
    }
  }
})
```

**What does the rule currently do for this code?** it forces a2 to come after a

**What will the rule do after it's changed?** it allows a2 since there is a newline above

**Additional context**

See ESLint: https://eslint.org/docs/latest/rules/sort-keys#allowlineseparatedgroups

rodindev:
Closes #2212

Adds an `allowLineSeparatedGroups` option to `vue/sort-keys`, mirroring the ESLint core option of the same name.
When enabled, a blank line between properties resets the sort order, so each group is sorted independently.
Default `false`, so existing behavior is unchanged.

Test cases are ported from ESLint core's suite.

### DIFF
--- a/.changeset/sort-keys-allow-line-separated-groups.md
+++ b/.changeset/sort-keys-allow-line-separated-groups.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Added `allowLineSeparatedGroups` option to [`vue/sort-keys`](https://eslint.vuejs.org/rules/sort-keys.html) rule.
--- a/.changeset/sort-keys-allow-line-separated-groups.md
+++ b/.changeset/sort-keys-allow-line-separated-groups.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": minor
 ---
 
-Added `allowLineSeparatedGroups` option to [`vue/sort-keys`](https://eslint.vuejs.org/rules/sort-keys.html) rule.
+Added `allowLineSeparatedGroups` option to [`vue/sort-keys`](https://eslint.vuejs.org/rules/sort-keys.html) rule
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -256,7 +256,6 @@ module.exports = {
             node
           ]
 
-          // check blank line between tokens
           isBlankLineBetweenNodes ||= tokens.some((token, index) => {
             const previousToken = tokens.at(index - 1)
             return (
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -248,7 +248,6 @@ module.exports = {
         let isBlankLineBetweenNodes = objectStack.prevBlankLine
 
         if (prevNode) {
-          // Collect the previous node, the tokens between, and the current node
           const tokens = [
             prevNode,
             ...sourceCode.getTokensBetween(prevNode, node, {
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -249,9 +249,13 @@ module.exports = {
 
         if (prevNode) {
           // Get tokens between current node and previous node
-          const tokens = sourceCode.getTokensBetween(prevNode, node, {
-            includeComments: true
-          })
+          const tokens = [
+            prevNode,
+            ...sourceCode.getTokensBetween(prevNode, node, {
+              includeComments: true
+            }),
+            node
+          ]
 
           // check blank line between tokens
           let previousToken
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -257,11 +257,10 @@ module.exports = {
           ]
 
           isBlankLineBetweenNodes ||= tokens.some((token, index) => {
-            const previousToken = tokens.at(index - 1)
-            return (
-              previousToken &&
-              token.loc.start.line - previousToken.loc.end.line > 1
-            )
+            if (index === 0) {
+              return false
+            }
+            return token.loc.start.line - tokens[index - 1].loc.end.line > 1
           })
         }
 
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -277,6 +277,9 @@ module.exports = {
         }
 
         if (allowLineSeparatedGroups && isBlankLineBetweenNodes) {
+          // A computed key (e.g. `[a+b]`) has no name to compare, so it's skipped and doesn't
+          // reset the sort. Carry the blank line over; otherwise the next key would be compared
+          // against the previous group and wrongly reported as out of order.
           objectStack.prevBlankLine = thisName === null
           return
         }
--- a/lib/rules/sort-keys.js
+++ b/lib/rules/sort-keys.js
@@ -96,6 +96,9 @@ module.exports = {
           },
           natural: {
             type: 'boolean'
+          },
+          allowLineSeparatedGroups: {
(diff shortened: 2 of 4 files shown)


lib/rules/sort-keys.js
```diff
@@ -234,10 +244,58 @@ module.exports = {
...
         const thisName = getPropertyName(node)
 
+        const prevNode = objectStack.prevNode
+        let isBlankLineBetweenNodes = objectStack.prevBlankLine
+
+        if (prevNode) {
+          // Get tokens between current node and previous node
+          const tokens = sourceCode.getTokensBetween(prevNode, node, {
+            includeComments: true
+          })
```
ota-meshi: I believe that lines L268-L285 can be removed by processing them as follows.

```suggestion
          const tokens = [
            prevNode,
            ...sourceCode.getTokensBetween(prevNode, node, {
              includeComments: true
            }),
            node
          ]
```

lib/rules/sort-keys.js
```diff
@@ -234,10 +244,58 @@ module.exports = {
...
+        }
+
+        objectStack.prevNode = node
+
         if (thisName !== null) {
           objectStack.prevName = thisName
         }
 
+        if (allowLineSeparatedGroups && isBlankLineBetweenNodes) {
+          objectStack.prevBlankLine = thisName === null
```
ota-meshi, line 277: I think it's difficult to understand what this is meant to be. Could you please write a comment explaining your intention?

ota-meshi (changes requested):
Thank you for this PR!

lib/rules/sort-keys.js
rodindev: This line comes from ESLint core's `sort-keys` (https://github.com/eslint/eslint/blob/7797c266977b0bc4971aa79721813d480de72cd1/lib/rules/sort-keys.js#L239), I kept it as-is. I'd add this comment:

```js
// A computed key has no name to compare, so carry the blank line to the next property.
```

Works for you?

ota-meshi: I don't just want you to explain the statement you wrote exactly as it is.
I want you to explain why it's being done, what would happen if it weren't there, etc.

rodindev: Makes sense! Added a comment

ota-meshi: Thank you for adding a comment!

I have a question.
How about using `objectStack.prevName = null` (as done in L232) instead?
What's the difference between `objectStack.prevName = null` and the current logic?


```suggestion
          objectStack.prevName = null
```

rodindev: I actually just tried it and one of the invalid tests failed. It happens because `prevName = null` resets the anchor completely, so after a blank line the first key in the new group has nothing to compare against. For example `{ a, b, <blank>, z(){}, y }` stops reporting that `y` should be before `z`. `prevBlankLine` only skips the comparison across the blank line and keeps `z` as the anchor, so the keys inside the new group are still sorted.

ota-meshi: Thanks for the explanation!

ota-meshi (approved):
LGTM! Thank you!

.changeset/sort-keys-allow-line-separated-groups.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Added `allowLineSeparatedGroups` option to [`vue/sort-keys`](https://eslint.vuejs.org/rules/sort-keys.html) rule.
```
FloEdelmann: For consistency:

```suggestion
Added `allowLineSeparatedGroups` option to [`vue/sort-keys`](https://eslint.vuejs.org/rules/sort-keys.html) rule
```

lib/rules/sort-keys.js
```diff
@@ -234,10 +244,46 @@ module.exports = {
         const numKeys = objectStack.numKeys
         const thisName = getPropertyName(node)
 
+        const prevNode = objectStack.prevNode
+        let isBlankLineBetweenNodes = objectStack.prevBlankLine
+
+        if (prevNode) {
+          // Collect the previous node, the tokens between, and the current node
```
FloEdelmann: I think this comment doesn't add much value: it exactly explains *what* the next statement does (not *why*) and the next statement is quite readable.


```suggestion
```

lib/rules/sort-keys.js
```diff
@@ -234,10 +244,46 @@ module.exports = {
...
+          for (const token of tokens) {
+            if (
+              previousToken &&
+              token.loc.start.line - previousToken.loc.end.line > 1
+            ) {
+              isBlankLineBetweenNodes = true
+            }
+            previousToken = token
+          }
+        }
```
FloEdelmann: `isBlankLineBetweenNodes` would be more readable with `array.some` (and stop the loop once one is found):


```suggestion
        isBlankLineBetweenNodes = tokens.some((token, index) => {
          const previousToken = tokens.at(index - 1)
          return (
            previousToken &&
            token.loc.start.line - previousToken.loc.end.line > 1
          )
        })
```

FloEdelmann (changes requested):
Thank you! I only have some suggestions for minor code style improvements.

lib/rules/sort-keys.js
rodindev: Thanks! Applied the first two suggestions.

For `isBlankLineBetweenNodes`: the `.some` version reads better, but with `=` two valid tests failed, because `isBlankLineBetweenNodes` starts from the carried-over `prevBlankLine`, and a plain `=` overwrites it. Switching to `||=` keeps the carry

lib/rules/sort-keys.js
```diff
@@ -234,10 +244,42 @@ module.exports = {
...
+        if (prevNode) {
+          const tokens = [
+            prevNode,
+            ...sourceCode.getTokensBetween(prevNode, node, {
+              includeComments: true
+            }),
+            node
+          ]
+
+          // check blank line between tokens
```
FloEdelmann: Now I think this comment is also superfluous:


```suggestion
```

lib/rules/sort-keys.js
```diff
@@ -234,10 +244,41 @@ module.exports = {
...
+            node
+          ]
+
+          isBlankLineBetweenNodes ||= tokens.some((token, index) => {
+            const previousToken = tokens.at(index - 1)
+            return (
+              previousToken &&
+              token.loc.start.line - previousToken.loc.end.line > 1
+            )
+          })
```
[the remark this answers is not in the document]
rodindev: @FloEdelmann Copilot flagged the `tokens.at(index - 1)` from your suggestion. It's harmless here, but I can switch to an explicit `index === 0` guard if you prefer. Both work for me, just let me know.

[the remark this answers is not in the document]
FloEdelmann: Copilot does have a point here. I'll apply its suggestion.

FloEdelmann (approved):
Thank you!

PR #3089, opened 2026-06-25
docs: More accurate example in `prefer-single-event-payload`

MorevM:
A small correction to the example in the rule description. 
After reading the current version, I was a little confused for a few seconds :)


FloEdelmann (approved):
Thank you!

PR #3097, opened 2026-07-11
Fix prefer-separate-static-class for template literals

### LINKED ISSUE #2128: Rule `vue/prefer-separate-static-class` does not support string interpolated static and dynamic classes.
**Checklist**

- **ESLint version:** 8.37.0
- **eslint-plugin-vue version:** 9.10.0
- **Node version:** 14.20.0
- **Operating System:** macOS Monterey

```json5
module.exports = {
    extends: [
        '@nuxtjs/eslint-config-typescript',
        'plugin:vue/recommended'
    ],
    rules: {
        indent: ['error', 4],
        semi: ['error', 'always'],
        'require-await': 'off',
        'no-tabs': 0,
        'prefer-destructuring': ['warn', { object: true, array: false }],
        'no-unused-vars': ['warn'],
        'vue/no-v-html': 0,
        'vue/require-default-prop': 0,
        'vue/html-indent': [
            'error',
            'tab',
            { alignAttributesVertically: true }
        ],
        'vue/prefer-separate-static-class': 'error',
        'vue/no-arrow-functions-in-watch': 'error',
        'vue/html-self-closing': [
            'error',
            {
                html: {
                    void: 'always',
                    normal: 'never',
                    component: 'any'
                }
            }
        ]
    }

};
```

**What did you do?**	

```vue
<template>
    <div class="component">
        <input type="text" :class="`form-item input ${isDisabled ? 'disabled' : ''}`">
    </div>
</template>
<script>
export default {
    data() {
        return {
            isDisabled: false
        }
    }
    // vuelidate logic etc
}
</script>
```

I would expect the linter to error out that there are static classes on the property. 

I recognize the complexity of trying to implement a solution that works for `--fix` since there is no clear way to know how many nested conditions or string class properties exist with the string interpolated ternary expression. 

**What actually happened?**

```
The linter is ignoring the dynamic :class="`${}`" and treating it as a valid static class.
```

[eslint-plugin-vue Online Playground](https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJyNV9uOpDYQ/ZVRK9K+rDVSHleTlSLlE/K4D2tMQXvH2MQ2fclo/j2nDBgMzWSeujlVlMtVpy68nZSr6fTt9BKp642M9P2HfXp6qYYYnX36powM4Y8fp58T8Nvbly/vP3+cvv9Z109/u7/cy/MowWsvz9nG6evJD4bC6dvb6TLQs3JdRzaKWntSUV9w4u9fk+RXuIkhUBAX6UNGu8FELa7O1wKv9s7yy1Z2tKhYJ6T37iqawcKks0FoK64yqnOhE+5WsYTtDJFq0XvXk4+6tKXO2vBhNuKoQrB/j83VMsq1Wk09riZZkUXCVb9wU8CIoJfs35E2hejdHf+MbkjdlaEjTWdgC2oBPpIPQvb6Y9Wg4HIdjIsfKdMFV/5IodEmkj8STuGXRszpP9A8x84IMpSYoMOBlrZGW/o/W5yMAKCRYIqI50Nz3iF5HDT7KqJs04sHqilaQsboNTh9dDJH83NKn7N3EZW2eAM0PdQ4vN1FOIu64IoSnat1ow/zNKoOXYVozKqHZgfiUmh0K17pzh2i1BxwLwj24AX5DSR0s5EYzYaXWBRv0q13PnJRBeV1j8BRHPqtBszKhlMpr1IXFZrr5kihGyJCZMfkF0f3aDosIHSSInCeGrQOwRUvbV1KAvkLfdiYss42SFmw8yTdOdMaLWFQcfAFacJZ+uLcbQ8KuiZBDaIRwyda3lxi7GWJ3yJOkohbiFKdCycGi269ciI8EBatnOFAO1/uSNsrSEvx7OqtMhjfiKuOZ/xrysRAjLkSFucfcyq9V9yQ6Z+afKnGl8VP6k3OLvdaa6Wh8pBcqJgGoO4SgxvvOq6dLPb0z4CJt0rZqpJnIQdExHufSm5M/OrKs5YnW+MkD5p4u5NOLT4sw3CnEl3bokS0TSyJoHXQxViaFcfQrTmRJdLoOjfdopOOfj0i3f2BDnU6htGcXF92zD3SQDepliiP5+ZEp3rbCOEUT41k+LGorLhRZDn5UavXDZ55453bujH26x2ojJNbM/tuuMZ34JrpM8a83IEPDOrt1S4orc7tQfTyvcEVCxZI7T3s/R4LZ3fdg+ZB3LjUErjiuQgYG+O1mXXTfjj501KMJefzQodsdhqxpRupYaYQb2sS4GrpClnAiyxeDrrNppiqaHSgQSv+Je8WHAN5qwqXH2ESRc36Sy2NEux0Bp63dJvBmqqhbckvz4YiF9WC8AiVvg0FMHqObFbzwJ4la3btxnIxeDksMwpuT0U5P6AvIlyKY50OK4W95Czku9FtEwZk1UtROWdIWj4plpKAopyRRhoTz9jJ2nOG0LE2FlvjKqyTJTj12A1osQg/zLe2I+1SCvqMejwPvBpfzxojo5cqx8U4hNk1THKlwyqbIBquVvOecBCoTt946WNr4BjIE2WVhZau2O66ypmMOOtUxA3huO7wS0HJPjuCTweE0ZhsISnPD2B+dGlcVAP2cr1cGctBCgQtAO4+uTVjGJ/NJojhUZlh2ajxceV5FdGK+/my5bCYNycmq5cL33gNFxWhkGlV12kvQBNeHpAQSibTZx6v+4sMeweCXJkVFGRD+AbBF4a5b1BUrlwX3gS7fvoigSltkbhFnhYUIytawrvdWlZbRiXVaxrxZBeizEKVvzVXYJlK3mHS/3mE3jVNGwjPOh2snFrvOHeQVrc092m7CNQzuxFTXmNVpt77+38clXyU)

cyphercodes:
## Summary

- report static class text inside dynamic `:class` template literals when the static text is bounded by class-name whitespace
- keep partial interpolated class names such as `` `foo-${bar}` `` valid to avoid false positives
- add regression coverage for issue #2128

Fixes #2128.

## Tests

- `npx vitest run --reporter=dot tests/lib/rules/prefer-separate-static-class.test.ts`
- `npx eslint lib/rules/prefer-separate-static-class.js tests/lib/rules/prefer-separate-static-class.test.ts`
- `npm run lint`
- `git diff --check`

### DIFF
--- a/.changeset/prefer-static-class-template-literal.md
+++ b/.changeset/prefer-static-class-template-literal.md
@@ -2,4 +2,4 @@
 "eslint-plugin-vue": patch
 ---
 
-Fixed `vue/prefer-separate-static-class` to report static class text inside template literal class bindings.
+Fixed `vue/prefer-separate-static-class` to report static text inside template literals
--- a/.changeset/prefer-static-class-template-literal.md
+++ b/.changeset/prefer-static-class-template-literal.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/prefer-separate-static-class` to report static class text inside template literal class bindings.
--- a/.changeset/prefer-static-class-template-literal.md
+++ b/.changeset/prefer-static-class-template-literal.md
@@ -1,5 +1,5 @@
 ---
-"eslint-plugin-vue": patch
+"eslint-plugin-vue": minor
 ---
 
-Fixed `vue/prefer-separate-static-class` to report static text inside template literals
+Added support for template literals to `vue/prefer-separate-static-class`
--- a/lib/rules/prefer-separate-static-class.js
+++ b/lib/rules/prefer-separate-static-class.js
@@ -57,15 +57,11 @@ function findStaticClasses(expressionNode) {
       if (
         property.type === 'Property' &&
         property.value.type === 'Literal' &&
-        property.value.value === true
+        property.value.value === true &&
+        (isStringLiteral(property.key) ||
+          (property.key.type === 'Identifier' && !property.computed))
       ) {
-        const key = /** @type {ASTNode | TemplateLiteral} */ (property.key)
-        if (isStringLiteral(key)) {
-          return [key]
-        }
-        if (key.type === 'Identifier' && !property.computed) {
-          return [key]
-        }
+        return [property.key]
       }
       return []
     })
--- a/lib/rules/prefer-separate-static-class.js
+++ b/lib/rules/prefer-separate-static-class.js
@@ -11,10 +11,37 @@ const {
 } = require('../utils')
 
 /**
- * @param {Expression | VForExpression | VOnExpression | VSlotScopeExpression | VFilterSequenceExpression} expressionNode
- * @returns {(Literal | TemplateLiteral | Identifier)[]}
+ * @param {TemplateElement} templateElement
+ * @param {number} index
+ * @param {TemplateLiteral} templateLiteral
+ * @returns {boolean}
+ */
+function isStaticClassTemplateElement(templateElement, index, templateLiteral) {
+  const value = templateElement.value.cooked
+  if (!value || value.trim() === '') {
+    return false
+  }
+
+  return (
+    (index === 0 || /^\s/u.test(value)) &&
+    (index === templateLiteral.expressions.length || /\s$/u.test(value))
+  )
+}
+
+/**
+ * @param {ASTNode | TemplateLiteral} expressionNode
+ * @returns {(Literal | TemplateLiteral | TemplateElement | Identifier)[]}
  */
 function findStaticClasses(expressionNode) {
+  if (expressionNode.type === 'TemplateLiteral') {
+    if (expressionNode.expressions.length === 0) {
+      return [expressionNode]
+    }
+    return expressionNode.quasis.filter((templateElement, index) =>
+      isStaticClassTemplateElement(templateElement, index, expressionNode)
+    )
+  }
+
   if (isStringLiteral(expressionNode)) {
     return [expressionNode]
   }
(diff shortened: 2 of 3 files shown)


FloEdelmann (changes requested):
There are utils for checking whether a node is a string literal and for getting its value, which support template literals. Please use them instead of recreating the logic in the rule.

cyphercodes:
Addressed the review by relying on the shared string literal helpers for static literal/template-literal detection, keeping the custom handling only for static segments inside interpolated template literals.\n\nVerification:\n- npx vitest run tests/lib/rules/prefer-separate-static-class.test.ts --reporter=dot\n- npx eslint lib/rules/prefer-separate-static-class.js tests/lib/rules/prefer-separate-static-class.test.ts\n- git diff --check

lib/rules/prefer-separate-static-class.js
```diff
@@ -33,11 +57,15 @@ function findStaticClasses(expressionNode) {
...
+        property.value.value === true
       ) {
-        return [property.key]
+        const key = /** @type {ASTNode | TemplateLiteral} */ (property.key)
+        if (isStringLiteral(key)) {
+          return [key]
+        }
+        if (key.type === 'Identifier' && !property.computed) {
+          return [key]
+        }
```
FloEdelmann: Is this change necessary for typing? Otherwise it looks unnecessary and needlessly duplicated.

FloEdelmann (changes requested):

cyphercodes:
Addressed the latest review nit by removing the unnecessary key temp/cast and restoring the compact object-key condition while keeping the shared string-literal helper path.

Verification:
- `npx vitest run tests/lib/rules/prefer-separate-static-class.test.ts --reporter=dot`
- `npx eslint lib/rules/prefer-separate-static-class.js tests/lib/rules/prefer-separate-static-class.test.ts`
- `git diff --check`

.changeset/prefer-static-class-template-literal.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/prefer-separate-static-class` to report static class text inside template literal class bindings.
```
FloEdelmann:
```suggestion
Fixed `vue/prefer-separate-static-class` to report static text inside template literals
```

FloEdelmann (approved):
Thank you!

cyphercodes:
Addressed the follow-up by applying the suggested changeset wording and adding a middle-quasi regression for the existing template-literal handling.

Verification:
- `npx vitest run tests/lib/rules/prefer-separate-static-class.test.ts --reporter=dot`
- `npx eslint lib/rules/prefer-separate-static-class.js tests/lib/rules/prefer-separate-static-class.test.ts`
- `git diff --check`

FloEdelmann (approved):

.changeset/prefer-static-class-template-literal.md
ota-meshi, line 1: Thank you for this PR!

I think this change to be a feature addition that supports template literals. 

What do you think?

ota-meshi (changes requested):

cyphercodes:
Agreed — treating this as feature support makes sense. I updated the changeset from patch to minor and reworded it as:

> Added support for template literals to `vue/prefer-separate-static-class`

Verification:
- `git diff --check`

ota-meshi (approved):
LGTM! Thank you!

PR #3099, opened 2026-07-17
fix: handle unresolved types in require-valid-default-prop

### LINKED ISSUE #2279: vue/require-valid-default-prop breaks when importing type with union
**Checklist**

- **ESLint version:** 8.45.0
- **eslint-plugin-vue version:** 9.15.1
- **Node version:** v20.3.1
- **Operating System:** Linux

```json5
/* eslint-env node */

const {defineConfig} = require('eslint-define-config')

module.exports = defineConfig({
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript',
    '@vue/eslint-config-prettier/skip-formatting'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  },
  rules: {
    'vue/require-valid-default-prop': 'error'
  }
})

```

**What did you do?**	

```vue
<script setup lang="ts">
import type { Foo } from '../types';

// type Foo = 'true' | 'false' | '' | boolean

const props = withDefaults(defineProps<{
  msg?: Foo | 'foo'
}>(), {
  msg: false, // Errors here `Type of the default value for 'msg' prop must be a string.`
})
</script>

<template>
  <div class="greetings">
  </div>
</template>
```

```ts
export type Foo = boolean
```

No error, this is valid syntax post Vue 3.3

**What actually happened?**

`Type of the default value for 'msg' prop must be a string.`

This is incorrect, as this is correct post Vue 3.3.

https://github.com/VividLemon/vue-eslint-require-default-prop-bug

awnoa:
Fixes #2279.

Unresolved type references now use a dedicated inference marker instead of sharing the `null` placeholder. `require-valid-default-prop` skips validation only for that marker, while real `null` unions continue to validate known runtime types.

Tests cover both the imported-union case and a `string | null` mismatch.

### DIFF
--- a/lib/utils/ts-utils/ts-ast.js
+++ b/lib/utils/ts-utils/ts-ast.js
@@ -1,6 +1,7 @@
 const { getScope } = require('../scope')
 const { findVariable } = require('@eslint-community/eslint-utils')
 const { inferRuntimeTypeFromTypeNode } = require('./ts-types')
+const UNKNOWN_TYPE = '__UNKNOWN__'
 /**
  * @typedef {import('@typescript-eslint/types').TSESTree.TypeNode} TSESTreeTypeNode
  * @typedef {import('@typescript-eslint/types').TSESTree.TSInterfaceBody} TSESTreeTSInterfaceBody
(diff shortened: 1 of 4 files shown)


lib/utils/ts-utils/ts-ast.js
```diff
@@ -1,6 +1,7 @@
 const { getScope } = require('../scope')
 const { findVariable } = require('@eslint-community/eslint-utils')
 const { inferRuntimeTypeFromTypeNode } = require('./ts-types')
+const UNKNOWN_TYPE = '__UNKNOWN__'
```
FloEdelmann, line 4: I'm not sure if such a magic value is the best solution here 😬 

Especially if it is repeated in the rule instead of being exported here and imported there.

@ota-meshi @waynzh Do you maybe have a better suggestion on how to achieve this in a cleaner way?

FloEdelmann (changes requested):

PR #3100, opened 2026-07-19
feat: support `@nuxtjs/composition-api`

ylc395:
`@nuxtjs/composition-api` is also vue package.

### DIFF
--- a/.changeset/perky-chairs-press.md
+++ b/.changeset/perky-chairs-press.md
@@ -1,5 +1,5 @@
 ---
-'eslint-plugin-vue': patch
+'eslint-plugin-vue': minor
 ---
 
-support `@nuxtjs/composition-api`
+Added support for SFCs using `@nuxtjs/composition-api`, which is now treated the same as `vue` or `@vue/composition-api`.
--- a/.changeset/perky-chairs-press.md
+++ b/.changeset/perky-chairs-press.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+support `@nuxtjs/composition-api`
--- a/.changeset/perky-chairs-press.md
+++ b/.changeset/perky-chairs-press.md
@@ -2,4 +2,4 @@
 'eslint-plugin-vue': minor
 ---
 
-Added support for SFCs using `@nuxtjs/composition-api`, which is now treated the same as `vue` or `@vue/composition-api`.
+Added support for SFCs using `@nuxtjs/composition-api`, which is now treated the same as `vue` or `@vue/composition-api`
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -2848,6 +2848,12 @@ function isGlobalOrImportedFromVue(callee, context) {
 
   if (variable !== null && variable.defs.length === 1) {
     const def = variable.defs[0]
+    /** @type {Set<string | null>} */
+    const vueLikePackageNames = new Set([
+      'vue',
+      '@vue/composition-api',
+      '@nuxtjs/composition-api'
+    ])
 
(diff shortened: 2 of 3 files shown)


.changeset/perky-chairs-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+support `@nuxtjs/composition-api`
```
FloEdelmann: Please include which rules or features change with this improvement. Also, please align the wording style to previous changelog entries: https://github.com/vuejs/eslint-plugin-vue/blob/master/CHANGELOG.md

.changeset/perky-chairs-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
```
FloEdelmann:
```suggestion
'eslint-plugin-vue': minor
```

This is an improvement rather than a bug fix.

lib/utils/index.js
```diff
@@ -2882,7 +2883,11 @@ function isGlobalOrImportedFromVue(callee, context) {
             ? getStringLiteralValue(firstArg)
             : null
 
-          return source === 'vue' || source === '@vue/composition-api'
+          return (
+            source === 'vue' ||
+            source === '@vue/composition-api' ||
+            source === '@nuxtjs/composition-api'
```
FloEdelmann: Those three string literals are repeated twice. Please extract them to a `Set` and then use `set.has(source)`/`set.has(def.node.parent.source.value)`.

lib/utils/index.js
```diff
@@ -2260,7 +2260,7 @@ module.exports = {
   iterateWatchHandlerValues,
 
   /**
-   * Wraps composition API trace map in both 'vue' and '@vue/composition-api' imports, or '#imports' from unimport
+   * Wraps composition API trace map in 'vue', '@vue/composition-api' and '@nuxtjs/composition-api' imports, or '#imports' from unimport
```
FloEdelmann, line 2263: Does this addition need to be mentioned in the docs somewhere?

FloEdelmann (changes requested):

lib/utils/index.js
ylc395: Maybe it's not necessary to specifically mention it in docs.

.changeset/perky-chairs-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': minor
+---
+
+Added support for SFCs using `@nuxtjs/composition-api`, which is now treated the same as `vue` or `@vue/composition-api`.
```
FloEdelmann:
```suggestion
Added support for SFCs using `@nuxtjs/composition-api`, which is now treated the same as `vue` or `@vue/composition-api`
```

lib/utils/index.js
```diff
@@ -2260,7 +2260,7 @@ module.exports = {
...
    * @param {import('@eslint-community/eslint-utils').TYPES.TraceMap} map
    */
   createCompositionApiTraceMap,
```
FloEdelmann, line 2266: Are there test cases for this function? Can you add a test case for `@nuxtjs/composition-api` there or in a rule implicitly relying on this function?

ylc395: I added a test case that would fail without this PR. This is the use case that triggered this issue.

FloEdelmann: Awesome, thank you!

FloEdelmann (approved):

waynzh (approved):
LGTM, thank you!

PR #3101, opened 2026-07-19
feat(define-props-declaration): check defineModel

### LINKED ISSUE #3069: define-model-declaration

Analogous to [define-props-declaration](https://eslint.vuejs.org/rules/define-props-declaration), there should be a rule define-model-declaration that enforces a type-based declaration.

[ ] Other (please specify:)

```vue
<script setup lang="ts">
defineModel();
</script>
```

```vue
<script setup lang="ts">
defineModel({ type: String });
</script>
```

**Workaround**
```ts
'no-restricted-syntax': [
  'error',
  {
    message:
      'defineModel() must use an explicit type parameter, e.g. defineModel<string>().',
    selector:
      "CallExpression[callee.name='defineModel']:not([typeArguments]):not([typeParameters])",
  },
],
```

ifer47:
## What changed

- extend `vue/define-props-declaration` to inspect `defineModel`
- require explicit type arguments in `type-based` mode and reject them in `runtime` mode
- preserve valid type-based model options such as `default` and `required`
- add named and unnamed regression cases, documentation, generated rule metadata, and a minor changeset

## Why

`defineModel` declares a model prop but was not covered by the declaration-style rule. This implements the accepted direction discussed in #3069 by extending the existing rule instead of adding a nearly overlapping rule.

Closes #3069

## How tested

- `npx vitest run --reporter=dot tests/lib/rules/define-props-declaration.test.ts` (14 passed)
- `npm test` (289 files, 11,525 tests passed)
- `npm run build`
- `npm run update`
- ESLint on the changed rule and test, with the newly introduced repository-wide `require-meta-languages` check disabled
- markdownlint on the changed rule documentation and generated index
- `git diff --check`

Current `master` baseline notes:

- `npm run lint` reports 220 repository-wide `meta.languages` errors after resolving `eslint-plugin-eslint-plugin@7.5.0`; this affects existing rules and internal rules, not only this change.
- `npm run tsc` reports four existing errors in `lib/rules/prefer-separate-static-class.js`, which is not changed here.

## Risk and rollout

- Risk: Low. This only changes an opt-in rule and only adds checks for `defineModel` calls inside `<script setup lang=ts>`.
- Rollback: Revert this commit to restore `defineProps`-only behavior.

## Notes for reviewers

- Duplicate search covered issue `#3069`, `define-model-declaration`, `define-props-declaration`, `defineModel` + runtime/type-based declaration, and `onDefineModelEnter` across open/closed/merged PRs, issues, comments, and commits.
- Open PR #2466 adds autofixing for `defineProps`; it does not inspect `defineModel` or implement this issue.
- Open PR #3032 adds `defineModel` checks to the default-prop rules; it does not enforce declaration style.
- In type-based mode, `defineModel<T>(name, options)` remains valid because runtime options such as `default` and `required` can accompany a type argument.

### DIFF
--- a/.changeset/kind-models-declare.md
+++ b/.changeset/kind-models-declare.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": minor
+---
+
+Extended `vue/define-props-declaration` to enforce the declaration style of `defineModel`
--- a/docs/rules/define-props-declaration.md
+++ b/docs/rules/define-props-declaration.md
@@ -2,17 +2,17 @@
 pageClass: rule-details
 sidebarDepth: 0
 title: vue/define-props-declaration
-description: enforce declaration style of `defineProps`
+description: enforce declaration style of `defineProps` and `defineModel`
 since: v9.5.0
 ---
 
 # vue/define-props-declaration
 
-> enforce declaration style of `defineProps`
+> enforce declaration style of `defineProps` and `defineModel`
 
 ## :book: Rule Details
 
-This rule enforces `defineProps` typing style which you should use `type-based` or `runtime` declaration.
+This rule enforces whether `defineProps` and `defineModel` should use `type-based` or `runtime` declarations.
 
 This rule only works in setup script and `lang="ts"`.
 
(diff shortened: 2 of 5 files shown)


FloEdelmann (changes requested):
See https://github.com/vuejs/eslint-plugin-vue/pull/3101#discussion_r3631153330:

> In `type-based` mode, `defineModel()` with no type argument currently reports `messageId: 'hasArg'` (“Use type-based declaration instead of runtime declaration.”). That message is misleading for the zero-argument form because it’s not a runtime declaration; it’s an untyped declaration. Consider emitting a dedicated message for the no-arg case (while keeping `hasArg` for runtime-style `defineModel({ ... })`).

But use a new `messageId` instead of inlining the message.

Issue #2777, opened 2025-07-03
Rewrite it in TypeScript


ota-meshi:
**Tell us about your environment**

- **ESLint version:** v9.30.0
- **eslint-plugin-vue version:** v10.3.0
- **Vue version:** v3.5.17
- **Node version:** v22.17.0

**The problem you want to solve.**

I would like this project to be rewritten in TypeScript.

**Your take on the correct solution to problem.**

Rewrite it in TypeScript.

**Additional context**

I think it would be difficult to rewrite all the code at once. It might be better to consider an infrastructure that allows we to rewrite it bit by bit.

Related to:

https://github.com/vuejs/eslint-plugin-vue/pull/2528#issuecomment-2334027137
#2523

FloEdelmann:
Related PRs and issues:

#### PRs released in [v10.6.0](https://github.com/vuejs/eslint-plugin-vue/releases/tag/v10.6.0)

- #2960

#### PRs released / issues introduced in [v10.7.0](https://github.com/vuejs/eslint-plugin-vue/releases/tag/v10.7.0)

- #2916
- #2967
- #2968
- #3004
- #3005
- #3010

#### PRs released in [v10.8.0](https://github.com/vuejs/eslint-plugin-vue/releases/tag/v10.8.0)

_None regarding TypeScript support_

#### PRs released in [v10.9.0](https://github.com/vuejs/eslint-plugin-vue/releases/tag/v10.9.0)

- #3002
- #3027
- #3028
- #3034
- #3037
- #3038
- #3039
- #3041
- #3042
- #3036
- #3043
- #3049
- #3048
- #3047
- #3046
- #3045
- #3051
- #3044
- #3053
- #3054
- #3057

#### PRs not yet released

- #3055
- #3007
- #3013
- #3016

ValentinYoushkevich:
I'd like to help with the rules that are still in JS (about 195 of them). Before starting I want to make sure I don't step on #3007 and #3055 by @9romise, which have been conflicting since March, or on the ESM work in #3102 that targets `next`.

Should rule migrations go to `master` or to `next`? And are there rules I should leave alone for now? If it's fine, I'd go in small PRs of a few related rules each, starting with ones that don't depend on the utils touched by those PRs.

PR #3102, opened 2026-07-20
feat!: migrate to pure ESM, drop Node.js < 22

sxzz:
## Summary

- publish `eslint-plugin-vue` as an ESM-only package with an import-only package export
- migrate the package source, internal rules, tools and scripts, tests, and documentation tooling from CommonJS module syntax to native ESM
- require Node.js `^22.19.0 || ^24.11.0 || >=26.0.0`

## Breaking changes

- CommonJS consumers can no longer load the package with `require('eslint-plugin-vue')`; use `import` instead
- Node.js versions outside the updated `engines.node` range are no longer supported

## Scope

Related to #2777. This PR intentionally handles only the ESM migration; rewriting the JavaScript source in TypeScript remains out of scope.

## Validation

- `npm run update`
- `npm run lint`
- `npm run tsc`
- `npm run build`
- `npm test`
- `npm run docs:build`
- `npm pack --dry-run`


package.json
```diff
@@ -2,8 +2,16 @@
...
+  "type": "module",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
```
FloEdelmann, line 14: There should be exports for utils and typings, see

- https://github.com/vuejs/eslint-plugin-vue/issues/3005
- https://github.com/vuejs/eslint-plugin-vue/issues/3010

eslint.config.js
```diff
@@ -77,6 +75,7 @@ export default typegen([
     rules: {
       'eslint-plugin/require-meta-default-options': 'off', // TODO: enable when all rules have defaultOptions
       'eslint-plugin/require-meta-docs-recommended': 'off', // use `categories` instead
+      'eslint-plugin/require-meta-languages': 'off',
```
FloEdelmann: It is already disabled for `lib/rules/*.{js,ts}` files and thus should not be needed here

FloEdelmann (changes requested):
I'd prefer migrating to TypeScript first, then we could maybe introduce an ESM bundle alongside CJS in a minor version and drop CJS in the next major version. This would give users a smoother migration path, especially those on older Node.js version that your PR drops as well.

Also note that in TypeScript files the ESM syntax is already supported (and transpiled to CJS), so the developer experience is best in TypeScript files.

What could be done already is migrating internal files (e.g. `tools/*.js`, `*.config.js`) to ESM by using the `.mjs` file extension.

sxzz:
Thanks for the detailed review. My original plan was to migrate the project to both ESM-only and TypeScript at the same time. However, that resulted in a change set that was too large to review effectively, so I decided to split the work into multiple PRs.

This only changes how the migration is reviewed, not how it will be released: the ESM-only and TypeScript migrations are still intended to be released together as part of the next major version.

Changing the order now and migrating to TypeScript first would require rewriting almost the entire PR again. I’m sorry, but I don’t think that additional churn is necessary when the intended final result and release timing remain the same.

To avoid disrupting maintenance of the current version, perhaps these migration PRs could be merged into a `next` branch and released together once the next major version is ready.

package.json
sxzz: I agree that these exports should be supported. However, since they are tracked by two separate issues and already have corresponding PRs, I would prefer not to expand the scope of this already large migration PR. We can incorporate those changes in a follow-up PR before the next major version is released.

FloEdelmann:
But when rewriting to TypeScript first, we don't *have* to release a major version immediately.

package.json
FloEdelmann: Without `exports`, those issues have workarounds. With adding `exports`, you make those files completely inaccessible, so it has to be done in one go. But adding `exports` (with those files) could be done in a separate PR that is merged before this one.

sxzz:
That’s true, but shipping the TypeScript migration first would make very little practical difference to end users, since the package would still be compiled to JavaScript and declaration files. To me, this is mainly a trade-off: is it worth discarding or substantially reworking the changes in this PR in order to first ship a TypeScript version that is almost identical from a consumer’s perspective?

My current proposal for the next major version would be:

- release it once all of the planned work is complete
- publish as pure ESM
- use TypeScript-only source code, while publishing compiled JavaScript and declaration files
- remove the legacy configs (flat configs only)
- require Node.js `^22.19.0 || ^24.11.0 || >=26.0.0`

Perhaps I should open a separate issue first so that we can discuss and agree on the scope and release strategy for the next major version before proceeding.

sxzz:
I opened #3105 to discuss the scope and release strategy for the next major version. Let’s continue the broader discussion there.

sxzz:
Based on the discussion in #3105, especially @ota-meshi's and @waynzh's feedback, it looks like we're close to broad agreement on moving directly to pure ESM in the next major, developing the breaking changes on a `next` branch, and reviewing the TypeScript and ESM migrations separately while releasing them together.

@ota-meshi @waynzh @FloEdelmann, if you're broadly aligned with this direction, could you please review/re-review this PR? We can continue discussing any remaining release-strategy details in #3105.

lib/utils/index.js
```diff
@@ -669,7 +672,7 @@ function wrapRuleModule(baseRule, ruleName, options) {
 // Exports
 // ------------------------------------------------------------------------------
 
-module.exports = {
+export default {
```
ota-meshi: Could you use named exports for each utility instead of `export default`?
I think that would likely make the code easier to read.

lib/utils/scope.js
ota-meshi: Could you use named export instead of `export default`?

```suggestion
/**
* Gets the scope for the current node
* @param {RuleContext} context The rule context
* @param {ESNode} currentNode The node to get the scope of
* @returns { import('eslint').Scope.Scope } The scope information for this node
*/
export function getScope(context, currentNode) {
```

lib/utils/ts-utils/typescript.js
```diff
@@ -11,7 +13,7 @@
 /** @type {TypeScript | undefined} */
 let cacheTypeScript
 
-module.exports = {
+export default {
```
ota-meshi: Could you use named exports for each utility instead of `export default`?

lib/utils/ts-utils/ts-types.js
```diff
@@ -28,7 +30,7 @@ const {
  * @typedef {import('../index').ComponentUnknownSlot} ComponentUnknownSlot
  */
 
-module.exports = {
+export default {
```
ota-meshi: Could you use named exports for each utility instead of `export default`?

lib/utils/ts-utils/ts-ast.js
```diff
@@ -73,7 +77,7 @@ const TYPE_NODE_TYPES = {
   TSVoidKeyword: true
 }
 
-module.exports = {
+export default {
```
ota-meshi: Could you use named exports for each utility instead of `export default`?

lib/utils/ts-utils/index.js
```diff
@@ -30,7 +33,7 @@ const {
  * @typedef {import('../index').ComponentUnknownSlot} ComponentUnknownSlot
  */
 
-module.exports = {
+export default {
```
ota-meshi: Could you use named exports for each utility instead of `export default`?

lib/rules/max-lines-per-block.js
```diff
@@ -2,10 +2,11 @@
  * @author lsdsjy
  * @fileoverview Rule for checking the maximum number of lines in Vue SFC blocks.
  */
-'use strict'
 
-const { SourceCode } = require('eslint')
-const utils = require('../utils')
+import * as SourceCodeModule from 'eslint'
```
ota-meshi: Could you use the name `eslint` instead of `SourceCodeModule`?

lib/rules/no-async-in-computed-properties.js
```diff
@@ -2,9 +2,10 @@
  * @fileoverview Check if there are no asynchronous actions inside computed properties.
  * @author Armano
  */
-'use strict'
-const { ReferenceTracker } = require('@eslint-community/eslint-utils')
-const utils = require('../utils')
+import * as ReferenceTrackerModule from '@eslint-community/eslint-utils'
```
ota-meshi: Could you use the name `eslintUtils` instead of `ReferenceTrackerModule`?

ota-meshi (changes requested):
Thank you for this PR!

lib/rules/no-async-in-computed-properties.js
sxzz: I refactored them with named imports:
```ts
import { ReferenceTracker } from '@eslint-community/eslint-utils'
```

lib/rules/html-quotes.js
```diff
@@ -3,11 +3,14 @@
...
  * See LICENSE file in root directory for full license.
  */
-'use strict'
 
-const utils = require('../utils')
+import {
+  defineTemplateBodyVisitor,
+  isVBindSameNameShorthand,
+  hasInvalidEOF as utilsHasInvalidEOF
+} from '../utils/index.js'
```
ota-meshi, line 11: I find `utilsHasInvalidEOF` a bit confusing. I think it would be better to use `* as utils` and keep the existing code instead. What do you think?

```js
import * as utils from '../utils/index.js'

// ...

          hasInvalidEOF = utils.hasInvalidEOF(node)
```

sxzz: The renamed `hasInvalidEOF` variable to `hasTemplateEOFError`.

lib/rules/prefer-separate-static-class.js
```diff
@@ -33,7 +32,11 @@ function isStaticClassTemplateElement(templateElement, index, templateLiteral) {
  * @returns {(Literal | TemplateLiteral | TemplateElement | Identifier)[]}
  */
 function findStaticClasses(expressionNode) {
-  if (isStringLiteral(expressionNode)) {
+  if (
+    Boolean(isStringLiteral(expressionNode)) &&
+    (expressionNode.type === 'Literal' ||
```
ota-meshi: I thought no additional checks were necessary, but was that not the case?

ota-meshi (changes requested):

lib/rules/prefer-separate-static-class.js
sxzz: You’re right, the extra runtime checks aren’t needed. They were only there because, after switching to a named ESM export, TypeScript started seeing the isStringLiteral type predicate. The predicate was too broad: it covered every TemplateLiteral, although the function only returns true for template literals without expressions. I tightened the predicate and restored the original code. The next PR will migrate the whole codebase to TypeScript, which should make this cleaner.

tools/new-rule.js
ota-meshi: I think it probably needs to be changed.

```diff
- import utils from '../utils/index.js'
+ import * as utils from '../utils/index.js'
```

ota-meshi (changes requested):
Thank you for making the changes to this PR.
I have left one comment.
Also, could you add two changeset files to explain the changes in this PR?
One should cover the switch to ESM, and the other should cover dropping support for older Node versions.

tools/new-rule.js
sxzz: Fixed, and added the two changesets as well. I went with `import { defineTemplateBodyVisitor } from '../utils/index.js'` here. I prefer named imports over namespace imports when we only need specific utilities, since they make the dependencies explicit and can help bundlers with tree-shaking.

ota-meshi (approved):
LGTM! Thanks!

PR #3103, opened 2026-07-20
chore: disable require-meta-languages rule to avoid breaking changes

ota-meshi:
This PR disables the `eslint-plugin/require-meta-languages` rule in our ESLint config.

It appears that the new `eslint-plugin-eslint-plugin` introduced the `eslint-plugin/require-meta-languages` rule, but we do not need it at this time.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -316,7 +316,8 @@ export default typegen([
         { pattern: 'https://eslint.vuejs.org/rules/{{name}}.html' }
       ],
       'internal/no-invalid-meta': 'error',
-      'internal/no-invalid-meta-docs-categories': 'error'
+      'internal/no-invalid-meta-docs-categories': 'error',
+      'eslint-plugin/require-meta-languages': 'off'
     }
   },
   {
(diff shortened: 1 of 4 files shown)


eslint.config.mjs
```diff
@@ -316,15 +316,17 @@ export default typegen([
...
     }
   },
   {
     files: ['eslint-internal-rules/*.js'],
     rules: {
       'eslint-plugin/require-meta-docs-url': 'off',
       'internal/no-invalid-meta': 'error',
-      'internal/no-invalid-meta-docs-categories': 'error'
+      'internal/no-invalid-meta-docs-categories': 'error',
+      'eslint-plugin/require-meta-languages': 'off' // We are intentionally disabling it because it is a breaking change.
```
FloEdelmann: We could probably enable it for our internal rules. I'm also fine with keeping it disabled for them, too, but then the comment is wrong here.

eslint.config.mjs
```diff
@@ -316,15 +316,17 @@ export default typegen([
         { pattern: 'https://eslint.vuejs.org/rules/{{name}}.html' }
       ],
       'internal/no-invalid-meta': 'error',
-      'internal/no-invalid-meta-docs-categories': 'error'
+      'internal/no-invalid-meta-docs-categories': 'error',
+      'eslint-plugin/require-meta-languages': 'off'
```
FloEdelmann: Please sort `eslint-plugin/require-meta-languages` after `eslint-plugin/require-meta-docs-url` to keep them together and ordered alphabetically.

FloEdelmann (changes requested):

ota-meshi:
Thanks for the review! I've updated the pull request.
Could you please check it again?

FloEdelmann (approved):
Looks good now, thanks!

PR #3108, opened 2026-07-31
fix: treat externally passed Vue props as used

### LINKED ISSUE #2062: "vue/no-unused-properties" work incorrect with useVModels from "@vueuse/core"
**Checklist**

- **ESLint version:** ^8.22.0
- **eslint-plugin-vue version:** ^9.3.0
- **Node version:** v16.17.1
- **Operating System:** macOS 13.0.1

```json5
/* eslint-env node */

module.exports = {
  root: true,
  'extends': [
    'plugin:vue/vue3-essential',
    'eslint:recommended',
    '@vue/eslint-config-typescript'
  ],
  parserOptions: {
    ecmaVersion: 'latest'
  },
  rules: {
    "vue/no-unused-properties": ["error", {
      "groups": ["setup", "props"],
    }]
  }
}

```

**What did you do?**	

```vue
<script lang="ts">
import { defineComponent } from 'vue'
import { useVModels } from '@vueuse/core'

  props: {
    msg: {
      type: String,
      required: true
    }
  },
  setup(props,{ emit }) {
    const {
      msg: msgModel
    } = useVModels(props, emit)

    return {
      msgModel
    }
  }
})
</script>
```

I expected no eslint errors, because I use msg props through useVModels

**What actually happened?**
image

https://github.com/galsim/no-unused-properties-use-v-models

raisulchowdhury:
## Summary

Treat props passed to functions outside the current file as potentially used.
This prevents `vue/no-unused-properties` from reporting false positives for
helpers such as `useVModels`, while retaining precise same-file analysis where
the callee can be traced.

Fixes #2062.

## Validation

- baseline false positives reproduced
- focused suite: 234 passing
- formatting and diff checks
- target lint still reports the two unchanged meta-schema errors present on
  the baseline branch

## AI assistance

AI assistance was used for investigation and drafting. I reviewed the property
reference contract and independently ran the focused validation, preserving
the pre-existing meta-schema errors in the report above.

### DIFF
--- a/.changeset/curvy-bars-vanish.md
+++ b/.changeset/curvy-bars-vanish.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fix false positives in `vue/no-unused-properties` when props are passed to functions imported from another module.
--- a/lib/rules/no-unused-properties.ts
+++ b/lib/rules/no-unused-properties.ts
@@ -306,7 +306,9 @@ export default {
     ) {
       if (
         (property.groupName === 'props' &&
-          propertyReferencesForProps.hasProperty(property.name)) ||
+          propertyReferencesForProps.hasProperty(property.name, {
+            unknownCallAsAny: true
+          })) ||
         propertyReferences.hasProperty('$props')
       ) {
         // used props
--- a/tests/lib/rules/no-unused-properties.test.ts
+++ b/tests/lib/rules/no-unused-properties.test.ts
@@ -1541,6 +1541,95 @@ tester.run('no-unused-properties', rule, {
         </script>
       `
     },
+    // Props passed to a function whose implementation is outside this file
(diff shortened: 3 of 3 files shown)


.changeset/curvy-bars-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fix false positives in `vue/no-unused-properties` when props are passed to functions imported from another module.
```
FloEdelmann, line 5: for consistency with other changelog entries:

```suggestion
Fixed false positives in `vue/no-unused-properties` when props are passed to functions imported from another module
```

FloEdelmann (commented):
I guess the change makes sense, but given we have test cases that explicitly mention those patterns as invalid, can you please research if the cases were already discussed somewhere?

PR #3111, opened 2026-08-05
refactor: extract common type in VNodeListenerMap

FloEdelmann:


FloEdelmann:
Lint violations are fixed in #3112.

PR #3113, opened 2026-08-11
fix(no-unused-properties): ignore props typed as `never`

### LINKED ISSUE #2545: vue/no-unused-properties -- reports properties of type 'never' as unused.
**What did you do?**	

```vue
<script setup lang="ts">
interface MyTextAreaInput extends /* @vue-ignore */ TextareaHTMLAttributes {
  // prevent accidentally using min/max instead of minlength/maxlength
  min?: never; // <- 'min' of property found, but never used.
  max?: never; // <- 'max' of property found, but never used.
}
defineProps<MyTextAreaInput>()

</script>
```

I realize it is very unusual to use type `never` as a prop type. So not very important issue.

ValentinYoushkevich:
Closes #2545

A prop explicitly typed as `never` is declared to *forbid* its usage, so it is never referenced by design. Reporting it as unused is noise:

```vue
<script setup lang="ts">
interface MyTextAreaInput extends /* @vue-ignore */ TextareaHTMLAttributes {
  // prevent accidentally using min/max instead of minlength/maxlength
  min?: never // <- 'min' of property found, but never used.
  max?: never // <- 'max' of property found, but never used.
}
defineProps<MyTextAreaInput>()
</script>
```

### Implementation note

`prop.types` cannot be used to detect this. `inferRuntimeType` has no `TSNeverKeyword` case, so it falls through to `inferRuntimeTypeFromTypeNode`, which returns `['null']` when no type information is available — making `never` indistinguishable from any other unresolved type.

The check is therefore done on the AST: the prop's `TSPropertySignature` type annotation is `TSNeverKeyword`. This only requires `@typescript-eslint/parser`, no type-aware setup, which matches @FloEdelmann's note on the issue.

### Deliberately out of scope

An aliased `never` (`type Impossible = never; min?: Impossible`) is **not** covered — that is a `TSTypeReference` and can only be resolved with full type information, which requires users to configure `parserOptions.project`. Happy to extend this if you would like it covered.

Props typed `never` are skipped regardless of optionality: a required `min: never` cannot be passed a value either, so reporting it as unused is equally unhelpful.

### DIFF
--- a/lib/rules/no-unused-properties.ts
+++ b/lib/rules/no-unused-properties.ts
@@ -415,10 +415,7 @@ export default {
           )
 
           for (const prop of props) {
-            if (!prop.propName) {
-              continue
-            }
-            if (isNeverTypeProp(prop)) {
+            if (!prop.propName || isNeverTypeProp(prop)) {
               continue
             }
             if (prop.type === 'object') {
--- a/lib/rules/no-unused-properties.ts
+++ b/lib/rules/no-unused-properties.ts
@@ -67,6 +67,19 @@ const PROPERTY_LABEL = {
   expose: 'expose'
 }
 
+/**
+ * Check whether the given prop is explicitly typed as `never`.
+ * Such props are declared to forbid their usage, so they are never
+ * referenced by design and reporting them as unused is noise.
+ */
+function isNeverTypeProp(prop: ComponentProp): boolean {
+  return (
+    prop.type === 'type' &&
+    prop.node.type === 'TSPropertySignature' &&
+    prop.node.typeAnnotation?.typeAnnotation.type === 'TSNeverKeyword'
+  )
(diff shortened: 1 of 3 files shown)


lib/rules/no-unused-properties.ts
```diff
@@ -405,6 +418,9 @@ export default {
             if (!prop.propName) {
               continue
             }
+            if (isNeverTypeProp(prop)) {
+              continue
+            }
```
FloEdelmann:
```suggestion
            if (!prop.propName || isNeverTypeProp(prop)) {
              continue
            }
```

FloEdelmann (changes requested):

FloEdelmann (approved):
Looks good to me, thanks!

PR #3114, opened 2026-08-11
fix(require-default-prop): no false positive for `withDefaults` defaults passed by reference

### LINKED ISSUE #2993: vue/require-default-prop: false positive when pass withDefaults argument by reference
**Checklist**

- **ESLint version:**  9.39.1
- **eslint-plugin-vue version:** 10.5.1 
- **Vue version:** 3.5.25
- **Node version:** 24.11.1
- **Operating System:** Windows 10 pro 22H2

```json5
import { globalIgnores } from 'eslint/config'
import { defineConfigWithVueTs, vueTsConfigs } from '@vue/eslint-config-typescript'
import pluginVue from 'eslint-plugin-vue'

export default defineConfigWithVueTs(
  {
    name: 'app/files-to-lint',
    files: ['**/*.{ts,mts,tsx,vue}'],
  },

  globalIgnores(['**/dist/**', '**/dist-ssr/**', '**/coverage/**']),

  pluginVue.configs['flat/recommended'],
  vueTsConfigs.recommended,
)
```

**What did you do?**	

```vue
<script setup lang="ts">
defineOptions({ name: 'ExampleComponent' });

interface Props {
  optionalProp?: string;
}

const defaultsObj = {
  optionalProp: 'hello!',
}

withDefaults(defineProps<Props>(), defaultsObj);
</script>
```

No "vue/require-default-prop" error for "optionalProp"

**What actually happened?**

**\src\ExampleComponent.vue
  5:3  warning  Prop 'optionalProp' requires default value to be set  vue/require-default-prop

✖ 1 problem (0 errors, 1 warning)

https://github.com/tilonorrinco/bug-vue-require-default-prop

Thanks in advance!

ValentinYoushkevich:
Closes #2993

## Problem

`withDefaults()` accepts any expression as its defaults argument: the SFC compiler inlines an object literal and falls back to the `mergeDefaults()` runtime helper for everything else. `vue/require-default-prop` only understood the inline literal, because `getWithDefaultsProps()` bails out unless the second argument is an `ObjectExpression` and skips `SpreadElement` properties.

All of these were reported as `Prop 'foo' requires default value to be set.` even though a default is provided:

```vue
<script setup lang="ts">
import { defaults, makeDefaults } from './defaults'

withDefaults(defineProps<{ foo?: string }>(), defaults)
withDefaults(defineProps<{ foo?: string }>(), { ...defaults })
withDefaults(defineProps<{ foo?: string }>(), makeDefaults())
withDefaults(defineProps<{ foo?: string }>(), { foo: 'foo' } as const)
</script>
```

## Fix

The rule now resolves the defaults argument itself: identifiers and spread sources are resolved to the object they refer to, reusing the existing `getObjectOrArray()` helper (already used for `defineProps(obj)`), which is now exported from `lib/utils`; if any part of the defaults cannot be determined statically — an import, a call expression, a computed key — no type prop is reported for that `defineProps()`, since any of them may receive a default at runtime.

Props that genuinely have no default are still reported, including when the referenced object resolves and does not contain them:

```vue
<script setup lang="ts">
const defaults = { foo: 'foo' }
// 'bar' is still reported
withDefaults(defineProps<{ foo?: string; bar?: number }>(), defaults)
</script>
```

Only `require-default-prop` changes behaviour. The shared `getWithDefaultsProps()` util is untouched, so the other rules using it are unaffected.

### DIFF
--- a/lib/rules/require-default-prop.js
+++ b/lib/rules/require-default-prop.js
@@ -47,62 +47,6 @@ function isValueNodeOfBooleanType(value) {
   return false
 }
 
-/**
- * Collects the names of the props that are given a default value by
- * `withDefaults()`. Vue accepts any expression as the defaults argument and
- * falls back to the `mergeDefaults()` runtime helper, so the defaults cannot
- * always be determined statically. `hasUnknown` is `true` in that case, which
- * means that any prop may have a default value.
- * @param {RuleContext} context The rule context object.
- * @param {CallExpression} node The node of defineProps
- * @return {{ names: Set<string>, hasUnknown: boolean }}
- */
-function getWithDefaultsPropNames(context, node) {
-  /** @type {Set<string>} */
-  const names = new Set()
-  if (!utils.hasWithDefaults(node)) {
-    return { names, hasUnknown: false }
-  }
-  const defaults = node.parent.arguments[1]
-  if (!defaults) {
-    return { names, hasUnknown: false }
-  }
-
-  let hasUnknown = false
-  /** @type {ESNode[]} */
-  const queue = [defaults]
-  /** @type {Set<ESNode>} */
-  const visited = new Set()
-  while (queue.length > 0) {
-    const target = /** @type {ESNode} */ (queue.shift())
-    if (visited.has(target)) {
-      continue
-    }
-    visited.add(target)
-
-    const object = utils.getObjectOrArray(context, target)
-    if (object?.type !== 'ObjectExpression') {
-      hasUnknown = true
-      continue
-    }
-    for (const property of object.properties) {
-      if (property.type === 'SpreadElement') {
-        // The spread source may itself be resolvable, so process it as well.
-        queue.push(property.argument)
-      } else {
-        const name = utils.getStaticPropertyName(property)
-        if (name == null) {
-          hasUnknown = true
-        } else {
-          names.add(name)
-        }
-      }
-    }
-  }
-
-  return { names, hasUnknown }
-}
-
 module.exports = {
   meta: {
     type: 'suggestion',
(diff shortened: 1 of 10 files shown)


lib/rules/require-default-prop.js
```diff
@@ -224,8 +280,7 @@ module.exports = {
       utils.defineScriptSetupVisitor(context, {
         onDefinePropsEnter(node, props) {
           const hasWithDefaults = utils.hasWithDefaults(node)
-          const defaultsByWithDefaults =
-            utils.getWithDefaultsPropExpressions(node)
+          const propsWithDefaults = getWithDefaultsPropNames(context, node)
```
FloEdelmann: > The shared `getWithDefaultsProps()` util is untouched, so the other rules using it are unaffected.

Does that really make sense? Shouldn't the cases mentioned in the issue be handled properly in the other rules as well?

ValentinYoushkevich: Fair point — I checked, and the same blind spot is in three other rules that go through `getWithDefaultsPropExpressions()`: `no-boolean-default`, `require-valid-default-prop` and `no-required-prop-with-default`. All of them get `{}` as soon as the defaults are passed by reference.

The difference is the direction of the failure. Those three under-report, while `require-default-prop` over-reports — it flags props that do have a default and breaks otherwise valid code, which is what #2993 is about. That's why I scoped the fix to it.

Pushing the resolution into `getWithDefaultsProps()` isn't a drop-in either: `no-restricted-props` uses the `Property` nodes it returns as edit targets for its suggestions, so once the lookup follows an identifier those edits land in an object literal declared elsewhere — possibly shared or imported — and renaming a key there is wrong. The shapes differ too: this rule only needs the set of names plus a "there may be defaults I can't see" flag, the others need the expression per prop.

Happy to cover them, either here or as a follow-up: add resolution to `getWithDefaultsPropExpressions()` with a way to signal the unresolvable case, adopt it in the three read-only consumers, and leave the suggestion path on the literal-only lookup. Which would you prefer?

FloEdelmann: I'd do it here in the same PR, so that the logic from `getWithDefaultsPropExpressions` doesn't need to be copied+modified into `require-default-prop` in this PR, and then removed from there again in the follow-up PR.

ValentinYoushkevich: Done in fac2e5d6. `getWithDefaultsPropExpressions()` now takes `(context, node)` and returns `{ props, hasUnknown }`, and all four rules go through it. `getWithDefaultsProps()` stays literal-only, since fixers edit the nodes it returns.

One part wasn't a straight move: collecting expressions instead of just names makes ordering matter, so in `{ ...base, a: 1 }` the literal `a` has to win over the spread. The queue-based walk got that wrong — spreads are expanded in place now.

FloEdelmann (approved):
Thanks!

PR #3115, opened 2026-08-11
fix(no-ref-object-reactivity-loss): report values gotten in an IIFE

### LINKED ISSUE #3056: no-ref-object-reactivity-loss: false negative when the accessing is wrapped in IIFE with the same scope
**Checklist**

- **ESLint version:** 10.0.2
- **eslint-plugin-vue version:** 10.8.0
- **Vue version:** 3.5.29
- **Node version:** 24.12.0
- **Operating System:** NixOS

https://github.com/n0099/open-tbm/blob/e48097c0ac96da264942f52d74bcd294855df35a/fe/eslint.config.ts

**What did you do?**	

```ts
const firstThread = computed(() => firstPostPage.value?.threads[0]); // https://github.com/n0099/open-tbm/blob/92fd7111a62c16f0f7ec29da5fae04a34532f1e2/fe/src/utils/post/seo/index.ts#L18
console.log(firstThread.value?.title);
//          ^ reporting violation of eslint rule `vue/no-ref-object-reactivity-loss`
(() => {
    const firstThreadTitle = computed(() => firstThread.value?.title); // https://github.com/n0099/open-tbm/blob/92fd7111a62c16f0f7ec29da5fae04a34532f1e2/fe/src/utils/post/seo/index.ts#L56
    console.log(firstThread.value?.title);
    //          ^ not reporting violation of eslint rule `vue/no-ref-object-reactivity-loss`
    console.log(firstThreadTitle.value);
    //          ^ reporting violation of eslint rule `vue/no-ref-object-reactivity-loss`
})();
```

`firstThread.value?.title` inside this IIFE should also report.

**What actually happened?**

Not reporting.

https://github.com/n0099/open-tbm/blob/e48097c0ac96da264942f52d74bcd294855df35a/fe

ValentinYoushkevich:
Closes #3056

The body of an immediately invoked function expression runs synchronously in the enclosing scope, so getting a value from a ref object there loses reactivity just like getting it directly:

```js
import { ref } from 'vue'
const count = ref(0)
const v1 = count.value // reported
;(() => {
  const v2 = count.value // not reported before this PR
})()
```

The rule pushes a new scope entry for every function it enters and compares the scope of the `ref()`/`computed()` call with the current one by identity. The callee of an IIFE got its own entry, so the check failed and the report was silenced.

This PR keeps the enclosing scope when entering a function that is the callee of a call expression. `async` functions and generators are excluded, because their bodies are not guaranteed to run synchronously to the end.

The second case from the issue — a ref passed as an argument to a plain function that is then called — is not covered, since it needs inter-procedural analysis.

### DIFF
--- a/docs/rules/no-ref-object-reactivity-loss.md
+++ b/docs/rules/no-ref-object-reactivity-loss.md
@@ -25,21 +25,7 @@ const v3 = computed(() => count.value /* ✓ GOOD */)
 const v4 = fn(count.value) /* ✗ BAD */
 const v5 = fn(count) /* ✓ GOOD */
 const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
-```
-
-</eslint-code-block>
-
-The body of an immediately invoked function expression (IIFE) is executed in the same scope, so the values gotten there also lose reactivity.
-
-<eslint-code-block :rules="{'vue/no-ref-object-reactivity-loss': ['error']}" language="javascript" filename="example.js" >
-
-```js
-import { ref } from 'vue'
-const count = ref(0)
-;(() => {
-  const v1 = count.value /* ✗ BAD */
-  const v2 = computed(() => count.value /* ✓ GOOD */)
-})()
+const v7 = (() => count.value)() /* ✗ BAD */
 ```
 
 </eslint-code-block>
--- a/docs/rules/no-ref-object-reactivity-loss.md
+++ b/docs/rules/no-ref-object-reactivity-loss.md
@@ -29,6 +29,21 @@ const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
 
 </eslint-code-block>
 
+The body of an immediately invoked function expression (IIFE) is executed in the same scope, so the values gotten there also lose reactivity.
+
+<eslint-code-block :rules="{'vue/no-ref-object-reactivity-loss': ['error']}" language="javascript" filename="example.js" >
+
+```js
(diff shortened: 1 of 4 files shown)


docs/rules/no-ref-object-reactivity-loss.md
```diff
@@ -29,6 +29,21 @@ const v6 = computed(() => fn(count.value) /* ✓ GOOD */)
...
+```js
+import { ref } from 'vue'
+const count = ref(0)
+;(() => {
+  const v1 = count.value /* ✗ BAD */
+  const v2 = computed(() => count.value /* ✓ GOOD */)
+})()
+```
+
+</eslint-code-block>
```
FloEdelmann: Not sure if it's worth to explain this in detail. Maybe just add an IIFE to the code example above?

FloEdelmann (changes requested):
The tests and implementation looks good to me, thank you! I just have a small suggestion for the rule docs.

docs/rules/no-ref-object-reactivity-loss.md
ValentinYoushkevich: Agreed, that was too much for one case. Dropped the section and added `const v7 = (() => count.value)() /* ✗ BAD */` to the example above instead — 3a37b56f.

FloEdelmann (approved):

waynzh (approved):
Thanks!

PR #3116, opened 2026-08-11
chore(tools): don't fail `npm run new` when VS Code CLI is missing

### LINKED ISSUE #2781: npm run new fails when VS Code CLI is not installed
---
name: Bug report
about: Create a report to help us improve
---

**Checklist**

- **ESLint version:** 9.17.0 (development environment issue, not rule-specific)
- **eslint-plugin-vue version:** 10.2.0
- **Vue version:** 3.x (development environment issue, not app-specific)
- **Node version:** v23.11.0
- **Operating System:** macOS (Darwin 24.5.0)

This is a development tooling issue with npm run new command.

**What did you do?**
```bash
npm run new <rule-name> <author>
```

The command should complete successfully without errors, providing a smooth experience for new contributors following the developer guide.

**What actually happened?**

The command generates files successfully but fails at the end when trying to automatically open files in VS Code:

```
/bin/sh: code: command not found
Error: Command failed: code "path/to/generated/file"
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at checkExecSyncError (node:child_process:882:11)
    at Object.execSync (node:child_process:954:15)
    at /path/to/tools/new-rule.js:143:6
```

The issue occurs because the script assumes VS Code CLI (`code` command) is installed, but many developers use other editors (like Cursor, Vim, etc.) or don't have VS Code command line tools available. This creates a poor first impression for new contributors.

This issue can be reproduced on any system where VS Code command line tools are not installed. The problem is in `tools/new-rule.js:143-145` where it executes `code` command without error handling to open generated files automatically.

**Suggested Solution**

Add graceful fallback when `code` command is not available:
```javascript
try {
  cp.execSync(`code "${ruleFile}"`, { stdio: 'ignore' })
} catch (err) {
  // Continue silently if VS Code is not available
  console.log('Files generated successfully')
}
```

ValentinYoushkevich:
Closes #2781

`npm run new <rule name> <author name>` writes the rule, test and docs files and then opens them with the VS Code CLI. When `code` is not on `PATH` — the contributor uses another editor, or VS Code is installed without the shell command — the first `execSync` throws:

```
/bin/sh: code: command not found
Error: Command failed: code "path/to/generated/file"
    at Object.execSync (node:child_process:954:15)
    at tools/new-rule.js:143:6
```

All three files are generated successfully at that point, but:

- the remaining two files are never opened, because the first `execSync` throws;
- the script exits with code 1 and a Node stack trace, so it looks like the generation itself failed;
- the paths of the generated files are never printed, so it is not obvious what was created.

This is the first command a new contributor runs from the developer guide, so it is worth failing softly.

### Changes

- print the paths of the generated files;
- wrap the editor launch in `try`/`catch` and ignore the failure — the files are already on disk, there is nothing to recover from.

### Verification

Without `code` on `PATH`:

```
$ npm run new some-rule Someone
Created:
  lib/rules/some-rule.ts
  tests/lib/rules/some-rule.test.ts
  docs/rules/some-rule.md
$ echo $?
0
```

With `code` on `PATH` the three files are still opened as before.

No changeset: `tools/` is not part of the published package (`"files": ["dist"]`), same as #3049

### DIFF
--- a/tools/new-rule.js
+++ b/tools/new-rule.js
@@ -139,7 +139,19 @@ Nothing.
 `
   )
 
-  cp.execSync(`code "${ruleFile}"`)
-  cp.execSync(`code "${testFile}"`)
-  cp.execSync(`code "${docFile}"`)
+  const createdFiles = [ruleFile, testFile, docFile]
+
+  logger.log('Created:')
+  for (const file of createdFiles) {
+    logger.log(`  ${path.relative(process.cwd(), file)}`)
+  }
+
+  try {
+    for (const file of createdFiles) {
+      cp.execSync(`code "${file}"`, { stdio: 'ignore' })
+    }
+  } catch {
+    // The VS Code CLI (`code`) is not available.
+    // The files have already been created, so there is nothing to recover from.
+  }
 })(process.argv[2], process.argv[3])


FloEdelmann (approved):
Makes sense to me, thank you!

PR #3119, opened 2026-08-12
feat: add `vue/no-content-outside-block` rule

### LINKED ISSUE #2209: no-code-outside-tag

[ ] Other (please specify:)

```vue
<template>
  <div>666</div>
</template>

console.log(1);

<script>
const num = 1;
console.log(num);
</script>

```

**Additional context**

the statement `console.log(1);` outside the tag will be ignored by vue compiler, and not work as expected

ValentinYoushkevich:
resolves #2209

Comments and custom blocks are left alone as you asked. Comments turned out to be free — `vue-eslint-parser` keeps them in `documentFragment.comments`, not in `children`.

Removal is a suggestion rather than an autofix: stray text is usually code that was meant to go inside a block, and having `--fix` silently delete it felt wrong.

### DIFF
--- a/tests/lib/rules/no-content-outside-block.test.ts
+++ b/tests/lib/rules/no-content-outside-block.test.ts
@@ -18,58 +18,62 @@ tester.run('no-content-outside-block', rule as RuleModule, {
     },
     {
       filename: 'test.vue',
-      code: `<template>
-  <div>666</div>
-</template>
+      code: `
+        <template>
+          <div>666</div>
+        </template>
 
-<script>
-export default {}
-</script>
+        <script>
+          export default {}
+        </script>
 
-<style>
-div {
-  color: red;
-}
-</style>
-`
+        <style>
+          div {
+            color: red;
+          }
+        </style>
+      `
     },
     {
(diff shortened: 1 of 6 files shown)


tests/lib/rules/no-content-outside-block.test.ts
```diff
@@ -0,0 +1,441 @@
...
+</style>
+`
+    },
+    {
+      // only a `<script>` block
+      filename: 'test.vue',
+      code: `<script>
+console.log(1)
+</script>
+`
```
FloEdelmann: Please indent code blocks:


```suggestion
      code: `
        <script>
          console.log(1)
        </script>
      `
```

ValentinYoushkevich: Done, indented all of them.

One side effect worth flagging: since the stray content is now indented, the removal suggestion leaves the line's indentation behind, so a few expected outputs have trailing whitespace on those lines. I can make the fixer drop the whole blank line instead if you'd prefer that.

waynzh (approved):
LGTM, thanks!

PR #3122, opened 2026-08-13
docs: explain how to apply the configs to a subset of files

### LINKED ISSUE #2448: Full example for flat config with object

ValentinYoushkevich:
The user guide only shows the flat configs applied to everything ESLint visits, so the common `.eslintrc` pattern of scoping a config to a few directories has no documented equivalent. The workaround people land on is mapping over the config array and overwriting `files`, which also overwrites the `**/*.vue` patterns the plugin declares — with a mixed scope such as `**/*.{ts,vue}` that applies the SFC processor to `.ts` files and skips them silently.

Adds a subsection after "Bundle Configurations" that shows `defineConfig()` with `extends`, side by side with the `.eslintrc` `overrides` block it replaces, and a warning box covering the ESLint 9.22 requirement and what to do below that version.

The examples were run against ESLint 10.8.1 to confirm the scoping actually holds and that the plugin's `.vue` restriction survives the merge.

Closes #2448

Two calls I would rather leave to you: `defineConfig` needs ESLint 9.22+ while the peer range still allows 8.57, which the warning box and the pre-9.22 fallback are meant to cover; and the section could equally sit next to the typescript-eslint example instead of after "Bundle Configurations".

### DIFF
--- a/docs/user-guide/index.md
+++ b/docs/user-guide/index.md
@@ -81,6 +81,73 @@ By default, all rules from **base** and **essential** categories report ESLint e
 Alternatively, there are versions of the configs with all rules set to error suffixed with `-error` that you can use.
 :::
 
+#### Applying the Configs to a Subset of Files (`eslint.config.js`)
+
+Apart from the entries that this plugin restricts to `.vue` files, the configs above apply to every file ESLint visits. To limit them to some directories only — the flat config equivalent of `overrides` in `.eslintrc` — put your patterns in a `files` key and pull the plugin's config in with `extends`. This requires the [`defineConfig()`](https://eslint.org/docs/latest/use/configure/combine-configs) helper:
+
+::: code-group
+
+```js [eslint.config.js]
+import { defineConfig } from 'eslint/config'
+import pluginVue from 'eslint-plugin-vue'
+import tsParser from '@typescript-eslint/parser'
+
+export default defineConfig([
+  {
+    files: ['packages/client/src/components/**/*.vue', 'stories/**/*.vue'],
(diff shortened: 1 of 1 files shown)


FloEdelmann (approved):
Thanks!

PR #3123, opened 2026-08-13
fix(require-typed-ref): don't report calls in explicitly typed positions

### LINKED ISSUE #3024: `vue/require-typed-ref` should not report an issue when being assigned to typed varible or parameter
`vue/require-typed-ref`

Fewer

I think this could be default behavior, but an option would also be ok if it would be disruptive to other use cases.

```vue
// Example 1
const value1: Ref<string | null> = ref(null)

// Example 2
const value2: Ref<string | undefined> = ref()

// Example 3
function test(x: Ref<string | null>) {
  // ...
}

test(ref(null));

// Example 4
function test(x: Ref<string | undefined>) {
  // ...
}

test(ref());
```

Each example reports a violation of `vue/require-typed-ref`, despite them being properly type-checked

No errors reported for these examples.

**Additional context**

It might be a separate bug that `ref(null)` is reported by this rule, since it would be properly type checked even without the type annotation since it is inferred to be `Ref<null, null>`. I see there was some [discussion](https://github.com/vuejs/eslint-plugin-vue/issues/2105#issuecomment-1573589557) about that when the rule was created, but since it does not infer to `any`, I don't think it should be reported by this rule.

ValentinYoushkevich:
`vue/require-typed-ref` skipped a `ref()`/`shallowRef()` call only when its
parent was a `VariableDeclarator` whose `id` carried a type annotation. Every
other parent fell through to an unconditional report, so a call whose type was
already pinned down by an annotation elsewhere was still flagged.

The check now covers the positions where an explicit annotation is visible in
the AST:

- `as` / `satisfies` / `<T>` assertions
- assignment to a variable or parameter declared with a type annotation
- default value of a typed parameter
- argument of a function declared in the same file whose parameter at that
  index is typed

```ts
// reported before, valid now
const count = ref() as Ref<number | undefined>

let count: Ref<number | undefined>
count = ref()

function useCount(count: Ref<number | null>) {}
useCount(ref(null))
```

Calls still get reported when the type is only known to the type checker: an
imported or member-expression callee, and contextual types such as
`return { count: ref() }` under an annotated return type.

Closes #3024

### DIFF
--- a/lib/rules/require-typed-ref.ts
+++ b/lib/rules/require-typed-ref.ts
@@ -2,14 +2,127 @@
  * @author Ivan Demchuk <https://github.com/Demivan>
  * See LICENSE file in root directory for full license.
  */
+import { findVariable } from '@eslint-community/eslint-utils'
 import { iterateDefineRefs } from '../utils/ref-object-references.ts'
 import utils from '../utils/index.js'
+
+type FunctionNode =
+  FunctionDeclaration | FunctionExpression | ArrowFunctionExpression
+
+/** `ref() as Ref<T>`, `ref() satisfies Ref<T>` and `<Ref<T>>ref()` */
+const TYPE_ASSERTIONS = new Set([
+  'TSAsExpression',
+  'TSSatisfiesExpression',
+  'TSTypeAssertion'
+])
 
 function isNullOrUndefined(node: Expression | SpreadElement) {
   return (
     (node.type === 'Literal' && node.value === null) ||
     (node.type === 'Identifier' && node.name === 'undefined')
   )
+}
+
+function hasTypeAnnotation(node: Pattern): boolean {
+  return node.type === 'AssignmentPattern'
+    ? hasTypeAnnotation(node.left)
+    : 'typeAnnotation' in node && node.typeAnnotation != null
+}
+
+/**
+ * Resolves the single definition of the given identifier reference.
+ */
+function findDefinition(context: RuleContext, node: Identifier) {
+  const variable = findVariable(utils.getScope(context, node), node)
+  return variable?.defs.length === 1 ? variable.defs[0] : null
+}
+
+/**
+ * Checks whether the given identifier refers to a variable or a parameter
+ * that was declared with an explicit type annotation.
+ */
+function isTypedVariable(context: RuleContext, node: Identifier): boolean {
+  const def = findDefinition(context, node)
+  return (
+    (def?.type === 'Variable' || def?.type === 'Parameter') &&
+    hasTypeAnnotation(def.name as Pattern)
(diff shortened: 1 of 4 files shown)


lib/rules/require-typed-ref.ts
```diff
@@ -2,16 +2,129 @@
...
+    ? init
+    : null
+}
+
+/**
+ * Checks whether the parameter receiving the argument at the given index has
+ * an explicit type annotation.
+ */
+function hasTypedParameter(fn: FunctionNode, index: number): boolean {
+  const param = fn.params[index] ?? fn.params.at(-1)
```
FloEdelmann: Why the fallback to the last parameter?

FloEdelmann (commented):
Thanks, this looks like a good improvement! I have just one question.

lib/rules/require-typed-ref.ts
ValentinYoushkevich: This is needed to handle rest parameters.

When a function is declared like:
```typescript
function useValues(...refs: Ref<number>[]) {}
```

And we call `useValues(ref())`, the argument at index 0 needs to be checked against the last parameter (rest parameter `...refs`).

The logic in `hasTypedParameter`:
1. `fn.params[index] ?? fn.params.at(-1)` — if the argument is outside the bounds of parameters, we take the last parameter
2. `if (fn.params[index] == null && param.type !== 'RestElement')` — if it's not a rest parameter, that's an explicit error (more arguments than parameters without `...`), so we return `false`
3. If it is a rest parameter with a type annotation (`Ref<number>[]`), then `hasTypeAnnotation(param)` will return `true`

This way, we correctly handle functions with rest parameters, where the type annotation applies to the rest parameter itself, not to individual arguments.

FloEdelmann: Makes sense, thanks!

lib/rules/require-typed-ref.ts
```diff
@@ -2,16 +2,129 @@
...
+  if (param == null) {
+    return false
+  }
+  if (fn.params[index] == null && param.type !== 'RestElement') {
+    return false
+  }
+  return hasTypeAnnotation(param)
```
FloEdelmann: I think this should be equivalent and is much more readable IMO:

```suggestion
  const param = fn.params[index]
  if (param != null) {
    return hasTypeAnnotation(param)
  }

  const lastParam = fn.params.at(-1)
  return lastParam?.type === 'RestElement' && hasTypeAnnotation(lastParam)
```

FloEdelmann (changes requested):

FloEdelmann (approved):

Issue #3070, opened 2026-04-15
`toRef` can pass `vue/no-dupe-keys` allowing duplicated reference in template


hanool:
**Checklist**

- [x] I have tried restarting my IDE and the issue persists.
- [x] I have read the [FAQ](https://eslint.vuejs.org/user-guide/#faq) and my problem is not listed.

**Tell us about your environment**

- **ESLint version:** 10.1.0
- **eslint-plugin-vue version:** 10.8.0
- **Vue version:** 3.5.31
- **Node version:** 24.14.1
- **Operating System:** Ubuntu24.04 (on WSL2)

**Please show your full configuration:**
I'm using default `@nuxt/eslint` module
```js
// @ts-check
import withNuxt from './.nuxt/eslint.config.mjs'

export default withNuxt(
  // Your custom configs here
)

```

**What did you do?**	

```vue
<template>
  <div>
    <!-- template will have reference for both ref foo and props.foo -->
    <p>{{foo}}</p>
  </div>
</template>
<script setup>
import { ref, toRef } from "vue"
const props = defineProps(['foo'])
// const foo = ref() // this will be errored with vue/no-dupe-keys
const foo = toRef(props, 'foo') // this passes vue/no-dupe-keys, and toRef foo.
</script>
```

**What did you expect to happen?**
`const foo = toRef(props, 'foo')` is marked as invalidation of `vue/no-dupe-keys`

**What actually happened?**
`const foo = toRef(props, 'foo')` passes `vue/no-dupe-keys`

**Repository to reproduce this issue**
[playground](https://ota-meshi.github.io/eslint-plugin-vue-demo/#eJxlUMtqxDAM/BXhy2Zhd30vbr6h9LruIY1lYppYRnZSSsi/148tLfSmkUYzGu1iJIPiSaiES5iHhL32AMq4rRa5DP2+W6LjUDK0oWxTJf/sqDiyCwkiprXQ3BKIE+zAaC+Q6BUtHGCZFtBiW1EL7UfyMUFgChGewaB1Hl8K6u6n7Hh6O2svJTRabmRSVuvOkJtpchE+3TzDOwIyE6PJOE2QxaWnq1kDXj/wK/74NIF6SVc9L1BdfuXCECPGfwL5/EfOZjgNG5ZDkNGPWHUtcctxK2jw5pE4o1t5VHtOL45vaEmAEg==)

PR #3124, opened 2026-08-13
fix(no-dupe-keys): detect locals aliasing a different prop

ValentinYoushkevich:
`onDefinePropsEnter` skips a local declaration that shadows a prop whenever the initializer mentions the props object anywhere (`isInsideInitializer`, added in #2189 so `const foo = toRef(props, 'foo')` stays valid). The check only asks whether props is touched, not which prop is read, so it also hides real collisions:

```vue
<script setup>
const props = defineProps(['foo', 'bar'])
const foo = toRef(props, 'bar')
</script>
```

The template's `foo` resolves to `props.bar` and prop `foo` becomes unreachable — exactly what the rule exists to catch.

The exemption is now skipped when the declaration provably aliases a *different* prop: `props.bar`, `props['bar']`, `toRef(props, 'bar')`, `const { bar: foo } = props` and `const { bar: foo } = toRefs(props)`. Same-name aliases keep passing, and declarations whose source prop can't be resolved (`toRef(props, key)`, `computed(() => props.bar + 1)`) are still left alone, so no new false positives.

The exact snippet from the issue (`const foo = toRef(props, 'foo')`) stays valid: it is a valid test from #2189, the setup binding wins in the template and unwraps to the same value the prop holds. Only the different-prop variants are reported. The Options API `setup()` path is untouched.

Refs #3070

The literal snippet from the report stays valid, so this does not close the issue by itself. Making that one error means reverting #2189 and reporting `const { foo } = toRefs(props)` and `const foo = props.foo` along with it — happy to redo it that way if that is what you want.

### DIFF
--- a/.changeset/no-dupe-keys-prop-alias.md
+++ b/.changeset/no-dupe-keys-prop-alias.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-vue": patch
+---
+
+Fixed `vue/no-dupe-keys` not reporting `<script setup>` variables that shadow a prop while aliasing a different one, such as `const foo = toRef(props, 'bar')`
--- a/lib/rules/no-dupe-keys.js
+++ b/lib/rules/no-dupe-keys.js
@@ -59,6 +59,92 @@ function isInsideInitializer(node, references) {
 }
 
 /**
+ * Gets the name of the prop that the given expression reads from the props object.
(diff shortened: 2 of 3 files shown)


FloEdelmann (approved):
Looks good to me, thank you! 🙂

PR #3126, opened 2026-08-15
feat: remove Reactivity Transform support from `vue/no-ref-object-reactivity-loss`

### LINKED ISSUE #3117: Remove Reactivity Transform support from `vue/no-ref-object-reactivity-loss`
The [vue/no-ref-object-reactivity-loss](https://eslint.vuejs.org/rules/no-ref-object-reactivity-loss.html) rule supports Reactivity Transform, but that proposal was cancelled, see https://vuejs.org/guide/extras/reactivity-transform.html.

I propose to remove support for Reactivity Transform in eslint-plugin-vue completely to simplify the code. That would probably only hurt users of [Vue Macros' Reactivity Transform feature](https://vue-macros.dev/features/reactivity-transform.html), but I guess it's okay to not support third-party libraries.

vaibhav8a:
Closes #3117.

Reactivity Transform was dropped from Vue, so `vue/no-ref-object-reactivity-loss` no longer needs to track reactive variables. This removes that half of the rule and the machinery behind it.

### What is removed

- The `getReactiveVariableInSameScope` report, and with it the `Identifier:exit` visitor that produced it. The rule now only reports ref objects (`refObj.value`, `{ value } = refObj`).
- `extractReactiveVariableReferences()` and its `ReactiveVariableReferenceExtractor` in `lib/utils/ref-object-references.ts`, together with the `$ref` / `$computed` / `$shallowRef` / `$customRef` / `$toRef` / `$` macro tracking, the `$$()` escape-hint handling, and the `extractIdentifier()` helper that only that extractor used. The rule was the only caller, so nothing else in the plugin loses functionality.
- The `reactive-vars` snapshot fixtures and the corresponding `describe.each` block in the util test.

`extractRefObjectReferences()` and everything ref-object related is untouched.

### Tests

The Reactivity Transform cases in the rule test are gone. I kept three of them as **valid** cases so the new behaviour is pinned rather than merely untested — code using the macros must now produce no reports:

```js
const count = $ref(0)
const value1 = count
const value2 = fn(count)
```

Full unit suite passes locally (`npm run test:base`): the 9 pre-existing file-level failures on `master` (missing `dist/` build) are unchanged, and the test count drops only by the removed Reactivity Transform cases. `npx prettier --check` is clean on every changed file, and `npm run tsc` reports no new errors (identical error count to `master` on my machine, none in the changed files).

### A note on the version bump

I marked the changeset `minor`, since users of [Vue Macros' Reactivity Transform](https://vue-macros.dev/features/reactivity-transform.html) would lose reports. Happy to switch it to `major` and hold this for the next major release instead if you would rather bundle it there — just say the word.


waynzh (changes requested):
I think dropping this support in a minor release is fine. 
I also found outdated “ref sugar is not yet supported” wording in `docs/user-guide/index.md:213`—could we update it in this PR to reflect that Reactivity Transform support has been dropped?

PR #3127, opened 2026-08-17
Add `simple-expressions-in-templates` rule

### LINKED ISSUE #253: Rule proposal: `simple-expressions-in-templates`
**Style guide:**
https://vuejs.org/v2/style-guide/#Simple-expressions-in-templates-strongly-recommended

**Description:**
This rule would enforce usage of simple expressions in templates. Instead of writing complex logic in templates it should be moved to computed properties and only referenced using them in templates.

**What should we check?**
- expressions (`{{}}`)
- bindings (`v-bind:foo=""` and `:foo=""`)

**When should we throw a warning?**
Simplest solution would be to define number of letters inside expressions, and to be honest I think it should work well for almost all cases. Let's say we throw error when there are at least 20 signs in expressions. Then we have the following cases:

```
{{todos.filter(todo => !todo.done)}}
```

piet-maier:
This PR adds the `simple-expressions-in-templates` rule. The rule corresponds to a [strongly recommended rule in the Vue Style Guide](https://vuejs.org/style-guide/rules-strongly-recommended.html#simple-expressions-in-templates). The PR also closes #253.

The Style Guide rule disallows complex expressions in "mustache" templates, but does not specify what constitutes a complex expression. Similar rules exist in other projects, including the [ESLint built-in "complexity" rule](https://eslint.org/docs/latest/rules/complexity), which uses cyclomatic complexity. However, that approach is not suitable for component templates, which consist of single expressions. This implementation instead calculates complexity by counting the number of function calls within an expression. With a default maximum of 1, the rule allows complex logic to be extracted into methods that can be called inside the template, thereby correctly allowing the good and disallowing the bad example from the Style Guide. The upper limit can also be configured if necessary.

Of course, not every expression that would be considered complex falls under this definition. Nevertheless, I believe this to be a suitable approach, in terms of both implementation and runtime complexity. It also has the advantage of enabling users to configure the allowed complexity according to their needs, in contrast to for example the `no-unreadable-*` rules of the `eslint-plugin-unicorn` package. The rule could also be extended to include `new` expressions or take complex callback functions passed to function calls into account. I welcome feedback and would be happy to advance the implementation if you have any suggestions.

### DIFF
--- a/lib/rules/simple-expressions-in-templates.ts
+++ b/lib/rules/simple-expressions-in-templates.ts
@@ -1,0 +1,37 @@
+/**
+ * @author Piet Maier
+ */
+
+'use strict'
+
+import utils from '../utils/index.js'
+
+export default {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow complex template expressions',
+      categories: ['vue3-strongly-recommended', 'vue2-strongly-recommended'],
+      url: 'https://eslint.vuejs.org/rules/simple-expressions-in-templates.html'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      simpleExpressions:
+        'Component templates should only include simple expressions, with more complex expressions refactored into computed properties or methods.'
+    }
+  },
+  create(context: RuleContext) {
+    return utils.defineTemplateBodyVisitor(context, {
+      // This pattern matches "mustache" interpolation expressions but not directive expressions.
+      'VElement > VExpressionContainer'(node: VExpressionContainer) {
+        if (node.expression?.type === 'CallExpression') {
+          context.report({
+            node,
+            messageId: 'simpleExpressions'
+          })
+        }
+      }
+    })
+  }
+}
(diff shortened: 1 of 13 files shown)


tests/lib/rules/simple-expressions-in-templates.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+const tester = new RuleTester({
+  languageOptions: {
+    parser: vueESLintParser,
+    ecmaVersion: 2020,
+    sourceType: 'module'
+  }
+})
+
+tester.run('simple-expressions-in-templates', rule, {
```
waynzh, line 19: `<template>{{ items.map(([, value]) => value) }}</template>`, `foo([, 1])`, `[foo(), , bar()]` etc. will cause a `TypeError`. Could we add some test cases for this destructuring element syntax?

lib/rules/simple-expressions-in-templates.ts
```diff
@@ -0,0 +1,92 @@
...
+'use strict'
+
+import utils from '../utils/index.js'
+
+export default {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow complex template expressions',
+      categories: ['vue3-strongly-recommended', 'vue2-strongly-recommended'],
```
waynzh, line 14: We can leave `categories` empty for now and revisit preset inclusion in a later release.
Please also add a minor changeset.

waynzh (changes requested):

Issue #3130, opened 2026-08-25
`vue/no-mutating-props`: extend to detect nested mutation of `defineModel()` refs


faizanu94:
**What rule do you want to change?**

`vue/no-mutating-props`

**Does this change cause the rule to produce more or fewer warnings?**

More. It adds detection for a case that currently passes silently.

**How will the change be implemented? (New option, new default behavior, etc.)?**

New default behavior (no new option). `defineModel()` is sugar for a `modelValue` prop + `update:modelValue` emit, so mutating it in place is the same category of bug the rule already catches for `defineProps()`. This mirrors the precedent set by #2360 and #3032, which taught other rules about `defineModel()`.

**Please provide some example code that this change will affect:**

```vue
<script setup>
const model = defineModel()

function bad() {
  model.value.name = 'x'       // nested property mutation
  model.value.items[0] = 'y'   // nested index assignment
  model.value.items.push('z')  // mutating array method
}

function good() {
  model.value = { ...model.value, name: 'x' } // correct: reassigns the ref
}
</script>
```

**What does the rule currently do for this code?**

Nothing — 0 errors. Verified by running the current rule (v10.10.0) via ESLint's `Linter` against the parsed SFC. For comparison, the *identical* mutation shape on a `defineProps()` object IS flagged:

| Case | Current result |
| --- | --- |
| `defineModel()` nested mutation (the 3 `bad()` lines) | 0 errors |
| `defineProps(['user'])` then `props.user.name = 'x'` / `.push()` | 2 errors |
| `defineModel()` correct pattern (`model.value = {...}`) | 0 errors |

Root cause: the rule only registers `onDefinePropsEnter`; there is no `onDefineModelEnter` handler, even though that visitor hook already exists.

**What will the rule do after it's changed?**

Flag the three `bad()` lines as `unexpectedMutation`, while leaving `good()` (and the `shallowOnly` option semantics) untouched.

**Additional context**

Why it's a real bug, not style: `model.value = next` goes through the ref setter and emits `update:modelValue`, syncing the parent. `model.value.foo = x` skips the setter, so no emit fires and the parent silently desyncs.

Scope note: this is a syntactic check on direct `model.value.<path>` mutation. It will not track mutation via an intermediate alias (`const i = model.value.items[0]; i.x = 1`) — that needs data-flow analysis and I'd suggest keeping it out of scope for v1.

I'm happy to open the PR (implementation + tests + docs + changeset) if this is accepted.

[...]

PR #3131, opened 2026-08-25
feat(no-mutating-props): report nested mutation of `defineModel()` refs

faizanu94:
Refs #3130

## Summary

`defineModel()` returns a ref that bridges a prop and its `update:modelValue` emit. Reassigning `model.value` goes through the ref setter and emits the update, but mutating a nested property in place bypasses the setter, never emits, and silently desyncs the parent:

```vue
<script setup>
const model = defineModel()

// ✗ bypasses update:modelValue — silently desyncs the parent
model.value.name = 'x'
model.value.items[0] = 'y'
model.value.items.push('z')

// ✓ goes through the ref setter, emits update:modelValue
model.value = { ...model.value, name: 'x' }
</script>
```

`vue/no-mutating-props` already reports this exact class of bug for `defineProps()`, but ignored `defineModel()` because it only registered `onDefinePropsEnter`. This wires `defineModel()` bindings into the same `utils.findMutating` pipeline used for destructured props, via the existing `onDefineModelEnter` hook (added in #2360). This follows the same "teach an existing rule about `defineModel()`" pattern as #2360 and #3032.

## Behavior

- **Reports** nested mutation of a `defineModel()` ref: property assignment, index assignment, arbitrary depth, `++`/`delete`, mutating array methods (`push`, `splice`, …), and `Object.assign(model.value, …)`.
- **Allows** reassigning the ref itself (`model.value = next`), which is the correct update pattern.
- **`shallowOnly: true`** continues to permit nested mutation, consistent with its meaning for props.
- **No change** to `defineProps()` / Options API behavior.

## Known limitation (deliberate, out of scope for this PR)

This is a syntactic check on direct `model.value.<path>` mutation. It does not track mutation through an intermediate alias (`const i = model.value.items[0]; i.x = 1`), which would require data-flow analysis. Happy to note this in the docs or track it separately if preferred.

## Test plan

- [x] Added valid + invalid cases to `tests/lib/rules/no-mutating-props.test.ts` (nested property/index/depth, mutating calls, `Object.assign`, named model, correct-reassign patterns, unrelated locals, `shallowOnly`, and a `defineProps` regression guard).
- [x] `npx vitest run tests/lib` → 288 files, 11628 tests pass.
- [x] `npx eslint` on changed files and `npx markdownlint "**/*.md"` clean.
- [x] `npm run build` succeeds.
- [x] Added a `minor` changeset.

### DIFF
--- a/lib/rules/no-mutating-props.js
+++ b/lib/rules/no-mutating-props.js
@@ -258,6 +258,62 @@ module.exports = {
       }
     }
 
+    /**
+     * Verifies that a `defineModel()` ref binding is not mutated through `.value`.
+     *
+     * `defineModel()` returns a ref that bridges a `modelValue` prop and an
+     * `update:modelValue` emit. Reassigning the ref (`model.value = next`) goes
+     * through the ref setter and emits the update, so it is the correct pattern and
+     * is intentionally allowed. Mutating a nested property in place
+     * (`model.value.foo = 1`, `model.value.items[0] = 1`, `model.value.items.push(1)`,
+     * `Object.assign(model.value, {...})`) bypasses the setter, never emits, and
+     * silently desyncs the parent, so it is reported.
+     * @param {Identifier} defId The identifier the model ref is bound to.
+     * @param {string} name The model's prop name, used in the report message.
+     */
+    function verifyModelVariable(defId, name) {
+      const variable = findVariable(utils.getScope(context, defId), defId)
+      if (!variable) {
+        return
+      }
+
+      for (const reference of variable.references) {
+        if (!reference.isRead()) {
+          continue
+        }
+        const id = reference.identifier
+        const valueNode = id.parent
+        if (
+          !valueNode ||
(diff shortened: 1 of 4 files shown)


lib/rules/no-mutating-props.js
```diff
@@ -351,6 +407,17 @@ module.exports = {
             }
             verifyPropVariable(prop, path)
           }
+        },
+        onDefineModelEnter(node, model) {
+          if (
```
waynzh, line 412: Thanks! This looks like a useful extension.
Could we also handle the array-destructured form of defineModel()? The `Identifier` check currently skips this standard usage: 
```
<script setup>
const [model, modifiers] = defineModel()
model.value.foo = 1
</script>
```

faizanu94: Great catch, thanks! Handled the array-destructured form `const [model, modifiers] = defineModel()` takes the first element as the model ref through the same findMutating pipeline, with a test matching your example and a docs note. Ready for review!

PR #3134, opened 2026-09-08
fix(no-undef-properties): support `...toRefs(reactive({}))` in setup() return

### LINKED ISSUE #1915: vue/no-undef-properties have some error when use toRefs
```js
    "eslint": "8.18.0",
    "eslint-plugin-prettier": "4.0.0",
    "eslint-plugin-vue": "9.1.1",
```
```js
module.exports = {
  root: true,
  globals: {
    uni: 'readonly',
    wx: 'readonly',
    App: 'readonly',
    ProcessEnvVueAppEnv: 'readonly',
    ProcessEnvWxPlatform: 'readonly',
    ProcessEnvBuildTime: 'readonly',
    getCurrentPages: 'readonly',
    getApp: 'readonly',
    PropType: 'readonly',
    AnyObject: 'readonly',
    Ref: 'readonly'
  },
  extends: ['plugin:vue/vue3-recommended'],
  parser: 'vue-eslint-parser',
  parserOptions: {
    createDefaultProgram: true,
  },
  plugins: ['@typescript-eslint', 'vue'],
  env: {
    browser: true,
    node: true,
  },
  rules: {
    'vue/no-undef-properties': 'error',
  },
  overrides: [
    {
      files: ['*.js', '*.vue'],
      rules: {
        'no-undef': 'error',
      },
    },
  ],
};
```

```vue
<template>
  <div>{{ a }}</div>
</template>
<script>
  setup() {
    const data = reactive({
      a: 1,
      b: 1,
    });
    return {
      ...toRefs(data),
    };
  },
});
</script>
```
image

chiliec:
Fixes #1915

## What

`vue/no-undef-properties` reported every template reference as undefined when a component exposes state via `...toRefs(data)` from `setup()`, where `data` is a `const` bound to `reactive({ ... })` (the exact example from the issue). The rule only collected explicit `Property` entries from the `setup()` return object and ignored spreads.

## Fix

In `onVueObjectEnter`, walk the `setup()` return object(s) for `SpreadElement`s whose argument is a `toRefs(...)` call (tracked via `ReferenceTracker`, same approach as `property-references.ts`) and resolve its argument to a `reactive({ ... })` object literal — either inline (`toRefs(reactive({...}))`) or through a single-definition `const` initialized with `reactive({...})`. The keys of that literal are then registered as defined setup properties. Anything not statically resolvable (non-`const`, reassigned, non-literal `reactive` arg, unknown call) is left untouched, so existing reporting is unchanged.

## Tests

- valid: `const data = reactive({a, b}); return { ...toRefs(data) }` (issue repro) and inline `setup: () => ({ ...toRefs(reactive({ a })), ...toRefs(reactive({ b, c })) })`
- invalid: same shape but the template references `c` which is not in the reactive object → still reported

## Validation

- `npx vitest run tests/lib/rules/no-undef-properties.test.ts` → 81 passed (78 before + 3 new)
- RED→GREEN: with the rule change stashed, the 3 new cases fail (`'a' is not defined.`, `'b' is not defined.`, and the invalid case reports extra `a`/`b`); with it restored, all pass
- `npm run build && npx vitest run tests/lib` → 288 files, 11624 tests passed
- `npx eslint lib/rules/no-undef-properties.ts tests/lib/rules/no-undef-properties.test.ts` → clean; `prettier --check` clean
- `npx tsc` → 27 pre-existing errors (all `../..`/`../../../dist` module resolution in tests before `npm run build`), none in the touched files; identical count on a clean checkout

Changeset added (`patch`). Happy to adjust.

### DIFF
--- a/.changeset/no-undef-properties-to-refs.md
+++ b/.changeset/no-undef-properties-to-refs.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+Fix `vue/no-undef-properties` false positives for properties returned via `...toRefs(reactive({ ... }))` from `setup()`
--- a/lib/rules/no-undef-properties.ts
+++ b/lib/rules/no-undef-properties.ts
@@ -8,6 +8,7 @@ import reserved from '../utils/vue-reserved.json' with { type: 'json' }
 import { toRegExpGroupMatcher } from '../utils/regexp.ts'
 import { getStyleVariablesContext } from '../utils/style-variables/index.ts'
 import { definePropertyReferenceExtractor } from '../utils/property-references.ts'
+import { ReferenceTracker } from '@eslint-community/eslint-utils'
 
 interface PropertyData {
   hasNestProperty?: boolean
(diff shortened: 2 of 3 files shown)


.changeset/no-undef-properties-to-refs.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-vue': patch
+---
+
+Fix `vue/no-undef-properties` false positives for properties returned via `...toRefs(reactive({ ... }))` from `setup()`
```
FloEdelmann, line 5:
```suggestion
Fixed `vue/no-undef-properties` false positives for properties returned via `...toRefs(reactive({ ... }))` from `setup()`
```

lib/rules/no-undef-properties.ts
```diff
@@ -101,6 +102,137 @@ export default {
...
+        toRefs: { [ReferenceTracker.CALL]: true }
+      })
+      for (const { node, path } of references) {
+        if (path.at(-1) === 'reactive') {
+          reactive.add(node)
+        } else {
+          toRefs.add(node)
+        }
+      }
+      return (reactiveApiNodes = { reactive, toRefs })
```
FloEdelmann, line 130: Please split this into two statements.

FloEdelmann (changes requested):
Makes sense to me. I have two small suggestions, see below.

